PR #1465, opened 2019-07-05
Handle inbox items dynamically

thiagodelgado111:
## Description

This PR adds different renderers for different inbox event types so we can simplify the InboxItem component

**New stuff** ✨
* TBD

**Changes** 🏗
* Extrapolate contract and store addresses based on the events we get for normalization

## TODO

- [x] Get  contract and store addresses right out of the event we're normalizing
- [ ] Hook event types to a specific rendered so we can keep the inbox item component dummy

### DIFF
--- a/src/data/normalizers/index.js
+++ b/src/data/normalizers/index.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import type { Event } from '../types';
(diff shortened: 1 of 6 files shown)


chmanie:
Closing this as most of this was handled/touched in #1816

Issue #1597, opened 2019-07-22
No check for ETH when displaying Fund Colony CTA on Colony Home


collinvine:
## Steps to reproduce

1. Create a new colony (BYOT in my case)
2. Fund your colony with Ether, but not with the native token
3. Claim eth into your pot
4. Go to Colony Home; the Funding CTA still appears and has copy reading "end any ETH or ERC20 tokens to your colony's address below"

## Expected behaviour

* This to disappear if the colony has received ETH as funding
* (Or for me to be able to X it out and dismiss it)

## Actual behaviour

* Doesn't check for ETH balance, and is still displayed even after I've funded my colony

![Screen Shot 2019-07-22 at 2 46 17 PM](https://user-images.githubusercontent.com/15755488/61667676-7f722a80-ac8f-11e9-8e80-db1a7fd0e53c.png)
![Screen Shot 2019-07-22 at 2 46 22 PM](https://user-images.githubusercontent.com/15755488/61667681-813bee00-ac8f-11e9-9aac-e12111a7ccfb.png)

## Suggestions (optional)
If possible, check for _any_ token that's been sent to the Colony.

If that's not possible, then check for ETH in addition to the native token:

1. Check for ETH
2. Allow users to `x` dismiss the section (🌮)
3. Update copy so it reads: "Send ETH or `${nativeToken}` tokens to your colony's address below:"

PR #1653, opened 2019-08-02
Validate that the token balance exists before setting a task worker payout

### LINKED ISSUE #1535: Funding task with token that isn't available
## Steps to reproduce

1. Create a new colony and BYOT (I used DAI)
  a. interestingly, I used the mainnet DAI address [here](https://etherscan.io/token/0x89d24a6b4ccb1b6faa2625fe562bdd9a23260359) instead of the Goerli one [here](https://goerli.etherscan.io/token/0x72560b830ced423fbb9ec1ae8d01b41f015a5f21).
2. Create a new task
3. Assign a worker and fund the task (I add 1 DAI)
4. The transaction went through, even though I haven't sent DAI to the Colony

## Expected behaviour

* This would fail because my colony doesn't have any DAI in its address

## Actual behaviour

* The transaction went through

Screen Shot 2019-07-16 at 2 30 47 PM

* After a refresh, the token amount is not displayed but finalize button is active:

Screen Shot 2019-07-16 at 2 36 31 PM

* Clicking `finalize` tx fail of course:
Screen Shot 2019-07-16 at 2 38 30 PM

* After the task is cancelled, the **DAI 1** reappears
Screen Shot 2019-07-16 at 2 41 24 PM

## Suggestions (optional)

* Could validate if the token is available to fund the task first?

JamesLefrere:
## Description

This PR carries out the suggestion in #1597 to validate that the given token balance exists in the colony pot before setting the task worker payout.

**Changes** 🏗

* Validate that the balance exists in the colony pot for the given token when attempting to set a task payout
* Ensure token addresses are of the `Address` type

## TODOs

- [ ] Check that this works

Resolves #1535

### DIFF
--- a/src/data/types/TaskEvents.js
+++ b/src/data/types/TaskEvents.js
@@ -59,7 +59,7 @@ export type TaskEvents = {|
     typeof PAYOUT_SET,
     {|
       amount: string,
-      token: string,
+      token: Address,
     |},
   >,
   PAYOUT_REMOVED: EventDefinition<typeof PAYOUT_REMOVED, void>,
--- a/src/modules/dashboard/data/commands/task.js
+++ b/src/modules/dashboard/data/commands/task.js
@@ -397,7 +397,7 @@ export const setTaskPayout: Command<
   TaskStoreMetadata,
   {|
     amount: BigNumber,
-    token: string,
+    token: Address,
   |},
   {|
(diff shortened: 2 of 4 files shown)


JamesLefrere:
Ok, I think I'll close this because it's basically symptom fighting and doesn't solve any real issue (i.e., it hasn't been possible to reproduce the issue since this was reported).

PR #1715, opened 2019-08-13
Add interactive InfoPopover

### LINKED ISSUE #1656: Interactive user avatar popovers
### Story

As a _task owner_, I want to _double-check if the assignee wallet address is the right one_, so that _I feel more confident completing a task_.

### Description

1. The general idea is to make avatars and usernames clickable (task UI and colony home) to copy the wallet address quickly. It is not convenient now. 
2. On figma, there is more detailed specification which areas are part of this task. It is not a global change. Edge cases are also included (f.e. if there is no username, but only the wallet address).
3. On click, the popover appears; on clicking off (i.e. somewhere on the background) it hides. It should be possible to copy the wallet address and highlight text inside a popover. It would be nice to add a subtle ease-in-out transition. 

### Design

[Figma link](https://www.figma.com/file/IsTzagr0TvzEMf3nwIcTthLm/Task-Workflow?node-id=4448%3A0)

image

- [x] The `UserInfo` component is replaced with this one
- [x] The user display name is shown when available (no biggie if not)
- [ ] The username is shown when available (no biggie if not) and links to the user profile
- [x] The user address is always shown and is copyable
- [ ] 🌮The popover is opened on hover (but this is quite complicated, so don't worry too much)

### Suggestions

* We should come up with a more meaningful name for the component (we like the sound of `InfoPopover`)
* We should open another issue to do this for colonies

### Subtasks

- [x] make InfoPopover work in `TaskFeedEvent`
- [x] make InfoPopover work in `UserInfo` component

karolinkas:
## Description

This adds a change to the UserInfo popover to allow a user the copy a wallet address right from the popover and renames the same component. We had thought a global change would be required but it's actually quite easy to show InfoPopover only on some screens just like required by Karol. 

## Changes
- [x] Rename UserInfo to InfoPopover
- [x] allow clicking on UserAvatar to get InfoPopover
- [x] allow clicking on username to get InfoPopover
- [x] add CopyableAddress component in tooltip content 
- [x] change trigger to clicking 
- [x] show InfoPopover on UserAvatar on some more occurences as suggested in the figma file: ( that includes TaskListItem, InboxItem, ColonyMeta, TaskFeedComment, TaskFeedEvent  )
https://www.figma.com/file/IsTzagr0TvzEMf3nwIcTthLm/Task-Workflow?node-id=4448%3A677
 
## Current State on ColonyHome
userInfoPopover

## Current State on TaskFeedEvents
taskFeedEvent

## Current State on Task Comments
popoverOnComment

Resolves #1656

### DIFF
--- a/src/modules/core/components/InfoPopover/InfoPopover.jsx
+++ b/src/modules/core/components/InfoPopover/InfoPopover.jsx
@@ -35,7 +35,7 @@ const renderTooltipContent = (user?: UserType) => {
       )}
       {username && (
         <span title={username} className={styles.userName}>
-          <UserMention username={username} hasLink={false} />
+          <UserMention username={username} hasLink />
         </span>
       )}
       {walletAddress && <CopyableAddress full>{walletAddress}</CopyableAddress>}
--- a/src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonyMeta.css
+++ b/src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonyMeta.css
@@ -39,6 +39,7 @@
 .userAvatar {
   display: inline-block;
   margin: 9px 9px 0 0;
+  cursor: pointer;
 }
 
 .colonyAvatar {
--- a/src/modules/dashboard/components/TaskFeed/TaskFeedEvent.jsx
+++ b/src/modules/dashboard/components/TaskFeed/TaskFeedEvent.jsx
@@ -142,7 +142,7 @@ const TaskFeedEventDomainSet = ({
   },
   intl: { formatMessage },
 }: *) => {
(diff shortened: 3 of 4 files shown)


auryn-macmillan:
Just testing this out in the dapp.
For addresses that do not have a username, the popover does not show up.

This could occur for one of two reasons; the data has not loaded yet or an address has been assigned as an admin or worker but has not yet created a colony account.

In this case, I think we should still show the popover so that people can see the address.

sprusr:
@auryn-macmillan yeah I noticed that too, also that the pointer cursor is shown even when clicking on it doesn't do anything (when there's no profile) - our bad! Could you open an issue for that please? Thanks!

PR #1726, opened 2019-08-19
Hide Colony Recovery for Non-Founders

rdig:
## Description

This PR is a simple fix to prevent showing the Colony DB Recovery button to non-founders 

_(Although they couldn't actually use that feature, since it checks for permissions prior to firing that action)_

**New stuff**

- [x] `dashboard` module `isFounder` check

**Changes**

- [x] `ColonyHome` check if founder before showing the recovery button

**Note**

Rebased on the TS `master` branch

### DIFF
--- a/src/modules/dashboard/checks.ts
+++ b/src/modules/dashboard/checks.ts
@@ -150,5 +150,5 @@ export const canFinalizeTask = (
 /*
  * Permissions
  */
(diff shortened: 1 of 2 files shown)


rdig:
@chmanie rebased on `master` TypeScript branch

Please have another look when you have time

PR #1747, opened 2019-08-21
Add colonyDomainUserRolesSet saga

### LINKED ISSUE #1697: Add granular domain-specific role setting
### Story

As a Colony developer, I want to be able to set a specific role for a certain user in a specific domain of a colony.

### Description

> tl;dr - new set permission saga which sets the roles of a user in a specific domain - accepts map of role to boolean and diffs with existing roles to determine which need updating with txs

Right, so now we can query permissions according to the new roles (#1696), but we're still setting them exclusively using the `OldRoles` extension contract. Now we need to be able to set the individual new roles.

We're going to want a new action/saga which accepts a `colonyAddress`, a `domainId` within that colony, and a map of `{ [roleId]: boolean }`. The saga should then work out which of those roles already apply to the current user, and create a group of transactions to make up the difference. It should also check to make sure that the current user (the one who's logged in, calling this function) has the appropriate permission to be calling this function - if not, throw a descriptive error.

Additionally, we're going to need a flag that we can set for a user's permissions per domain of a colony while there are transactions in progress for altering them. The saga should set this flag when the transaction group starts, and remove it once they're all done.

### Design

Still no designs here!

- [ ] You can set individual roles for a user in a specific domain!
- [ ] The redux state has some kind of pending flag set for that user's permissions in that domain while the transactions are happening

### Suggestions

* This _pending_ flag could be put alongside the permissions, so the redux state would be like `{ [domainId]: { [roleId]: boolean, pending: boolean } }`

### Subtasks

- [ ] Populate if you need!

sprusr:
## Description

We can set individual roles for an address in a domain now!

**New stuff** ✨

* The `colonyDomainUserRolesSet` saga added, along with actions
* Added `pending` state to user domain roles
* Added intl messages for role set transactions

**Changes** 🏗

* -

**Deletions** ⚰️

* -

## TODO

- [ ] Decide if it needs any additional error handling (if the current user doesn't have permission to set the roles it won't break, but transactions will just fail)

Resolves #1697


src/modules/dashboard/sagas/roles.ts
```diff
@@ -54,6 +62,82 @@ function* colonyDomainUserRolesFetch({
...
+
+    default:
+      throw new Error(`Invalid role to set: ${role}`);
+  }
+};
+
+function* colonyDomainUserRolesSet({
+  payload: { colonyAddress, domainId, userAddress, roles },
+  meta,
+}: Action<typeof ActionTypes.COLONY_DOMAIN_USER_ROLES_SET>) {
```
JamesLefrere: With TS, `typeof` is no longer necessary here 👍

src/modules/dashboard/sagas/roles.ts
```diff
@@ -54,6 +62,82 @@ function* colonyDomainUserRolesFetch({
...
+          group: {
+            key: 'setRoles',
+            id: meta.id,
+            index,
+          },
+        }),
+      ),
+    );
+
+    yield put<Action<typeof ActionTypes.COLONY_DOMAIN_USER_ROLES_SET_SUCCESS>>({
```
JamesLefrere: Likewise, you can remove `typeof`. In fact, it's also possible to do `yield put<AllActions>` and it's smart enough to work it which action by the type.

src/modules/dashboard/sagas/roles.ts
```diff
@@ -54,6 +62,82 @@ function* colonyDomainUserRolesFetch({
...
+      throw new Error(`Invalid role to set: ${role}`);
+  }
+};
+
+function* colonyDomainUserRolesSet({
+  payload: { colonyAddress, domainId, userAddress, roles },
+  meta,
+}: Action<typeof ActionTypes.COLONY_DOMAIN_USER_ROLES_SET>) {
+  try {
+    const existingRoles = yield* executeQuery(getColonyDomainUserRoles, {
```
JamesLefrere: With TS, you can now do `yield execute`

PR #1752, opened 2019-08-22
Don't show Mint Tokens link for Colonies that brought they're own tokens

### LINKED ISSUE #1750: Mint new tokens is visible for BYOT users
## Steps to reproduce

1. Create a new colony and BYOT
2. Navigate to the tokens tab in the colony's settings

## Expected behaviour

There should be no "mint new tokens" button.

## Actual behaviour

There is a "mint new tokens" button.
At 35:06 on [this video](https://drive.google.com/open?id=1FyLbsmA4qoMP-vC7zAGIHZd0uabgnj_e) you can see this in the wild.

rdig:
## Description

This PR fixes the logic that displays the _Mint Tokens_ dialog link in the admin section, for users that created the Colony using an external token (byot).

Before it was just using the colony's `canMintNativeToken` value for this check, but this PR also adds the check for the `isExternal` value from the token reference.

It's now using the same logic to display this, as we do on the Colony Home mint tokens widget.

**Changes**

- [x] admin `Tokens` also check the token reference when deciding to show the mint tokens dialog

**Screenshot**

Screenshot from 2019-08-22 16-50-40

Resolves #1750

### DIFF
--- a/src/modules/dashboard/data/queries/colonies.ts
+++ b/src/modules/dashboard/data/queries/colonies.ts
@@ -533,7 +533,7 @@ export const getColonyCanMintNativeToken: Query<
   async execute(colonyClient) {
     // wrap this in a try/catch since it will fail if tokens can't be minted
     try {
-      await colonyClient.mintTokens.call({ amount: new BigNumber(1) });
+      await colonyClient.mintTokens.estimate({ amount: new BigNumber(1) });
     } catch (error) {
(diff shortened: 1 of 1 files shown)


rdig:
@sprusr 
- Removed previous changes
- Reverted the `getColonyCanMintNativeToken` query to use `estimate`

Have another look when you have to time

PR #1764, opened 2019-08-26
A little bit of summer cleaning

chmanie:
## Description

This PR removes some unused packages to look more shiny for the outside world.

**Deletions** ⚰️

* Remove unused `flow-typed` dir
* Remove plop (including templates)
* Remove greenkeeper badge


plopfile.js
```diff
@@ -1,96 +0,0 @@
-const prompts = {
```
sprusr: Wow I didn't know we had this! Why remove it - does literally nobody use it?

chmanie: I think it's outdated now and I don't think anyone uses it.

PR #1770, opened 2019-08-27
Add new permissions dialog and wire it up

### LINKED ISSUE #1709: Create DLP add/edit role dialog component
### Story

As a colony admin, I want to be able to edit the roles for a user in a certain domain of my colony, so that I can control who has permissions.

### Description

This is an interesting dialog component. We're going to want to be using it for both adding new permissions to a user, as well as editing existing roles. What the component displays varies based on how it was opened, and also based on the permissions of the user interacting with it.

### Design

[Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=998%3A24864)

We have an added link in the sidebar "Add Role", which when clicked will open the dialog.

You can also access the dialog by clicking on a user in the list who already has a role. The user will already be selected in the dialog, and the field will not be editable. The dialog title should change accordingly.

Checkboxes should be populated with the permissions the selected user already has. Checkboxes should be disabled if the current user doesn't have permission to change them in this domain. The arbitration role should be always disabled with "Coming soon".

- [ ] Clicking "Add Role" in the sidebar opens the add/edit dialog in the "add" configuration
- [ ] The dialog features a user picker (populated with users currently in state)
- [ ] Dialog is disabled if opened in "edit" configuration
- [ ] Dialog returns/dispatches/something an object ~`{ colonyAddress, domainId, roles: { [roleId]: boolean } }`

### Suggestions

* Can probably use a `DialogActionButton` with fake actions for now

### Subtasks

- [ ] ???

### LINKED ISSUE #1710: Wire DLP add/edit roles dialog
### Story

As a colony admin, I want to be able to edit the roles of a user in a certain domain of my colony, so that I can control what permissions users have in my colony.

### Description

Time to wire up this dialog! We will already have the saga and actions by now to dispatch, and also the selector for the data about which roles a user has.

When selecting the user from the picker (or when the dialog is opened with a user already selected), we should dispatch the action to refetch the roles data for that user in the selected domain to ensure we have fresh data (this should be done regardless - TTL of zero).

When selecting the data for what roles the selected user already has, we should use the selector that only selects roles/permissions in that specific domain (not including parent domains). When we select the data for whether the current user is allows to give/remove certain permissions in the domain, we should include parent domain roles too.

Submitting the form should dispatch the action to call the saga which diffs the selected roles with those that the user already has in that domain, and create transactions accordingly. This will update the state of that user's roles to pending (which shows the pending indicator on the permissions/roles list).

We also need to make it so that clicking a user in the permissions list causes the edit roles dialog to appear with the appropriate data pre-filled.

### Design

No more designs for this!

- [ ] Clicking on a user in the permissions list opens the edit roles dialog with that user already selected
- [ ] The roles that a user has are checked accordingly
- [ ] Checkboxes are disabled if the current user is unable to change that role in that domain
- [ ] Submitting the form causes transactions to be created and the user's permissions set to pending in the list

### Suggestions

* 🤷‍♀️

### Subtasks

- [ ] ???

karolinkas:
## Description

Adding a new dialog that receives data grom the permissions screen while allowing to edit but also to create new permissions. When a user is being passed in to the dialog, which is optional, the dialog is in edit mode. 

This issue includes also wiring of the dialog.

## Achtung!
It should be merged after the permissions tab in #1772, because only then I can trigger the dialog when clicking on users as well as specified in  #1710, but this takes a moment and is prepared in the dialog, just commented out. This branch is currently rebased on top of feature/1699-dlp-list-component.
 
## Figma link
https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF

## TODO

- [x] Clicking "Add Role" in the sidebar opens the add/edit dialog in the "add" configuration
- [x] The dialog features a user picker (populated with users currently in state)
- [x] Dialog is disabled if opened in "edit" configuration
- [x] Dialog returns/dispatches/something an object ~{ colonyAddress, userAddress, domainId, roles: { [roleId]: boolean } }

Resolves #1709
Resolves #1710


src/modules/admin/components/Permissions/Permissions.tsx
```diff
@@ -11,36 +15,58 @@ const MSG = defineMessages({
...
-    </aside>
-  </div>
-);
+// eslint-disable-next-line no-unused-vars
+const Permissions = ({ colonyAddress, openDialog }: Props) => {
+  const handleEditPermissions = useCallback(
+    () =>
+      openDialog('ColonyPermissionEditDialog', {
+        colonyAddress,
+        domain: { name: 'Cooking', id: 5 },
```
karolinkas: This `domain: { name: 'Cooking', id: 5 },` is just hardcoded until the Permissions screen is ready and merged. Will replace with real values before merging.

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+              </div>
+              <InputLabel label={MSG.permissionsLabel} />
+              {availableRoles.map((role, id) => (
+                <div className={styles.permissionChoiceContainer}>
+                  <Checkbox
+                    className={styles.permissionChoice}
+                    key={role}
+                    value={role}
+                    name="roles"
+                    disabled={role === 'Arbitration' || role === 'Root'}
```
JamesLefrere: I guess this should probable use `checkIfCanBeSet`?

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+        .filter(({ data }) => !!data)
+        .map(({ data, key }) => ({
+          id: key,
+          ...data,
+        })),
+    [userData],
+  );
+
+  const availableRoles = [
+    'Root',
```
JamesLefrere: Are these defined in a constant anywhere?

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+
+  // When updating the selected user fetch that user's Roles
+  const [currentUser, setCurrentUser] = useState(null);
+
+  const updateCurrentUser = useCallback(({ profile: { walletAddress } }) => {
+    setCurrentUser(walletAddress);
+  }, []);
+
+  // Get current user roles to populate the checkboxes
+  const useCurrentUserForRoles = (): Array<string> => {
```
JamesLefrere: The name of this function uses the convention for hooks, but it's a function. I also find it a little difficult to read, but maybe it can be much simpler with a mapping instead.

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+          return accumulator;
+        }, {}),
+      })),
+      mergePayload({ colonyAddress }),
+    ),
+    [colonyAddress],
+  );
+
+  // When updating the selected user fetch that user's Roles
+  const [currentUser, setCurrentUser] = useState(null);
```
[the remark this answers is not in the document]
karolinkas: Actually it refers to the user that is currently selected in the SingleUserPicker, but I can rename it anyway.

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
karolinkas: I kind of believed I made a custom hook here, that's why I called it like that. :-) I must be wrong here. I am still not sure what turns it into a hook actually.

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+  };
+  const selectedRoles = useCurrentUserForRoles();
+
+  return (
+    <Dialog cancel={cancel}>
+      <ActionForm
+        initialValues={{
+          domainId: domain.id,
+          colonyAddress,
+          roles: selectedRoles,
```
karolinkas: @sprusr I just noticed that the checkboxes in the dialog are being initialised before the `useUserDomainRoles` hook even returns the user roles. Any idea how to work around that?

sprusr: Maybe `selectedRoles && <ActionForm`...? So that it doesn't render until there is a value there. Otherwise can use `setValues` which is in `FormikProps` (available line 236 here).

karolinkas: Will try, thank you for thinking about it!

karolinkas: I found out what was missing, a`enableReinitialize` property on the ActionForm.

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -0,0 +1,300 @@
...
+      colonyAddress,
+      domain.id,
+      // selectedUser ||
+      currentUser,
+    );
+    if (data) {
+      return Object.keys(data).reduce((accumulator, role) => {
+        if (data[role]) {
+          // Role is capitalised in the ddb and needs to be readable
+          const roleLow = role.toLowerCase();
```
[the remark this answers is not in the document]
karolinkas: Could you elaborate here? Wouldn't an enum mapping be useful to get from one object to another object? We want an array though.

For example, we wanna get from 
`{ ADMIN:  true, ROOT: false, FUNDING: true}`
to 
["ADMIN", "FUNDING"]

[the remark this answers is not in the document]
JamesLefrere: Sure, but you can also use the mapping to get to an array: http://www.typescriptlang.org/play/?alwaysStrict=false&target=3#code/FDCmDsFcFsAICUD2AbUBnWBvYtYEEARAWQEkA5WAXlgHI8ATaAS3BoBod94BhACRIAqAUW4CAqvCFVaeAE4BjABZMALqHkrIs0O0554AIUHw8AkgHkK1OrIBGq2QEMVTRKw65458wOk0kiCq6uABiYmQE5ADifiGQ4PQsAOa6AL4gKgCeAA6gCCh51ADWoJmIAGawWbkV+ahoANwgAPTNsAI56PKyTNkqsAAG8OqIsvQDbLDggbDlsohwTNDQkCqOtqgNVcoYqCoYkBjdoM55juCwiLYAVuoqwK1VnbAA7qqKsI6w8qCyayxPXKXSolTKwAAU03610O-TQKh64CSl1kUxgtl+AEpgNU8gBlUCoDSgehIerSAAKjj+TEcyAAPMN5KN6IyCpNbIgCucAHw8poPNoAAX2AFomElptpgMzwPDPgA3RxMZDrVBk9DScw3O4AOiVyEg6HBGrQmK2jwA2nRGCx2DIFMo1BotDpJrqPQBdECy+VJUAqU3ScHzeoALlgBKJalJBTNEdNls9VB5nG1tw0utBaBDcexuFwuvKKrUslzqBTsFD6Et1c+GA1nvzBd10Ec2XLhR5dRrdccDYKTYFvoKuuQiCS4P9gbj4Mw+GI5AjCKNkzCEWiEfKdLQeVSmPNsCtNuYrG9QA

I think you get the best of both worlds this way, and you can also map the other way.

PR #1771, opened 2019-08-27
Cache purser identities

### LINKED ISSUE #1296: Keep identities on indexedDB so the dapp can pick them up
### Story

Example: As a *user creating a colony account*, I want to *sign my identity once or every once in x time*, so that *I can save some time and have a better experience*

Currently, we have to sign new identities every time we connect. So it should be possible to reuse identities to avoid having to sign every time we connect a wallet. We only have to serialize the identity and store it somewhere safe. If it's lost (e.g: I delete the indexedDB containing it) then we can create a new one. if we log in using a different device, we would still need to create a new one though

- [ ] The identity is serialized and stored on indexedDB safely
- [ ] When you connect, if there's a valid identity for your wallet, we load it back from indexedDB instead of asking the user to sign again every time

thiagodelgado111:
## Description

This PR adds a cache for purser identitie so users doesn't have to generate a new orbit key every time they log in. They create a key the first time they log in then after that they just have to unlock their wallet.

**Changes** 🏗

* `PurserIdentityProvider` now uses `localforage` for a identity cache

## TODO
- [x] Add `localforage` to PurserIdentityProvider
- [x] Check if there is a serialized identity for the current wallet address
- [x] Load that identity instead of creating a new one

Closes #1296

### DIFF
--- a/src/data/PurserIdentityProvider.ts
+++ b/src/data/PurserIdentityProvider.ts
@@ -1,6 +1,7 @@
 /* eslint-disable class-methods-use-this */
 
 import OrbitDBKeystore from 'orbit-db-keystore';
+import localForage from 'localforage';
 import { IdentityProvider } from './types';
 
 import PurserIdentity from './PurserIdentity';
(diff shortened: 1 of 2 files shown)


chmanie:
Do we want to rebase and eventually also merge this?

Issue #1711, opened 2019-08-13
Remove unused components, sagas, etc


sprusr:
This is just a tidy-up issue for things that we will now have replaced! This should be the last thing to do in stage 1 of DLP/F.

## To remove

- [ ] Organization tab and components within that are not used elsewhere (specifically admin/roles part)
- [ ] Deploying the `OldRoles` extension for new colonies (?)
- [ ] Old set roles saga(s) (?)
- [ ] ???

PR #1772, opened 2019-08-27
Add UserList to admin Permissions tab

### LINKED ISSUE #1193: Move keyTypes to a global location
Move this to a global location, once more parts of the dApp use keyboard interactions

---

https://github.com/JoinColony/colonyDapp/blob/108d4fdbed31bc8a47c8819b0c6865ae222fe8a6/src/modules/dashboard/components/TaskComments/keyTypes.js#L4-L8

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `@todo` comment in 108d4fdbed31bc8a47c8819b0c6865ae222fe8a6 when #1095 was merged. cc @JoinColony.

### LINKED ISSUE #1699: Create DLP permissions list components
### Story

As a user I want to be able to see a list of users with roles for each domain of my colony along with which roles they have, so that I have greater control over who has permissions in my colony.

### Description

It's time for some components to be made! This issue is for the permissions list, which will have a filter dropdown to select the domain, and then display a list of all users who have permissions in that domain, along with what roles it is they have permission for.

This issue doesn't include full wiring, however you should definitely consider what the data will look like that will be provided to these components (or fetched by them).

### Design

[Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=998%3A26740)

image

In cases where the user doesn't have a profile, we use a blockie and just don't show a name or username. The address is just shifted up in their place.

Upon clicking an item in the list, when wired it will open the permissions add/edit dialog.

Users with the root role permission should be shown at the top of the list. The order of the rest of the list doesn't matter.

There should be support for a "pending" state for users whose permissions are in the process of being altered.

- [ ] A permissions list item component has been created according to the design - this should fetch user info (since we already have this available, no need to wait for separate wiring) but the roles info can be a mock for now
- [ ] The colony admin permissions tab contains a filterable list of permission list items! Probably all mock data for now
- [ ] Users with root role permission are shown at the top of the list

### Suggestions

* You might want to use some mock data for these components, but keep the committed mocks to a minimum, and also easy to remove! They won't be around for long 😄 also carefully consider the structure of that data, it should at least very closely match what the real data will be like

### Subtasks

- [ ] ???

### LINKED ISSUE #1708: Wire DLP permissions list
### Story

As a colony admin, I want to see who has what roles in a certain domain of my colony, so that I can better administer that colony.

### Description

So by this point we have the component for the DLP permissions/role list done and we want to fill it with real data. This should just be a matter of connecting the dots that already exist.

If a user's roles for the selected domain are pending, they should be shown with the pending indicator.

### Design

None here!

- [ ] The permissions tab loads with the root domain selected by default
- [ ] A list of users with roles is shown for the selected domain
- [ ] Changing the selected domain causes the data shown to change to that of the selected domain

### Suggestions

* Some wiring should have already been done in the component issue

### Subtasks

- [ ] ???

sprusr:
## Description

This PR adds a filterable (by domain) user list to the permissions tab.

**New stuff** ✨

* Colony roles i18n messages
* Global key types definitions

**Changes** 🏗

* Refactor user list
* Improved typings for table & table cell core components

**Deletions** ⚰️

* Redundant key types definitions

## TODO

- [x] a11y for user list item when `onClick` exists

**Demo** 🎥 

###### Second half of demo shows keyboard interactions - pressing `Enter` calls the callback on the given row (will open the edit dialog once that's completed)

Permissions List

Resolves #1699 
Resolves #1708 
Resolves #1193


JamesLefrere:
@sprusr are you aware `pinion` seems to have been updated here?

ceolson01:
Ah, that looks to be my fault. I'll revert that. Thanks for the heads up 👍

ceolson01:
@karolinkas I've added you as a reviewer here since one of your PRs will be closely tied with the work here

src/modules/admin/components/UserList/UserList.tsx
```diff
@@ -6,7 +6,7 @@ import { pipe, mergePayload, withKey } from '~utils/actions';
 import { useAsyncFunction } from '~utils/hooks';
 import { Table, TableBody } from '~core/Table';
 import Heading from '~core/Heading';
-import UserListItem from './UserListItem';
+import UserListItem from '../UserListItem/UserListItem';
 import styles from './UserList.css';
```
ceolson01: I don't think this component will be needed anymore - it's only being used in the `Organizations` tab which I assume will be replaced by `Permissions`? For increased content flexibility, `UserListItem` is being used directly in `Permissions`.

Thus, I believe `UserList` can be removed via #1711

ceolson01:
@karolinkas there's a little documentation about the roles in the [colonyJS Colony Roles section](https://colony.io/dev/docs/colonyjs/topics-colony-roles).

Regarding the cutoff on the smaller viewport - each row is individually scrollable (if there's overflow from a large number of roles). Do you think there's a better way to achieve this?

src/modules/admin/components/UserListItem/UserListItem.tsx
```diff
@@ -79,8 +91,35 @@ const UserListItem = ({
...
+    : {};
+
   return (
-    <TableRow className={styles.main}>
+    <TableRow
+      className={getMainClasses({}, styles, { hasCallbackFn: !!callbackFn })}
+      {...rowProps}
+    >
       <TableCell className={styles.userAvatar}>
         <UserAvatar size="xs" address={address} user={user} />
```
karolinkas: According to the design in Figma `size="xs"`, but maybe we don't wanna change that since it should be a global decision. Up to you!

PR #1778, opened 2019-08-29
Properly react on message signing errors

### LINKED ISSUE #1718: Pending Action Persists after Cancelling to Sign A Task Comment
## Steps to reproduce

1. Log in to Colony with metamask
2. Comment on a task
3. Confirm the message in the wallet popover
4. Reject signing the message in metamask
5. Try confirming the message again in the popover

## Expected behaviour

Confirming the message again should re-prompt metamask to sign the message.

## Actual behaviour

Clicking `confirm` after first having rejected signing the message in metamask has no effect.
on0Lk2cFSseCLVse8Xxx0w

## Suggestions (optional)

Some way to cancel unconfirmed messages in the wallet popover might be useful.

## Feedback Portal Posts
https://help.colony.io/hc/en-us/community/posts/360033700854-Pending-Action-Persists-after-Cancelling-to-Sign-A-Task-Comment

chmanie:
## Description

Errors yielded in the comment signing process are now properly handled (including MetaMask rejections). The message to sign will be marked as `FAILED` when that happens and both of the forms are reset. The user can then try again to sign the same message using another gas station entry.

See also this [comment](https://github.com/JoinColony/colonyDapp/issues/1718#issuecomment-526381075).

**New stuff**

* Snuck in a little fix for selecting IPFS environments.

**Changes** 🏗

* Extracted `signMessage* function into `core`
* Fix form errors reset for action form
* Use `ActionForm` in the task comment form component
* Properly manage meta ids to be able to use the asycListeners and reset the forms accordingly
* Properly handle the `MESSAGE_CANCEL` action
* Remove form status form TaskComments (as it didn't make a lot of sense since errors are handled in the gas station)

Resolves #1718

### DIFF
--- a/src/modules/core/sagas/messages.ts
+++ b/src/modules/core/sagas/messages.ts
@@ -1,0 +1,55 @@
+import nanoid from 'nanoid';
+import { call, put, take } from 'redux-saga/effects';
+
+import { putError } from '~utils/saga/effects';
+import { ActionTypes } from '~redux/index';
+import { AllActions } from '~redux/types/actions';
+import { Context, getContext } from '~context/index';
+
+export function* signMessage(purpose, message) {
+  const wallet = yield getContext(Context.WALLET);
+  const messageId = `${nanoid(10)}-signMessage`;
+  /*
+   * @NOTE Initiate the message signing process
+   */
+  const messageString = JSON.stringify(message);
+  yield put<AllActions>({
+    type: ActionTypes.MESSAGE_CREATED,
+    payload: {
+      id: messageId,
+      purpose,
+      message: messageString,
+      createdAt: new Date(),
+    },
+  });
+
+  /*
+   * @NOTE Wait (block) until there's a matching action
+   * and get its generated id for the async listener
+   */
+  const {
+    meta: { id },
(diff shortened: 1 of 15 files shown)


src/modules/core/sagas/messages.ts
```diff
@@ -0,0 +1,55 @@
...
+      message: messageString,
+    });
+
+    yield put<AllActions>({
+      type: ActionTypes.MESSAGE_SIGNED,
+      payload: { id: messageId, signature },
+      meta: { id },
+    });
+    return signature;
+  } catch (caughtError) {
```
rdig: This works fine for Metamask and Hardware wallets, which can cancel signing a transaction _"externally"_:

Screenshot from 2019-08-30 10-07-50

But won't work for software wallet, which rely on our Gas Station interface to cancel said message signing

Screenshot from 2019-08-30 10-08-18
Screenshot from 2019-08-30 10-08-26

I think this is a case where we need to `throw` when cancelling a message signing from the gas station

chmanie: Nice catch. I've fixed it and will push soon

Issue #1760, opened 2019-08-26
Re-enable hardware wallets after purser fixes them


auryn-macmillan:
## Steps to reproduce

1. Authenticate to the dapp with ledger hardware wallet with an address that has not been used on the dapp.
2. Attempt claim an ENS username

## Expected behaviour

* You should be able to claim your username.

## Actual behaviour

When claiming Username during Signing in with Hardware wallet Ledger Nano S, it fails when signing the transaction. It is possible to sign the transaction from MEW with the data from the error message.

The JSON in the error messages holds the contract address, data, gas limit and so on. So I was able to use MEW (Myetherwallet.com) for sending the transaction for claiming username.

So it does not seems to be any hardware problem more of an integration bug somewhere.

Web browser: Brave Browser (Version 0.68.131 Chromium: 76.0.3809.100 (Official Build) (64-bit)
OS: Windows 10 (Version 10.0.17134 Build 17134)
Wallet: Ledger Nano S (Hardware firmware: v1.5.5)

 

![image](https://user-images.githubusercontent.com/8453294/63699216-63156080-c7ee-11e9-84f3-8fd3ea00c89c.png)

## Feedback Portal Posts
https://help.colony.io/hc/en-us/community/posts/360034522394-Incompatible-EIP155-based-V1-Getting-Started-ENS-registration-HW-Ledger-Nano-S-?page=1#community_comment_360008613954

rdig:
The fix for this will come from the `purser` side and work is ongoing on this already:

https://github.com/JoinColony/purser/pull/271

ceolson01:
Putting on-hold for now

PR #1781, opened 2019-08-30
Disable hardware wallets (temporarily)

JamesLefrere:
## Description

This PR temporarily disables hardware wallets because #1760 does not have a working solution yet. 

**Note: pending a discussion with product**

**Changes** 🏗

* Add a `disabled` prop to the `DecisionOption` component, so that options can be functionally and visually disabled
* Convert the class component to a functional component with hooks
* Temporarily disable hardware wallets because of colonyDapp#1760, and move them down the list of wallets

### DIFF
--- a/src/modules/core/components/DecisionHub/DecisionOption.css
+++ b/src/modules/core/components/DecisionHub/DecisionOption.css
@@ -19,6 +19,11 @@
   }
 }
 
+.stateDisabled {
+  opacity: 0.5;
+  cursor: not-allowed;
+}
+
 .themeAlt {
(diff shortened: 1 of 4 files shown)


collinvine:
Given this is temporary, and that we are already working on the fix, I am okay with this.

For a user, it may be confusing. Why is it listed be disabled? What do I need to do to re-enable it? Is it a problem on my end?

Can we add a tooltip on hover that says it is only temporarily disabled and full support is coming soon? The copy can be something like, "Hardware wallet support coming soon"

PR #1789, opened 2019-08-30
Firesale fixes for production

### LINKED ISSUE #1777: Task assigning an address without an account, breaks loading the task route
## Steps to reproduce

1. Create a Colony
2. Create a Task
3. Assign as worker to that task, a wallet address that hasn't claimed a profile yet

## Expected behaviour

* It should allow you to add that non-claimed address as a worker, it should work fine to make the payout

## Actual behaviour

* A bunch of components break:

Screenshot from 2019-08-29 14-47-06

rdig:
## Description

This PR was used to fix, then deploy a bunch of fixes to issues that were causing production to break.

**Changes**

- [x] Fix `TaskEventFeed` proper defaults for the user selector _(this was the culprit that was breaking the tasks)_
- [x] `InboxContainer` import proper inbox items record
- [x] `InboxItem` fix relative links _(now that we are using a Popover, the base path is not static / known anymore)_
- [x] `InboxItem` remove hidden overflow to be able to display the `InfoPopover` compoent

Resolves #1777

### DIFF
--- a/src/modules/users/components/Inbox/InboxContainer/InboxContainer.tsx
+++ b/src/modules/users/components/Inbox/InboxContainer/InboxContainer.tsx
@@ -51,7 +51,7 @@ const allReadActions = {
 };
 
 const InboxContainer = ({ full, close }: Props) => {
-  const inboxItems = useSelector(inboxItemsSelector);
+  const { record: inboxItems } = useSelector(inboxItemsSelector);
   const markAllRead = useAsyncFunction(allReadActions);
   return (
     <div
@@ -77,7 +77,7 @@ const InboxContainer = ({ full, close }: Props) => {
           full ? styles.inboxContainerFull : styles.inboxContainerPopover
         }
       >
-        {inboxItems.length === 0 ? (
+        {!inboxItems || (inboxItems && inboxItems.length === 0) ? (
           <div className={styles.loadingText}>
             <FormattedMessage {...MSG.loadingInbox} />
             <DotsLoader />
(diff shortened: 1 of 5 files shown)


src/modules/users/components/Inbox/InboxContainer/InboxContainer.tsx
```diff
@@ -51,7 +51,7 @@ const allReadActions = {
 };
 
 const InboxContainer = ({ full, close }: Props) => {
-  const inboxItems = useSelector(inboxItemsSelector);
+  const { record: inboxItems } = useSelector(inboxItemsSelector);
```
sprusr: Yeah actually here's a case where we don't convert to a JS object I think? Appears to work fine in this component though.

chmanie: I always thought `useSelector` would do the conversion and pick the `record`

sprusr: Oh wow it does! Okay that's really confused me 😵 the react-redux hook is also called `useSelector` (and that doesn't do any extra processing).

chmanie: So why doesn't it do it in this case?

sprusr: No it does, I was thinking `useSelector` was the `react-redux` one.

chmanie: But then why do we need to do it here? Is it a nested `record` for some reason?

sprusr: Oh right, no it doesn't pick `record` just does the `toJS` conversation. Sorry, I was reading/replying too fast!

PR #1797, opened 2019-09-03
Fix description field (for good this time?)

### LINKED ISSUE #1740: Task descriptions not displaying
## Steps to reproduce

1. Create a task with a title and description.
2. Refresh

## Expected behaviour

The text description text should be visible.

## Actual behaviour

image

Note: it is only on tasks that I created that I cannot view the description and I cannot reproduce this on QA.
Here is the task from the screenshot: https://colony.io/colony/speaketh/task/jGagdIUyHUgxxeaqu3WiG

chmanie:
## Description

This PR aims to fix the description field sometimes not rendering properly. We stored the state of the field as an `EditorState` object in formik which caused all sorts of troubles (when comparing it to the data before, etc.). With the change we're storing it as a `string` now when the user blurs out of the field. This should give formik the opportunity to compare the old and new values properly.

**Changes** 🏗

* Store `MultiLineEdit` state as a string in formik
* Convert `MultiLineEdit` into a functional component

**Deletions** ⚰️

* Remove `useInitEditorState` hook

Resolves #1740

### DIFF
--- a/src/modules/core/components/Fields/InlineEdit/MultiLineEdit/MultiLineEdit.tsx
+++ b/src/modules/core/components/Fields/InlineEdit/MultiLineEdit/MultiLineEdit.tsx
@@ -1,6 +1,5 @@
-import React, { ReactNode, Component, FocusEvent } from 'react';
+import React, { ReactNode, useCallback, useState } from 'react';
 import {
-  Editor as EditorType,
   EditorState as EditorStateType,
   ContentState,
   Editor,
@@ -13,6 +12,8 @@ import 'draft-js/dist/Draft.css';
 import { asField, InputLabel } from '~core/Fields';
 import InputStatus from '~core/Fields/InputStatus';
 import { getMainClasses } from '~utils/css';
+import { usePrevious } from '~utils/hooks';
+
 import styles from './MultiLineEdit.css';
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
@@ -65,7 +66,7 @@ interface Props {
   $error?: string;
 
   /** Will be injected by `asField`, or must be supplied if unconnected */
-  $value: EditorStateType;
+  $value: string;
 
   /** @ignore Will be injected by `asField` */
   formatIntl: (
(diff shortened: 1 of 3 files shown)


src/modules/core/components/Fields/InlineEdit/MultiLineEdit/MultiLineEdit.tsx
```diff
@@ -92,112 +90,86 @@ interface Props {
...
+  setValue,
+  status,
+  statusValues,
+  elementOnly = true,
+  readOnly = false,
+  spellCheck = false,
+}: Props) => {
+  const prevValue = usePrevious($value);
+  const initialEditorState =
+    $value && prevValue !== $value
```
chmanie: Even though it seems to work, I just thought about this and I think the condition here is wrong. Why should it be empty when the previous value is the same as the current one?

chmanie:
Would you mind to try again with the revised version?

src/modules/core/components/Fields/InlineEdit/MultiLineEdit/MultiLineEdit.tsx
```diff
@@ -92,112 +89,87 @@ interface Props {
...
+    () => (!allowReturns ? HANDLED : NOT_HANDLED),
+    [allowReturns],
+  );
+  const onBlur = useCallback(
+    (evt: FocusEvent) => {
+      if (onEditorBlur) {
+        const content = editorState.getCurrentContent();
+        const newValue =
+          (content.hasText() && content.getPlainText().trim()) || '';
+        if (newValue !== $value) setValue(newValue);
```
ceolson01: Shouldn't this execute even if `onEditorBlur` is `undefined`?

chmanie: You're right, I'll adjust!

chmanie:
> This doesn't seem to protect against a user entering nothing but spaces

OK, let's see whether I can fix this

src/modules/core/components/Fields/InlineEdit/MultiLineEdit/MultiLineEdit.tsx
```diff
@@ -92,112 +89,87 @@ interface Props {
...
+  const handleReturn = useCallback(
+    () => (!allowReturns ? HANDLED : NOT_HANDLED),
+    [allowReturns],
+  );
+  const onBlur = useCallback(
+    (evt: FocusEvent) => {
+      if (onEditorBlur) {
+        const content = editorState.getCurrentContent();
+        const newValue =
+          (content.hasText() && content.getPlainText().trim()) || '';
```
chmanie: I actually thought the `.trim()` would take care of the spaces issue. I'll look into it more

PR #1803, opened 2019-09-04
Add moveFundsBetweenPots saga

### LINKED ISSUE #1779: Add move funds between domain pots saga
### Story

As a Colony developer, I want to be able to move funds between two domain funding pots, so that I can allow the user to control where funds are allocated.

### Description

We're going to need a way of moving funds between domain pots. This should be as simple as creating a saga which puts a `moveFundsBetweenPots` transaction into the gas station.

The saga should take something like `fromDomainId, toDomainId, amount` as params. Since `moveFundsBetweenPots` wants pot ids (rather than domain ids), we'll first need to call `getDomain` for each of the domain ids to get their pots.

### Design

None here!

- [ ] Saga/actions exist for creating a `moveFundsBetweenPots` transaction

### Suggestions

* Do you know what the difference is between a domain and a pot? Probably not essential for completing this issue, but a domain has a pot. Here's some reading :) https://colony.io/dev/docs/colonynetwork/whitepaper-tldr-pots-and-funding

### Subtasks

- [ ] ???

chmanie:
## Description

This PR adds a `moveFundsBetweenPots` saga which can be called to move funds between two domains. It actually takes two `domainIds` (`fromDomain`, `toDomain`) as parameters as the dapp is only concerned about domains. It will then figure out the corresponding `potIds`.

**New stuff** ✨

* Add `moveFundsBetweenPots` saga
* Create files for domain commands and queries

**Notes**

* The `TRANSACTION_SUCCEEDED` action should be adjusted as needed for the UI to do the right thing.

Resolves #1779


src/modules/dashboard/data/commands/domains.ts
```diff
@@ -0,0 +1,68 @@
+import * as yup from 'yup';
```
sprusr: Nice refactor into here 👌

src/modules/dashboard/sagas/domains.ts
```diff
@@ -143,8 +140,50 @@ function* domainEdit({
...
 }
 
+function* moveFundsBetweenPots({
+  payload: { colonyAddress, fromDomain, toDomain, amount, tokenAddress },
+  meta,
+}: Action<ActionTypes.MOVE_FUNDS_BETWEEN_POTS>) {
+  let txChannel;
+  try {
+    txChannel = yield call(getTxChannel, meta.id);
+    const [fromPot, toPot] = yield all([
```
sprusr: These are actually the domains though right? It's an object like `{ potId, skillId }`

chmanie: Yes! Thank you, I'll adjust!

src/modules/dashboard/sagas/domains.ts
```diff
@@ -143,8 +140,50 @@ function* domainEdit({
...
 }
 
+function* moveFundsBetweenPots({
+  payload: { colonyAddress, fromDomain, toDomain, amount, tokenAddress },
+  meta,
+}: Action<ActionTypes.MOVE_FUNDS_BETWEEN_POTS>) {
+  let txChannel;
+  try {
+    txChannel = yield call(getTxChannel, meta.id);
+    const [{ potId: fromPot }, { potId: toPot }] = yield all([
```
chmanie: @sprusr is this better?

chmanie:
Let's add these in the UI issue for that!

PR #1804, opened 2019-09-04
Enable `strictNullCheck` compiler option

chmanie:
## Description

We found that we want to enable the `strictNullCheck` compiler option as it turned out to be quite helpful. This required a lot of adjustments in the types but during that I also fixed a lot (!) of errors. Furthermore this adds the `@types/yup` types which will eventually spare us some type / schema duplication.

You could also see this as a cleanup PR supporting the TS move.

**New stuff** ✨

* Add `@types/yup`

**Changes** 🏗

* Lots of fixes and improvements

- [x] ~Infer yup types for commands~ (we'll do another PR for this)


src/immutable/FetchableData.ts
```diff
@@ -0,0 +1,27 @@
...
+import { DefaultValues } from '~types/index';
+
+interface Shared<R> {
+  record: R;
+  error?: string | void;
+  isFetching?: boolean;
+  lastFetchedAt?: Date;
+}
+
+export type FetchableDataType<R> = $ReadOnly<Shared<R>>;
```
JamesLefrere: FYI there is a `Readonly` built-in TS type, so we could replace `$ReadOnly` if we wanted to 🤷‍♂

src/immutable/FetchableData.ts
```diff
@@ -0,0 +1,27 @@
...
+  error: undefined,
+  isFetching: false,
+  lastFetchedAt: new Date(0),
+};
+
+export class FetchableDataRecord<R> extends Record<Shared<any>>(
+  defaultValues,
+) {}
+
+export const FetchableData = <R>(p?: Shared<R>) =>
```
JamesLefrere: I approve of this rename!

src/immutable/state/index.ts
```diff
@@ -16,11 +16,11 @@ export * from './dashboard';
...
-  admin: AdminStateRecord;
-  core: CoreStateRecord;
-  dashboard: DashboardStateRecord;
-  router: RouterState;
-  users: UsersStateRecord;
+  admin?: AdminStateRecord;
+  core?: CoreStateRecord;
+  dashboard?: DashboardStateRecord;
+  router?: RouterState;
+  users?: UsersStateRecord;
```
JamesLefrere: I'm surprised that these need to be optional. Why is that?

src/modules/core/components/TransactionList/TransactionDetails.tsx
```diff
@@ -268,7 +272,7 @@ const TransactionDetails = ({
     : OutgoingTransaction;
   return (
     <TransactionComponent
-      colony={colony}
+      colony={colony as ColonyType}
```
JamesLefrere: For this kind of case, perhaps it's better to use:

```typescript
colony={colony!}
```

Because that way we don't have to specify the type again, only assert that it's defined.

src/modules/dashboard/components/CreateColonyWizard/CreateColonyWizard.ts
```diff
@@ -78,6 +77,6 @@ const CreateColonyContainer = compose(
   withWizard({
     steps: stepFunction,
   }),
-)(CreateColony);
+)(WizardTemplate);
```
JamesLefrere: Right, because the unnecessary component was removed 👍

src/redux/types/actions/colony.ts
```diff
@@ -226,13 +225,27 @@ export type ColonyActionTypes =
   | UniqueActionType<ActionTypes.COLONY_MINT_TOKENS_SUBMITTED, object, object>
   | UniqueActionType<
       ActionTypes.COLONY_PROFILE_UPDATE,
-      $Shape<ColonyType>,
+      {
```
JamesLefrere: I think you could also use:

```typescript
Pick<ColonyType, 'colonyAddress' | 'colonyName' | 'description' | 'displayName' | 'guideline' | 'website'>,
```

src/types/DefaultValues.ts
```diff
@@ -0,0 +1,4 @@
+/*
+ Pick all the values from an object and make them `any` (handy for immutable default props)
+ */
+export type DefaultValues<V> = Record<keyof V, any>;
```
JamesLefrere: Practical solution! 👍

src/utils/web3/eventLogs/events.ts
```diff
@@ -82,14 +82,49 @@ export const decorateLog = async (
...
+  };
+  timestamp: number;
+  transaction: EventTransaction;
+}
+
 /*
  * Get logs using a logFilter and decorate them with a transaction, a timestamp and parsed event data
  */
-export const getDecoratedEvents = async (
+export const getDecoratedEvents = async <T, P>(
```
JamesLefrere: Good job with the generics 👍

src/modules/core/components/TransactionList/TransactionDetails.tsx
JamesLefrere: That said, it seems like there's an eslint rule we're using that would disallow this.

JamesLefrere:
That branch name, huh? 😅

src/modules/core/components/TransactionList/TransactionDetails.tsx
chmanie: This is actually an issue in the component itself. I fixed it by using a guard

src/immutable/state/index.ts
chmanie: Because they didn't have a default value. Now they do!

src/immutable/FetchableData.ts
chmanie: I replaced all of them!

chmanie:
This should be good to review now. Quite some work, basically half a deep-dive into TS, half Fleißarbeit.

src/immutable/state/core/index.ts
```diff
@@ -1,32 +1,35 @@
...
   ipfsData: IpfsDataType;
   messages: CoreMessagesRecord;
-  connection: ConnectionRecordType;
+  network: FetchableDataRecord<NetworkRecord>;
+  transactions: CoreTransactionsRecord;
 };
 
-export type CoreStateRecord = CollectionType<any, any> &
-  RecordOf<CoreStateProps>;
+export class CoreStateRecord extends Record<CoreStateProps>({
```
JamesLefrere: I thought `combineReducers` would set these default values, but this is fine too.

chmanie:
I think it’s borked currently, so I need to get back in and fix it

src/modules/dashboard/reducers/coloniesReducer.ts
```diff
@@ -124,14 +125,14 @@ const coloniesReducer: ReducerType<AllColoniesMap> = (
...
+        : state.set(colonyAddress, FetchableData<ColonyRecord>({ record }));
     }
     default:
       return state;
   }
 };
 
-export default withDataRecordMap<AllColoniesMap, ColonyRecordType>(
+export default withFetchableDataMap<AllColoniesMap, ColonyRecord>(
   // @ts-ignore
```
JamesLefrere: Maybe this `@ts-ignore` can be removed, but no biggie if not

src/styleguide/Wrapper.tsx
```diff
@@ -31,6 +28,7 @@ const MockState = Record({
   users: undefined,
 });
 
+// @ts-ignore
```
JamesLefrere: Styleguide, ist mir egal

src/types/roles.ts
```diff
@@ -0,0 +1,5 @@
+export type UserRolesObject = { [role: string]: boolean };
```
JamesLefrere: You can do `Record<string, boolean>` but of course you don't get the label that way.

src/utils/hooks/index.ts
```diff
@@ -233,7 +235,10 @@ export const useDataMapFetcher = <T>(
   const memoizedKeys = useMemoWithFlatArray(() => keys, keys);
 
   const dispatch = useDispatch();
-  const allData: ImmutableMapType<string, DataRecordType<any>> = useMappedState(
+  const allData: ImmutableMapType<
```
JamesLefrere: You _could_ if you wanted to rename all of these instances to `ImmutableMap`

src/immutable/Colony.ts
```diff
@@ -40,13 +31,13 @@ export type ColonyType = $ReadOnly<
...
-  tokens?: ImmutableMapType<Address, TokenReferenceRecordType>;
+  tokens?: ImmutableMap<Address, TokenReferenceRecord>;
 };
 
-const defaultValues: ColonyRecordProps = {
+const defaultValues: DefaultValues<ColonyRecordProps> = {
   avatarHash: undefined,
-  canMintNativeToken: undefined,
-  canUnlockNativeToken: undefined,
+  canMintNativeToken: false,
```
ceolson01: This change seems to prevent me from being able to mint new tokens in a colony (in `TabContribute.tsx`)

src/modules/dashboard/reducers/allTokens.ts
```diff
@@ -1,19 +1,15 @@
...
-import { withDataRecordMap } from '~utils/reducers';
+import { withFetchableDataMap } from '~utils/reducers';
+
+import { AllTokensMap } from '../state/index';
 
 const INITIAL_STATE: AllTokensMap = ImmutableMap({
-  [ZERO_ADDRESS]: DataRecord<TokenRecordType>({
-    record: TokenRecord(
+  [ZERO_ADDRESS]: FetchableData<TokenRecord>({
+    record: Token(
```
ceolson01: Something seems to be askew with eth tokens... 🤷‍♂ 

Here's the redux state:

```
    allTokens: {
      '0x0000000000000000000000000000000000000000': {
        isFetching: false,
        lastFetchedAt: '2019-09-18T14:31:00.404Z'
      },
      '0xcBA6e1D695fe0A4078A884902D3B5532c60Ad18E': {
        record: {
          address: '0xcBA6e1D695fe0A4078A884902D3B5532c60Ad18E',
          decimals: 18,
          isVerified: false,
          name: 'Cool Token',
          symbol: 'COOL'
        },
        isFetching: false,
        lastFetchedAt: '2019-09-18T14:30:52.735Z'
      }
    },
```

JamesLefrere:
Thanks Curtis, these are the exact things we found :)

chmanie:
OK, you two, I'm quite confident I've addressed all the issues mentioned above. Let's hope for the best 🤞

PR #1810, opened 2019-09-05
Prevent names with periods (dots) when registering a ENS domain

### LINKED ISSUE #1798: Periods in ENS domain cause 404
## Steps to reproduce

1. Create a colony with a period in the ENS domain
SharedScreenshot

## Expected behaviour

Colony should be created as normal

## Actual behaviour

The colony becomes inaccessible.
Moreover, the domain registered on the contract appears to be omit everything after and including the period.
`centrifuge.test` was created, but it shows up as `centrifuge` in the developer dashboard.
image
`0x0a1a7E2F398d2C394411d2839F896B9D5fF7C6d7`

rdig:
## Description

This PR fixes the Colony Wizard and the User wizard by preventing the validation of usernames that contain a dot `.` in them.

While at it, I also simplified the ENS domain regex to be more read-able:
- `^` start matching
- `[A-Za-z0-9]` allow upper  case, lower case, numerals
- `[^.]` negate to not allow dots / periods
- `{1,255}` match at least `1` and at most `255` chars
- `$` end matching

**Changes** 

- [x] `ensAddress` yup validator regex doesn't allow dots `.` any more

**Screenshots**

Screenshot from 2019-09-05 17-18-31

Screenshot from 2019-09-05 17-18-39

Resolves #1798

### DIFF
--- a/src/lib/TextDecorator/schemas.ts
+++ b/src/lib/TextDecorator/schemas.ts
@@ -1,8 +1,7 @@
 /* eslint import/prefer-default-export: 0 */
 
-import { ENS_DOMAIN_REGEX } from '../../modules/validations';
-
 const USER_PREFIX = '@';
+const ENS_DOMAIN_REGEX = '^([A-Za-z0-9](?:[A-Za-z0-9-.]{0,255}[A-Za-z0-9])?)';
 
 export const USERNAME_SCHEMA = {
   prefix: USER_PREFIX,
--- a/src/modules/validations.ts
+++ b/src/modules/validations.ts
@@ -14,8 +14,7 @@ import en from '../i18n/en-validation.json';
 yup.setLocale(en);
 
 // eslint-disable-next-line import/prefer-default-export
-export const ENS_DOMAIN_REGEX =
-  '^([A-Za-z0-9](?:[A-Za-z0-9-.]{0,255}[A-Za-z0-9])?)';
+export const ENS_DOMAIN_REGEX = '^[A-Za-z0-9][^.]{1,255}$';
 
 /* Custom validators */
 function equalTo(ref, msg) {
--- a/src/modules/validations.ts
+++ b/src/modules/validations.ts
@@ -13,6 +13,14 @@ import en from '../i18n/en-validation.json';
 
 yup.setLocale(en);
 
+/*
+ * The ens domain regex is composed of
+ * ^ start match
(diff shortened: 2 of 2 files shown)


src/modules/validations.ts
```diff
@@ -14,8 +14,7 @@ import en from '../i18n/en-validation.json';
 yup.setLocale(en);
 
 // eslint-disable-next-line import/prefer-default-export
-export const ENS_DOMAIN_REGEX =
-  '^([A-Za-z0-9](?:[A-Za-z0-9-.]{0,255}[A-Za-z0-9])?)';
+export const ENS_DOMAIN_REGEX = '^[A-Za-z0-9][^.]{1,255}$';
```
rdig: Arg! Apparently this is also used in `TextDecorator`, that's why this was set up with a capture group.

I'll fix it...

rdig: So the only way I managed to properly solve this, is to have separate regexes for the ENS yup validator and the `TextDecorator` schema.

This is because the text decorator needs to match multiple ones, while the validator only matches the first.

src/lib/TextDecorator/schemas.ts
```diff
@@ -1,8 +1,7 @@
 /* eslint import/prefer-default-export: 0 */
 
-import { ENS_DOMAIN_REGEX } from '../../modules/validations';
-
 const USER_PREFIX = '@';
+const ENS_DOMAIN_REGEX = '^([A-Za-z0-9](?:[A-Za-z0-9-.]{0,255}[A-Za-z0-9])?)';
```
sprusr: Could this go in the `validations.ts` file also, but with a different name? Probably also with comments about what the difference is between them - I am confused what this one is for! :)

src/modules/validations.ts
sprusr: It would be really nice to have a comment just explaining why this regex is the way it is. Of course I am though perfectly literate in reading regex 😗 (this is a lie)

src/lib/TextDecorator/schemas.ts
rdig: I really can't tell what this is all supposed to do, especially with the capture groups and the negated inner capture group.

As far as I can tell it needs to validate multiple occurrences of a ens name in a string _(as opposed to my version, which just validates the first occurrence)_

As far as I remember @chmanie wrote this, so maybe he has more details on it ?

PR #1818, opened 2019-09-08
Add colony admin "add domain" hint

### LINKED ISSUE #1814: Add hint to colony admin add domains
### Story

As a user, I want to be aware of the fact that domains cannot be removed, so that I don't accidentally create them.

### Description

We're adding a small hint to the right of the label for the "Add New Domain" input on the colony admin domains tab. It should read "This cannot be undone".

### Design

[Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=1578%3A26560)

image

- [ ] This hint is added to the right of the "Add New Domain" label

### Suggestions

* ???

### Subtasks

- [ ] ???

ceolson01:
## Description

This PR adds some hint text to the "Add domain" input in the colony admin.

**New stuff** ✨

* `helpAlign` appearance prop for `Input`

**Changes** 🏗

* Update unit test `.expect`
* Increase size of `Input` and `Button` to match figma

Resolves #1814

### DIFF
--- a/src/modules/admin/components/Organizations/OrganizationAddDomains.tsx
+++ b/src/modules/admin/components/Organizations/OrganizationAddDomains.tsx
@@ -18,6 +18,10 @@ const MSG = defineMessages({
     id: 'admin.Organizations.OrganizationAddDomains.labelAddDomain',
     defaultMessage: 'Add New Domain',
   },
+  helpText: {
+    id: 'admin.Organizations.OrganizationAddDomains.helpText',
+    defaultMessage: 'This cannot be undone',
+  },
   placeholderAddAdmins: {
     id: 'admin.Organizations.OrganizationAddDomains.placeholderAddAdmins',
     defaultMessage: 'Search for a user or paste a wallet address',
@@ -60,11 +64,16 @@ const OrganizationAddDomains = ({ colonyAddress }: Props) => {
         {({ status, isSubmitting, isValid }) => (
           <div className={styles.inputWrapper}>
             <div className={styles.domainInput}>
-              <Input name="domainName" label={MSG.labelAddDomain} />
+              <Input
+                appearance={{ helpAlign: 'right', theme: 'fat' }}
+                help={MSG.helpText}
+                label={MSG.labelAddDomain}
+                name="domainName"
+              />
             </div>
             <div className={styles.submitButton}>
               <Button
-                appearance={{ theme: 'primary', size: 'medium' }}
+                appearance={{ theme: 'primary', size: 'large' }}
                 style={{ width: styles.wideButton }}
                 text={MSG.buttonAddDomain}
                 type="submit"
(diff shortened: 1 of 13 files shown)


src/modules/admin/components/Organizations/OrganizationAddDomains.tsx
```diff
@@ -60,11 +64,16 @@ const OrganizationAddDomains = ({ colonyAddress }: Props) => {
...
+                appearance={{ helpAlign: 'right', theme: 'fat' }}
+                help={MSG.helpText}
+                label={MSG.labelAddDomain}
+                name="domainName"
+              />
             </div>
             <div className={styles.submitButton}>
               <Button
-                appearance={{ theme: 'primary', size: 'medium' }}
+                appearance={{ theme: 'primary', size: 'large' }}
```
sprusr: Nice one fixing this too 👍

src/modules/core/components/Fields/InputLabel/InputLabel.tsx
```diff
@@ -73,7 +75,13 @@ const InputLabel = ({
       htmlFor={inputId || null}
     >
       <span className={styles.labelText}>{labelText}</span>
-      {helpText && <span className={styles.help}>({helpText})</span>}
+      {helpText && (
+        <span className={styles.help}>
+          <span className={styles.paren}>(</span>
```
sprusr: What's up with these hidden parens? Do we need them?

ceolson01: My thought was, to provide context (that this **is in fact** help text) to screen reader users, we should keep the parens in the DOM, while visually hiding them for sighted users (since the right alignment provides the context for this group of users). The parens will still be displayed if the help text is left aligned.

☝️ that said, I'm realizing that screen readers won't read the parens with `display: none`, so I'll need to adjust them to be hidden via `text-indent` or another method.

ceolson01: I made all the parens parts of the messages, which I think is a bit safer for purposes of i18n.

ceolson01:
@sprusr this is ready for another look.

I've also rebased with the updates from [`feature/dlp`](https://github.com/JoinColony/colonyDapp/tree/feature/dlp).

src/modules/core/components/Fields/InputLabel/InputLabel.tsx
```diff
@@ -73,7 +75,7 @@ const InputLabel = ({
       htmlFor={inputId || null}
     >
       <span className={styles.labelText}>{labelText}</span>
-      {helpText && <span className={styles.help}>({helpText})</span>}
+      {helpText && <span className={styles.help}>{helpText}</span>}
```
sprusr: Nice, yeah I think this is a better approach 👌

PR #1819, opened 2019-09-09
TransactionDetails defaults for not-yet-loaded data

### LINKED ISSUE #1812: The transactions tab crashes the dapp
## Steps to reproduce

1. navigate to colony.io/colony/{colonyName}/admin
2. Select the transactions tab

## Expected behaviour

The transaction tab should open

## Actual behaviour

the dapp crashes

rdig:
## Description

This PR is a simple fix for instances where the Transactions Lists wants to render, but the data for either a user, a colony or a task hasn't loaded yet from the store.

To fix it, we're just defaulting the destructuring assignment to it's respective immutable record:
- `UserRecord`
- `ColonyRecord`
- `TaskRecord`

**Changes**

- [x] `TransactionDetails` default colony, user, task to they're respective immutable records

Resolves #1812

### DIFF
--- a/src/modules/core/components/TransactionList/TransactionDetails.tsx
+++ b/src/modules/core/components/TransactionList/TransactionDetails.tsx
@@ -1,17 +1,21 @@
 import React from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
-import { useDataSubscriber } from '~utils/hooks';
-import { colonySubscriber } from '../../../dashboard/subscribers';
-import MaskedAddress from '~core/MaskedAddress';
-import Link from '~core/Link';
-import { Address, ENSName } from '~types/index';
 import {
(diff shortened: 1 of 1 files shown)


src/modules/core/components/TransactionList/TransactionDetails.tsx
```diff
@@ -68,7 +76,10 @@ const UserDetails = ({
 );
 
 const ColonyDetails = ({
-  colony: { displayName: colonyDisplayName, colonyAddress },
+  colony: {
+    displayName: colonyDisplayName,
+    colonyAddress,
+  } = ColonyRecord().toJS(),
```
rdig: If I don't call `.toJS()` on the record, TS complains about not being able to destructure it, but the code itself works even without this call.

Issue #1783, opened 2019-08-30
Add domain funding to colony home


sprusr:
### Story

As a user with funding role in a domain (but not admin in the colony), I want to be able to use the "move funds" dialog, so that I can control the funding for domains in which I have the funding role.

### Description

We've already created a "move funds" dialog in #1780, now we want to add a button to open it from the colony home page, as well as a section displaying the currently allocated funds for the selected domain.

The "Fund" button to open the dialog should be shown if the current user has the funding role in any domain of the colony, otherwise it's hidden. The current domain funds should show the amount for any tokens which have been selected/enabled in the colony admin tokens screen.

If "All" domains is selected, hide the whole available funds section and "Fund" button.

🌮 If this functionality was done in #1780 (or you fancy adding it now!), then the dialog should open with the currently selected domain already selected as the "to" domain. If the current user doesn't have permission, then default back to nothing selected.

### Design

[Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=998%3A28357)

![image](https://user-images.githubusercontent.com/7497084/64027563-3be9c680-cb41-11e9-911c-1a8d2cc18777.png)

![image](https://user-images.githubusercontent.com/7497084/64027832-d518dd00-cb41-11e9-8dd4-8659259d5d93.png)

### This issue is complete when...

- [ ] "Available Funds" section added to colony home sidebar
- [ ] "Fund" button shown as part of that section, but only if the current user has funding role in at least one domain of the colony - this should open the dialog

### Suggestions

* ???

### Subtasks

- [ ] ???

PR #1823, opened 2019-09-11
Update home screen

### LINKED ISSUE #1785: Colony home screen updates
### Story

As a user viewing a colony, I want a clearer layout, so that I can more easily understand the current state of the colony.

(This is totally a fluff story, sorry)

### Description

There's a whole load of movings around and design changes to be made to the colony home! They can be summed up as:

* `ColonyMeta` has a new layout, and is shifted upwards to overlap with the nav bar
* The domains list has been moved to right below the new colony meta on the left
* The "Contribute" tab is renamed to "Tasks", with the tasks filter moved to within the tab itself (below tabs), and the "New task" button is also moved to within the tab, on the right of the filter
* A currently selected domain breadcrumb is added to the top of the main section of the page (above tabs)

### Design

[Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=1013%3A43)

image

Description is limited to 88 characters, clicking on "More" will expand to full length.

image

- [x] New `ColonyMeta` layout has been created
- [x] Domains list is on the left of the page
- [x] Tabs component has been shuffled about
- [x] Domain breadcrumb has been added to top of page

### Suggestions

* ???

### Subtasks

- [x] Expandable paragraph component needs creating

karolinkas:
## Description

This implements the redesign of the ColonyHome page in combination with the addition of two new simple core components: an expandable paragraph and a breadcrumb component. 

- [x] New ColonyMeta layout has been created
- [x] Domains list is on the left of the page
- [x] Tabs component has been shuffled about
- [x] Domain breadcrumb has been added to top of page
- [x] extract ExtendParagrpah component
- [x]  added available funds in sidebar which gets wired in #1783

## TODO
- [x] edit .md files for styleguide
- [x] move more button a tiny bit
- [x] extract BreadCrumb completely

## New UI Screenshot
Screen Shot 2019-09-12 at 15 23 19

## ExpandableParagraph component in action
expandableParagraph

## BreadCrumb component in action
breadCrumb

Resolves #1785


src/modules/dashboard/components/ColonyHome/ColonyHome.css
```diff
@@ -3,19 +3,37 @@
 .main {
   composes: stretchVertical stretchHorizontal from '~styles/layout.css';
   display: grid;
-  padding: 130px 80px; /* vertical padding of 60px + nav height (70px) */
+  padding: 16px 18px; /* vertical padding of 60px + nav height (70px) */
```
rdig: The comment values should be changed as well, right ?

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -220,21 +240,73 @@ const ColonyHome = ({
...
+  const crumbs = (domains || [])
+    .sort((a, b) => a.id - b.id)
+    .reduce(
+      (accumulator, element) => {
+        if (element.id <= filteredDomainId) {
+          accumulator.push(element.name);
+        }
+        return accumulator;
+      },
+      ['Root'],
```
rdig: There's two things here:
- This should be a message descriptor _(But this might be a little hard, please look into it)_
- As everywhere else, the root domain should be lowercase, ie: `root`

src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonySubscribe.css
```diff
@@ -1,13 +1,13 @@
...
   padding: 0;
-  height: 26px;
-  width: 26px;
+  height: 15px;
+  width: 15px;
   position: relative;
-  top: -10px;
-  left: -26px;
+  top: 19px;
+  left: -11px;
```
rdig: This should have a comment as to why it's a negative value

src/modules/dashboard/components/ColonyHome/ColonyDomains/ColonyDomains.tsx
```diff
@@ -35,6 +36,7 @@ const ColonyDomains = ({
   colonyAddress,
   setFilteredDomainId,
   filteredDomainId,
+  noTitle,
```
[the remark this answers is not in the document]
karolinkas: Can do, so far there's only one occurrence, so it's hard to say what's better.

karolinkas:
@rdig I applied all changes you mentioned above, please check one more time.

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -220,21 +240,79 @@ const ColonyHome = ({
...
-          canAdminister={!isInRecoveryMode && canAdminister(permissions)}
-        />
+        <div className={styles.metaContainer}>
+          <ColonyMeta
+            colony={colony}
+            canAdminister={!isInRecoveryMode && canAdminister(permissions)}
+          />
+        </div>
+        <div className={styles.domainContainer}>
+          <ColonyDomains
```
rdig: I just noticed this:

`ColonyDomains` need to be refactored to be part of the `ColonyMeta` component, not external to it

src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonyMeta.css
```diff
@@ -13,10 +13,15 @@
 
 .description {
   margin-top: 15px;
+  margin-left: 52px;
+  padding-bottom: 15px;
+  min-height: 75px;
+  border-bottom: 1px solid var(--temp-grey-5);
 }
 
 .dynamicSection {
```
rdig: The `.dynamicSection` class is not used in `ColonyMeta`

Please clean up the styles file

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
karolinkas: @rdig Could you tell me a reason or an argument to cause more work?

karolinkas: Also I just realised a little icon I missed in the DomainsList I assume, it's to collapse the list, so it will have it's own state, saying that I would say it's better to be in it's own component. 
Screen Shot 2019-09-19 at 14 18 02

rdig: > @rdig Could you tell me a reason or an argument that justifies this, because it is causing more work.

Because both from a logical point of view, and from a code point of view they're part of the meta information regarding a colony.

From a code point of view, after they're moved into the sidebar component (aka: Colony Meta), you actually simply a bunch of styles, as they can now be applied globally (you won't need to individually set margins and padding on sections)

> because it is causing more work.

If you don't want to do it here, I'll do it in my downstream branch #1839, but either way this is happening

rdig: > Also I just realised a little icon I missed in the DomainsList I assume, it's to collapse the list, so it will have it's own state, saying that I would say it's better to be in it's own component.

I'm not 100% sure that is to collapse / show the list. It might be there to add new domains strait from the new Colony Home.

Better ask @JoinColony/product to make sure

Karol-colony: please ignore this icon

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -220,21 +240,79 @@ const ColonyHome = ({
...
+        />
+      )}
+    </>
+  );
+
+  const crumbs = (domains || [])
+    .sort((a, b) => a.id - b.id)
+    .reduce(
+      (accumulator, element) => {
+        if (element.id <= filteredDomainId && element.name) {
```
rdig: I think this is breaking the production builds, based on what Circle is telling us: https://app.circleci.com/jobs/github/JoinColony/colonyDapp/20579

Please check the production build manually locally

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
karolinkas: > If you don't want to do it here, I'll do it in my downstream branch #1839, but either way this is happening

I'll do it just wanted to understand the reasoning behind.

karolinkas:
Ok, so I have double-triple check everything and weirdly enough I couldn't reproduce some things in Chrome:
- the empty states of the taskList are showing well, independent of the filter option
Screen Shot 2019-09-19 at 17 25 42

- the dividers are perfectly well aligned for me
Screen Shot 2019-09-19 at 17 26 18

Everything else is adjusted and ready!

rdig:
> the dividers are perfectly well aligned for me

@ceolson01 if you some time, can you please check how the separator alignment is rendering for you ?

Me and Karolin get very different results.

ceolson01:
Firstly, let me say this is looking awesome @karolinkas and @JoinColony/product 👏 

### Alignment

I was able to reproduce the divider alignment issue with the viewport width at ~1125px and below, but it *is* aligned at greater widths. But the alignment does seem fragile / coincidental. I think we need to play with the `grid-row` property a bit, or maybe even adjust how these elements are grouped.

Here's a demo:

Divider Alignment

### More / Hide Link

I also found an interesting little bug 🐛  - the `More` / `Hide` link seems to jump around a little bit for me.

**Short text demo:**

More Jump

**Long text demo:**

More Jump Long

rdig:
@karolinkas please have a look / fix the issues found by Curtis, then I'll re-review this.

karolinkas:
@rdig I could reproduce the button jumping bug and fixed it but not the alignment issue with small screen sizes, but made some adjustments that should fix it. Please check if it's fixed for you too.

karolinkas:
@rdig Ok, I simplified ColonyHome the annoying by removing that complicated MessageDescriptor creation and adjusted the outstanding UI issues. This should be ready to merge now.

Karol-colony:
"Currently, it displays all domains as a sub-domain of the previous one, which clearly is wrong as they are all top-level domains (under root, of course)" right, as you said. 

We don't have intention to display nested domains at the current stage. We are not ready with implications of this one. I guess, there wouldn't be so many intends, only: core domain / nested domain; It would be good if also someone else from @JoinColony/product could confirm that.

karolinkas:
@rdig I get your point of not being able to display nesting with the current breadcrumb component, but also currently it's not possible to change the way they are being displayed since  all share the same parentId. Maybe it goes out of the scope of this PR?

rdig:
> since all share the same parentId

But if they all have the same parent id, shouldn't they all have the same 1-level deep nesting ? That's what I'm asking for.

karolinkas:
@rdig Yes, completely, but it seems impossible to visualize that with a breadCrumb component or do you have a specific idea?

rdig:
All domains that you select should basically render like the first one _("Public Ballot" in this case)_

Screenshot from 2019-09-26 15-01-09

So no matter what you select, since they're all under root should show:
```
Root > Domain Name
```

itspatch:
I think it is fine if we implement this the way Raul has suggested and see how it goes. It is the most "correct" to have Root > Domain Name. 

Originally, it was designed so that it would just say All Domains > Domain Name. It didn't matter if that Domain Name was Root or any other domain. So the breadcrumb would be All Domains > Root or All Domains > Public Ballot. This isn't technically correct because Public Ballot is not a subdomain of All Domains, but with this breadcrumb would allow the user to get back to where they came from (All Domains). 

However, I like educating the user about where they are in the domain hierarchy, so I am ok with going with Root > Domain Name.

karolinkas:
@rdig Thanks for explaining, that was actually easy to implement. I commited it.

PR #1825, opened 2019-09-12
Reset task comment field on success

### LINKED ISSUE #1822: Comment field is not being reset when adding a comment
I'll edit this later.

## Steps to reproduce

1. Example: log in with a MetaMask wallet

## Expected behaviour

* Example: The user is logged in without errors

## Actual behaviour

* Example: A `WALLET_CREATE_ERROR` action is dispatched and the user cannot log in

## Suggestions (optional)

* Example: A new version of MetaMask was released, maybe there are some breaking changes

## Screenshot

image

JamesLefrere:
## Description

This PR uses the `onSuccess` prop to reset the comment form. It's not clear how this behaviour was originally working, because other fields (e.g. task title, task description) operate on the basis of the form not being reset on success, so that the value is always visible.

**Changes** 🏗

* Reset the task comment field on success with the `onSuccess` prop
* Export the `OnSuccess` and `OnError` types from `ActionForm`

Resolves #1822

### DIFF
--- a/src/modules/core/components/Fields/Form/ActionForm.tsx
+++ b/src/modules/core/components/Fields/Form/ActionForm.tsx
@@ -19,8 +19,16 @@ const MSG = defineMessages({
 
 const displayName = 'Form.ActionForm';
 
-type OnError = (error: any, bag: FormikBag<any, any>, values?: any) => void;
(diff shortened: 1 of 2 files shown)


rdig:
This one can be merged as well :+1:

PR #1827, opened 2019-09-12
Automatically build and push dapp images on successful builds

area:
Builds on all branches except master generate a build of a docker image targeted at goerli, and push it to GCR.

Builds on master generate a build of a docker image targeted at mainnet and push it to GCR.

Unfortunately, this build seems a bit spotty ([green](https://circleci.com/gh/JoinColony/colonyDapp/20232), [red](https://circleci.com/gh/JoinColony/colonyDapp/20231) on the same commit hash), and I'm not sure  why...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,7 +220,7 @@ jobs:
       - run: sudo apt-get update && sudo apt-get install google-cloud-sdk
       - run: bash ./scripts/authGoogle.sh
       - run: gcloud docker -- push eu.gcr.io/${PROJECT_NAME}/dapp-goerli:$CIRCLE_SHA1
-  build-deploy-mainnet-image:
+  build-mainnet-image:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -236,6 +286,23 @@ workflows:
...
               only: master
+      - build-goerli-image:
+          requires:
+            - build-with-lockfile
+            - integration-testing
+            - e2e-testing
+          filters:
+            branches:
+              ignore: master
+      - build-deploy-mainnet-image:
```
rdig: Do we still need to deploy the mainnet image ?

Given that we now have !staging and !production in addition to !qa

area: We still need to build it and push it (which is all this does now - it just didn't get renamed when I removed the deploy).

PR #1828, opened 2019-09-13
Remove old admin things

karolinkas:
## Description

A bunch of things can be removed after implementing domain level permissions. 
- [x] Organization tab and components within that are not used elsewhere (specifically admin/roles part)
 - [x] Deploying the OldRoles extension for new colonies (?)
 - [x] Old set roles saga(s) (?)

I tested colony creation to make sure I didn't delete the wrong thing to make sure it still works and it seems all fine. 

What about the `useOldRoles` does it have to updates in some way since it proxies old to new roles like you described in the inline comment?


src/modules/admin/components/Organizations/OrganizationAddDomains.tsx
```diff
@@ -18,8 +18,8 @@ const MSG = defineMessages({
     id: 'admin.Organizations.OrganizationAddDomains.labelAddDomain',
     defaultMessage: 'Add New Domain',
   },
-  placeholderAddAdmins: {
-    id: 'admin.Organizations.OrganizationAddDomains.placeholderAddAdmins',
+  placeholderAddDomain: {
```
sprusr: Nice catch 👍 I think also we should probably move this component into the `Domains` dir, but can be left like this for now.

src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -194,21 +190,6 @@ function* colonyCreate({
       });
     }
 
-    yield createGroupedTransaction(deployOldRoles, {
```
sprusr: Have you tested creating a colony after removing these? I think it will be broken now, since `setOneTxRole` uses `setAdminRole` which is dependent on `OldRoles` being deployed for the colony.

We'll need to replace that one transaction with two different ones: `setAdministrationRole` and `setFundingRole`, both with `domainId: 1`. Not ideal that we will have multiple transactions for this, but we're still overall reducing the number of transactions.

karolinkas: Yes, as I wrote in the PR description I tested after making changes in the PR and it worked well, but maybe it would lead to problems later, so I'll see how I get these two things in.

src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -194,34 +189,33 @@ function* colonyCreate({
...
-      methodName: 'addExtension',
-      params: { contractName: 'OldRoles' },
-      ready: false,
+      methodName: 'setAdministrationRole',
+      params: {
+        address: walletAddress,
+        domainId: 1,
+        setTo: true,
+      },
+      ready: true,
```
sprusr: I think all of these ones that have been changed should be `ready: false`.

src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -194,34 +189,33 @@ function* colonyCreate({
...
 
-    yield createGroupedTransaction(deployOldRoles, {
+    yield createGroupedTransaction(setAdministrationRole, {
       context: COLONY_CONTEXT,
-      methodName: 'addExtension',
-      params: { contractName: 'OldRoles' },
-      ready: false,
+      methodName: 'setAdministrationRole',
+      params: {
+        address: walletAddress,
```
sprusr: This shouldn't be set here for this one, or for `setFundingRole`, they should be set to the `OneTxPayment` extension address below.

src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -413,59 +406,11 @@ function* colonyCreate({
...
-    const {
-      payload: {
-        transaction: {
-          receipt: {
-            logs: [deployOneTxLog],
-          },
-        },
-      },
-    } = yield takeFrom(deployOneTx.channel, ActionTypes.TRANSACTION_SUCCEEDED);
-    const oneTxAddress = parseExtensionDeployedLog(deployOneTxLog);
```
sprusr: We still need all of this section here to wait for the transaction, and then get the `oneTxAddress` so that we can set the params of the roles transactions below.

src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -413,59 +406,11 @@ function* colonyCreate({
...
-
-    /*
-     * Set OneTx role
-     */
-    yield put(transactionAddParams(setOneTxRole.id, { address: oneTxAddress }));
```
sprusr: We need to do this `transactionAddParams` for both of the new role setting transactions, so this line should be replaced by two (one for each role that's set). The lines below this should be replaced with the same thing for each new role setting transaction.

src/modules/dashboard/sagas/colonyCreate.ts
karolinkas: Ok, now I see what went wrong, reading this comment I thought you wanted me to remove `setOneTxRole` and replace it. I'll bring it back now.

rdig:
I think this missed a bunch of things, since accessing the admin area still relies on the old roles and checks: https://github.com/JoinColony/colonyDapp/blob/master/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx#L231

Because of this, #1845 was introduced

PR #1829, opened 2019-09-16
Fix colony admin permissions dialog

sprusr:
## Description

Various fixes for DLP phase 1.

**Changes** 🏗

* It wasn't possible to unset roles
* It wasn't possible to set admin role
* Role names are now from `react-intl` messages

**Deletions** ⚰️

* Removed an unwanted console.log

### DIFF
--- a/src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
+++ b/src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
@@ -29,7 +29,6 @@ import {
   useDataMapFetcher,
   useUserDomainRoles,
 } from '~utils/hooks';
-import { capitalize } from '~utils/strings';
 import { filterUserSelection } from '~utils/arrays';
 
 import { userSubscriber } from '../../../users/subscribers';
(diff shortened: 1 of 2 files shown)


src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -149,44 +176,52 @@ const ColonyPermissionEditDialog = ({
...
-          if (checkIfCanBeSet(role)) {
-            accumulator[role.toUpperCase()] = true;
-            return accumulator;
-          }
-          return accumulator;
-        }, {}),
+        domainId: domain.id,
+        colonyAddress,
+        roles: availableRoles.reduce(
+          (acc, role) => ({ ...acc, [role]: !!p.roles.find(r => r === role) }),
```
JamesLefrere: `p.roles.includes(role)` ?

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -149,44 +176,52 @@ const ColonyPermissionEditDialog = ({
...
-  /* Currently arbitration should appear in the list of roles
-    but can not be set yet
-   */
-  const checkIfCanBeSet = role => {
-    if (role === 'Arbitration' || role === 'Root') {
-      return false;
-    }
-    return true;
-  };
+  // Get the current user's roles in the selected domain
```
JamesLefrere: Same :-/

sprusr: Huh?

JamesLefrere: Huh, that appeared on the wrong line... I was rather attempting to comment about the enum.

PR #1831, opened 2019-09-16
Improve Colony transactions styles

### LINKED ISSUE #1539: Colony Settings - Styling Improvements
## Steps to reproduce

1. Create new colony (optional)
2. Mint tokens (optional)
3. Go to colony settings via cog wheel icon
4. Go to transactions

## Expected behaviour

1. There should be slide in animation instead of sudden hover state. 
So content is pushed to the left and button appears. While it appears
suddenly it is kind of suprising in a negative way. 

2. Double check while claiming transactions (minted funds) the contract address for pending transaction is not breaking the card layout. I saw many '0x000...' zeroes. Possible bug. I recommend checking it while creating a new colony from the ground. 

3. Double check on figma the distance between the settings menu and a back button, at first glance it appears few pixels lower as it is on a design. 

## Actual behaviour

Screen Shot 2019-07-17 at 09 06 10
(hover on the transaction card) 

## Suggestions (optional)

Not a super priority, but we can do better with hover animation

JamesLefrere:
## Description

This PR makes some UI changes to the colony admin transactions, contributing to #1539.

**Changes** 🏗

* Improve Colony admin transaction styles
* Correct the left menu top margin to reflect the design
* Use a slide-in transition for the Etherscan button

Resolves #1539

### DIFF
--- a/src/modules/admin/components/AdminDashboard/AdminDashboard.css
+++ b/src/modules/admin/components/AdminDashboard/AdminDashboard.css
@@ -9,6 +9,6 @@
 }
 
 .headingWrapper {
-  margin-top: 70px;
+  margin-top: 50px;
   padding-left: 50px;
 }
--- a/src/modules/core/components/TransactionList/TransactionDetails.css
+++ b/src/modules/core/components/TransactionList/TransactionDetails.css
@@ -1,11 +1,21 @@
+
+.overflowPrevention {
+  width: 130px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+}
+
 .primaryText {
+  composes: overflowPrevention;
   font-weight: var(--weight-medium);
   color: var(--grey-1);
 }
 
 .secondaryText {
+  composes: overflowPrevention;
   margin-top: 2px;
   font-size: var(--size-small);
+  white-space: nowrap;
   color: var(--grey-3);
   letter-spacing: 0.5px;
 }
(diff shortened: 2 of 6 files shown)


JamesLefrere:
Ah, my OS was hiding the scroll bars. Ok, I'll give that a go 👍

rdig:
Hey,

I know this was not introduced here, but while you're at it...

If the username is very long, the transaction list gets broken, since no ellipsis is applied.

_(there's both an inline ellipsis and multiline ellipsis available for `compose` in `~styles/text.css`)_

See this:

Screenshot from 2019-09-19 12-37-24

rdig:
Can I re-review this, or do you need more time ?

chmanie:
Maybe consider doing everything that's stated in #1539?

ceolson01:
@rdig 

For the first screenshot - is that with a shorter viewport?

As for the `overflow-x`... can you tell if it's being applied to the `tr`?

I can't replicate either of these issues in any of my browsers, so debugging this is a tad difficult 🤦‍♂

ceolson01:
@rdig your screenshots & descriptions were super helpful - thank you for those.

I was unable to replicate those issues on my local machines, but I've blindly attempted a fix for larger viewports... hopefully that does it 🤞

ceolson01:
Shoooot I think I originally applied the overflow change to the wrong selector - I've `amend`ed my most recent commit to _hopefully_ fix this... 🙄

Thanks for iterating through this with me. Good for another quick look 🤞

Issue #1780, opened 2019-08-30
Add move funds between pots dialog


sprusr:
### Story

As a colony admin, I want to be able to move funds between domain pots, so that I have greater control over where funds are allocated.

### Description

This issue is the main event of the DLF/P phase 2 work! In this, we're adding the dialog component which allows selecting a "from" and "to" domain, as well as entering an amount and token. Submitting the dialog will dispatch the action created in #1779 to create the actual transaction.

Since we already have all the means to wire this thing up, we'll include that here too. The domain pot amounts for the selected token should be displayed below the domain dropdowns. If ETH is selected, then the USD conversion should also be shown (as in Figma).

Domains for which the current user does not have the effective funding role should be shown as greyed out in the domains dropdown, along with a tooltip. Form validation should be run to check that the selected from domain has enough funds to transfer the entered amount.

A "Move funds" button should be added to the admin tokens tab sidebar to open this dialog.

🌮 The dialog should accept a "to domain" as a param when opening, and should be opened with the currently selected domain preselected (or nothing selected if the user does not have permission.

### Design

Dialog: [Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=998%3A27269)

![image](https://user-images.githubusercontent.com/7497084/64004077-de885200-cb0d-11e9-8ebf-077aa63ebd30.png)

![image](https://user-images.githubusercontent.com/7497084/64004198-12fc0e00-cb0e-11e9-9f4d-3fa6f5fcd26e.png)

![image](https://user-images.githubusercontent.com/7497084/64004660-f8766480-cb0e-11e9-87b1-f927715df614.png)

### This issue is complete when...

- [ ] Dialog exists with from/to dropdowns containing a list of domains, amount input, and token dropdown
  - [ ] Domains in dropdown are disabled if user does not have funding role to move from that pot (⚠️ need clarification on whether to domain also needs funding role)
  - [ ] Domain pot balances for selected token are displayed
  - [ ] Form is validated invalid if not enough funds in from pot
  - [ ] USD amount is displayed if ETH selected
- [ ] "Move funds" button added to colony admin tokens tab sidebar to open dialog
  - [ ] 🌮 Opens dialog with currently selected domain from tokens tab preselected as "to" domain

### Suggestions

* Getting whether the current user has funding permission for each domain shouldn't be too bad, but there might be some trickiness there, especially for greying out domain dropdown items.

### Subtasks

[...]

PR #1834, opened 2019-09-17
Refactor token balances [DLF/P]

### LINKED ISSUE #1768: Refactor colony token balances
### Story

As a Colony developer, I want to differentiate between the different funding pots (domains) where my funds are allocated, so that I have a greater overview of my domain-specific budgets.

### Description

This issue is the start of the group of "phase 2", in which we add the ability to move funds between different domain funding pots. In order to do that, we first need to have the distinction between different domain pots where token balances are held.

Token balances should be stored in state per domain pot. The new state could look something like `{ [domainId: number]: { [tokenAddress: Address]: BigNumber } }`. The saga which fetches token balances should be changed to also include a `domainId` in its params. Any places in the dApp which currently fetch colony token balances should use the root domain as a default.

### Design

Here's [the Figma](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=998%3A27269) for what we're aiming for with this phase. Give it at least a glance over 😄 

- [ ] Refactor colony token balances reducer to take a `domainId` and store using this
- [ ] Refactor colony token balance fetch saga/actions to include `domainId` as params, fetch token balance for just that domain's pot
- [ ] Refactor any existing uses of the colony token balance fetcher/selector to use root domain

### Suggestions

* ???

### Subtasks (optional)

- [ ] ???

ceolson01:
## Description

This PR refactors the token references for colonies (and users) for domain-level (pots, really) balances.

This PR introduces 2 new types:

`ColonyTokenReference`:
```jsx
[colonyAddress: Address]: {
    ...
    tokens: {
        [tokenAddress: Address]: {
            address: Address,
            ...
            balances?: {
                [domainId: number]: BigNumber,
            },
        },
    },
},
```

`UserTokenReference`:
```jsx
[tokenAddress: Address]: {
    address: Address,
    balance?: BigNumber,
    iconHash?: string,
}
```

**New stuff** ✨

* `ColonyTokenReference` type
* `UserTokenReference` type
* `FetchColonyTokenBalances` saga for fetching balances of all domains for a colony token
* `tokenBalanceSelector` (will be helpful for both #1780 and #1799)

**Changes** 🏗

* Fix `colonyRecentTokensSelector`
* Validate pot funding per domain
* `FetchColonyTokenBalance` now requires a `domainId` to fetch balances for
* Validate domain selection has enough funds
  * Disable domains that don't have enough funds

**Deletions** ⚰️

* `TokenReference` type (er... moved to `ColonyTokenReference` and `UserTokenReference`)

## TODO

- [ ] ~Refactor types to match #1804~ Will pair with @chmanie on a rebase when the time comes

Resolves #1768


src/modules/dashboard/data/queries/colonies.ts
```diff
@@ -492,14 +492,20 @@ export const getColonyTokenBalance: Query<
...
       balance: rewardsPotTotal,
-    } = await colonyClient.getFundingPotBalance.call({ potId: 0, token });
-    return new BigNumber(nonRewardsPotsTotal.add(rewardsPotTotal).toString(10));
+    } = await colonyClient.getFundingPotBalance.call({ potId, token });
+    if (domainId === 0) {
+      const {
+        total: nonRewardsPotsTotal,
+      } = await colonyClient.getNonRewardPotsTotal.call({ token });
+      return new BigNumber(
+        nonRewardsPotsTotal.add(rewardsPotTotal).toString(10),
```
sprusr: I think you can avoid converting these to string, BN constructor should take a BN.

src/modules/dashboard/data/queries/colonies.ts
```diff
@@ -492,14 +492,20 @@ export const getColonyTokenBalance: Query<
...
+    const { potId } = await colonyClient.getDomain.call({ domainId });
     const {
       balance: rewardsPotTotal,
-    } = await colonyClient.getFundingPotBalance.call({ potId: 0, token });
-    return new BigNumber(nonRewardsPotsTotal.add(rewardsPotTotal).toString(10));
+    } = await colonyClient.getFundingPotBalance.call({ potId, token });
+    if (domainId === 0) {
+      const {
+        total: nonRewardsPotsTotal,
+      } = await colonyClient.getNonRewardPotsTotal.call({ token });
```
sprusr: Ahhh this is clearer to me now what domain 0 is :)

src/immutable/ColonyTokenReference.ts
```diff
@@ -0,0 +1,34 @@
+import { $ReadOnly } from 'utility-types';
```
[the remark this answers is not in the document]
ceolson01: Yes, agreed. Their uses seem much more obvious now since their properties have diverged a bit. I also like that `TokenReference`, in the context of users (now `UserTokenReference`) no longer has props that'll _never_ be used within that context (`isNative`, `isExternal`, etc.) 🙂

src/modules/admin/components/Tokens/TokenCard.tsx
```diff
@@ -31,7 +36,8 @@ const MSG = defineMessages({
 });
 
 const TokenCard = ({
-  token: { address, isNative, balance },
+  domainId,
```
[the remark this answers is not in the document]
ceolson01: I decided instead to go the route of making the typing more strict. This way, if a `ColonyTokenReferenceType` is passed, `domainId` is required. For a user token, the opposite is true.

src/modules/core/checks.ts
```diff
@@ -17,11 +20,35 @@ export const isPendingMultisig = (tx: TransactionType) =>
 /*
  * Tokens
  */
-export const tokenBalanceIsPositive = ({ balance }: TokenReferenceType) =>
-  !!balance && balance.gten(0);
+export const tokenBalanceIsPositive = (
+  tokenReference: ColonyTokenReferenceType | UserTokenReferenceType,
+  domainId: number,
+) => {
+  let balance = new BigNumber(0);
```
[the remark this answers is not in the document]
ceolson01: Nice call 👍

src/immutable/state/users/index.ts
```diff
@@ -33,7 +33,7 @@ export type CurrentUserColoniesType = ImmutableSet<Address>;
 export type CurrentUserTasksType = ImmutableSet<[Address, TaskDraftId]>;
 
 export type CurrentUserTokensType = DataRecordType<
-  List<TokenReferenceRecordType>
+  List<ColonyTokenReferenceRecordType>
```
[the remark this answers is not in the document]
ceolson01: 🤦‍♂

ceolson01:
@sprusr I believe I've addressed all of the feedback above. Ready for another look

src/modules/dashboard/data/queries/colonies.ts
```diff
@@ -501,11 +501,9 @@ export const getColonyTokenBalance: Query<
       const {
         total: nonRewardsPotsTotal,
       } = await colonyClient.getNonRewardPotsTotal.call({ token });
-      return new BigNumber(
-        nonRewardsPotsTotal.add(rewardsPotTotal).toString(10),
-      );
+      return new BigNumber(nonRewardsPotsTotal.add(rewardsPotTotal));
```
sprusr: Ahhh I'm sorry, it seems I was mistaken about the BN constructor taking another BN 🙇 this is throwing an error now :(

src/modules/admin/components/Tokens/TokenCard.tsx
```diff
@@ -14,16 +14,17 @@ import { SpinnerLoader } from '~core/Preloaders';
...
 import { tokenFetcher } from '../../../dashboard/fetchers';
 
 import styles from './TokenCard.css';
 
-interface Props {
-  domainId?: number;
-  token: ColonyTokenReferenceType | UserTokenReferenceType;
+interface Props<T> {
+  token: T;
+  domainId: T extends ColonyTokenReferenceType ? number : never;
```
sprusr: Wow this is amazing 🤩

src/modules/dashboard/data/queries/colonies.ts
ceolson01: Shoooot I swear I tested it with this... But you're right, broken indeed!

I've dropped this commit.

src/modules/dashboard/sagas/colony.ts
```diff
@@ -348,14 +355,57 @@ function* colonyTokenBalanceFetch({
...
+        fetchColonyTokenBalance(
+          colonyAddress,
+          tokenAddress,
+          COLONY_TOTAL_BALANCE_DOMAIN_ID,
+        ),
+      ),
+      // fetch balances for root domain
+      put(fetchColonyTokenBalance(colonyAddress, tokenAddress, 1)),
+      // fetch balances for other domains
+      ...domains.map(({ id: domainId }) =>
```
sprusr: This is causing an error when `domains` is undefined (still being loaded). I can reproduce by directly loading `/colony/mycolony/admin`.

PR #1854, opened 2019-09-19
Move Transactions and Advanced tab

### LINKED ISSUE #1806: Move colony transactions list to colony home
### Story

As a user of a colony, I want to be able to see transactions relevant to the colony, so that I know what activity has been happening in the colony.

### Description

The transaction list currently in the colony admin pages should be moved to a publicly accessible location (not hidden in admin settings). That location is just right to the colony home screen!

It's going to show up as a tab next to the tasks in the colony and its domains. Initially we wanted the currently selected domain to filter the list of transactions, but right now our "transaction" list is quite limited, and doesn't really include things that can be filtered by domain.

As a result, this issue should be a matter of just moving components around.

While we're here, we're also moving the "Advanced" sub-tab out of the "Profile" tab of the colony admin pages, and into its own top-level tab.

### Design

[Figma link](https://www.figma.com/file/L6scZfliLKnDt95nwan4jwzg/DLP-DLF?node-id=998%3A31392)

Transaction list, as it is.
image

"Advanced" has its own top-level tab now:
image

Also has a tip about the fact that the domain filter doesn't apply at the moment:
image

- [x] Colony transactions list is now a tab on colony home, removed from colony admin
- [x] Tip is added that this is for "All Transactions in Colony"
- [x] Colony admin "Advanced" is now a top-level tab

### Suggestions

* Good luck I guess?

### Subtasks

- [ ] ???

karolinkas:
## Description

Move the transaction tab from the AdminDashboard to ColonyHome and the Advanced tab from the Profile to it's own tab within the AdminDashboard.

Resolves #1806


src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -116,6 +126,7 @@ const ColonyHome = ({
   const [filteredDomainId, setFilteredDomainId] = useState(0);
   const [isTaskBeingCreated, setIsTaskBeingCreated] = useState(false);
   const [showRecoverOption, setRecoverOption] = useState(false);
+  const [activeTab, storeActiveTab] = useState('tasks');
```
ceolson01: Let's follow the same convention here - `setActiveTab` instead of `storeActiveTab`

Also, could we make this typing more strict? Something like `... = useState<'tasks' | 'transactions'>('tasks');`

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -298,10 +316,13 @@ const ColonyHome = ({
           <BreadCrumb elements={crumbs} />
         </div>
         <Tabs>
-          <TabList>
-            <Tab>
+          <TabList extra={activeTab === 'tasks' ? filterSelect : noFilter}>
```
ceolson01: I don't think there's supposed to be a `filterSelect` here when `'tasks'` is the `activeTab`... according to the figma spec it should just be `null`.

Notice the duplicate one in the `extra` position in this screenshot:

Screen Shot 2019-09-20 at 4 11 01 PM

karolinkas: Of couurse 🥇

karolinkas:
@ceolson01 I targeted everything you mentioned.

src/modules/admin/components/Profile/Profile.tsx
```diff
@@ -31,19 +26,13 @@ const Profile = ({ colony }: Props) => (
         <Tab>
```
ceolson01: This shouldn't have any tabs, it should now **just** be the profile component. If you look at the figma spec, you'll notice this is simply a heading now:

Screen Shot 2019-09-25 at 12 57 16 PM

src/modules/core/components/BreadCrumb/BreadCrumb.tsx
```diff
@@ -9,7 +8,7 @@ interface Props {
    * array of elements outside of this components and just an array
    * of strings gets passed in. The last active element gets highlighted.
    */
-  elements: MessageDescriptor[];
+  elements: string[];
```
ceolson01: I think it'd be wise to allow a mixed array of `strings` **or** `MessageDescriptor`s. We can then use `injectIntl` here and perform the `formatMessage` conditionally on each item.

src/modules/core/components/BreadCrumb/BreadCrumb.tsx
```diff
@@ -19,23 +18,19 @@ const BreadCrumb = ({ elements }: Props) => {
...
-                </b>
+                <span className={styles.breadCrumble}>{crumb}</span>
+                <span className={styles.arrow}>&gt;</span>
               </>
             ) : null}
           </>
+          <>
+            {i === elements.length - 1 || elements.length === 1 ? (
+              <b className={styles.breadCrumble}>{crumb}</b>
+            ) : null}
```
ceolson01: I think this can be simplified a little bit. Since the falsy expression in each of these ternaries is `null`, these don't need to be ternaries.

Also, those surrounding fragments aren't necessary 🙂

Something along the lines of:

```javascript
<div className={styles.element} key={`breadCrumb_${crumb}`}>
  {elements.length > 1 && i < elements.length - 1 && (
    <>
      <span className={styles.breadCrumble}>{crumb}</span>
      <span className={styles.arrow}>&gt;</span>
    </>
  )}
  {(i === elements.length - 1 || elements.length === 1) && (
    <b className={styles.breadCrumble}>{crumb}</b>
  )}
</div>
```

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -97,10 +108,20 @@ const MSG = defineMessages({
     id: 'dashboard.ColonyHome.recoverColonyCancelButton',
     defaultMessage: 'Nope! Take me back, please',
   },
+  root: {
+    id: 'root',
+    defaultMessage: 'root',
+  },
```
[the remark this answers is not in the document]
karolinkas: Oh nice, this is merged now then.

src/modules/core/components/BreadCrumb/BreadCrumb.tsx
karolinkas: I agree with the suggestions for the breadcrumb component, but I would prefer to do them in the other PR.

karolinkas:
@ceolson01 This is ready for another round.

src/modules/core/components/BreadCrumb/BreadCrumb.tsx
ceolson01: Sure, makes sense!

> I am not sure why these changes even show up here.

Me neither 🤷‍♂🙂

src/modules/dashboard/components/ColonyHome/TabTransactions/TabTransactions.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { ColonyType } from '~immutable/index';
+
+import Transactions from '~admin/Transactions';
+
+interface Props {
+  colony: ColonyType;
+}
+
+const TabTransactions = ({ colony }: Props) => (
+  <Transactions colonyAddress={colony.colonyAddress} />
```
ceolson01: Since this `TabTransactions` component is simply returning the `Transactions` component, it seems a little unnecessary. Maybe we should just remove this `TabTransactions` component and use `Transactions` directly in the `ColonyHome` component. Thoughts?

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
```diff
@@ -69,7 +69,7 @@ const navigationItems = (colony: ColonyType): NavigationItem[] => [
   {
     id: 1,
     title: MSG.tabProfile,
-    content: <Profile colony={colony} />,
+    content: <ProfileEdit colony={colony} />,
```
ceolson01: `ProfileEdit` is still missing the heading & underline bit.

Here's how it is now:

Screen Shot 2019-09-27 at 11 09 40 AM


And here's the figma:

Profile

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
```diff
@@ -93,8 +93,8 @@ const navigationItems = (colony: ColonyType): NavigationItem[] => [
   },
   {
     id: 5,
-    title: MSG.tabTransaction,
-    content: <Transactions colonyAddress={colony.colonyAddress} />,
+    title: MSG.tabAdvanced,
+    content: <ProfileAdvanced colony={colony} />,
```
ceolson01: Since `ProfileAdvanced` no longer has tabs at the top, I think it's deserving of a heading there (just like `ProfileEdit`). This isn't spec'd in figma, though, so maybe get with product to confirm

src/modules/admin/components/Transactions/Transactions.css
```diff
@@ -1,5 +1,6 @@
 .main {
-  padding: 0 20% 40px 0;
+  margin-top: -25px; /* This is an exception, other tabs require more space here */
```
ceolson01: 👏 thanks for the comment here

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
karolinkas: That's a bit strange, maybe I committed after you reviewed, every tab has a heading by now. What I will confirm with product though is if they want me to adjust their layout a bit since they are currently jumping a bit.

PR #1860, opened 2019-09-20
Add root to domains list

### LINKED ISSUE #1700: Add "Root" domain to the colony home screen and domain list
### Story

As a *user interacting with domains*, I want to *be able to select the root domain from the colony home screen*, so that *I can filter by tasks in that domain*

### Description

Ok, that user story isn't very helpful. But we want to add the "root" domain to the colony home screen and the list of domains in the admin screen. This means a user will be able to see "all" domains or select "root" to just see the tasks in the root domain. 

### Design

Add "Root" to domains tab. This means the domain tab will never have an empty state. 
Colony

Add "Root" below "All" on the colony home screen.
Colony

karolinkas:
## Description

This PR is rebased on top of the #1785, that's why it seems a bit bigger then it actually is.

The changes are adding the "root" domain to the DomainList by default, while making sure the root domain can not be edit, because it wouldn't make sense to edit it. Also all newly added domains are now all underscore. 
Screen Shot 2019-09-20 at 18 47 29

To achieve this I changed the `allDomains.ts` reducer by adding a root domain if there's none, so that it gets fetched with the domainsFetcher.

Consequently, it also shows on Colony Home. 
Screen Shot 2019-09-20 at 18 50 58

Resolves  #1700

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
@@ -105,7 +105,7 @@ interface MatchParams {
   colonyName: string;
 }
 
-interface Props extends RouteComponentProps<MatchParams> {}
+type Props = RouteComponentProps<MatchParams>;
 
 const COLONY_DB_RECOVER_BUTTON_TIMEOUT = 20 * 1000;
 
--- a/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
@@ -1,5 +1,5 @@
-import { Redirect } from 'react-router';
 import React, { useState, useCallback, useEffect, useMemo } from 'react';
+import { Redirect, RouteComponentProps } from 'react-router';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { subscribeActions as subscribeToReduxActions } from 'redux-action-watch/lib/actionCreators';
 import { useDispatch } from 'redux-react-hook';
@@ -101,9 +101,11 @@ const MSG = defineMessages({
   },
 });
 
-interface Props {
-  match: any;
+interface MatchParams {
+  colonyName: string;
 }
+
+interface Props extends RouteComponentProps<MatchParams> {}
 
 const COLONY_DB_RECOVER_BUTTON_TIMEOUT = 20 * 1000;
 
--- a/src/modules/dashboard/reducers/allDomains.ts
+++ b/src/modules/dashboard/reducers/allDomains.ts
@@ -84,10 +84,6 @@ const allDomainsReducer: ReducerType<AllDomainsMap> = (
         meta: { key },
         payload: { domains },
       } = action;
-      /* Add root domain to response which is the parent of all domains  */
-      if (domains.find(domain => domain.name !== 'root')) {
-        domains.push({ name: 'root', id: 0 });
-      }
       return state.set(
         key,
         FetchableData({
--- a/src/modules/dashboard/reducers/allDomains.ts
+++ b/src/modules/dashboard/reducers/allDomains.ts
@@ -84,6 +84,10 @@ const allDomainsReducer: ReducerType<AllDomainsMap> = (
         meta: { key },
         payload: { domains },
       } = action;
+      /* Add root domain to response which is the parent of all domains  */
+      if (domains.find(domain => domain.name !== 'root')) {
+        domains.push({ name: 'root', id: 0 });
+      }
       return state.set(
         key,
         FetchableData({
--- a/src/modules/admin/components/Domains/DomainList.tsx
+++ b/src/modules/admin/components/Domains/DomainList.tsx
@@ -46,7 +46,7 @@ const DomainList = ({
     <div className={styles.listWrapper}>
       <Table scrollable>
         <TableBody>
-          {domains && (
+          {domains &&
             Object.keys(domains).map(domainId => (
               <DomainListItem
                 key={domainId}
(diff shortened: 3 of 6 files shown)


src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -220,21 +239,72 @@ const ColonyHome = ({
...
+    domains
+      .sort((a, b) => a.id - b.id)
+      .reduce((accumulator, element) => {
+        if (element && element.name && element.id <= filteredDomainId) {
+          const message = {};
+          message[element.name] = {
+            id: element.name,
+            defaultMessage: element.name,
+          };
+          const messageDescriptor = defineMessages(message);
```
chmanie: `defineMessages` is not supposed to be used in a loop (like here). Messages have to be statically defined in the head of the file (like we always do).

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -220,21 +239,72 @@ const ColonyHome = ({
...
+          (accumulator as any).push(messageDescriptor);
```
chmanie: Please don't use `as any`. In this case we know the type so you can just do `accumulator as MessageDescriptor[]`.

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
karolinkas: Mh, then I don't know how to do it since that's specifically what Raul asked me to do, to create messageDescriptors.

chmanie: Yes you should create messageDescriptors, just not in a loop.

We have to statically define them (we know what to expect here, otherwise translations won't be possible). Maybe @rdig can also chime in?

karolinkas: Could we just add new MessageDescriptors to the statically defined one?

chmanie: These ones? https://github.com/JoinColony/colonyDapp/pull/1860/files/d4d0ba5a98050a540ff96f5d0611462d85084986#diff-d800e9f821d34745e078f8a1a78854f6R49

I don't see why not!

karolinkas: @chmanie Ok, I see now why not. You can not assign anything to the MSG object.

chmanie: Yeah, because it has to be _statically_ defined. What do you want to assign?

rdig: Looking at this from another point of view:

Does the `Root` domain actually need a translation? We can just assume it's going to be called `Root` in all languages.

karolinkas: Nobody actually gave a useful suggestion here. I need to dynamically create new translations, fine it's static, so I can't but what's could be a workaround then?

ceolson01: The only copy that needs to be (or can be) translated is static copy that's hardcoded within the dapp. User-entered data (such as domains, in this case) cannot be translated, and shouldn't have i18n messages.

> Does the Root domain actually need a translation? We can just assume it's going to be called Root in all languages.

I think it'd be wise to translate root wherever we can. There's a default `root` message provided in the global i18n file as of #1799 being merged.

karolinkas: Ok, true, so we'll use the global root message and keep all other domains as they are.  That will simplify the BreadCrumb component to the way it was like a week ago. Objections anyone?

chmanie: To me that seems like a good plan. It’s like Curtis said, thank you for the explanation. We can’t translate dynamic values, as when you think about it we have to expect beforehand which translations there _can_ be. Think about a person tasked with translating the whole app to Spanish. We need all the texts there can possibly be in English first, like in a document (just in this case it’s a JSON file). That’s what statically means in this case.

src/modules/dashboard/reducers/allDomains.ts
```diff
@@ -43,6 +43,10 @@ const allDomainsReducer: ReducerType<AllDomainsMap> = (
         meta: { key },
         payload: { domains },
       } = action;
+      /* Add root domain to response which is the parent of all domains  */
```
JamesLefrere: I think this should be added to the query, because that's the source of truth, and it should make things clearer. The root domain object could be added to the start of this array: https://github.com/JoinColony/colonyDapp/blob/175e6f72bc6875551c00d8f74ec856343e70a52b/src/modules/dashboard/data/queries/colonies.ts#L468

karolinkas: @JamesLefrere Just moved the change to the query! :white_check_mark:

ceolson01:
After rebasing, this PR took a bit of a turn in a different direction. That said, I've had to add a few commits, so I've added you as a reviewer again @chmanie.

Most notably, because the `root` domain is now included in the colony domains query [per this suggestion](https://github.com/JoinColony/colonyDapp/pull/1860#discussion_r328547114), the resulting redundant `root` domains have been removed.

There's also a slight type improvement to the `match` prop from react-router.

ceolson01:
I'm putting this `on-hold` until #1877 is merged. This will require a bit of a rebase, and the scope may also be slightly reduced, as #1877 contributes to #1700 in a few ways 🎉

ceolson01:
Closing, as #1907 covered this

Issue #1846, opened 2019-09-18
React DOM violations for `TaskListItem`


JamesLefrere:
## Steps to reproduce

1. Go to the task list (colony home) for a colony with tasks
2. Change the items shown on the view (e.g. change the task filter)

## Expected behaviour

* No errors, everybody had a great day

## Actual behaviour

* Errors logged, someone had just a so-so day

```
backend.js:1 Warning: validateDOMNesting(...): <tr> cannot appear as a child of <a>.
    in tr (created by TableRow)
    in TableRow (created by dashboard.TaskList.TaskListItem)
    in a (created by Link)
    in Link (created by Link)
    in Link (created by InjectIntl(Link))
    in InjectIntl(Link) (created by dashboard.TaskList.TaskListItem)
    in dashboard.TaskList.TaskListItem (created by InjectIntl(dashboard.TaskList.TaskListItem))
    in InjectIntl(dashboard.TaskList.TaskListItem) (created by TaskList)
    in tbody (created by TableBody)
    in TableBody (created by TaskList)
    in table (created by Table)
    in Table (created by TaskList)
    in TaskList (created by dashboard.ColonyTasks)
    in dashboard.ColonyTasks (created by TabContribute)
    in TabContribute (created by dashboard.ColonyHome)
    in div (created by TabPanel)
    in TabPanel (created by TabPanel)
    in TabPanel (created by dashboard.ColonyHome)
    in div (created by UncontrolledTabs)
    in UncontrolledTabs (created by Tabs)
    in Tabs (created by Tabs)
    in Tabs (created by dashboard.ColonyHome)
    in main (created by dashboard.ColonyHome)
    in div (created by dashboard.ColonyHome)
    in dashboard.ColonyHome (created by Route)
    in main (created by pages.NavigationWrapper)
    in div (created by pages.NavigationWrapper)
    in div (created by pages.NavigationWrapper)
    in pages.NavigationWrapper (created by Route)
    in Route (created by ConnectedOnlyRoute)
    in ConnectedOnlyRoute (created by Routes)
    in Switch (created by Routes)
    in Routes (created by Context.Consumer)
    in Connect(Routes) (created by Route)
    in Route (created by withRouter(Connect(Routes)))
    in withRouter(Connect(Routes)) (created by App)
    in div (created by App)
    in DialogProvider (created by Route)
    in Route (created by withRouter(DialogProvider))
    in withRouter(DialogProvider) (created by App)
[...]

PR #1867, opened 2019-09-23
Link to each task without an <a> tag

karolinkas:
## Description

An <a> tag can not wrap a <tr> component, therefore we are now linking to each task programmatically. 

Contributes to #1846

### DIFF
--- a/src/modules/dashboard/components/TaskList/TaskListItem.tsx
+++ b/src/modules/dashboard/components/TaskList/TaskListItem.tsx
@@ -1,10 +1,13 @@
 import {
-  IntlShape,
+  InjectedIntlProps,
   defineMessages,
   FormattedMessage,
   injectIntl,
 } from 'react-intl';
 import React from 'react';
+
+import { RouteComponentProps, withRouter } from 'react-router-dom';
+import compose from 'recompose/compose';
 
 import { Address, ENSName } from '~types/index';
 import { TaskType } from '~immutable/index';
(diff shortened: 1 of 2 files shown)


src/modules/dashboard/components/TaskList/TaskListItem.tsx
```diff
@@ -76,40 +79,46 @@ const TaskListItem = ({ data, intl: { formatMessage } }: Props) => {
...
   );
 };
 
 TaskListItem.displayName = displayName;
 
-export default injectIntl(TaskListItem);
+const enhance = compose(
+  withRouter,
+  injectIntl,
+) as any;
```
chmanie: I don't know if that _can_ be solved properly right now but it definitely is a code smell. We introduced a lot of `as any`s during the typescript refactor but that doesn't mean we can add more now. If any we should try to remove as many of those as possible. What's the typescript error if you remove it? Let's find a better solution here.

karolinkas: Yeih, found a type, after trying for a while, I will add it in, in other occurrences as well.

chmanie:
There's https://www.npmjs.com/package/@types/recompose, are we using this?

And also this guide (for flow though): https://github.com/acdlite/recompose/tree/master/types

ceolson01:
@chmanie I've narrowed the scope of this PR. I think the type changes to the HOCs should come separately, as the changes are a bit more involved than this PR should see.

This is good for another look.

PR #1858, opened 2019-09-20
Add move tokens dialog

sprusr:
## Description

This PR adds a dialog which acts as a front for the `moveFundsBetweenPots` method on the colony contract.

image

**New stuff** ✨

* `TokensMoveDialog` and `TokensMoveDialogForm`
* `Select` core component now supports field status

**Changes** 🏗

* Ensure colony token balances are keyed by number and stored as `ImmutableMap` in `coloniesReducer`
* Refetch affected token balances after transaction in `colonyDomainsFetch` saga

**Deletions** ⚰️

* ???

## TODO

- [x] Wire up domain balance fetching (make sure the from domain actually has enough balance)
- [x] In `Tokens`, `canMoveTokens` should actually be as a result of some kind of check
- [x] Preselected `toDomain` when opening via "Move funds"?
- [x] Open `TokensMoveDialog` with selected domain as `toDomain` once #1799 is merged

Resolves #1780

PR #1868, opened 2019-09-24
Add domain funding to colony home

### LINKED ISSUE #1783: Add domain funding to colony home

ceolson01:
## Description

This PR adds a domain funding (both balances and "Fund" modal button) widget to the colony home.

**New stuff** ✨

* `ColonyFunding` component

**Changes** 🏗

* Use `ColonyFunding` component in `ColonyHome`

## TODO

- [x] Wire up to the transfer funds between pots dialog once #1858 is complete

Closes #1783

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
@@ -9,7 +9,6 @@ import { useSelector, useRoles } from '~utils/hooks';
 import TokenItem from './TokenItem';
 import { useColonyTokens } from '../../../hooks/useColonyTokens';
 import { canMoveTokens as canMoveTokensCheck } from '../../../../admin/checks';
-import { COLONY_TOTAL_BALANCE_DOMAIN_ID } from '../../../../admin/constants';
 import { currentUserSelector } from '../../../../users/selectors';
 
 import styles from './ColonyFunding.css';
@@ -44,7 +43,8 @@ const ColonyFunding = ({ colonyAddress, currentDomainId }: Props) => {
     [currentUserWalletAddress, roles],
   );
 
-  return currentDomainId === COLONY_TOTAL_BALANCE_DOMAIN_ID ? null : (
+  // hide for root domain
+  return currentDomainId === 0 ? null : (
     <div>
       <Heading appearance={{ size: 'normal', weight: 'bold' }}>
         <FormattedMessage {...MSG.title} />
--- a/src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
@@ -1,0 +1,76 @@
+import React, { useMemo } from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+
+import Button from '~core/Button';
+import Heading from '~core/Heading';
+import { Address } from '~types/index';
+import { useSelector, useRoles } from '~utils/hooks';
+
+import TokenItem from './TokenItem';
+import { useColonyTokens } from '../../../hooks/useColonyTokens';
+import { canMoveTokens as canMoveTokensCheck } from '../../../../admin/checks';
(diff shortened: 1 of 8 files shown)


rdig:
So I think this should be based on Karolin's #1823 since funding goes on the refreshed Colony Dashboard.

But most likely that will be merged before this, so I guess it's all good :)

ceolson01:
Yes, good call @rdig. I'll add that as a blocker 👍

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -125,7 +113,7 @@ const ColonyHome = ({
   intl: { formatMessage },
 }: Props) => {
   const [filterOption, setFilterOption] = useState(TasksFilterOptions.ALL_OPEN);
-  const [filteredDomainId, setFilteredDomainId] = useState(0);
+  const [filteredDomainId, setFilteredDomainId] = useState(1);
```
sprusr: Should it be all or root selected by default? Was there a product decision?

rdig: But isn't `root` being selected show all tasks in all it's child domains? Meaning all tasks are displayed regardless ?

sprusr: So actually selecting domain `1` will show only tasks from the root domain, and not children. You have to select `0` to show all tasks.

src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonyMeta.tsx
```diff
@@ -146,9 +142,9 @@ const ColonyMeta = ({
           <li>
             <Button
               className={getActiveDomainFilterClass(0, filteredDomainId)}
-              onClick={() => setFilteredDomainId()}
+              onClick={() => setFilteredDomainId(1)}
             >
-              <FormattedMessage {...MSG.allDomains} />
+              <FormattedMessage {...{ id: 'domain.root' }} />
```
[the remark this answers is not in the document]
ceolson01: 🤦‍♂

No clue why I did that, nice catch!

ceolson01:
@sprusr might be good to give this another quick look - some changes after rebase, including using `0` for `root` domain.

This `0` & `1` `domainId` business is getting tedious 😅

src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
```diff
@@ -0,0 +1,97 @@
...
+    () =>
+      openDialog('TokensMoveDialog', {
+        colonyAddress,
+        toDomain: currentDomainId,
+      }),
+    [openDialog, colonyAddress, currentDomainId],
+  );
+
+  // hide for root domain
+  return currentDomainId === 0 ? null : (
```
sprusr: Hm yeah this is actually kind of confusing behaviour now that I think about it more, but it's what was agreed with product at least for now. I think the whole selecting of domain + hierarchy + "all domains" behaviour needs another look from dev + product point of view.

PR #1869, opened 2019-09-24
Fix table hover issue

### LINKED ISSUE #1840: User permissions table UI broken on hover
As the title says, when hovering the user permissions table, some parts of the row aren't highlighted:

Screenshot from 2019-09-17 17-57-39

While at it, this should also fix the domains table, who also suffers from this issue, but in a different way:

Screenshot from 2019-09-17 17-59-01

ceolson01:
## Description

This PR fixes the table row hover issue, and ensures the entire background of the row is changed.

**Changes** 🏗

* Target both `td`s and the entire `tr` of the hovered instance.

**Demo** 🎥 

Table Hover

Closes #1840

### DIFF
--- a/src/modules/admin/components/Permissions/Permissions.css
+++ b/src/modules/admin/components/Permissions/Permissions.css
@@ -23,10 +23,6 @@
   padding: 10px 0;
 }
 
-.userPermissionsCell {
-  margin-left: auto;
-}
-
 .main table .tableBody {
   overflow-y: auto;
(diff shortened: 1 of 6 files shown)


ceolson01:
Wow, great eye!

I've amended the commit. I think my solution works - it's a bit cheeky, but it seems to work in all of our current cases (I can't think of a reason the logic wouldn't work in future cases, too), and is easy to override. 

Good for another look!

PR #1872, opened 2019-09-26
Update environment variables examples

chmanie:
## Description

This PR updates the `.env.example` file to account for all environment variables that are needed to run the Dapp locally.

**Changes** 🏗

* Add `PINNING_ROOM` env variable
* Add some more comments

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,18 +1,26 @@
-# local, rinkeby, goerli, etc
+# rinkeby, trufflepig
+LOADER=trufflepig
+
+# local, goerli, mainnet
 NETWORK=local
+
+# chain id
+# local - 1337
+# goerli - 5
+# mainnet - 1
 CHAIN_ID=1337
 
-# trufflepig, network
-LOADER=network
+# pinning service pubsub room
+PINNING_ROOM=PINION_DEV_ROOM
 
-# .eth for main/devnet, .test for testnet
-COLONY_NETWORK_ENS_NAME=joincolony.eth
-
-SKIP_GAS_STATION_CONFIRM=true
+# verbose logger
 VERBOSE=true
 
-# API keys for development
+# ethplorer key
 ETHPLORER_API_KEY=freekey
 
-# Infura ID
-INFURA_ID=blahblah
+# ENS suffix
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -1,18 +1,26 @@
...
-SKIP_GAS_STATION_CONFIRM=true
+# verbose logger
 VERBOSE=true
 
-# API keys for development
+# ethplorer key
 ETHPLORER_API_KEY=freekey
 
-# Infura ID
-INFURA_ID=blahblah
```
sprusr: Where did this go? It's not required, but useful. Maybe should be in here but commented out by default?

.env.example
```diff
@@ -1,18 +1,26 @@
...
+# ENS suffix
+COLONY_NETWORK_ENS_NAME=joincolony.eth
```
sprusr: Could benefit from similar explanation to `CHAIN_ID`. `.eth` is used for local and mainnet, `.test` used for goerli.

.env.example
chmanie: OK!

.env.example
chmanie: Will add

chmanie:
@sprusr better?

.env.example
```diff
@@ -1,18 +1,31 @@
...
+# verbose logger
 VERBOSE=true
 
-# API keys for development
+# ethplorer key (use `freekey` for rate limited dev trial)
 ETHPLORER_API_KEY=freekey
 
-# Infura ID
-INFURA_ID=blahblah
+# Uncomment to be rate limited with INFURA
```
sprusr: Haha this comment! Well either way they don't enforce the rate limit. But just to clarify, if you don't set this in `.env` it uses our default key from colonyJS

chmanie: We have a default key in colonyJS?

sprusr: Sure do! https://github.com/JoinColony/colonyJS/blob/develop/packages/colony-js-client/src/lib/InfuraProvider/index.js#L8

It's our "public" key, but if Infura start enforcing the daily request limits, then we're probably going to want a private one for the dApp.

PR #1874, opened 2019-09-26
Transaction List long values wrapping cell

### LINKED ISSUE #1741: Styling - Value breaks into two rows on transaction list
## Steps to reproduce

1. Create a new colony
2. Mint new tokens
3. Claim them

## Expected behaviour

Display value in a single line, push content to the left if needed, if other hand content breaks something use css animation to decrease opacity or add slide in animation

Also: improve how address are rendered - se 0x01000000000 ... 

## Actual behaviour

On hover, values break into two rows, it doesn't look nice

bug-1
bug-2

Clipboard01

## Suggestions (optional)

Use css animations to push content and hide if needed.
For "From" value use shortened path, f.e. 0x1234 ... 12345 (use global pattern)
It has fairly low priority.

rdig:
## Description

This PR attempts a fix for a issue with the Colony's Transaction table, where, if a transaction amount was very long, when hovering _(and the etherscan button would render)_ it would be to much for the width of that cell and wrap the contents onto a second line.

The only _"sane"_ way to fix this that I found was to make the Etherscan button `absolute` as to render above the value itself _(as that can grow indefinitely)_

As I've mentioned above, this value can grow indefinitely, so while my attempt fixes this table for the vast amount of cases, it will still break in edge cases.

But I guess that's what we have to live with for now.

Ie: 
```
123(...)n.123123123123123123
```

**Changes** 
- [x] `TransactionListItem` render etherscan button above transaction value

**Screenshot**:

Before:
Screenshot from 2019-09-26 15-37-46

After:
Screenshot from 2019-09-26 15-53-13

Resolves #1741

### DIFF
--- a/src/modules/core/components/TransactionList/TransactionListItem.css
+++ b/src/modules/core/components/TransactionList/TransactionListItem.css
@@ -2,6 +2,9 @@
 
 tr.main:hover .etherscanButtonWrapper {
   display: block;
+  position: absolute;
+  background-color: var(--colony-white);
+  padding: 10px 15px 10px 20px;
 }
 
 .main td {
@@ -48,7 +51,7 @@ td.transactionDetails {
 td.transactionAmountActions {
   flex-direction: row-reverse;
   padding-right: 30px;
-  width: 50%;
+  width: 60%;
 }
 
 .customButton {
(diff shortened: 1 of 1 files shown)


ceolson01:
Clever solution! Looks good 👍

PR #1878, opened 2019-09-27
Fix admin permissions hover

### LINKED ISSUE #1852: Hover state missing for admins without permission
## Steps to reproduce

1. Permissions tab
2. Open a user who you do not have permission to edit
3. No idea why this is happening

## Expected behaviour

* We have a few designs for hover states to show the user that they do not have permission. We can use this hover state for any role (i.e. "You do not have permission to set the administrative role.")

You do nothave permission

## Actual behaviour

* There is no hover state so when I go to use it, it feels like the modal is just frozen or broken. 

Colony

ceolson01:
## Description

This PR adds a tooltip (er... `Popover`, to be technical) when a user hovers over a disabled permission in the colony permissions edit dialog.

**New stuff** ✨

* `PermissionCheckbox` component

**Changes** 🏗

* Use `PermissionCheckbox` component in `ColonyPermissionsEditDialog`
* Improve clarity of some logic regarding state updates
* Use global i18n messages for buttons and some roles.

**Demo** 🎥 

Permissions Hover

Closes #1852

### DIFF
--- a/src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
+++ b/src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
@@ -267,9 +267,9 @@ const ColonyPermissionEditDialog = ({
 
   // When selected user gets updates get that user's roles
   // to populate the checkboxes
-  const useSelectedUserForRoles = (): Array<string> => {
-    const { data } = useUserDomainRoles(colonyAddress, domain.id, selectedUser);
+  const { data } = useUserDomainRoles(colonyAddress, domain.id, selectedUser);
 
+  useEffect(() => {
     // Avoid too many rerenders when no new data has loaded with the following condition
     if (
       selectedRoles &&
@@ -280,9 +280,7 @@ const ColonyPermissionEditDialog = ({
 
       setUserRoles(getRoles(data));
     }
-    return [];
-  };
-  useSelectedUserForRoles();
+  }, [data, selectedRoles, selectedUser]);
 
   // Set user whose roles should be edited
   const {
--- a/src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
+++ b/src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
@@ -59,33 +59,6 @@ const MSG = defineMessages({
     id: 'core.ColonyPermissionEditDialog.permissionsLabel',
     defaultMessage: 'Permissions',
   },
-  roleDescription0: {
-    id: 'core.ColonyPermissionEditDialog.roleDescription0',
-    defaultMessage:
-      'The highest permission, control all aspects of running a colony.',
-  },
-  roleDescription1: {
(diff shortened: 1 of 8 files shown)


chmanie:
I should base my 1850 on this

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -267,9 +236,9 @@ const ColonyPermissionEditDialog = ({
 
   // When selected user gets updates get that user's roles
   // to populate the checkboxes
-  const useSelectedUserForRoles = (): Array<string> => {
-    const { data } = useUserDomainRoles(colonyAddress, domain.id, selectedUser);
+  const { data } = useUserDomainRoles(colonyAddress, domain.id, selectedUser);
 
+  useEffect(() => {
```
sprusr: Nice improvement, definitely wanted doing 👍

src/modules/admin/components/Permissions/ColonyPermissionEditDialog.tsx
```diff
@@ -340,34 +307,25 @@ const ColonyPermissionEditDialog = ({
...
-                    <span className={styles.permissionChoiceDescription}>
-                      <Heading
-                        text={MSG[`role${id}`]}
-                        appearance={{ size: 'small', margin: 'none' }}
-                      />
-                      <FormattedMessage {...MSG[`roleDescription${id}`]} />
-                    </span>
-                  </Checkbox>
+                    role={role}
+                    roleDescription={MSG[`roleDescription${idx}`]}
```
sprusr: How come move the message handling for the role name to `PermissionCheckbox` but not for the description?

ceolson01: So the role names are actually in the `en/i18n.json` file, but `PermissionCheckbox` does indeed have the _logic_ to piece together the message id strings.

That said, perhaps it would make sense to move the description messages into `PermissionCheckbox` and use the role name to access each message... I think I'll do that

ceolson01:
@sprusr I moved the role description messages into the `PermissionCheckbox` component, and also added styles to use the `pointer` cursor for checkboxes.

Before I merge this, could you please check out those last couple of changes?

src/modules/admin/components/Permissions/PermissionCheckbox.tsx
```diff
@@ -33,9 +60,11 @@ const PermissionCheckbox = ({
   disabled,
   intl: { formatMessage },
   role,
-  roleDescription,
 }: Props) => {
   const roleNameMessage = { id: `role.${role.toLowerCase()}` };
+  const roleDescriptionMessage = MSG[
```
sprusr: Could this be inside the `useMemo`?

ceolson01: Yessir! Good catch, thank you 🙏

PR #1881, opened 2019-09-30
Fix TaskList height on Dashboard and ColonyHome

### LINKED ISSUE #1570: Long task list breaks UI consistency
## Steps to reproduce

1. You have a lot of tasks in your Colony Home tasks list (more then fit the vertical height of your screen)

## Expected behaviour

- Just the task list should scroll

## Actual behaviour

- Whole page scrolls

## Demo

demo-lotsa-tasks

sprusr:
## Description

The `TaskList` now scrolls when it would otherwise cause the page to scroll. This is done in `UserTasks` and` ColonyTasks`.

Some of the CSS for these components is quite overwhelming, but I hope I haven't done anything crazy!

**Changes** 🏗

* `UserTasks` has a max height
* `ColonyTasks` has a max height

Resolves #1570

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyTasks/ColonyTasks.css
+++ b/src/modules/dashboard/components/ColonyHome/ColonyTasks/ColonyTasks.css
@@ -29,3 +29,8 @@
 .noTaskAddition {
   margin-top: 20px;
 }
+
+.taskList {
+  display: flex;
+  max-height: calc(100vh - 350px);
+}
--- a/src/modules/dashboard/components/ColonyHome/ColonyTasks/ColonyTasks.css
+++ b/src/modules/dashboard/components/ColonyHome/ColonyTasks/ColonyTasks.css
@@ -32,5 +32,6 @@
 
 .taskList {
   display: flex;
+  min-height: 88px;
   max-height: calc(100vh - 350px);
 }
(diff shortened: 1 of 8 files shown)


src/modules/dashboard/components/ColonyHome/ColonyTasks/ColonyTasks.css
```diff
@@ -29,3 +29,8 @@
 .noTaskAddition {
   margin-top: 20px;
 }
+
+.taskList {
+  display: flex;
+  max-height: calc(100vh - 350px);
```
ceolson01: This one produces the following result:

Task List

src/modules/dashboard/components/Dashboard/UserTasks.css
```diff
@@ -7,3 +7,13 @@
...
 }
+
+.filter {
+  display: flex;
+  justify-content: flex-end;
+}
+
+.taskList {
+  display: flex;
+  max-height: calc(100vh - 450px);
```
ceolson01: This one produces the following result:

Task List 2

sprusr:
@ceolson01 I've added a `min-height` equal to the height of one task list item. The problem was just that the screen height was too small to show any list. Now it will at least keep its height of a single item and cause the page to scroll.

PR #1882, opened 2019-09-30
Github integration UI

karolinkas:
## Description

This is the UI work from the github hackaton we had. 
I have agreed with Chris that I'll make a PR so the work doesn't get lost.


rdig:
@chmanie if you want I can take over this

rdig:
@chmanie  Also, I think it's a good idea to create a feature branch _(protected)_ for the Github Integration Project, so that when ready, this, can be merged into it.

PR #1896, opened 2019-10-04
Replace the use of blocksBack with fromBlock=1

### LINKED ISSUE #1893: Founder/root user unable to move funds
## Steps to reproduce

1. Log into a colony as the founder / user with all the permissions
2. Navigate to `admin > tokens`

## Expected behaviour

Should be able to move funds

## Actual behaviour

Can't move tokens
[video of this issue in action](https://drive.google.com/file/d/1mG21c5OAuChfiCzTA0OWTUbEUZOiGY5Q/view?usp=sharing)

sprusr:
## Description

We were previously only fetching event logs up to a certain number of blocks ago. This was due to performance reasons and nodes taking a long time to respond if we include all blocks.

The way in which nodes index events has apparently been improved now (at least for Infura), and so we can query for all blocks! This fixed weird things happening for colonies older than the blocksBack value.

Probably best to test this on goerli or mainnet (otherwise you won't see any potential performance hits).

**Changes** 🏗

* Any instance of `blocksBack` replaced with `fromBlock: 1` in the raw log filter

Resolves #1893

### DIFF
--- a/src/modules/users/data/queries.ts
+++ b/src/modules/users/data/queries.ts
@@ -305,7 +305,6 @@ export const getUsername: Query<
 
 /**
  * This query gets all tokens sent from or received to this wallet since a certain block time
- * @todo Use a meaningful value for `blocksBack` when getting past transactions.
  */
 export const getUserColonyTransactions: Query<
   ColonyClient,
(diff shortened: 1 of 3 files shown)


src/modules/users/data/queries.ts
```diff
@@ -367,13 +367,12 @@ export const getUserColonyTransactions: Query<
       events: { Transfer },
```
rdig: There's a comment on line `350` that references `blocksBack`, but it's all good :)

https://github.com/JoinColony/colonyDapp/blob/33029846ca73bfe616c1330663e0132c73ba3afb/src/modules/users/data/queries.ts#L348-L351

sprusr: Ah nice catch! I've removed that

sprusr:
@rdig trying on mainnet is a good call, I just tried it locally connected to Infura mainnet and all is working 🙌

PR #1905, opened 2019-10-28
Add `IconButton` component

### LINKED ISSUE #1795: Add icon to a button whenever gas station is called out
## Steps to reproduce

1. Create a task.
2. Fill in details such as assignee and funding. Click confirm. 

## Expected behaviour

Whenever button call out gas station it should be easy to distinguish from other buttons. 
Users would like to know when to expect a transaction fee. This is **global** change. 

We suggest to use wallet icon as it is on design:
https://www.figma.com/file/DjVCPq7LpkjEMEfQZtKgBdC1/Med-Fi-Style-Guide?node-id=0%3A1
artboard: buttons / to call out gas station - please check distance, margins and so on Figma "code" tab

image

## Actual behaviour

There is no distinction currently, all buttons look the same. 

## Suggestions (optional)

1. Use the current wallet icon, just add white background. 
2. Note, button with icon should appear always when gas station is called out. The purpose is to inform when to expect transaction. Do not include icon inside a gas station popover. 
3. Consider to add extra style to developers style guide for a large button with an icon.

rdig:
## Description

This PR adds in a new core button component, `IconButton` that is to be used when a user needs to sign any transaction that requires actual wallet interaction.

**New stuff** 

- [x] Added core `IconButton` component
- [x] Added `IconButton` style guide entry

**Changes**

- [x] `GasStationPrice` now uses `IconButton` for confirmations

**Screenshot**

Screenshot from 2019-10-30 17-09-13

Resolves #1795

### DIFF
--- a/src/modules/core/components/Button/IconButton.md
+++ b/src/modules/core/components/Button/IconButton.md
@@ -1,0 +1,60 @@
+
+The icon(ed) button is a twist on the _classic_ `<Button>` where in front of the button's normal text you can now render an icon from the available icons library.
+
+The use case this was created for was to provide the user with a UX confirmation that the action that's about to be done requires a contract interaction / signing.
+
+```js
+import { defineMessages } from 'react-intl';
+
+const MSG = defineMessages({
+  buttonText: {
+    id: 'styleguide.buttonText',
+    defaultMessage: 'Click Me',
+  },
+});
+```
+
+```js
+<IconButton text={buttonText} />
+```
+
+### Icons
+
+All icons available to the `<Icon />` component are also supported here.
+
+```js
+<IconButton text={buttonText} icon="circle-plus" />
+```
+
+### Themes
+
+This component supports all the themes the `<Button />` component, on which is based on, supports.
+
+```js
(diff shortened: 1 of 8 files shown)


src/modules/core/components/Button/IconButton.md
```diff
@@ -0,0 +1,60 @@
+
+The icon(ed) button is a twist on the _classic_ `<Button>` where in front of the button's normal text you can now render an icon from the available icons library.
```
ceolson01: Ayyy styleguide! 😍

src/modules/core/components/Button/IconButton.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const IconButton = ({
+  appearance = { theme: 'primary', size: 'medium' },
+  icon = 'wallet',
+  text,
+  textValues,
+  /*
+   * @NOTE Prevent children from being passed both to this component, or to
+   * the underlying `<Button />
+   */
```
ceolson01: 👍

PR #1912, opened 2019-11-04
Add tooltip for insufficiently funded domains in task domains list

### LINKED ISSUE #1884: Show text on hover for insufficiently funded domain (Task Screen)
### Story

As a *user funding a task **then** setting the task's domain*, I want to *see text on hover of a disabled domain* (whose pot has insufficient funds), so that *I know I should move funds to that domain's pot*.

### Description

When selecting a domain for a task (which already has funding set), we check that each domain has enough funds in its pot to be able to fund the given task. For domains that **do not** have enough funds, they're disabled (greyed out in this case). Unfortunately, there's no indication as to **why** a particular domain is disabled. Let's fix this!

Here's how it currently looks:

can't choose domain

- [ ] Hovering over a disabled domain (or any type of list item) opens a tooltip
- [ ] The tooltip contains a message: `Insufficient funds`
- [ ] The tooltip should have a black background with white text (use our already-established dark theme)
- [ ] The tooltip should be placed below the actively hovered domain

### Suggestions

It's probably a good idea to require that any disabled list item has a tooltip message (we're dealing with core components here). This will better force us to keep our users in-the-know.

### LINKED ISSUE #1888: Cleanup `Popover` placement styles
&#x60;react-popper&#x60; uses the &#x60;placement&#x60; prop, while our styles uses &#x60;appearance.placement&#x60;. Let&#x27;s consolidate this into one or the other.

---

https://github.com/JoinColony/colonyDapp/blob/2b4d11e2b1b7d074c903ff96661d25edfd767c83/src/modules/admin/components/Permissions/PermissionCheckbox.tsx#L95-L100

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `@todo` comment in 2b4d11e2b1b7d074c903ff96661d25edfd767c83 when #1878 was merged. cc @JoinColony.

ceolson01:
## Description

This PR adds a tooltip on `hover` in the `ItemsList`, specifically for the case of insufficiently funded domains when assigning a task to a particular domain. If the task has funding set already, and a domain in the list doesn't have the funding required, the tooltip will display on hover.

**Changes** 🏗

* Add `disabledText` to `ConsumableItem` for the `ItemsList`
* Add message descriptor for insufficiently funded domain
* Use the above ☝️
* Add `placement` modifiers (`*-start` and `*-end`) for popovers
* Improved popover and arrow placement

## TODO

- [x] Make tooltip disappear when necessary (currently, once hovered, tooltip remains open all the time). I think this has to do with the fact that the `Button` that the popper `ref`s is `disabled`, and react doesn't call leave events for disabled buttons.

See: https://github.com/facebook/react/issues/4251

Resolves #1884 
Resolves #1888


src/modules/core/components/ItemsList/ItemsList.tsx
```diff
@@ -232,15 +248,23 @@ class ItemsList extends Component<Props, State> {
...
-            title={decoratedName}
-          >
-            {decoratedName}
-          </button>
+          <Tooltip content={tooltipContent} placement="bottom">
+            {/*
+             * Must use a `div` here, as `mouseleave` event isn't fired on buttons that are `disabled`.
+             * See: https://github.com/facebook/react/issues/4251
+             */}
+            <div>
```
rdig: As an optional improvement _(again, optional, so up to you...)_ we can make this `div` element be displayed as `inline-block`, so that the tooltip appears centered to the actual domain name:

Screenshot from 2019-11-08 14-57-21

ceolson01:
@rdig this is ready for another look.

You made this wonderful suggestion:

> As an optional improvement (again, optional, so up to you...) we can make this div element be displayed as inline-block, so that the tooltip appears centered to the actual domain name

Unfortunately, this meant the entire row wouldn't be clickable (nor would the entire row open a tooltip) - a user would have to click directly on the text to trigger the click event.

Thus, I've expanded our `Popover` component to support placement modifiers (`*-start` and `*-end`) for added flexibility and precision. These modifiers match that of the `react-popper` API, and, after a bit of wrestling with our arrow styles, seem to integrate quite nicely.

These modifiers behave just like flex box directives:
- `*-start` modifier positions the popover (and arrow) to the left or top, given a horizontal or vertical placement, respectively.
- `*-end` modifier positions the popover (and arrow) to the right or bottom, given a horizontal or vertical placement, respectively.

The below example utilizes `bottom-start`. Additionally, it ensures the entire width of each `ItemsList` item is still clickable (for the `!disabled` ones, of course).
 
Screen Shot 2019-11-12 at 4 28 29 PM


#### Note

`auto-start` and `auto-end` work just like `auto` in that they'll supply the appropriate placement value to the popper, but have the added benefit of carrying the modifier suffix over to the resulting placement value.

i.e. `auto-start` -> `left-start` or `bottom-start`

PR #1916, opened 2019-11-08
More DLx Fixes!

ceolson01:
## Description

This PR fixes more DLx issues discovered following deployment of #1907 to QA, including broader improvements to domain-level permissions.

**New stuff** ✨

* Migrations for task events to add root domain
* `getAllUserRoles` transformer
* `canArchitect` check

**Changes** 🏗

* Adjust task permissions to allow admins
* Use `domainId`s everywhere in commands & sagas
* Use current `domainId` in task access controller

## TODO

Colony Admin Permissions
- [x] Fix asterisks for inherited roles OR remove asterisk text at bottom of list / @rdig 
- [x] ~SingleUserPicker only works (on first click) if the selected user is already in the redux store. Doesn’t work when pasting an (unloaded) user’s address and selecting it. Even still, the form won’t pass validation, thus can’t submit~ / @rdig 
- [x] Add user permissions on a subdomain does nothing after signing TXs / @chmanie 
- [x] User with only the architecture domain cannot see the admin panel, and thus cannot modify domains. We may have to rethink the architecture of the colony admin area because of the large number of user role permutations @chmanie 

Tasks
- [x] ~Finalizing a task in a subdomain pays with funds from the root domain~ This should be an own issue as it requires colonyJS to be altered as well, while probably fixing a whole bunch of permissioning issues on the way
- [x] Can request to work on a task that's been discarded
- [x] Actually use `makePaymentFundedFromDomain`
- [x] Use inherited roles to check task permissions in access controller

### DIFF
--- a/src/data/permissions/task.ts
+++ b/src/data/permissions/task.ts
@@ -1,37 +1,12 @@
-import {
-  COLONY_ROLE_ADMINISTRATION,
-  COLONY_ROLE_ROOT,
-} from '@colony/colony-js-client';
-
-import { ROOT_DOMAIN } from '~constants';
-import { EventTypes } from '~data/constants';
 import { TaskEvents } from '~data/types/TaskEvents';
-import { Address, ColonyClient, PermissionsManifest } from '~types/index';
-import { makeUserHasRoleFn } from './utils';
+import { Address, PermissionsManifest } from '~types/index';
 
-export default function loadModule(
-  colonyClient: ColonyClient,
-): PermissionsManifest<{
+export default function loadModule(): PermissionsManifest<{
   domainId?: number;
   colonyAddress?: Address;
   event?: TaskEvents;
   heads?: object[];
 }> {
-  const isAdmin = makeUserHasRoleFn(colonyClient, COLONY_ROLE_ADMINISTRATION);
-  const isFounder = makeUserHasRoleFn(colonyClient, COLONY_ROLE_ROOT);
-
-  const isAdminOrFounder = async (
-    userAddress: Address,
-    domainId: number,
-  ): Promise<boolean> => {
-    // I don't know why this can be string or number? Maybe because we're dealing with the raw (not migrated) event here.
-    // Should we apply migrations here as well?
-    const hasAdminRole = await isAdmin(userAddress, domainId);
-    if (hasAdminRole) return hasAdminRole;
-
-    return isFounder(userAddress, ROOT_DOMAIN);
-  };
-
   return {
     COMMENT_STORE_CREATED: { inherits: 'create-task' },
     DUE_DATE_SET: { inherits: 'set-task-due-date' },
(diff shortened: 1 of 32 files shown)


src/data/permissions/task.ts
```diff
@@ -49,24 +24,26 @@ export default function loadModule(
...
-        ? isAdminOrFounder(userAddress, event.payload.domainId)
-        : false;
-    },
-    'set-task-due-date': { inherits: 'is-founder-or-admin' },
-    'set-task-payout': { inherits: 'is-founder-or-admin' },
-    'set-task-skill': { inherits: 'is-founder-or-admin' },
-    'update-task': { inherits: 'is-founder-or-admin' },
+    'set-task-assignee': { inherits: 'is-admin' },
+    // disallow changing of the domain for now
+    'set-task-domain': async () => false,
```
ceolson01: I have a feeling this is going to frustrate a lot of folks. Is it worth exploring a solution with product team that makes it more obvious during task creation that the task will be "permanently" (at least for now) assigned to the currently selected domain? Perhaps that would require less effort than a long-term DLP solution?

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
```diff
@@ -172,15 +193,20 @@ const AdminDashboard = ({
...
   return (
     <div className={styles.main}>
       <VerticalNavigation
-        navigationItems={navigationItems(colony, domains, rootRoles)}
+        navigationItems={navigationItems(
+          colony,
+          domains,
+          rootUserRoles,
+          allUserRoles,
+        )}
```
ceolson01: Would it make sense to `useMemo` with this fn call?

src/modules/admin/components/Permissions/UserPermissions.tsx
```diff
@@ -40,12 +30,10 @@ const UserPermissions = ({ userAddress, domains, domainId }: Props) => {
 
   return (
     <div className={styles.main}>
-      {/* @TODO restore pending role indicator */}
```
ceolson01: I think this `todo` is still useful - the `pending` role indicator refers to the yellow dot that should be visible until a tx is signed / confirmed as a result of an optimistic UI (issue reference: https://github.com/JoinColony/colonyDapp/issues/1708)

src/data/permissions/task.ts
rdig: As as stop gap solution for this... how about?

We have a _New Task_ button, which, upon clicking, will open a dropdown/tooltip, with all the available domains for you to select from.

That will make it a tiny bit more obvious to the user the domain that task is going to be created in.

It's not the best solution, but it's better then what we have now.

Thoughts? @chmanie @ceolson01 and @JoinColony/product

src/modules/admin/components/Permissions/UserPermissions.tsx
chmanie: I'm in favour of just ditching that indicator for now as it is quite hard to do. And getting back to it once we have a more general solution to present the status of transactions (we don't do it anywhere else)

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
chmanie: I don't think so. It doesn't do a very heavy calculation. Just some conditions, that should be very fine.

src/data/permissions/task.ts
chmanie: I agree that this is not ideal, but we also want to be done with this at some point. I suggest to leave it be for now and figure out a better solution together with @JoinColony/product, then implement it later.

chmanie:
> I can't seem to create a task in a subdomain.

Can you expand on what situation that was exactly? What permissions did the user have?

ceolson01:
> Can you expand on what situation that was exactly? What permissions did the user have?

Sorry, I wasn't very specific there. It seems you've figured it out... but in case not, this user account had all permissions in the root domain.

src/data/permissions/task.ts
collinvine: Actually, that could be interesting. I could imagine we just add an ellipses to the button's copy so it's more clear that there are other options:

`New Task …` and that opens the menu with

`in root`
`in development`
`in product`

I'd be okay with that as an initial solution.

---

I'm also okay with leaving it as is for now and:

a. Letting users discover this, delete it, then recreate it in the correct domain
b. Letting users complain to us that it sucks so we can fix it then

PR #1932, opened 2019-11-26
Server Based Colonies

rdig:
## Description

This PR aims at adding, at least in the initial stages, the queries, mutations and types for the Apollo GraphQL server to be able to store Colony related data.

Alongside that, this branch will also touch Components, Actions, Action Creators and Saga, in order to make user of the above GraphQL data.

Rebased on `feature/graphql-generation`


chmanie:
> ## Re-add domains once we decide what we're going to do with the
> https://github.com/JoinColony/colonyDapp/blob/f9a35e94764e9732c31f1266ad28a1662b2d09cc/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx#L140-L145
> 
> ###### This comment was generated by [todo](https://todo.jasonet.co) based on a `TODO` comment in [f9a35e9](https://github.com/JoinColony/colonyDapp/commit/f9a35e94764e9732c31f1266ad28a1662b2d09cc) in #1932. cc @rdig .

There's no need to recover Colonies anymore (at least not an immediate need)

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
```diff
@@ -152,11 +161,12 @@ const AdminDashboard = ({
...
   );
 
-  const { error: colonyError, data: colony } = useDataSubscriber(
-    colonySubscriber,
-    [colonyAddress],
-    [colonyAddress],
-  );
+  const {
+    data: { colony } = {},
+    loading: colonyDataLoading,
```
chmanie: loading isn't really necessary here. We can just rely on the fact that there is data there.

src/modules/admin/components/AdminDashboard/AdminDashboard.tsx
```diff
@@ -152,11 +161,12 @@ const AdminDashboard = ({
     [colonyName],
   );
 
-  const { error: colonyError, data: colony } = useDataSubscriber(
-    colonySubscriber,
-    [colonyAddress],
-    [colonyAddress],
-  );
+  const {
+    data: { colony } = {},
```
chmanie: Please do not use defaults here, just do a check whether `data` is defined and if not, show a loading spinner

src/modules/admin/components/Profile/ProfileEdit.tsx
```diff
@@ -78,12 +99,19 @@ const ProfileEdit = ({ colony }: Props) => {
...
     website,
   } = colony;
-  const transform = useCallback(
-    pipe(
-      withKey(colonyAddress),
-      mergePayload({ colonyAddress }),
-    ),
-    [colonyAddress],
+
+  const [editColony] = useMutation(EditColonyProfileDocument);
```
chmanie: There should be a hook in the generated file (`useEditColonyProfileMutation`, I think) which you can immediately use. Furthermore you can define the parameters directly in the mutation and you don't need to use `useCallback`

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -119,7 +112,11 @@ const ColonyHome = ({
     COLONY_TOTAL_BALANCE_DOMAIN_ID,
   );
   const [isTaskBeingCreated, setIsTaskBeingCreated] = useState(false);
-  const [showRecoverOption, setRecoverOption] = useState(false);
+  /*
+   * @TODO Re-add domains once we decide what we're going to do with the
```
chmanie: As stated in the other comment, let's just delete this and everything it's using (saga, command, etc)

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -138,66 +135,79 @@ const ColonyHome = ({
...
     [colonyName],
   );
 
-  const colonyArgs: [Address | undefined] = [colonyAddress || undefined];
-  const {
-    data: colony,
-    isFetching: isFetchingColony,
-    error: colonyError,
-  } = useDataSubscriber(colonySubscriber, colonyArgs, colonyArgs);
+  const { data: { colony } = {}, loading: colonyDataLoading } = useColonyQuery({
```
chmanie: I am honestly wondering why this even works. The colony address in this query should not be optional (and apparently isn't in the query). You should use a combination of `useEffect` and useColonyLazyQuery (https://github.com/JoinColony/colonyDapp/pull/1932/files#diff-dc754e48e7cb7cefc8a267996180cd40R957) here:

```
  const [loadColony, { data }] = useColonyLazyQuery();

  useEffect(() => {
    if (colonyAddress) {
      loadColony({
        variables: { address: colonyAddress },
      });
    }
  }, [loadColony, colonyAddress]);
```
I know this doesn't look the nicest but it's the safest way for now to ensure the query is using the correct address in cases where it's not defined initially (or can change at some point).

Please also do not use defaults but rather use a loading spinner.

src/modules/admin/components/Profile/ProfileEdit.tsx
rdig: This won't work in this case, as you still need the values from the form. _(ie: `profile` in this case)_

And as far as I can see, you can't specify a custom method to `useEditColonyProfileMutation`, just the values to be passed into the mutation

This is the exact same way you are editing the user profile:

https://github.com/JoinColony/colonyDapp/blob/712d0d025cba3a5095f14b32961f267690396c80/src/modules/users/components/UserProfileEdit/UserProfileEdit.tsx#L79-L83

But maybe I'm missing something... ?

chmanie: Oh yeah, sorry you're right, scratch the last part, you'll need `useCallback`. But you _can_ use the mutation hook (and I should, too, I just forgot about it in the edit user case)

chmanie: I updated it to use the generated code. See here: https://github.com/JoinColony/colonyDapp/blob/feature/user-metadata-mutations/src/modules/users/components/UserProfileEdit/UserProfileEdit.tsx#L77

rdig: Yup, just saw your commits.

I've updated as well.

rdig:
@chmanie 

- Integrated token data in the colony creation mutation
- Integrated the fixes you suggested
- Removed everything related to colony profile metadata recovery
- Added / integrated the `AnyColony` query return type

You can have a look / merge it when you like :+1:

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -210,46 +182,40 @@ const ColonyHome = ({
     [colonyAddress, filteredDomainId],
   );
 
-  if (colonyError || addressError) {
+  if (!colonyName || addressError) {
```
chmanie: 👍

src/modules/admin/components/Profile/ColonyAvatarUploader.tsx
```diff
@@ -49,8 +37,14 @@ const ColonyAvatarUploader = ({ colony: { colonyAddress }, colony }: Props) => {
     ),
     [colonyAddress],
   );
-  const upload = useAsyncFunction({ ...uploadActions, transform }) as any;
-  const remove = useAsyncFunction({ ...removeActions, transform }) as any;
+  const upload = useAsyncFunction({
```
chmanie: We still need the success and error actions here, to have the async function complete

src/modules/dashboard/sagas/colony.ts
```diff
@@ -211,29 +163,17 @@ function* colonyAvatarUpload({
...
-      type: ActionTypes.COLONY_AVATAR_UPLOAD_SUCCESS,
-      meta,
-      payload: { hash: ipfsHash },
+    yield apolloClient.mutate({
+      mutation: EditColonyProfileDocument,
+      variables: { input: { colonyAddress, avatarHash: ipfsHash } },
     });
   } catch (error) {
-    return yield putError(ActionTypes.COLONY_AVATAR_UPLOAD_ERROR, error, meta);
+    return yield putError(error.message, error, meta);
```
chmanie: We still need the UPLOAD_SUCCESS and UPLOAD_ERROR actions here for the async function creator to work

src/modules/dashboard/sagas/colony.ts
```diff
@@ -243,28 +183,15 @@ function* colonyAvatarRemove({
...
-      payload: undefined,
+    const apolloClient: ApolloClient<any> = yield getContext(
+      Context.APOLLO_CLIENT,
+    );
+    yield apolloClient.mutate({
+      mutation: EditColonyProfileDocument,
+      variables: { input: { colonyAddress, avatarHash: null } },
     });
   } catch (error) {
-    return yield putError(ActionTypes.COLONY_AVATAR_REMOVE_ERROR, error, meta);
```
chmanie: Same as above

src/data/mutations.graphql
```diff
@@ -23,3 +23,22 @@ mutation EditUser($input: EditUserInput!) {
...
+    avatarHash
+    description
+    displayName
+    guideline
+    website
+  }
+}
+
+mutation EditColonyProfile($input: EditColonyProfileInput!) {
+  editColonyProfile(input: $input) {
```
chmanie: Can you return all the fields this mutation can possibly change (alongside the `id`)? Then the Apollo cache will automatically pick it up.

So that's probably `avatarHash`, `description`, `displayName`, `guideline`, `website`

rdig:
@chmanie 
- Reverted the `COLONY_AVATAR_UPLOAD_SUCCESS` and `COLONY_AVATAR_UPLOAD_ERROR` actions logic
- Fixed the `editColonyProfile` mutation so that it's returning the whole colony profile props

This is ready to merge when you want to

PR #1940, opened 2019-12-11
Remove colony subscribers

chmanie:
## Description

This PR gets rid of all colony subscribers and also takes care of some typing issues around colonies. In some cases it's enough to get a colony purely from the database (colony avatar, colony grid, etc). In other cases we need more information that only is available on the blockchain. For that we use the good old `colonyFetcher` which also replaces the `colonySubscriber` in these situations.


src/modules/dashboard/components/ColonyGrid/ColonyGridItem.tsx
```diff
@@ -1,54 +1,46 @@
...
-    colonySubscriber,
-    [colonyAddress],
-    [colonyAddress],
-  );
-  const { data: colonyName, isFetching: isFetchingColonyName } = useDataFetcher(
-    colonyNameFetcher,
-    [colonyAddress],
-    [colonyAddress],
-  );
+  const { data, loading } = useColonyQuery({
```
rdig: Shouldn't this use the `colonyFetcher` asa well ?

src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonyMeta.tsx
```diff
@@ -3,6 +3,8 @@ import { defineMessages, FormattedMessage } from 'react-intl';
 
 import { COLONY_TOTAL_BALANCE_DOMAIN_ID } from '~constants';
 import { DomainsMapType } from '~types/index';
+import { ColonyType } from '~immutable/index';
+import { AnyColony } from '~data/index';
```
rdig: Do we still need `AnyColony` here ?

src/modules/dashboard/components/HookedColonyAvatar/HookedColonyAvatar.ts
```diff
@@ -1,10 +1,9 @@
 import ColonyAvatar, { Props as ColonyAvatarProps } from '~core/ColonyAvatar';
-import { useDataFetcher, useDataSubscriber } from '~utils/hooks';
+import { useDataFetcher } from '~utils/hooks';
 import { withHooks } from '~utils/hoc';
-import { AnyColony } from '~data/index';
+import { AnyColony, useColonyQuery } from '~data/index';
```
rdig: Shoul `AnyColony` here be replaced back with `ColonyType` ?

src/modules/dashboard/sagas/colony.ts
```diff
@@ -175,6 +242,8 @@ function* colonyAvatarUpload({
 
     yield put<AllActions>({
       type: ActionTypes.COLONY_AVATAR_UPLOAD_SUCCESS,
+      meta,
```
rdig: Do we still need these (meta and payload)?

I removed them since I've also removed the reducer, so nothing is processing the data that's being passed from here

chmanie: `payload` not necessarily but I'd suggest to just leave it in for simplicity's sake. `meta` we do need as it's passed through to the async function generator

src/modules/dashboard/components/ColonyHome/ColonyMeta/ColonyMeta.tsx
chmanie: Yeah, as we use ColonyMeta in both cases, one where we have a "full colony" and one where we have just the colony from the db

src/modules/dashboard/components/HookedColonyAvatar/HookedColonyAvatar.ts
chmanie: No, in this case it's fine, we just need the colony from the DB

src/modules/dashboard/components/ColonyGrid/ColonyGridItem.tsx
chmanie: It's not always clear, that I am sorry for. But here we just need the information that's in the db

PR #1945, opened 2019-12-26
Server Based Notifications

rdig:
## Description

This PR wires in graph-ql based notifications _(via queries)_ and all the related interactions with them _(mark as read, via mutations)_

We also remove any traces that are left from the orbitdb user inbox stores.

**Screenshots**

Screenshot from 2019-12-26 05-06-56
Screenshot from 2019-12-26 15-18-15
Screenshot from 2019-12-26 15-39-23
Screenshot from 2019-12-26 15-40-05

Rebased on `feature/colony-gql-state`


src/data/generated.ts
```diff
@@ -757,9 +757,9 @@ export type Task = {
   dueDate?: Maybe<Scalars['DateTime']>,
   finalizedAt?: Maybe<Scalars['DateTime']>,
   title?: Maybe<Scalars['String']>,
-  colony: Colony,
```
chmanie: The colony and the creator must not be undefined on a task, otherwise we'll run into a lot of trouble. I guess this is connected to this:
https://github.com/JoinColony/colonyDapp/pull/1945/files#diff-e9686509e232d1e2104f9c9c12b7ffcaR162

@rdig can you try to find a solution that fixes the @FIXME and doesn't require this change? Maybe we can use stub data.

rdig: I'm actually not sure where this change comes from (I'll look into it) but this is not related to my fixme

That fixme is actually correct, as for the first notification ever on every user, unless they also created a colony, there isn't a colony or a token set, so it doesn't make sense to wait for that data to load

src/data/constants.ts
```diff
@@ -55,7 +55,7 @@ export enum EventTypes {
   UNSUBSCRIBED_FROM_TASK = 'UNSUBSCRIBED_FROM_TASK',
   USER_AVATAR_REMOVED = 'USER_AVATAR_REMOVED',
   USER_AVATAR_UPLOADED = 'USER_AVATAR_UPLOADED',
-  USER_PROFILE_CREATED = 'USER_PROFILE_CREATED',
+  // USER_PROFILE_CREATED = 'USER_PROFILE_CREATED',
```
rdig: hmmm... why is this here ?

src/data/generated.ts
rdig: So as far as I could dig up, this is coming from the code generator and the way it decides which resolvers are optional (when building the types)

The closest thing I could find is this https://github.com/dotansimha/graphql-code-generator/issues/893#issuecomment-474924086 which suggest that in order to prevent this we need to declare all resolvers in the same file _(which wouldn't really work for us in the way the server code is set up currently)_

Let me know how you want to handle this, as I'm kinda stuck on this. Maybe try to pair and find a solution for this ?

chmanie: Oh this comes from the server. Yeah OK, we'll change it there

PR #1947, opened 2020-01-07
Notifications cleanup and fixes

rdig:
This PR adds in a couple of fixes, removes a `@FIXME` and some unused code from the server based notifications logic.

The gist of it:
- Disable on-click route for notifications _(for the initial launch. This also fixes a React DOM nesting violation)_
- Check if colony and tokens are in context before loading that data _(inside `InboxItem`)_
- Remove unused `decorateColonyEventPayload`
- Add a check for the domains array before attempting to load data from it _(inside `InboxItem`)_

### DIFF
--- a/src/modules/users/components/Inbox/InboxItem/InboxItem.tsx
+++ b/src/modules/users/components/Inbox/InboxItem/InboxItem.tsx
@@ -176,13 +176,8 @@ const InboxItem = ({
       <TableCell
         className={full ? styles.inboxRowCellFull : styles.inboxRowCellPopover}
       >
-        {/*
-         * @FIXME The first ever notification for every user (user profile claimed)
-         * does not have a colony name or a token, so in that case the spinner
-         * will always render even though it shouldn't at that point
-         */
-        // !colonyName ||
-        // !token ||
+        {(colonyAddress && !colonyName) ||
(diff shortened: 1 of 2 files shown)


src/modules/users/components/Inbox/InboxItem/InboxItem.tsx
```diff
@@ -189,7 +186,7 @@ const InboxItem = ({
             />
           </div>
         ) : (
-          <WithLink to={onClickRoute}>
+          <WithLink>
```
chmanie: If there's no link, can we remove this?

chmanie: Well, if it's not hurting anyone we can keep it as an implementation hint

rdig: I wouldn't, as we want this back at some point.

Plus it already has the logic to just render a `div` in case no route is provided:

https://github.com/JoinColony/colonyDapp/blob/ab02a52dc58b4ef1de6860088bc49dbdb0a71fcd/src/modules/users/components/Inbox/InboxItem/InboxItem.tsx#L73-L80

chmanie:
Shall we merge this?

PR #1955, opened 2020-01-16
Add suggestions tab & "create suggestion" form

ceolson01:
## Description

This PR adds the required components for suggestions, along with the create new suggestion form (sans mutations)

**New stuff** ✨

* `Suggestions` component
* `SuggestionsList` component
* `SuggestionCreate` component

**Changes** 🏗

* Add suggestions tab to `ColonyHome`

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
@@ -46,7 +46,11 @@ const MSG = defineMessages({
   },
 });
 
-type TabName = 'suggestions' | 'tasks' | 'transactions';
+enum TabName {
+  SuggestionsTab = 'suggestions',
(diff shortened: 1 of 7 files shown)


src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -41,6 +46,12 @@ const MSG = defineMessages({
   },
 });
 
+enum TabName {
+  SuggestionsTab = 'suggestions',
+  TasksTab = 'tasks',
+  TransactionsTab = 'transactions',
```
ceolson01: @chmanie These enum keys are a bit wordy (i.e. `SuggestionsTab` vs `Suggestions`) as I was experiencing a false-positive from the eslint `no-shadow` rule. Not really sure why that rule is applied to keys... but it is.

PR #1965, opened 2020-01-17
Feature: Suggestions

### LINKED ISSUE #1926: Suggestions Feature Epic

ceolson01:
## Description

This PR serves as the base branch for all work towards suggestions.

Closes #1926 (epic)


src/modules/dashboard/components/SuggestionCreate/SuggestionCreate.tsx
```diff
@@ -0,0 +1,97 @@
...
+const displayName = 'Dashboard.SuggestionCreate';
+
+const SuggestionCreate = ({ colonyAddress, domainId, openDialog }: Props) => {
+  const inputRef = useRef<HTMLInputElement>(null);
+
+  const [createSuggestion] = useCreateSuggestionMutation();
+
+  const handleSubmit = useCallback(
+    ({ title }: FormValues, { resetForm }: FormikHelpers<FormValues>) => {
+      openDialog('ConfirmDialog')
```
rdig: So @collinvine mentioned in Asana that we don't need a confirmation dialog.

Pressing "return" should just create the suggestion

See here: 
Screenshot from 2020-01-21 09-38-36
https://app.asana.com/0/1156749902660977/progress

So we should either remove it, or take this up with @JoinColony/product

ceolson01: Ah, somehow I missed that comment. Thanks for the update! I'm in favor of removing it - we can just submit by pressing `return` 🙂 

I'll remove it in a new PR 👍

ceolson01: Update: removed via #1984

PR #1966, opened 2020-01-17
Suggestions List

### LINKED ISSUE #1956: Suggestions List
### Story

As a user viewing a colony's "Suggestions" tab, I want to see a list of suggestions below the "new suggestion" form.

### Design

68651679-2b35d380-0528-11ea-882f-70273f5d6522

### This issue is complete when:

- [ ] Suggestions are filtered by domain (or all domains), as selected in the left-hand sidebar

### This issue does not include:

- Filtering the suggestions by status (#1923)
- Upvoting suggestions (#1922)
- Changing a suggestion status (#1921)

ceolson01:
## Description

This PR adds the suggestions list.

**New stuff** ✨

* `ListGroup` core component for easier-to-manage table-esque styles
  * `ListGroupItem` child component
* `SuggestionsListItem` component

**Changes** 🏗

* Add `creator` to `SuggestionFields` fragment
* Filter by domains
* Sort by `createdAt` (descending)

**Demo** 🎥 

Suggestions

Closes #1956


ceolson01:
@rdig Circle doesn't seem to like something about this... it fails trying to resolve a `SuggestionsListItem.css` import. Maybe you can spot my issue? 🤷‍♂

src/modules/core/components/ListGroup/ListGroup.tsx
```diff
@@ -0,0 +1,19 @@
...
+
+import { getMainClasses } from '~utils/css';
+
+import styles from './ListGroup.css';
+
+interface Props {
+  children: ReactNode;
+}
+
+const displayName = 'ListGroup';
```
rdig: Will the `ListGroup` eventually replace all other instances of lists? Ie: tasks, domains, permissions, etc...

As it seems to me these are basically a generalized list which can be filled with whatever content

src/modules/dashboard/components/SuggestionsListItem/SuggestionsListItem.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+const MSG = defineMessages({
+  byAuthorText: {
+    id: 'Dashboard.SuggestionsListItem.byAuthorText',
+    defaultMessage: 'by {creator}',
+  },
+});
+
+interface Props {
+  suggestion: OneSuggestion;
```
rdig: Why is this called "one" suggestion ? :)

I'm assuming it's referring to a singular suggestion, but then why not use the "suggestion/suggestions" pair ?

Yeah, yeah, I know... this is nit-picking... :)

src/modules/dashboard/components/Suggestions/Suggestions.tsx
```diff
@@ -15,7 +15,7 @@ const displayName = 'dashboard.Suggestions';
 const Suggestions = ({ colonyAddress, domainId }: Props) => (
   <div>
     <SuggestionCreate colonyAddress={colonyAddress} domainId={domainId} />
```
rdig: Not related to this PR, but I've noticed it, so I'll mentioned it.

There's no way to un-focus the "Suggest features" input field without causing a validation error.

demo-cannot-unfocus-suggestions

The only way to do it is to change tabs, but otherwise every place you click, the field will scream at you that it's not valid anymore.

I know this is a polish thing and not something major, but I think this can be solved via yup's validation. 

Granted, at some later point.

rdig: While on the "Create suggestion" subject, you can submit the suggestion title by pressing "Return", but you can't close the confirmation modal by pressing "Return".

So to create a new suggestion, you can't double-tap "return", you can only tap "return" then mouse-click "ok", which is a bit of an UX annoyance.

Screenshot from 2020-01-20 14-50-08

Again, this is very much a polishing stage thing, and no where related to this PR, but I'm thinking you can open a issue for it when you're do so we remember to take care of it.

src/modules/core/components/ListGroup/ListGroup.tsx
ceolson01: > Will the ListGroup eventually replace all other instances of lists? Ie: tasks, domains, permissions, etc...

I was imagining this to be the case - it's just much more simple to work with than a `table`, especially as we will (eventually) shift to support mobile. I think I'd like to develop it to be a little more robust first, though, and make this transition gradually.

src/modules/dashboard/components/SuggestionsListItem/SuggestionsListItem.tsx
ceolson01: `type Suggestion` **is** indeed created by the code generator, but I'm using `OneSuggestion` to describe  a single suggestion from a query, considering the fact that we might have multiple suggestion queries in the future.

It basically gives us the freedom to say something along the lines of:

```typescript
type OneSuggestion =
  | ColonySuggestionsQuery['colony']['suggestions'][number]
  | UserSuggestionsQuery['user']['suggestions'][number];
```

src/modules/dashboard/components/Suggestions/Suggestions.tsx
ceolson01: > There's no way to un-focus the "Suggest features" input field without causing a validation error.

Yes, this is driving me nuts 🙂

Perhaps we can blur the input before resetting the form? I'll look into this.

> So to create a new suggestion, you can't double-tap "return", you can only tap "return" then mouse-click "ok", which is a bit of an UX annoyance.

This is a really good point. I'll open an issue for this, as I think this could be a nice improvement everywhere it's used.

ceolson01: Update: I've fixed both of these:

1) Blur the input on success **before** resetting the form
2) When the confirm dialog opens, `OK` will be autofocused for easy (and obvious) submission

PR #1967, opened 2020-01-17
Use client as primary token source

### LINKED ISSUE #1959: Can't create colony with external native token: Token Data Not Found
## Steps to reproduce

1. Try creating a colony with a an external native token

## Expected behaviour

You should be able to progress through colony creation without issue

## Actual behaviour

When you enter the token address, you get this error message.
`Token data not found. Please type in token details`
image

If you continue on, you are able to sign the first transaction, but the interface does not register when the transaction is confirmed and progress to the next step.

### LINKED ISSUE #1971: Token name and symbol aren't displayed properly when creating a new colony
If you use an existing token when creating a colony, the symbol and name are not displayed properly in the preview

image (2)

image

### LINKED ISSUE #1978: DAI and other external tokens are not displayed in the dapp
## Steps to reproduce

1. Send an external token (one that was not created by a colony) to a colony

## Expected behaviour

A transaction with the token should show up in the transaction feed. Once it has been claimed, the token should show up in the colony's balance.

## Actual behaviour

DAI and other external tokens are not displayed in the dapp at all. They can't be claimed in incoming transactions and and existing balances are not displayed.

My guess is that this is also related to #1967.

chmanie:
## Description

This PR adds local resolvers for token data which query the token contracts directly instead of relying on the server. The colony and user tokens are still stored on the server but just as addresses. Furthermore we changed the way the tokens are added to colonies and users (via entering a token address in the edit tokens dialog).

**New stuff** ✨

* Token resolvers for users, colonies and payouts

**Changes** 🏗

* Refined token edit modals

Resolves #1959
Resolves #1971 
Resolves #1978


src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.tsx
```diff
@@ -190,17 +183,6 @@ const StepCreateToken = ({
                 help={MSG.helpTokenSymbol}
               />
             </div>
-            <div className={styles.inputFieldWrapper}>
-              <ActionFileUpload
-                name="tokenIcon"
```
rdig: If we removed the Token info from the native token creation screen, shouldn't we also remove it from the external token "creation" page?:

https://github.com/JoinColony/colonyDapp/blob/2f02621bbb7fda94b9d452b49647609660abc64e/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx#L187-L197

rdig: Also, I think we need to make @JoinColony/product aware of this change

src/data/token.ts
```diff
@@ -31,26 +34,97 @@ const getBalanceForTokenAndDomain = async (
...
+      ...ETHER_INFO,
+    };
+  }
+
+  const tokenClient = await colonyManager.getTokenClient(tokenAddress);
+  const chainData: TokenInfo = await tokenClient.getTokenInfo.call();
+
+  let serverDataResult;
+  try {
+    const { data } = await client.query({
```
rdig: This query always fails on my machine claiming it can't the `iconHash` prop on some `undefined` object:

Screenshot from 2020-01-22 17-09-53

src/modules/core/components/TokenEditDialog/TokenItem/TokenItem.tsx
```diff
@@ -0,0 +1,57 @@
...
+  token: AnyToken;
+}
+
+const TokenItem = ({ nativeTokenAddress, removeTokenFn, token }: Props) => {
+  const canRemoveToken =
+    nativeTokenAddress !== token.address && !tokenIsETH(token);
+  return (
+    <div className={styles.main}>
+      <div className={styles.tokenChoice}>
+        <TokenIcon token={token} name={token.name || undefined} />
```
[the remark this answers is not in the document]
chmanie: Yeah, I honestly have no idea what causes this. `$$typeof` comes from react so it should be OK

src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.tsx
chmanie: Yes, we should remove that, good catch!

src/data/token.ts
chmanie: Thanks, this is a good find!

src/data/token.ts
```diff
@@ -31,26 +34,97 @@ const getBalanceForTokenAndDomain = async (
...
+      query: TokenInfoDocument,
+      variables: { address: tokenAddress },
+    });
+    serverDataResult = data;
+  } catch (e) {
+    console.warn(`Token with address ${tokenAddress} not found on server`);
+  }
+
+  const serverData: TokenInfo = serverDataResult
```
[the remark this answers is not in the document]
chmanie: Fixed! Thanks!

PR #1974, opened 2020-01-20
Disable message signing when Commenting on a Task

### LINKED ISSUE #1954: Skip task comment signature step for centralized comments
I don't want to do the whole spec thing here.

As we a storing task comments in a central place now (with some sort of identity check), we can skip the signing of comments for now. Just send off the comment immediately without adding the signature to it.

rdig:
## Description

This PR disables _(just disables, the underlying message signing system is still there and perfectly functional)_ signing every single comment message before posting it on a task.

This makes the task comment system more fluid and easy to use.

**Changes** 

- [x] removed message signing from the `taskCommentAdd` saga

**Demo** 

demo-no-signature-comments

Resolves #1954


src/modules/dashboard/sagas/task.ts
```diff
@@ -196,10 +196,6 @@ function* taskCommentAdd({
   meta,
 }: Action<ActionTypes.TASK_COMMENT_ADD>) {
   try {
-    yield call(signMessage, 'taskComment', {
-      comment,
-      author,
-    });
```
ceolson01: I know there's a chance that we'll bring this feature back at some point, but I think we can remove this saga (for now at least), no? And just handle the mutation directly in the component?

chmanie: Yes! When we bring it back it won't be handled in a saga

rdig: Good point. I'll make the change

rdig:
@chmanie @ceolson01 

Refactored this:
- Removed the task add saga
- Moved the mutation directly in the component

But this introduced a bad, but expected behaviour in the task feed (where comments are displayed).

Because we currently use a poll interval and not a subscription, it takes time, from the time you submit the comment for it to actually show up. 

As you can see in the demo below:

demo-task-feed-poll-interval

Again, this is expected, and will be solved in the future when we introduce subscriptions on the server.

But until then I expect a number of users to complain about this. _(I guess we'll just have to deal with that)_

On the other hand, there's a trick we can use here: a visual one. Delay the the 'isSubmitting' flag by about 4 seconds, meaning it will "take longer" (in the eyes of the users) to submit a comment. And by the time the flag changes to `false`, the task feed would have already been updated.

But I would feel iffy implementing it like that to be honest...

Thoughts ? 

Also, please give it another review

rdig:
After taking @chmanie 's advice, I've added in a `refetchQueries` option to the task comment mutation, and now the comment system works properly again. 

Thanks for that Chris!

demo-comment-refreshed-feed

@ceolson01 @chmanie this is ready for another review then

src/modules/dashboard/components/TaskComments/TaskComments.tsx
```diff
@@ -72,26 +76,27 @@ const handleKeyboardSubmit = (
...
+        variables: {
+          input: {
+            id: draftId,
+            message,
+          },
+        },
+        refetchQueries: [
+          {
+            query: TaskFeedEventsDocument,
+            variables: { id: draftId, },
```
ceolson01: The type of `variables` isn't inferred from the value of the `query` property, so we should probably use some type assertion here:

```suggestion
            variables: { id: draftId, } as TaskFeedEventsQueryVariables,
```

src/modules/dashboard/components/TaskComments/TaskComments.tsx
```diff
@@ -72,26 +76,27 @@ const handleKeyboardSubmit = (
...
-
-  const transform = useCallback(
-    mergePayload({ colonyAddress, author: walletAddress, draftId, taskTitle }),
-    [colonyAddress, draftId],
+const TaskComments = ({ draftId, history }: Props) => {
+  const { username } = useLoggedInUser();
+
+  const [sendComment] = useSendTaskMessageMutation();
+  const onSubmit = useCallback(
+    ({ comment: message }: FormValues, { resetForm }: FormikBag<any, any>) =>
```
ceolson01: I think this can be a little more tightly typed:

```suggestion
    ({ comment: message }: FormValues, { resetForm }: FormikBag<object, FormValues>) =>
```

PR #1976, opened 2020-01-20
BYOT Display proper Token preview values

### LINKED ISSUE #1971: Token name and symbol aren't displayed properly when creating a new colony

rdig:
## Description

This PR fixes a small issue with the "user external" create colony wizard path, where the Token's preview values would not be displayed.

This was due to the fact that the return object changed after refactoring the metadata storage, so this needed to be adjusted as well.

**Changes**

- [x] `TokenSelector` use the proper object for diplay Token details

**Demo**

Screenshot from 2020-01-20 17-00-10

Resolves #1971


rdig:
Closing this in favor of @chmanie 's #1967 which apparently already fixed this

PR #1984, opened 2020-01-21
Upvote a suggestion

### LINKED ISSUE #1922: Upvote a suggestion
#1598  Specification

### Story

As an end user, I want to upvote a suggestion, so I can give support to good ideas.

Suggestions can be upvoted by any colony user. Clicking the upvote button adds an upvote and increments the counter. Clicking it again removes the upvote and decrements the counter.

Suggestions can be upvoted / un-upvoted no matter what the status is.

### Design

[Figma link](https://www.figma.com/file/6UoE1xwGiEIiDyyxPpotqR/Community-Work-Platform-Sprint?node-id=531%3A37334&viewport=2134%2C-6255%2C1)

**Upvote**
Upvote@2x (1)

**Hover state** 
upvote@2x

**Upvoted**
Upvote-Upvoted@2x

**Focus (accessability/keyboard support)**
Upvote - Focus@2x

- [ ] Users can upvote a suggestion
- [ ] Users can remove their upvote
- [ ] The counter increases or decreases appropriately

### LINKED ISSUE #1981: upvote button goes here */}^
https://github.com/JoinColony/colonyDapp/blob/2df1ec96f7da5b9ef4bb67b9ed08b41f5e04bb67/src/modules/dashboard/components/SuggestionsListItem/SuggestionsListItem.tsx#L49-L54

---

###### This issue was generated by [todo](https://todo.jasonet.co) based on a `todo` comment in 2df1ec96f7da5b9ef4bb67b9ed08b41f5e04bb67 when #1966 was merged. cc @JoinColony.

ceolson01:
## Description

This PR adds the ability to upvote a suggestion.

**New stuff** ✨

* `@formatjs/intl-unified-numberformat` polyfill for i18n numeral formatting
* `upvote` icon
* `AbbreviatedNumeral` core component for displaying the upvote count (i.e. instead of `1,600` it displays `1.6k`)
  * For cases where precision isn't necessary
  * Supports i18n
* `SuggestionUpvoteButton`

**Changes** 🏗

* Improve `Icon` prop typing
* Remove `ConfirmDialog` step when creating a suggestion
* Redirect to claim profile flow when a user with an unclaimed profile tries to upvote.
* Redirect to claim profile flow when a user with an unclaimed profile focuses on the suggestion create input

**Demo** 🎥 

Upvote

Closes #1922 
Closes #1981


src/img/icons/upvote.svg
```diff
@@ -0,0 +1 @@
+<svg width="10" height="6" fill="none" xmlns="http://www.w3.org/2000/svg"><g clip-path="url(#clip0)"><path fill-rule="evenodd" clip-rule="evenodd" d="M9.65 5.06c.1-.1.15-.21.1-.26L4.97.01c-.017-.017-.048-.02-.086-.01-.035-.005-.065 0-.084.02l-.083.083A.544.544 0 0 0 4.71.11l-.006.006L.02 4.81c-.04.04 0 .16.1.26l.35.34c.1.1.21.15.26.1l4.15-4.16 4.16 4.16c.04.04.16 0 .26-.1l.35-.35z" fill="#3C444D" fill-opacity=".7"/></g><defs><clipPath id="clip0"><path transform="rotate(90 9.77 0)" fill="#fff" d="M9.77 0h5.53v9.77H9.77z"/></clipPath></defs></svg>
```
rdig: This could be further optimized and made the standard 30x30 size, but It's good enough for now.

src/modules/core/components/Icon/Icon.tsx
```diff
@@ -21,10 +21,13 @@ type Appearance = {
   size?: 'tiny' | 'small' | 'normal' | 'medium' | 'large' | 'huge';
 };
 
-interface Props {
+interface InProps extends HTMLAttributes<HTMLElement> {
```
rdig: :+1:

src/modules/core/components/Icon/icons.json
```diff
@@ -46,6 +46,7 @@
     "tag",
     "three-dots-row",
     "unlock",
+    "upvote",
```
rdig: This is a vertically flipped version of our `caret-down-small`:

Screenshot from 2020-01-22 15-22-52

So by my reasoning this should be called `care-up-small`

src/modules/dashboard/checks.ts
```diff
@@ -122,3 +122,12 @@ export const canFinalizeTask = (task: AnyTask, roles: ROLES[]) =>
   canFund(roles);
 
 export const canRecoverColony = hasRoot;
+
+/*
+ * Suggestions
+ */
+
+export const hasUpvotedSuggestion = (
```
rdig: So I think we should actually change the logic behind your own upvotes.

Screenshot from 2020-01-22 15-30-11

We have two ways to go about it:
- Either you upvote your submission automatically when you create it
- Either you can't upvote your own suggestion (reddit style)

This comes down to what is the number of upvotes we want the suggestion to start with: 0 (you can't upvote yours) or 1 (you automatically upvote)

What are your thoughts on this ?

Maybe even @JoinColony/product can chime in ....

src/modules/dashboard/components/SuggestionCreate/SuggestionCreate.tsx
```diff
@@ -82,16 +79,18 @@ const SuggestionCreate = ({ colonyAddress, domainId, openDialog }: Props) => {
...
       validationSchema={validationSchema}
     >
-      <Input
-        appearance={{ theme: 'fat' }}
-        label={MSG.inputLabelTitle}
-        name="title"
-        innerRef={inputRef}
-      />
+      <div onFocus={handleUnclaimedProfile}>
+        <Input
```
rdig: This is another UI change suggestion.

Screenshot from 2020-01-22 15-19-05

I suggest we show either a tooltip saying something like: "Press return to create a new suggestion"

Or just add a button at the end of the input, like we have on domains:

Screenshot from 2020-01-22 15-32-42

Either way, I think that for "some" users the current suggestion creation flow will be confusing.

Thoughts ?

And again, maybe @JoinColony/product can chime in here...

ceolson01: You're absolutely right - I think the user would benefit by having some sort of instructive text or symbol.

Another solution would be to use the `InputStatus` (just below the input) to display some text (when it's valid, of course). It's already there, so I think it'd be the easiest & fastest solution to implement.

Screen Shot 2020-01-22 at 8 04 37 AM

ceolson01: That said, I think @JoinColony/product should make this call 🙂

src/modules/dashboard/checks.ts
ceolson01: Hmmm interesting ideas....

> Either you can't upvote your own suggestion (reddit style)

I think this makes the most sense to me. But once again, I'm curious about @JoinColony/product's input here.

Great call, @rdig

collinvine: This is not how Reddit works. Reddit is the opposite: you automatically upvote your own post, and have the option of removing it or even down voting it. (I just tested this.)

I prefer that: automatically add a +1 from the person who posted, and allow them to remove if they want.

ceolson01: Great, thanks guys!

ceolson01: Update: solved via https://github.com/JoinColony/colonyServer/pull/37

src/modules/dashboard/components/SuggestionCreate/SuggestionCreate.tsx
```diff
@@ -26,54 +27,50 @@ interface FormValues {
...
-}
-
 const validationSchema = yup.object({
   title: yup.string().required(),
 });
 
 const displayName = 'Dashboard.SuggestionCreate';
 
-const SuggestionCreate = ({ colonyAddress, domainId, openDialog }: Props) => {
+const SuggestionCreate = ({ colonyAddress, domainId, history }: Props) => {
```
chmanie: I'm guessing you're changing that to use hooks as well :)

src/modules/dashboard/components/SuggestionCreate/SuggestionCreate.tsx
chmanie: I wouldn't use the status field for help text. For that we have the `help` property. Let's keep it consistent!

src/modules/dashboard/components/SuggestionCreate/SuggestionCreate.tsx
ceolson01: Yep!

src/modules/dashboard/components/SuggestionCreate/SuggestionCreate.tsx
ceolson01: We decided on the button 🙂

chmanie: Ah, even better!

src/modules/core/components/Icon/icons.json
ceolson01: Done! Great call 🍻

PR #1986, opened 2020-01-22
Upgrade redux-router to v5

chmanie:
# Description

This PR sorts out some dependency chaos around react-router and upgrades it to v5.x. We're switching out `withRouter` with hooks wherever possible. Furthermore we're removing connected-react-router and use the history directly in the components to redirect (two occasions).

**Changes** 🏗

- Upgrade react-router to v5
- Use hooks where possible
- Redirect directly in components

**Deletions** ⚰️

- Remove connected-react-router


src/routes/Routes.tsx
```diff
@@ -173,9 +173,4 @@ const Routes = () => {
   );
 };
 
-// We need `withRouter` here because (surprisingly) react-router-dom is not
-// using the router context property (available to e.g. each `Switch`):
-// https://github.com/ReactTraining/react-router/issues/4671
-// We are using `withRouter` to get around `connect()`'s `shouldComponentUpdate`
-// function blocking updates when the route location changes.
-export default withRouter(Routes);
+export default Routes;
```
chmanie: I think this was needed because we used `react-router` directly sometimes? Or the connected redux-router? I don't know, it does work now.

PR #1991, opened 2020-01-23
Upgrade IPFS to v0.40.0

chmanie:
## Description

Exactly what it says in the title. A few adjustments had to be made due to the changed API. Nothing big though.

**Changes** 🏗

* Upgrade IPFS to v0.40.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -203,7 +203,7 @@
         "history": "^4.7.2",
         "hoist-non-react-statics": "^3.3.0",
         "immutable": "^4.0.0-rc.12",
-        "ipfs": "^0.36.4",
+        "ipfs": "^0.40.0",
         "jwt-decode": "^2.2.0",
         "linkify-it": "^2.0.3",
         "localforage": "^1.7.3",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -203,7 +203,7 @@
         "history": "^4.7.2",
         "hoist-non-react-statics": "^3.3.0",
         "immutable": "^4.0.0-rc.12",
-        "ipfs": "^0.36.4",
+        "ipfs": "^0.40.0",
```
rdig: Using this version of ipfs, the production build runs out of memory:

Screenshot from 2020-01-23 16-22-57

The only way I could get this to build, is to increase node's memory limit to 4 Gb:
```
node --max-old-space-size=4096 ...
```

So, if you want us to use this version of IPFS, please add this as an runtime variable to the production build script.

Something like:
```
    ...
    "scripts": {
        ...
        "webpack:build:prod": "NODE_OPTIONS="--max-old-space-size=4096" NODE_ENV=production webpack-cli --config webpack.prod.js",
    },
    ...
```

chmanie:
@rdig added the flag

PR #1992, opened 2020-01-23
Look what I removed! And it still works!

chmanie:
## Description

I removed packages that we don't need (anymore)

**Deletions** ⚰️

* @types/node
* bs58
* empty
* extfs
* fork-ts-checker-webpack-plugin
* gh-pages
* happypack
* http-proxy
* node-fetch
* redux-mock-store
* simple-git (sic!)
* tap-xunit
* core-js
* history
* hoist-non-react-statics

Maybe there's even more


ceolson01:
Nice, the dapp's going on a diet! 🎉

PR #1993, opened 2020-01-23
Fix token resolvers in queries and mutations

chmanie:
## Description

This fixes two bugs reported by @rdig in connection with querying and expanding tokens. The issue here in both cases was that the original value was not queried from the server and hence not passed into the resolvers. This is now fixed.

**Changes** 🏗

* Pass `tokenAddress` and `tokenAddresses` to the respective resolvers


chmanie:
Cool, let's merge this

PR #2001, opened 2020-01-26
Prevent Users w/o a profile from creating new Tasks

### LINKED ISSUE #1935: User without username can create tasks
## Steps to reproduce

1. Assign admin permission to an address that has not claimed a username
2. Navigate to the colony where permissions were assigned and authenticate with the newly permissioned account
3. Click the "New Task"

## Expected behaviour

Users who have not created a username should not be able to create tasks (or write any other data to orbitDB)

## Actual behaviour

Accounts with the admin permission, but without an ENS subdomain, can create tasks in the colony where they are permissioned. If the account claims a username after the fact, then they can no longer see any of the tasks they created in their "my tasks" section.

*note: I have not been able to replicate this on QA as the dapp crashes whenever I paste an ethereum address into the "new role" selector.

rdig:
## Description

This is a simple PR that prevents users that haven't yet claimed a profile from creating new tasks, even if those users were granted admin permissions in the colony.

**Changes**

- `ColonyHome`, `canCreateTask` also check if the user has claimed a profile

**Screenshot**

Screenshot from 2020-01-26 15-43-10

Resolves #1935

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
@@ -114,11 +114,6 @@ const ColonyHome = ({
 
   const { colony } = data;
 
-  /*
-   * Eventually this has to be in the proper domain. There's probably going to be a different UI for that
-   * Also prevent users w/o a claimed profile from creating tasks
-   * (even if the address got administrative permissions in the colony)
-   */
   const canCreateTask = canAdminister(currentDomainUserRoles) && !!username;
 
   const noFilter = (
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -114,8 +114,12 @@ const ColonyHome = ({
 
   const { colony } = data;
 
-  // Eventually this has to be in the proper domain. There's probably going to be a different UI for that
-  const canCreateTask = canAdminister(currentDomainUserRoles);
+  /*
+   * Eventually this has to be in the proper domain. There's probably going to be a different UI for that
+   * Also prevent users w/o a claimed profile from creating tasks
```
chmanie: This sounds like we would like to do that in the future. I don't think this comment is necessary (right now it's more confusing than helpful IMHO :) )

rdig: Ok, I'll remove it then

PR #2002, opened 2020-01-26
Don't Display UserInfo on Userlist Avatars

### LINKED ISSUE #1913: Both the roll model and profile popover open when you click on profile image in permissions list
## Steps to reproduce

1. click on a profile image in the permissions list

## Expected behaviour

Profile popover should open.

## Actual behaviour

Both the profile popover and the rolls modal open.

profile popover

rdig:
## Description

This PR disables the `UserInfo` tooltip on items displayed via a `UserList` in the admin panel. This was due to both actions firing at the same time: showing userinfo and opening the permission modal resulting in a broken UI state.

While I was at it, I also added a simple logic check to display the parent (root) inheritance description text, only if a different domain that root is selected.

**Changes**

- `UserListItem` don't show `UserInfo` on avatars
- `Permissions` only show parent inheritance tip on non-root domains

**Screenshots**

_`UserInfo` still available on the permissions edit modal:_
Screenshot from 2020-01-26 15-57-01

_Parent inheritance tip only shown on non-root domain:_
Screenshot from 2020-01-26 15-56-53
Screenshot from 2020-01-26 15-57-20

Resolves #1913

### DIFF
--- a/src/modules/admin/components/Permissions/Permissions.tsx
+++ b/src/modules/admin/components/Permissions/Permissions.tsx
@@ -163,6 +163,7 @@ const Permissions = ({ colonyAddress, domains, openDialog }: Props) => {
                     showDisplayName
                     showMaskedAddress
                     showUsername
+                    showInfo={false}
                   >
                     <TableCell>
                       <UserPermissions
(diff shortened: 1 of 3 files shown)


src/modules/admin/components/Permissions/Permissions.tsx
```diff
@@ -174,19 +175,21 @@ const Permissions = ({ colonyAddress, domains, openDialog }: Props) => {
...
-                  learnMore: (
-                    <ExternalLink
-                      text={MSG.learnMore}
-                      href={DOMAINS_HELP_URL}
-                    />
-                  ),
-                }}
-              />
-            </p>
+            {selectedDomain && selectedDomain.id !== ROOT_DOMAIN && (
```
chmanie: 👍

PR #2015, opened 2020-01-31
Whitelist Token Locking address when creating a new Colony

rdig:
## Description

This PR adds in the token locking address (fetched from the network) to the `deployTokenAuthority` transaction's `allowToTransfer` argument, when creating a new colony.

This way users will be allowed be allowed to deposit tokens, even while the token is locked.

Original report: https://joincolony.slack.com/archives/C02R9SP3J/p1580399575113400

**Changes** 

- [x] `colonyCreate` saga, import `networkClient`, fetch `tokenLockingAddress`, pass it to the `deployTokenAuthority` tx

### DIFF
--- a/src/modules/dashboard/sagas/colonyCreate.ts
+++ b/src/modules/dashboard/sagas/colonyCreate.ts
@@ -52,6 +52,8 @@ function* colonyCreate({
     Context.APOLLO_CLIENT,
   );
 
+  const { networkClient } = yield getContext(Context.COLONY_MANAGER);
+
   const TOKEN_DECIMALS = 18;
 
   /*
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -153,10 +155,16 @@ function* colonyCreate({
     });
 
     if (createToken) {
+      const { address: tokenLockingAddress } = yield call([
+        networkClient.getTokenLockingAddress,
+        networkClient.getTokenLockingAddress.call,
```
chmanie: Could this fail? Could this be not defined? Just making sure we're handling all the cases.

rdig: I don't think It can be `undefined`, unless colonyJS crashes completely.

It can however fail. I'm thinking infura connectivity troubles. But if that's the case the whole creation process fails, since everything is wrapped in big try/catch block.

The only alternative I can think of, is re-wrapping just this in another try/catch block, and if it fails, just _silently_, leave `allowToTransfer` as a empty `[]` array. But that takes us back to our original issue.

PR #2023, opened 2020-02-04
Fix task attributes cache

### LINKED ISSUE #2012: Changing task assignee does not update those values in the task list
### Description

This issue encompasses 2 bugs. Changing a tasks assignee:

1. Does not update those values until after a page refresh.
2. Does not add task to "my tasks" list if assigned to one's self

### Suggestion

This might be solved by adding in a `refetchQueries` directive, updating the cache directly, adjusting mutation response parameters, or some combination of the aforementioned solutions.

### Demo

bug-assignee-not-updating

ceolson01:
## Description

This PR fixes two particular issues, and likely prevents a couple of related cache bugs from surfacing in the future.

Most notably - in cases where an expanded user and their address are both fields on query types, we now accept both fields in mutation results. This solves some cache issues related to task worker assignment not updating immediately in task lists.

Also, this PR ensures that a task is added or removed from the user task list when they've assigned or unassigned themselves, respectively.

Closes #2012

### DIFF
--- a/src/modules/dashboard/components/TaskEditDialog/TaskEditDialog.tsx
+++ b/src/modules/dashboard/components/TaskEditDialog/TaskEditDialog.tsx
@@ -24,8 +24,11 @@ import {
   useUnassignWorkerMutation,
   useSetTaskPayoutMutation,
   useRemoveTaskPayoutMutation,
+  useLoggedInUser,
+  UserTasksDocument,
+  UserTasksQueryVariables,
 } from '~data/index';
-import { Address } from '~types/index';
+import { Address, createAddress } from '~types/index';
 import HookedUserAvatar from '~users/HookedUserAvatar';
 
 import Payout, { FormPayout } from './Payout';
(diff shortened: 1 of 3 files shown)


src/modules/dashboard/components/TaskEditDialog/TaskEditDialog.tsx
```diff
@@ -212,15 +216,31 @@ const TaskEditDialog = ({
...
         );
 
+      const refetchSelfTasks = [
+        {
+          query: UserTasksDocument,
+          variables: {
+            address: walletAddress,
+          } as UserTasksQueryVariables,
+        },
+      ];
```
ceolson01: In theory, this could be done as a `cacheUpdates` function, but would be quite messy, as it would require us to expand some of our existing queries unnecessarily, just for when this case occurs (user assigns or unassigns themselves). I can't imagine this is going to happen often enough to warrant such implementation.

PR #2030, opened 2020-02-06
Prevent finalizing a Task with insufficient domain pot balance

### LINKED ISSUE #1952: Finalizing a task in a domain without funds yields weird error messages
## Steps to reproduce

1. Create a task in a subdomain with no funds
2. Assign funding and worker
3. Try to finalize the task

## Expected behaviour

* It should not be possible to finalize task when the domain has no funding

## Actual behaviour

* See image

image

## Suggestions (optional)

* Check domain funding when the user is clicking "finalize task" and open a modal window explaining why finalizing isn't possible at this point in time

### LINKED ISSUE #1997: Can allocate more ETH/tokens to a task than the colony has in its balance
## Steps to reproduce

1. create a task
2. allocate more tokens than are in the colony

## Expected behaviour

The confirm buttin should be disabled and you should be warned that the colony/domain does not have enough tokens to fund the task.

## Actual behaviour

You can allocate as many tokens as you like.
image

rdig:
## Description

This PR fixes the bad UX pattern where a task can be _finalized_ with a payout larger then available funds in the domain pot.

The user would be allowed to press _Finalize Task_, generate the transaction, sign it, but as expected, the transaction would fail _(with a rather cryptic error message)_

The fix for this implies calculating the that task payout relative to the domain pot balance, and in the funds are not sufficient, show a dialog informing the user of that fact, otherwise proceed as nromal.

**New stuff** 

- [x] `TaskFinalize` component
- [x] `TaskFinalizeDialog` dialog component

**Changes** 

- [x] Refactor `Task` to extract the finalize button into it's own component (`TaskFinalize`)
- [x] Fix `Task` to use the `ROOT_DOMAIN` constant, rather then the number `1`

**Demo**

demo-prevent-insufficient-funds-payouts

Resolves #1952
Resolves #1997

### DIFF
--- a/src/modules/dashboard/components/TaskFinalize/TaskFinalize.tsx
+++ b/src/modules/dashboard/components/TaskFinalize/TaskFinalize.tsx
@@ -1,13 +1,16 @@
-import React, { useCallback, useState } from 'react';
+import React, { FC, useCallback, useState } from 'react';
 import { defineMessages } from 'react-intl';
 import moveDecimal from 'move-decimal-point';
 
 import Button from '~core/Button';
+import { withDialog } from '~core/Dialog';
 
 import { Address } from '~types/index';
+import { OpenDialog } from '~core/Dialog/types';
 import {
   AnyTask,
   Payouts,
+  Domain,
   TokenWithBalances,
   useTokenBalancesForDomainsQuery,
 } from '~data/index';
@@ -25,11 +28,16 @@ const MSG = defineMessages({
 
 const displayName = 'dashboard.TaskFinalize';
 
-interface Props {
+interface InProps {
   draftId: AnyTask['id'];
   colonyAddress: Address;
-  ethDomainId: number,
-  payouts: Payouts,
+  ethDomainId: Domain['ethDomainId'];
+  payouts: Payouts;
+}
+
+interface Props extends InProps {
+  // Injected via `withDialog`
+  openDialog: OpenDialog;
 }
 
 const TaskFinalize = ({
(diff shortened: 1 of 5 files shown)


src/modules/dashboard/components/TaskFinalize/TaskFinalize.tsx
```diff
@@ -0,0 +1,127 @@
...
+         */
+        const {
+          // @ts-ignore
+          balances: domainBalances,
+          decimals,
+        } = tokenBalances.tokens.find(
+          ({ address: domainTokenAddress }) =>
+            domainTokenAddress === tokenAddress,
+        ) as TokenWithBalances;
+        return domainBalances.every(
```
ceolson01: Because some of the members of the `TokenWithBalances` union don't have `balances` (I believe `UserToken` is one example), the compiler can't guarantee it'll exist. 

Also, this query (`useTokenBalancesForDomainsQuery`) will always return the same type. 🙂 

Instead, it might be better to check that `domainBalances` exists later:

```typescript
      const enoughFundsAvailable = payouts.every(({ amount, tokenAddress }) => {
        const domainBalances =
          tokenBalances &&
          tokenBalances.tokens.find(
            ({ address: domainTokenAddress }) =>
              domainTokenAddress === tokenAddress,
          );
        if (!domainBalances) {
          return false;
        }
        const { balances, decimals } = domainBalances;
        return balances.every(
          ...
```

rdig:
@ceolson01 

Yup, took your advice on the domains balances logic and now works like a charm. Thanks!

This is ready for another look.

PR #2031, opened 2020-02-06
Colony Community Tab

### LINKED ISSUE #2022: Add 'Community' tab with members list to each Colony
### Story

When contributor in a colony, I want to know what other users are in that colony, so my experience is more social and interesting.

As an admin of a colony, I want to see a list of members, so I can more effectively manage my colony.

### Description 
We want to surface a list of "members" within a colony. Right now, there's no way to see what other users—admins or contributors—are "in" that colony.

Do do this, we want to create a new **Community** tab (to the right of Suggestions, before Transactions) that shows a list of Members and their role:

- Founder (the user who created the colony)
- Admins (anyone with a permission in that colony)
- Members (anyone who has starred that colony and is not a founder or admin)

### Design

image

- [ ] The Community tab is implemented
- [ ] There's a list of users displayed
- [ ] Each user has their Role displayed on the list (`Founder`, `Admin`, `Member`)

### Suggestions

I'm not sure what the ideal ordering would be. Should the Founder and Admin be at the top of the list? Probably. 

If you're not a member, or not sure what that means, maybe we should provide copy saying something like "Star this colony to become a member" directly above the list (or elsewhere?). 

For an empty state we could have: "No members yet. Star this colony to become a member."

Here's a design as inspiration for that CTA when a user isn't a member yet:

image

and after clicking:

image

then it could disappear. These last parts are 🌮 by the way.

rdig:
## Description

This PR adds in a new Community tab to the colony home, which displays the list of users that are "members" of that colony (subscribed to), along with they're respective role.

Note: for the "Admin" role we consider a user with any form of permissions, be they in the root domain, or in any subdomain

**New stuff**

- [x] `Community` dashboard component

**Changes** 

- [x] `ColonyHome` add `Community` tab
- [x] `ColonySubscribe` refetch colony subscribers query
- [x]  `transformers` refactor for reusability
- [x]  `transformers` add `getCommunityRoles`

**Demo**

Community tab, subscribe / unsubscribe:
demo-community-tab

Community tab, empty state:
demo-community-tab-empty

Resolves #2022

### DIFF
--- a/src/data/cacheUpdates.ts
+++ b/src/data/cacheUpdates.ts
@@ -259,7 +259,6 @@ const cacheUpdates = {
               address: subscribedUserWalletAddress,
             },
           });
-          // console.log(subscribedUserProfileFromCache);
           if (
             subscribedUserProfileFromCache &&
             subscribedUserProfileFromCache.user
--- a/src/data/cacheUpdates.ts
+++ b/src/data/cacheUpdates.ts
@@ -19,6 +19,9 @@ import {
   ColonySubscribedUsersQuery,
   ColonySubscribedUsersQueryVariables,
   ColonySubscribedUsersDocument,
+  UserQuery,
+  UserQueryVariables,
+  UserDocument,
 } from './generated';
 
 type Cache = typeof apolloCache;
@@ -199,6 +202,9 @@ const cacheUpdates = {
           const { subscribedUsers } = cacheData.colony;
           const updatedColonySubscription = subscribedUsers.filter(
             ({ id: userWalletAddress }) =>
+              /*
+               * Remove the unsubscribed user from the subscribers array
+               */
               userWalletAddress !== unsubscribedUserWalletAddress,
           );
           cache.writeQuery<
@@ -219,7 +225,78 @@ const cacheUpdates = {
         }
       } catch (e) {
         log.verbose(e);
-        log.verbose('Not updating colony subscriptions cache - not loaded yet');
+        log.verbose(
+          'Cannot update the colony subscriptions cache - not loaded yet',
+        );
+      }
+    };
+  },
+  subscribeToColony(colonyAddress: Address) {
+    return (cache: Cache, { data }: any) => {
+      try {
+        const cacheData = cache.readQuery<
+          ColonySubscribedUsersQuery,
+          ColonySubscribedUsersQueryVariables
+        >({
+          query: ColonySubscribedUsersDocument,
+          variables: {
+            colonyAddress,
+          },
+        });
+        if (cacheData && data && data.subscribeToColony) {
+          const { id: subscribedUserWalletAddress } = data.subscribeToColony;
+          const { subscribedUsers } = cacheData.colony;
+          /*
+           * The subscribed to colony mutation, only returns the user wallet address,
+           * but we also need the user's profile to update the subscribers array
+           */
+          const subscribedUserProfileFromCache = cache.readQuery<
+            UserQuery,
+            UserQueryVariables
+          >({
+            query: UserDocument,
+            variables: {
+              address: subscribedUserWalletAddress,
+            },
+          });
+          // console.log(subscribedUserProfileFromCache);
+          if (
+            subscribedUserProfileFromCache &&
+            subscribedUserProfileFromCache.user
+          ) {
+            const {
+              user: { profile: subscribedUserProfile },
+            } = subscribedUserProfileFromCache;
+            const updatedColonySubscription = [...subscribedUsers];
+            /*
+             * Add the subscribed user to the subscribers array
+             */
+            updatedColonySubscription.push({
+              ...data.subscribeToColony,
+              profile: subscribedUserProfile,
+            });
+            cache.writeQuery<
+              ColonySubscribedUsersQuery,
+              ColonySubscribedUsersQueryVariables
+            >({
+              query: ColonySubscribedUsersDocument,
+              data: {
+                colony: {
+                  ...cacheData.colony,
+                  subscribedUsers: updatedColonySubscription,
+                },
+              },
+              variables: {
+                colonyAddress,
+              },
+            });
+          }
+        }
+      } catch (e) {
+        log.verbose(e);
+        log.verbose(
+          'Cannot update the colony subscriptions cache - not loaded yet',
+        );
       }
     };
   },
--- a/src/data/cacheUpdates.ts
+++ b/src/data/cacheUpdates.ts
@@ -16,6 +16,9 @@ import {
   SuggestionStatus,
   CreateTaskMutationResult,
   CreateTaskFromSuggestionMutationResult,
+  ColonySubscribedUsersQuery,
+  ColonySubscribedUsersQueryVariables,
+  ColonySubscribedUsersDocument,
 } from './generated';
 
 type Cache = typeof apolloCache;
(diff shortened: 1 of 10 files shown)


src/modules/dashboard/components/Community/Community.tsx
```diff
@@ -0,0 +1,211 @@
...
+  const {
+    colony: { subscribedUsers },
+  } = colonySubscribedUsers;
+
+  const communityUsers = subscribedUsers
+    .map(user => {
+      const {
+        profile: { walletAddress: userAddress },
+      } = user;
+      let communityRole = 'member';
```
ceolson01: Maybe use an `enum` for these different values?

src/modules/dashboard/components/Community/Community.tsx
```diff
@@ -0,0 +1,211 @@
...
+        communityRole = 'admin';
+      }
+      return {
+        ...user,
+        communityRole,
+      };
+    })
+    .sort(
+      sortObjectsBy({
+        name: 'communityRole',
```
ceolson01: Do we also want to sort these community members by a secondary property beyond the community role? Such as the user's display name, wallet address, etc.?

Also, maybe a question for @JoinColony/product?

src/modules/transformers.ts
```diff
@@ -116,35 +116,78 @@ export const getAllUserRoles = (
...
+};
+
 /*
  * Eventually we'll switch all of the dApp to using new roles, but until then
  * we still need to be able to get the old roles `admins` and `founder`. This
  * util can be removed once the DLP project is completed.
  */
 export const getLegacyRoles = (
-  domains,
+  domains: Record<string, DomainType>,
```
ceolson01: 👍

src/modules/transformers.ts
```diff
@@ -116,35 +116,78 @@ export const getAllUserRoles = (
...
-  const admins = Object.keys(rootDomainRoles).reduce(
-    (acc, address) =>
-      rootDomainRoles[address].includes(ROLES.ADMINISTRATION) &&
-      address !== founder
-        ? acc.add(address)
-        : acc,
-    new Set(),
-  );
+  const founder = getLegacyFounder(rootDomainRoles);
+  const admins = getLegacyAdmins(domains, ROOT_DOMAIN, founder);
```
ceolson01: ❤️

src/modules/transformers.ts
```diff
@@ -116,35 +116,78 @@ export const getAllUserRoles = (
...
+  return {
+    founder,
+    admins: Array.from(admins) as string[],
+  };
+};
 
+/*
+ * @NOTE This differs from the above transformer as it considers roles in any domain (root + subdomains)
+ * to be an admin role
```
ceolson01: 👍

src/modules/dashboard/components/Community/Community.tsx
```diff
@@ -0,0 +1,211 @@
...
+    id: 'dashboard.Community.subscribedReward',
+    defaultMessage: `{star} Subscribed!`,
+  },
+});
+
+interface Props {
+  colonyAddress: Address;
+}
+
+const Community = ({ colonyAddress }: Props) => {
```
ceolson01: `displayName` please!

src/modules/dashboard/components/Community/Community.tsx
collinvine: I had only briefly pondered on that and didn't come up with a great answer.

Ideally, we'd sort on "contribution" or "reputation" but I don't think we'll have that anytime soon.

Any other sorting variable—name, join date—I see pros and cons for. (Name: easy to scan; irrelevant sort | date: oldest at top would be like a kudos, but newest at bottom would push them down)

If there was a forced choice, maybe display name or oldest first (newest last) (that's probably the default, no?). But I'm not sold on any one.

Thoughts?

ceolson01: I tend to agree with the fact that no other properties are really relevant until reputation comes into the picture

src/modules/dashboard/components/Community/Community.tsx
rdig: Granted it's wrong, and an older format introduced by copy-paste, but it is there...

https://github.com/JoinColony/colonyDapp/blob/31e41bdb30de8a4b21461ffe58daaa861a0bbd31/src/modules/dashboard/components/Community/Community.tsx#L209

rdig:
@ceolson01 

- Added cache update functions _(although I like this way of saving on server requests, the level of boilerplate code is not something I enjoy, and we need to find a better solution for this)_
- clear the timeout on component unmount
- updated the display name
- added an `enum` to better manage roles within the component

Read for another look.

src/modules/dashboard/components/Community/Community.tsx
```diff
@@ -0,0 +1,227 @@
...
+  FOUNDER = 'founder',
+  ADMIN = 'admin',
+  MEMBER = 'member',
+}
+
+const displayName = 'dashboard.Community';
+
+const Community = ({ colonyAddress }: Props) => {
+  const [justSubscribed, setJustSubscribed] = useState<boolean>(false);
+  const subscribedMessageTimer = useRef<any>(null);
```
ceolson01: **Optional** This could be typed as a `number`

src/modules/dashboard/components/Community/Community.tsx
```diff
@@ -0,0 +1,227 @@
...
+      );
+    }
+  }, [subscribeToColonyMutation, colonyAddress, setJustSubscribed]);
+
+  /*
+   * We need to wrap the call in a second function, since only the returned
+   * function gets called on unmount.
+   * The first one is only called on render.
+   */
+  useEffect(() => () => clearTimeout(subscribedMessageTimer.current), [
```
ceolson01: 👍

src/modules/dashboard/components/Community/Community.tsx
```diff
@@ -0,0 +1,227 @@
...
+
+interface Props {
+  colonyAddress: Address;
+}
+
+enum ROLES {
+  FOUNDER = 'founder',
+  ADMIN = 'admin',
+  MEMBER = 'member',
+}
```
ceolson01: It seems like we've been tending to prefer pascal case with enums elsewhere. Definitely a nit-pick, and I know we have a couple other instances that are uppercase (likely leftover from js -> ts migration). Totally your call here!

src/data/cacheUpdates.ts
```diff
@@ -177,6 +183,123 @@ const cacheUpdates = {
...
+          const subscribedUserProfileFromCache = cache.readQuery<
+            UserQuery,
+            UserQueryVariables
+          >({
+            query: UserDocument,
+            variables: {
+              address: subscribedUserWalletAddress,
+            },
+          });
+          // console.log(subscribedUserProfileFromCache);
```
ceolson01: Leftover comment 🙂

src/data/cacheUpdates.ts
```diff
@@ -177,6 +183,123 @@ const cacheUpdates = {
...
+          variables: {
+            colonyAddress,
+          },
+        });
+        if (cacheData && data && data.subscribeToColony) {
+          const { id: subscribedUserWalletAddress } = data.subscribeToColony;
+          const { subscribedUsers } = cacheData.colony;
+          /*
+           * The subscribed to colony mutation, only returns the user wallet address,
+           * but we also need the user's profile to update the subscribers array
```
ceolson01: 👍

PR #2233, opened 2020-07-23
Upgrade ApolloClient to v3

chmanie:
## Description

This is tracking the upgrade to `@apollo/client` v3. We have a few things that need investigation and might need a migration path.

This is an ongoing investigation

**New stuff** ✨

- Upgraded `@apollo/client` to v3.
- Upgraded all `@graphql-codegen` packages

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

- Removed all old `apollo` related dependencies:
```
    "@apollo/react-hoc": "^3.1.3",
    "@apollo/react-hooks": "^3.1.3",
    "apollo-cache-inmemory": "^1.6.3",
    "apollo-client": "^2.6.4",
    "apollo-link-context": "^1.0.19",
    "apollo-link-http": "^1.5.16",
```

## TODO

- [x] Change the base branch again after the purser changes are in
- [x] See if the InMemoryCache rewrite affects us. See if we can benefit from it
- [x] `dataIdFromObject` is deprecated (but still works). Find a new solution: https://www.apollographql.com/docs/react/caching/cache-configuration/


codegen.yml
```diff
@@ -9,9 +9,6 @@ generates:
     schema: 'src/data/graphql/typeDefs.ts'
     config:
       skipTypename: true
-      withComponent: false
```
chmanie, line 12: These are now the default values

PR #2238, opened 2020-07-27
Experimental Subgraph Integration

rdig:
**Do not** merge this PR!

This PR is an exploration of integrating a subgraph node alongside our infrastructure and pulling data from the chain, via graphql queries directly.

### Setup

This PR works alongside its server branch companion: [colonyServer/feat/the-graph](https://github.com/JoinColony/colonyServer/tree/feat/the-graph)

There are a lot of moving parts involved in setting this up in our current environment:
- manual server and mongo DB instances
- manual ganache instance
- manual deployment of the contracts
- starting the dapp without ganache, without the server and the database, and without truffle deploying contracts
- a manual instance of the graph node

Screenshot from 2020-07-27 15-25-57

Now that I've scared you enough, @area has set up a nice step by step guide to get this started: https://github.com/JoinColony/colonyServer/blob/feat/the-graph/Readme.md#setup

### Limitations

1. The [event handlers](https://github.com/JoinColony/subgraph/blob/master/src/mappings/colony.ts) listening for contract events get compiled into Web Assembly before being deployed to the graph node. The source from which they get compiled is Typescript. This means that you need **uncompiled** TS code in order to properly compile. This limits the number of external libraries we can use server-side, as most just ship the TS definitions rather then the actual code.
2. During migration, we will have to use two graphql endpoints. While I got this working _(see: "Where the magic happens")_ it is not officially supported.
3. During migration, we will most likely run into graphql query naming collisions.
4. Since we couldn't use external libraries on the server-side, I couldn't checksum addresses when creating the event handlers and had to move that client-side. This prevents me from using query filtering, and I had to fetch all the domains at once. If we can't find a solution for this, it going to be very problematic on `mainnet`

Screenshot from 2020-07-27 15-27-45

### Where the magic happens

The only thing special about this is using two endpoints for the apollo client:

https://github.com/JoinColony/colonyDapp/blob/feat/the-graph/src/context/apolloClient.ts#L36-L40

While I couldn't find specific documentation references to this, I found a bunch of GH issues where contributors pointed out that this is not officially supported. So if we decide to go this route, it's something to keep in mind.

### Current implementation

The current implementation is pretty naive and very basic, as the query itself is hard-coded inside the component bypassing resolvers and code generation. But I guess it's good enough for demo purposes.

https://github.com/JoinColony/colonyDapp/blob/feat/the-graph/src/modules/admin/components/Domains/DomainList.tsx#L49-L67

### Demo

demo-subgraph-domains

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,3 +23,5 @@ NODE_ENV=development
 
 # HTTP endpoint for graphql server
 SERVER_ENDPOINT=http://127.0.0.1:3000
+
+GRAPH_ENDPOINT=http://127.0.0.1:8000/subgraphs/name/joinColony/subgraph
--- a/.env.example
+++ b/.env.example
@@ -24,4 +24,4 @@ NODE_ENV=development
 # HTTP endpoint for graphql server
 SERVER_ENDPOINT=http://127.0.0.1:3000
 
-GRAPH_ENDPOINT=http://127.0.0.1:8000/subgraphs/name/joinColony/subgraph
+SUBGRAPH_ENDPOINT=http://127.0.0.1:8000/subgraphs/name/joinColony/subgraph
--- a/src/context/apolloClient.ts
+++ b/src/context/apolloClient.ts
@@ -15,7 +15,7 @@ const httpLink = createHttpLink({
 });
 
 const subgraphHttpLink = createHttpLink({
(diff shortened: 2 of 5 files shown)


rdig:
This is no longer needed as #2323 implemented the proper version

PR #2240, opened 2020-07-30
purser/colony-js aftermath fixes

chmanie:
## Description

This PR fixes a few issues/bugs we found when deploying to staging.

**Changes** 🏗

* ColonyJS fixes for external and Colony tokens
* ColonyJS fixes for SAI (`bytes32`)
* Fix community roles not being shown
* Fix transfers not showing
* Fix token authority deployment issues
* Fix Permission domain filter


auryn-macmillan:
The domain name doesn't show in the dropdown when you load the permissions tab.
If you select a domain, no permissions are shown.

Screen record from 2020-07-30 10 41 02

rdig:
This is just a follow up to @auryn-macmillan 's previous comment. After selecting the domain, you can't add permissions to it.

bug-cannot-assign-permissions

As stated it the `#qa` channel, this does not happen on the production version, so it's most likely related to the latest updates.

PR #2261, opened 2020-08-25
Fix task finalize from Transaction hash

rdig:
## Description

This is a small change that fixes the resolver which checks, upon loading a task, if it was already finalized, extracting the required data from the transaction, then firing the appropriate mutation to update/create database entries. 

The issue at play here was that the user address used to get the user's roles in the colony was no checksummed, so it would always fail the check.

**Changes**

- [x] `finalizeTaskWithTxHash` the uses the checksummed version of `userAddress`


chmanie (approved):
Urgh, I was working under the assumption that the ethers `signer` always returns the checksummed address. Well boooo.

Thanks for the fix

PR #2262, opened 2020-08-26
Fix wizard initial values

ceolson01:
## Description

This PR allow `initialValues` in the `withWizard` HOC to accept an array **or** a function in the case of non-linear wizards.

**Changes** 🏗

* Update `withWizard` to accept a function for non-linear wizards


rdig (approved):
Looks good to me! :100:

PR #2263, opened 2020-09-03
Temporarily Disable Colony Creation

rdig:
## Description

This PR prevents users from creating new colonies, by disabling the submit button on the first step of the create colony flow. This change came about the high gas costs currently floating about on the Ethereum network.

However, this should be rolled back when the times is right, or for the xDai deployment.

**Reviewer Note**: @ceolson01 I've added you as a reviewer, but in the interest of time, most likely this will have been already merged.

**Changes**

- [x] `CreateColonyWizard` disabled the step array, but still allow username creation
- [x] `StepColonyName` disable submit button, and wrap it in a tooltip

**Screenshots**

Screenshot from 2020-09-03 12-26-28


ceolson01 (approved):
Looks good, let's get this xDai implementation going! 💪

PR #2283, opened 2020-10-29
Colony Home Funds Display

rdig:
## Description

This PR adds the colony funds display component to the Colony Home screen

**Changes**

- [x] Added `ColonyTotalFunds` component
- [x] Added `ColonyTotalFundsPopover` sub-component
- [x] Added `MuliWeb` light font
- [x] Renamed `Muli` to `MuliWeb`
- [x] Updated the `Popover` drop shadow specs

**Reviewers Note**

This needed to be updated since the design specs changed during development. Notice the _old_ and _current_ implementations.

**Update**: This went through yet another round of spec updates, but this time only minor style changes. One of those changes involved us serving a new font variant -- light, as well as renaming our existing fonts _(apparently there's a naming rights issue)_

**Demo**

Old Implementation

cs-colony-funds-display

Current Implementation

cs-colony-funds-display-new

Resolves DEV-15


src/modules/dashboard/components/ColonyTotalFunds/ColonyTotalFunds.tsx
```diff
@@ -0,0 +1,126 @@
...
+          className={styles.selectedTokenAmount}
+          unit={getTokenDecimalsWithFallback(currentToken.decimals)}
+          value={currentToken.balances[COLONY_TOTAL_BALANCE_DOMAIN_ID].amount}
+        />
+        <ColonyTotalFundsPopover
+          tokens={data.tokens}
+          onSelectToken={setCurrentTokenAddress}
+          currentTokenAddress={currentTokenAddress}
+        >
+          <span className={styles.selectedTokenSymbol}>
```
ceolson01, line 98: I'm sure you know what I'm going to say here... 😏 

Keyboard focus states! And should be keyboard navigable. Maybe a `button` here would be better practice?

<sup>(Can compose the `.button` class from `reset.css`)

src/modules/dashboard/components/ColonyTotalFunds/ColonyTotalFunds.tsx
```diff
@@ -0,0 +1,126 @@
...
+                title={MSG.tokenSelect}
+              />
+            </span>
+          </span>
+        </ColonyTotalFundsPopover>
+      </div>
+      <div className={styles.totalBalanceCopy}>
+        <FormattedMessage {...MSG.totalBalance} />
+        <Button
+          appearance={{ theme: 'blue', size: 'small' }}
```
ceolson01, line 113: Hmm can you also check the `Button` component to make sure it's showing keyboard focus states? Odd that this seems to be the only instance where the outline isn't visible 🤔

ceolson01 (approved):
Nice nice, great stuff ✅ 

Could use a couple a11y improvements, but nothing blocking!

src/modules/dashboard/components/ColonyTotalFunds/ColonyTotalFunds.tsx
rdig: This was done deliberately. I've manually disabled the outline.

Sadly I've lost the war trying to defend keyboard navigation :(

src/modules/dashboard/components/ColonyTotalFunds/ColonyTotalFunds.tsx
rdig: Sure

PR #2284, opened 2020-10-30
Colony Members Avatars

### LINKED ISSUE #2270: Colony Members List
_Note: This issue relies on #2267 to provide the grid layout_

## Specification

Discernible features:
- If no members are available to display, then don't display the component at all _(although that shouldn't be the case)_
- It should display at max 12 avatar positions _(see below about the last one)_ 
- The list should gradually fill up from 1 up until 12 positions
- 12 Members should be displayed all in the list
- 13 Members should be displayed as 11 entries while the last avatar entry should display a +2
- Other member count values above 13 will just increment the last display entry counter
- Note that each individual avatar has an tooltip display on click, but we are keeping the current existent one, rather then implementing the one displayed in the screenshot below
- The members list will be filtered by reputation inside a specific domain, but that will not be implemented in this issue, just be aware that most likely, the members array will need to be passed in as a prop

### Design

See Figma section: Home UI / Actions, Default View and Additional UI / Connect wallet button: [Colony Simplified Figma Screens](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified)

Details view

Screenshot from 2020-10-01 12-46-02

Screenshot from 2020-10-01 12-46-12

Screenshot from 2020-10-01 12-46-31

Screenshot from 2020-10-01 18-03-48

Context view

Screenshot from 2020-09-30 20-17-43

rdig:
## Description

This PR adds in the colony members avatars display on the right aside, which will show the users in that colony with the most reputation.

Currently the reputation filter feature is not implemented _(as it requires changes to the reputation oracle)_, as well as not featuring the community page route link, which will only be implemented in DEV-13

**Changes**

- [x] Refactored code `UserAvatar` and `InfoPopover` to support configuring the underlying `Popover` component via simple props
- [x] Added `ColonyMembers` component

**Demo**

cs-colony-members

Resolves #2270
Resolves DEV-6

### DIFF
--- a/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
+++ b/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
@@ -5,7 +5,7 @@ import NavLink from '~core/NavLink';
 import Heading from '~core/Heading';
 import HookedUserAvatar from '~users/HookedUserAvatar';
 import { DASHBOARD_ROUTE } from '~routes/index';
-import { Colony, useLoggedInUser } from '~data/index';
+import { Colony, useColonySubscribedUsersQuery, AnyUser } from '~data/index';
 
 import styles from './ColonyMembers.css';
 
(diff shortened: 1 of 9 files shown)


src/styles/variables.css
```diff
@@ -90,6 +92,7 @@
   --size-unreadable: 10px;
   --size-tiny: 11px;
   --size-small: 12px;
+  --size-smallish: 13px; /* This shows up more and more in Colony Simplified */
```
ceolson01, line 95: 👍

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+const UserAvatar = HookedUserAvatar({ fetchUser: false });
+const MAX_AVATARS = 15;
+
+const displayName = 'dashboard.ColonyMembers';
+
+const ColonyMembers = ({ colony: { colonyAddress } }: Props) => {
+  /*
+   * @TODO This will most likely be replaces with a request to the reputation
+   * oracle, in order to fetch the users sorted by reputation
```
ceolson01, line 40: ❤️

ceolson01 (approved):
LGTM ✅

rdig:
Note that this is waiting for JoinColony/colonyJS#477 so that I can add in colony level member reputation fetching

PR #2285, opened 2020-11-01
Fix Readme Startup w/ Skipped Processes

rdig:
## Description

This is a small PR that just updates a small section of the readme file to correct the proper way to pass arguments to the dev startup script in order to skip certain processes from starting up _(eg: `webpack`)_.

**Changes**

- [x] Updated `Readme.md`


ceolson01 (approved):
Nice catch 👏

PR #2287, opened 2020-11-02
Colony Actions List

rdig:
## Description

This PR adds in the core events, actions, and logs list component, as well as the individual components that will display both actions and events on the dashboard, based on the previously mentioned core component.

**Reviewer's Notes**

This still uses mock data as I have no idea how the final one will look. Because of this:
- expected type has not been set (has been however suggested in a comment)
- filtering is just a boilerplate skeleton, still needs actual logic, based on the final data format
- the infini-loader functionality is very rudimentary as it just splits the array in chunks -- based on the final data format, if it doesn't make sense, it could even be removed

All of the above have been noted in code comments as well

**Changes**

- [x] Added core `ActionsList` component
- [x] Added core `ActionsListItem` sub-component
- [x] Added dashboard `ColonyActions` component
- [x] `ColonyHome` changed grid sizings
- [x] `ColonyHome` right aside padding

**Core Changes**
- [x] core `UserAvatar` , `InfoPopover` control popover arrow -- cherry picked from #2284 
- [x] core `UserMention` to display the user info popover as well
- [x] Added core `UserInfoNotAvailable` simple popover sub-component
- [x] core `InfoPopover` can display even if a user's info could not be fetched
- [x] core `Popover` prevent click event bubbling

**Demo**

cs-colony-actions-list

Wide-ass screenshot _(notice the full text of the long name actions showing in full)_
Screenshot from 2020-11-06 21-46-19

Resolves DEV-9

### DIFF
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -1,0 +1,3 @@
+.main {
+  display: block;
+}
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -1,3 +1,27 @@
+/*
+ * @NOTE Unique values
+ */
+@value mainTextSize: 13px;
+
 .main {
-  display: block;
+  display: flex;
+  align-items: center;
+  margin-bottom: 1px;
+  padding: 26px 20px;
+  border-radius: var(--radius-large);
+  background-color: var(--colony-white);
+  font-size: mainTextSize;
+  font-weight: var(--weight-bold);
+  color: var(--dark);
+  letter-spacing: var(--spacing-medium);
+  box-shadow: 0px 2px 11px var(--drop-shadow);
+}
+
+.avatar {
+  padding-right: 15px;
+}
+
+.avatar figure {
+  height: 37px;
+  width: 37px;
 }
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -74,3 +74,24 @@
   font-size: var(--size-tiny);
   color: var(--temp-grey-blue-7);
 }
+
+.separator::after {
+  display: inline-block;
+  margin: 0px 10px -2px;
+  height: 12px;
+  width: 1px;
+  background-color: var(--temp-grey-13);
+  content: '';
+}
+
+.date {
+  composes: separator;
+}
+
+.domain {
+  composes: separator;
+}
+
+.commentCount {
+  display: block;
+}
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -70,9 +70,11 @@
 }
 
 .meta {
-  margin-top: 2px;
-  font-size: var(--size-tiny);
+  margin-top: 3px;
+  font-size: var(--size-small);
+  font-weight: var(--weight-normal);
   color: var(--temp-grey-blue-7);
+  letter-spacing: var(--spacing-normal);
 }
 
 .separator::after {
@@ -86,6 +88,10 @@
 
 .date {
   composes: separator;
+}
+
+.day {
+  margin-right: 4px;
 }
 
 .domain {
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -18,6 +18,7 @@
   letter-spacing: var(--spacing-medium);
   box-shadow: 0px 2px 11px var(--drop-shadow);
   cursor: pointer;
+  outline: 0;
 }
 
 .main:hover {
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -17,6 +17,11 @@
   color: var(--dark);
   letter-spacing: var(--spacing-medium);
   box-shadow: 0px 2px 11px var(--drop-shadow);
+  cursor: pointer;
+}
+
+.main:hover {
+  background-color: var(--temp-grey-blue-8);
 }
 
 .avatar {
--- a/src/modules/core/components/ActionsList/ActionsListItem.css
+++ b/src/modules/core/components/ActionsList/ActionsListItem.css
@@ -8,6 +8,7 @@
   align-items: center;
   margin-bottom: 1px;
   padding: 26px 20px;
+  max-height: 89px;
   border-radius: var(--radius-large);
   background-color: var(--colony-white);
   font-size: mainTextSize;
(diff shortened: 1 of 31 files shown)


src/modules/core/components/ActionsList/ActionsListItem.css
```diff
@@ -0,0 +1,134 @@
...
+ */
+@value mainTextSize: 13px;
+@value popoverWidth: 400px;
+@value popoverDistance: 10px;
+
+.main {
+  display: flex;
+  margin-bottom: 1px;
+  padding: 26px 20px;
+  max-height: 89px;
```
alicjakujawa, line 12: I would suggest to set height instead of max-height, to always keep those items same height (even if some of them won't have avatar)

src/modules/core/components/ActionsList/ActionsListItem.css
```diff
@@ -0,0 +1,134 @@
...
+  composes: status;
+}
+
+.stateYellow::before {
+  background-color: var(--golden);
+}
+
+.content {
+  flex: 1 1 100%;
+  min-width: 0;
```
alicjakujawa, line 81: What is the reason behind setting min-width to 0? 🤔

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -0,0 +1,207 @@
...
+                    },
+                  },
+                ],
+              }}
+            />
+          )}
+        </div>
+        <div className={styles.content}>
+          <div className={styles.title} title={item.title}>
+            <Decorate>{item.title}</Decorate>
```
alicjakujawa, line 148: I would make sure that item.title exist - if title is null, then app is blowing

src/modules/core/components/ActionsList/ActionsListItem.css
rdig: Sure

src/modules/core/components/ActionsList/ActionsListItem.css
rdig: This is minor "hack" in order to get text ellipsis to work properly.

You are right I should have left a comment explaining this here. And I will :)

src/modules/core/components/ActionsList/ActionsListItem.tsx
rdig: Nice catch! I'll make the change.

Also:
> app is blowing

:rofl:

src/modules/core/components/InfoPopover/MemberInfoPopover.tsx
```diff
@@ -49,7 +49,11 @@ const MSG = defineMessages({
 
 const displayName = 'InfoPopover.MemberInfoPopover';
 
-const MemberInfoPopover = ({ colonyAddress, domainId, user }: Props) => {
+const MemberInfoPopover = ({
+  colonyAddress,
+  domainId,
+  user = { id: '', profile: { walletAddress: '' } },
```
alicjakujawa, line 55: what is the reason behind those defaults? (if we check for existence below)

src/modules/core/components/InfoPopover/UserInfoPopover.tsx
```diff
@@ -2,19 +2,23 @@ import React from 'react';
...
 
 const displayName = 'InfoPopover.UserInfoPopover';
 
-const UserInfoPopover = ({ user }: Props) => (
+const UserInfoPopover = ({ user, userNotAvailable = false }: Props) => (
   <div className={styles.main}>
     <div className={styles.section}>
-      <UserInfo user={user} />
+      {userNotAvailable && <UserInfoNotAvailable />}
+      {!userNotAvailable && user && <UserInfo user={user} />}
```
alicjakujawa, line 21: {!userNotAvailable && user ? <UserInfo user={user} /> : <UserInfoNotAvailable />
Just to handle weird edgecase when userNotAvailable is false but user = null

src/modules/dashboard/components/ColonyHome/ColonyHome.css
```diff
@@ -22,9 +37,17 @@
 }
```
alicjakujawa, line 37: do we want to keep empty class name?

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -161,7 +162,14 @@ const ColonyHome = ({ match, location }: Props) => {
               path={COLONY_EXTENSIONS_ROUTE}
               component={() => <>Extensions</>}
             />
-            <Route path={COLONY_HOME_ROUTE} component={() => <>Actions</>} />
+            <Route
+              path={COLONY_HOME_ROUTE}
+              component={() => (
+                <div className={styles.actions}>
+                  <ColonyActions />
```
alicjakujawa, line 169: Not sure about this wrapper div with styles here? Maybe we should put it inside ColonyActions?

src/modules/core/components/InfoPopover/MemberInfoPopover.tsx
rdig: That's needed to provide basic defaults, in case the `user` prop is not passed in, since just below this, we are destructuring the user object.

See: 
https://github.com/JoinColony/colonyDapp/blob/b4502176d21c36a3f7c5dae08e00ec269ce7b213/src/modules/core/components/InfoPopover/MemberInfoPopover.tsx#L58-L60

Even if the `user` value is set, it will still fail on the first component render, until all props have loaded their correct value.

src/modules/core/components/InfoPopover/UserInfoPopover.tsx
rdig: Nice catch. I've made the change.

Also, you can either format your code suggestions manually using a [code block](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code-and-syntax-highlighting), or you can use Github's built-in suggestion feature: 
Screenshot from 2020-11-12 19-59-15

src/modules/dashboard/components/ColonyHome/ColonyHome.css
rdig: Haha! That's not empty, just that Github tricked you :)

See this:
Screenshot from 2020-11-12 20-00-33

It means it hid lines of code it deemed not important. 

If you actually look at the file, you'll see that the left aside has some `padding`:
https://github.com/JoinColony/colonyDapp/blob/b4502176d21c36a3f7c5dae08e00ec269ce7b213/src/modules/dashboard/components/ColonyHome/ColonyHome.css#L31-L33

rdig:
@alicjakujawa made the changes that you requested in your review. 

You can have another pass at it when you have the time.

If you are good with the changes, please approve it so that I can merge it.

alicjakujawa (approved):
Tested locally, works great 🔥 
(The only thing which seems to be missing is to sorting list when changing value here:
Screenshot 2020-11-12 at 19 51 56

But those actions are from mockData, so i believe that its ok for now?

rdig:
> But those actions are from mockData, so i believe that its ok for now?

You are correct. I don't actually know how the final actions interface will look like, so I can't really set up sorting properly

PR #2292, opened 2020-11-16
Core Permission Labels Component

alicjakujawa:
## Description

- This PR creating Permission core component and adding needed icons

**New stuff** ✨

* Permission component
* Icons for permissions
* Default permissions object

**How to use** ✨
You can either just use default icon/label component for Permission example:
<Permission permission="root"/>
Or you can change icon|name or both with:
<Permission permission="root" icon="other-permission-icon" />

All permissions with defaults:
Screenshot 2020-11-17 at 17 51 42
Permission with sample tooltip:
Screenshot 2020-11-17 at 17 53 41

Resolves DEV-53


src/img/icons/emoji-alarm-lamp.svg
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
rdig, line 1: Very nice work on the `SVG`s :100:

src/modules/core/components/Permission/Permission.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import { MessageDescriptor, useIntl } from 'react-intl';
+import Icon from '~core/Icon';
+import { Tooltip } from '../Popover';
+import styles from './Permission.css';
+import { permissionsObject } from './permissions';
+
+interface Props {
+  /** Permission name */
+  permission:
```
rdig, line 11: Please replace these types with the official ones imported from `colonyJS`
```js
import { ColonyRole } from '@colony/colony-js'
```
And `ColonyRole` will contain all the roles we need:

Screenshot from 2020-11-17 21-09-04

src/modules/core/components/Permission/Permission.tsx
```diff
@@ -0,0 +1,79 @@
...
+    typeof infoMessage === 'string'
+      ? infoMessage
+      : infoMessage && formatMessage(infoMessage);
+
+  return (
+    <Tooltip
+      placement="bottom"
+      content={tooltipText || null}
+      trigger={inherited && infoMessage ? 'hover' : 'disabled'}
+    >
```
rdig, line 60: Please add an offset to the tooltip, so it won't be rendered over the label.

Something like:
```suggestion
    <Tooltip
      placement="bottom"
      content={tooltipText || null}
      trigger={inherited && infoMessage ? 'hover' : 'disabled'}
      popperProps={{
        modifiers: [
          {
            name: 'offset',
            options: {
              offset: [0, 20],
            },
          },
        ],
     }}
    >
```

You can read more about the offset here: https://popper.js.org/docs/v2/modifiers/offset/

src/modules/core/components/Permission/permissions.ts
```diff
@@ -0,0 +1,55 @@
...
+    id: 'core.Permission.arbitrationLabel',
+    defaultMessage: 'Arbitration',
+  },
+  recoveryLabel: {
+    id: 'core.Permission.recoveryLabel',
+    defaultMessage: 'Recovery',
+  },
+});
+
+export const permissionsObject = {
```
rdig, line 30: Please type this object

src/modules/core/components/Permission/Permission.tsx
```diff
@@ -0,0 +1,79 @@
...
+  /** Permission name. If empty, default will be taken from permissions object */
+  name?: MessageDescriptor | string;
+
+  /** Optional info message about permission being inherited. */
+  infoMessage?: MessageDescriptor | string;
+}
+
+const displayName = 'Permission';
+
+const Permission = ({
```
rdig, line 34: Please rename this to `PermissionsLabel`

src/modules/core/components/Permission/Permission.css
```diff
@@ -0,0 +1,26 @@
+.wrapper {
+  display: inline-block;
+  padding: 1px 5px;
+  height: 20px;
+  border: 1px solid var(--temp-grey-blue-7);
+  border-radius: 5px;
+  cursor: pointer;
```
rdig, line 7: Only show the pointer cursor over them if the tooltip text is set.

For this use can use the `getMainClasses` util from `~utils/css`

src/modules/core/components/Permission/Permission.tsx
```diff
@@ -0,0 +1,79 @@
...
+  const tooltipText =
+    typeof infoMessage === 'string'
+      ? infoMessage
+      : infoMessage && formatMessage(infoMessage);
+
+  return (
+    <Tooltip
+      placement="bottom"
+      content={tooltipText || null}
+      trigger={inherited && infoMessage ? 'hover' : 'disabled'}
```
rdig, line 59: This logic is not correct. This will only show a tooltip if inherited. We might need to show a tooltip even if the permission is not inherited.

Please change this to:
```suggestion
      trigger={infoMessage ? 'hover' : 'disabled'}
```

src/modules/core/components/Permission/Permission.css
```diff
@@ -0,0 +1,26 @@
+.wrapper {
```
rdig, line 1: I think the labels should have some small `margin-right` value applied to them. Something like:
```css
margin-right: 4px;
```

If you really are in the mood, you can add an `Apperance` prop to this, which just have the margin option:
```js
interface Appearance {
  margin: 'small' | 'none';
}
```

You can look up an example of how to set this up in one of the other core components

rdig (approved):
Overall this is exactly what I was expecting. Nice work! :taco: 

With that said, I did find stuff for you to improve in this PR. 

Since I'll be away, I've pre-approved this, but **please make the required changes before merging it**.

src/modules/core/components/Permission/Permission.css
rdig: This can be used, inside the component something like this:
```js
<div className={getMainClasess(styles, {}, noPointer: !infoMessage)}>...</div>
```

then create a `noPointer` class that just removes the `pointer`:
```css
.noPointer {
  cursor: default;
}
```

The way this works is that it will apply the `main` class by default, and if `infoMessage` is not set, then it will also apply the `noPointer` class

src/modules/core/components/Permission/Permission.css
alicjakujawa: Being super honest, i think it shouldn't be permissions responsibility, it should be wrapper component responsibility - so in case when we display permissions as a list - list will make sure that children have correct space between. If we display them alone - they just don't have

PR #2293, opened 2020-11-17
Colony Version Upgrade Banner

rdig:
## Description

This PR adds in a banner on the `ColonyHome` page that is displayed when a new version of the network contracts is available so that the colony can be upgraded.

Right now it functions in "dumb" mode, meaning it will show the banner for each new version of the contracts. In the future, we might refactor this to keep track of the individual contract version, and only display this banner for specific versions.

I created a local apollo resolver that now fetches all the values we need from the network, currently just `version` and `feeInverse`, and since the redux setup was no longer necessary I **removed** the `core` `Network` redux store.

So dear reviewer, sorry that this PR got so long, as it's mostly cleanup and refactoring, rather than actually adding new features in.

**New stuff**

- [x] `NetworkContracts` local resolver and helper functions
- [x] Added colony upgrade alert banner to `ColonyHome`

**Changes** 

- [x] refactored `setupUserContext` saga to use the new `NetworkContracts` resolvers
- [x] refactored `admin` `ProfileAdvanced` to use the new `NetworkContracts` resolvers
- [x] refactored `dashboard` task `FeeInverse` to use the new `NetworkContracts` resolvers

**Deletions** 

- [x] Removed all `redux` logic related to the `core` `Network` store

**Demo**

cs-colony-upgrade

Resolves DEV-11


src/data/graphql/typeDefs.ts
```diff
@@ -18,6 +18,16 @@ export default gql`
     networkId: Int
   }
 
+  input NetworkContractsInput {
+    version: String
+    feeInverse: String
```
alicjakujawa, line 23: are those for sure optional?

src/modules/core/components/Alert/Alert.css
```diff
@@ -58,3 +58,15 @@
 .borderRadiusRound {
   border-radius: 50px;
 }
+
+.borderRadiusNone {
+  border-radius: 0;
+}
+
+.marginNormal {
```
alicjakujawa, line 66: naming.. :P Maybe marginDefault ?

src/data/generated.ts
```diff
@@ -1326,6 +1334,16 @@ export type LoggedInUser = {
   networkId?: Maybe<Scalars['Int']>;
 };
 
+export type NetworkContractsInput = {
+  version?: Maybe<Scalars['String']>;
+  feeInverse?: Maybe<Scalars['String']>;
```
alicjakujawa, line 1339: I have small talk with @area about feeInverse. As fee inverse is coming straight from the contract and we cannot change it from FE. So the question is, if we passing it as a props because we want save it in cache? 🤔  or there is different reason behind that?

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -54,6 +60,12 @@ const MSG = defineMessages({
     id: 'dashboard.ColonyHome.noFilter',
     defaultMessage: 'All Transactions in Colony',
   },
+  upgradeRequired: {
+    id: `dashboard.ColonyHome.upgradeRequired`,
+    defaultMessage:
+      /* eslint-disable-next-line max-len */
+      'This colony uses a version of the network that is no longer supported. You must upgrade to continue using this application.',
+  },
```
alicjakujawa, line 68:
```suggestion
  upgradeRequired: {
    id: `dashboard.ColonyHome.upgradeRequired`,
    defaultMessage: `This colony uses a version of the network that is no
    longer supported. You must upgrade to continue using this application.`,
  },
```

src/data/generated.ts
rdig: Nope, you are correct. It just because we save it in the local resolver's cache.

This is just for us to be able to set it locally.

src/data/graphql/typeDefs.ts
rdig: Yes.

Since this is coming from the contracts, the might have issues connecting, through the provider, to the network, in which case these values will be undefined.

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
rdig: Sure. I'll make the change.

src/modules/core/components/Alert/Alert.css
rdig: You are right. I'll make the change.

alicjakujawa (approved):
Just rebase and good to go 🔥

PR #2295, opened 2020-11-20
Add Colony Members Page

alicjakujawa:
## Description

- This PR create Members Page which display all members of colonies. Each item include information about reputation and permissions. 

**New stuff** ✨

* ColonyMembersPage
* UseTokenInfo hook
* MEMBERS_ROUTE and MEMBERS_BY_DOMAIN_ROUTE

**Changes** 🏗

* Updated UserPermissions with new PermissionsLabel component
* Updated UserMention with smaller font size
* Updated MemberListItem component
* Removed outline from Select component button
* Updated PermissionsLabel component - added minimal bool

**Deletions** ⚰️

* Community component
* Permissions component

## TODO

- Test useTokenInfoHook with user reputation data
- Test displaying reputation with user reputation data 
- Add filtering based on domain dropdown
- Add filtering based on domain from route (MEMBERS_BY_DOMAIN_ROUTE)

Full page:
Screenshot 2020-11-23 at 17 52 27

Members page permissions with mocked reputation:
Screenshot 2020-11-23 at 14 53 07

Updated permissions page:
Screenshot 2020-11-23 at 14 52 58

(Resolves | Contributes to) #31415

PR #2302, opened 2020-11-26
Wire Domain and Reputation Filtering

rdig:
## Description

This PR wires in, mostly in `ColonyHome` domain filtering, as well as member by reputation filtering _(which are related)_.

Alongside this, it will properly set up the members by domains route _(once #2295 lands)_ with the proper members filtered.

**Changes** 
- [x] `ColonyHome` set up domain filtering by either location query or state
- [x] Added `colonyMembersWithReputation` client resolver
- [x] Fix `UserPermissions` to only show permissions if available
- [x] Various fixes to the `Members` page
- [x] `Members` page now can be accessed based on a domain id
- [x] Some minor fixes that I've found along the way

**Demo**

cs-filter-domain-reputation

### Testing Notes

Note, that while testing this locally you will run into an edge case where the number of users displayed in a particular domain is bigger the the "total" number of subscribed users. That is because we're using a [mock reputation oracle](https://github.com/JoinColony/mock-oracle).

I've described this issue in more detail on Discord:
- https://discord.com/channels/562263648173555742/611581703026245689/781904708053893130
- https://discord.com/channels/562263648173555742/611581703026245689/781904759372120074

Resolves DEV-55


rdig:
I have no idea why the tests suddenly decided to fail. It's not something introduced here. Not even the underlying packages were updated. Honestly, this is just bizarre.

Screenshot from 2020-11-28 00-39-53

So the `node-proper-lockfile` library that `jest` uses under the hood fails. 

Now the owners of `node-proper-lockfile` claim the issue lies with `jest` as they didn't implement the spec correctly, and because of this, refuse to fix it: https://github.com/moxystudio/node-proper-lockfile/issues/90

There's even a PR to fix this, but apparently, they don't want to merge it because of the above reason _(which, frankly, is understandable)_: https://github.com/moxystudio/node-proper-lockfile/pull/91

I'll have a look on Monday to see if I can't get to the bottom of this, or, at the very least, find a temporary solution for this. _(well... I know of one, but I want a "proper" one)_

rdig:
[This](https://github.com/JoinColony/colonyDapp/pull/2302#issuecomment-735005308) gets even weirder: https://github.com/JoinColony/colonyDapp/pull/2302#issuecomment-735005308 since this is not reproducible on the base branch.

So that makes me want to reach the conclusion that it's something I introduced?

But I didn't touch anything related to either those tests or the components those tests, well... test.

Truly weird

src/data/resolvers/colony.ts
```diff
@@ -49,6 +80,45 @@ export const colonyResolvers = ({
...
+      /*
+       * @NOTE About zero reputation an decay
+       *
+       * Initially a user has 0 reputation in the colony.
+       * By default that user won't be returned in the below query.
+       *
+       * Let say that then, a user aquires some reputation, which, after a time, decays back to zero.
+       * That user will be returned in the below query.
+       *
+       * Maybe at some point we want to consider filering them out.
```
alicjakujawa, line 113: filtering #typo 😈

src/modules/core/components/MembersList/MembersListItem.tsx
```diff
@@ -110,21 +112,27 @@ const MembersListItem = <U extends AnyUser = AnyUser>(props: Props<U>) => {
...
             showInfo={!onRowClick || showUserInfo}
             domainId={domainId}
             notSet={false}
           />
         </div>
         <div className={styles.usernameSection}>
-          {displayName && (
-            <span className={styles.displayName} title={displayName}>
-              {displayName}
+          {userProfile.profile.displayName && (
```
alicjakujawa, line 122: Are we totally sure that userProfile won't be null?

src/modules/core/components/MembersList/MembersListItem.tsx
```diff
@@ -110,21 +112,27 @@ const MembersListItem = <U extends AnyUser = AnyUser>(props: Props<U>) => {
...
+              {userProfile.profile.displayName}
             </span>
           )}
-          {username && (
+          {userProfile.profile.username && (
             <span className={styles.username}>
-              <UserMention hasLink={false} username={username} />
+              <UserMention
+                hasLink={false}
+                username={userProfile.profile.username}
```
alicjakujawa, line 134: i feel that it would be better to destruct this object once at the top
{username, displayName} = userProfile.profile

src/modules/core/components/MembersList/MembersListItem.tsx
rdig: I think so...? But I'll check anyway :)

src/modules/core/components/MembersList/MembersListItem.tsx
rdig: Sure

alicjakujawa:
@rdig So the thing which makes node build filing is the change with exporting/importing useTokenInfo
If you revert it old state - so importing like import { useTokenInfo } from '~utils/hooks/useTokenInfo';
Tests are green again

rdig:
> @rdig So the thing which makes node build failing is the change with exporting/importing useTokenInfo
If you revert it to old state - so importing like that: import { useTokenInfo } from '~utils/hooks/useTokenInfo';
Tests are green again

Oh. Wow. Nice going finding this out! I owe you one :tumbler_glass:

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -83,11 +83,16 @@ const ColonyHome = ({ match, location }: Props) => {
...
   };
-  const filteredDomainId = domainFilter
-    ? parseInt(domainFilter, 10) || COLONY_TOTAL_BALANCE_DOMAIN_ID
+
+  const [domainIdFilter, setDomainIdFilter] = useState<number>(
+    Number(queryDomainFilterId),
+  );
+
+  const filteredDomainId = domainIdFilter
+    ? domainIdFilter || COLONY_TOTAL_BALANCE_DOMAIN_ID
```
alicjakujawa, line 95:
```suggestion
   const filteredDomainId = domainIdFilter || COLONY_TOTAL_BALANCE_DOMAIN_ID
```

src/modules/dashboard/components/ColonyHome/ColonyMembers/ColonyMembers.tsx
```diff
@@ -22,64 +24,68 @@ const MSG = defineMessages({
...
   });
 
+  const membersPageRoute = useMemo(() => {
+    const baseRoute = `/colony/${colonyName}/members`;
+    if (currentDomainId === COLONY_TOTAL_BALANCE_DOMAIN_ID) {
+      return baseRoute;
+    }
+    return `${baseRoute}/${currentDomainId}`;
+  }, [currentDomainId, colonyName]);
+
```
alicjakujawa, line 65: i feel that we don't need useMemo here, its quite inexpensive operation, we're just returning correct string

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
rdig: Nice catch. 

This had other logic in it initially, but apparently, I failed to clean it up properly.

src/modules/dashboard/components/ColonyHome/ColonyMembers/ColonyMembers.tsx
rdig: Ok

src/modules/core/components/MembersList/MembersListItem.tsx
rdig: Apparently, I was correct. `useUser` will always return a correct user object, with **at least** the wallet address populated.

Eg:
```js
{
  profile: {
    walletAddress: "0x0....."
  }
}
```

So that means we can always expect that `userProfile` is set, and we can always destructure from it, without either TS screaming at us or the React component throwing an error.

Here's two `userProfile`'s logged, one has a username registered, the otherone doesn't:
Screenshot from 2020-11-29 13-43-56

rdig:
@alicjakujawa this is ready for another look:
- fixed the things you pointed out in the review
- fixed the `tokenInfo` hook imports/exports that also fixed the tests. Thanks for this again!

alicjakujawa:
Code looks great. I have only w few doubts about functionality itself. Tested locally a few scenario and im not sure if this working as expected. 
1. Going from Colony Home by clicking on Members redirect me to Members page with All domains selected. This is **CORRECT**:
Screenshot 2020-11-30 at 07 12 33
2. Selecting "root" from dropdown list is showing above item + random users (i believe its because of our oracle mock server) So im guessing its still **CORRECT**
Screenshot 2020-11-30 at 07 14 24
3. Writing random number in url ex. "/colony/a/members/3" will display all mock-oracle data, and keep dropdown empty, this seems to be a bug:
Screenshot 2020-11-30 at 07 15 46
4. When we write letters in url as param ex. "/colony/a/members/ab" - we're fallback to "all domains" so its correct

What i think we're missing is, to check if param is a number, and if SO then check if this domain number actually exits, if not -> fallback to all domains

rdig:
> 3. Writing random number in url ex. "/colony/a/members/3" will display all mock-oracle data, and keep dropdown empty, this seems to be a bug:
Yes this is a bug, I'll fix it

> What i think we're missing is, to check if param is a number, and if SO then check if this domain number actually exits, if not -> fallback to all domains
Hmmm. I'm reluctant about this since it's kinda an extra operation just for one edge case. But I guess you're right, it's an edge case and we need to cover it.

src/data/resolvers/colony.ts
```diff
@@ -21,10 +29,33 @@ import {
...
+    ClientType.ColonyClient,
+    colonyAddress,
+  );
+  const { skillId } = await colonyClient.getDomain(domainId);
+  /*
+   * @TODO This is coming from colonyJS as I forgot to adjust the proper types when
+   * creating this common extension.
+   * I'll put in a PR to fix this at some point
+   */
+  // @ts-ignore
```
rdig, line 49: I've put a PR to fix in this colonyJS: https://github.com/JoinColony/colonyJS/pull/480

rdig:
@alicjakujawa fixed via 9b964e5

rdig:
Refixed via e4c8622

alicjakujawa (approved):
Perfect! 🔥

PR #2303, opened 2020-11-26
Create Default Route Layout

rdig:
## Description

As the title implies this is a simple PR that just adds a new route layout, called `Complex`. It is meant to be used in places where you need the subscribed colonies list displayed in the left sidebar of your component route, as well as a backlink.

Note that it also includes the `SimpleNav` components as those are vital to actual navigation. _(eg: User wallet controls)_

**Changes**

- [x] Added `Complex` route layout
- [x] Move `SimpleNav` into it's own folder
- [x] Move `Plain` into it's own folder
- [x] Move `NavBar` into it's own folder
- [x] Refactored `NavBar` to support passing in custom `className` prop
- [x] Refactor `LandingPage` to use the `Complex` route layout
- [x] Refactor `ColonyHome` to use the `Complex` route layout
- [x] Refactor `Members` to use the `Complex` route layout

**Screenshots**

Screenshot from 2020-11-26 18-14-31

Screenshot from 2020-11-26 18-28-59

Screenshot from 2020-11-27 15-40-17

Screenshot from 2020-11-27 15-40-34


src/modules/pages/components/RouteLayouts/Complex/Complex.tsx
```diff
@@ -0,0 +1,62 @@
...
+    hasBackLink = true,
+    backRoute,
+    backText,
+    backTextValues,
+    hasSubscribedColonies = true,
+  } = {},
+}: Props) => {
+  const location = useLocation<{ hasBackLink?: boolean }>();
+  const backLinkExists =
+    hasBackLink === undefined
```
alicjakujawa, line 30: !hasBackLink ?  
isn't enough?

src/routes/Routes.tsx
```diff
@@ -217,10 +217,11 @@ const Routes = () => {
           exact
           path={MEMBERS_ROUTE}
           component={ColonyMembers}
-          layout={NavBar}
+          layout={Complex}
           routeProps={({ colonyName }) => ({
             backText: ColonyBackText,
             backRoute: `/colony/${colonyName}`,
+            hasSubscribedColonies: true,
```
alicjakujawa, line 224: Didn't we decide that ColonyMembers shouldn't have subscribedColonies list?

alicjakujawa (approved):
Tested and works great 🔥  Two things for consideration:
1. If we want keep ColonySubsribedList on ColonyMembers page - then we need to update margin-left to calc(26.5% - 50px). 
2. Naming.. Not sure about complex layout... Maybe DefaultLayout or something like that? As this will be used quite often.  But this is just thought, naming is always hard xD

src/modules/pages/components/RouteLayouts/Complex/Complex.tsx
rdig: No, because you also set this, not just by props, but also by `react-router`'s state, so if it's `undefined` check the state, otherwise, use whatever the prop value is

src/routes/Routes.tsx
rdig: Just because it looked weird, but now it looks good :)

rdig:
> Naming.. Not sure about complex layout... Maybe DefaultLayout or something like that? As this will be used quite often. But this is just thought, naming is always hard xD

I don't like `DefaultLayout` either. I think I'll just go with `Default`.

But yeah, in this case, naming really is hard.

src/routes/Routes.tsx
rdig: But yeah, you're right. I think I got carried away. I'll remove it.

PR #2305, opened 2020-11-27
Add IndexModal

alicjakujawa:
## Description

- This PR adds in core IndexModal which will be used across the app for all modals which include action items - ex ActionsDialogModal

**New stuff** ✨

* IndexModal core component
* IndexModalItem component

**Changes** 🏗

* ColonyActionsDialog updates to use new IndexModal

**Deletions** ⚰️

* ColonyActionsDialogItem
* ColonyActionsDialog styles

**Modal with optional back button:**
Screenshot 2020-11-30 at 13 12 51
**Actions modal:**
Screenshot 2020-11-30 at 13 13 15

Resolves DEV-78


src/modules/core/components/IndexModal/IndexModal.tsx
```diff
@@ -0,0 +1,77 @@
...
+            appearance={{ theme: 'secondary' }}
+            onClick={back}
+            className={styles.backButton}
+          >
+            <Icon
+              appearance={{ size: 'normal' }}
+              name="caret-left"
+              title={MSG.backButton}
+            />
+            <FormattedMessage {...MSG.backButton} />
```
rdig, line 67: We have an id for this: https://github.com/JoinColony/colonyDapp/blob/e633b3b76d5e39ec4c6063f21dbdeb7c8de33392/src/i18n/en.json#L31

No need to declare another message descriptor

src/modules/core/components/IndexModal/IndexModalItem.tsx
```diff
@@ -2,37 +2,45 @@ import React from 'react';
...
   description,
-  disabled,
   icon,
+  onClick,
+  comingSoon,
 }: Props) => {
   return (
-    <div className={`${disabled ? styles.disabled : styles.content}`}>
+    <div
+      className={`${comingSoon ? styles.disabled : styles.content}`}
```
rdig, line 31: This can be achieved more cleanly with `getMainClasses`, but it's not major. If you don't have time, just leave it like this.

src/modules/core/components/IndexModal/IndexModalItem.tsx
```diff
@@ -42,7 +50,7 @@ const ColonyActionsDialogItem = ({
           <FormattedMessage {...description} />
```
rdig, line 50: So I think the description doesn't have the correct color:

Screenshot from 2020-11-30 19-48-25

And in the Figma file it looks like this:

Screenshot from 2020-11-30 19-48-29

rdig (approved):
Yep, this is exactly what I imagined when I spec'd this issue. Way to go reading my mind :smile: 

I've marked some very minor things, but this is good to go!

PR #2306, opened 2020-11-27
Actions Page

rdig:
## Description

This PR adds in the `ActionsPage` component which will be the main driver to explorer the actions that are created using the _Universal Actions Creator_

This PR just adds in the minimal needed code and logic to display the transactions and a few UI elements. The rest are all split up into their own smaller issues.

To access the page go to `/colony/<colony-name/tx/<transaction-hash>`

**Changes**

- [x] Added actions page route
- [x] Added `ActionsPage`
- [x] Added `ActionsPage` `Hash` sub-component
- [x] Added `ActionsPage` `TransactionHash` sub-component
- [x] Added `ActionsPage` `TransactionMeta` sub-component
- [x] Added `ActionsPage` `TransactionStatus` sub-component
- [x] Added `transaction` client resolver and related queries
- [x] Added `isTransactionFormat` web3 util

**Screenshots**

Basic actions "detection" via an event:
Screenshot from 2020-11-29 01-47-11

Screenshot from 2020-11-30 18-01-30

Transaction available, but not an action
Screenshot from 2020-12-01 17-05-12

Transaction failed:
Screenshot from 2020-11-29 01-49-26

Hash is not a transaction:
Screenshot from 2020-11-29 02-09-47

Transaction pending:
Screenshot from 2020-11-30 21-35-28
Screenshot from 2020-11-30 21-11-40

**Demo**

cs-demo-actions-page

### Notes on testing 

Here are some tips on how to extract or otherwise generate the required transactions to access the various states of the `ActionsPage`.

**1. Getting a transaction hash that registers as an "action":**

Go to the events list and using your React Dev Tools, extract one of the events' transaction hash. If it's an event, it will register as an action as well _(for now, at least)_. Use that transaction hash to access the page.

Screenshot from 2020-11-29 02-14-17

**2. Getting a successful transaction hash, but one that's not an event**

Send some token from one of your Metamask accounts to the other, but make sure you are connected to the local RPC endpoint _(`localhost:8545`)_. You can use your ganache accounts for this. Get the transaction hash from Metamask, then use that to access the page.

Screenshot from 2020-11-29 00-48-05

**3. Getting a failed transaction hash**

As before, use Metamask to send a transaction, but this time, enter your colony's native token address. That address can't handle receiving token transfers and so the transaction will fail. As before, get the transaction's hash and use that to access the page.

**4. Getting a wrongly formatted transaction hash**

Basically, enter anything that doesn't register as a 64-char hex string. Eg: `/colony/<colony-name>/tx/123`

**5. Getting a "pending" transaction**

You need to:
- stop the miner
- create a transaction
- get it's hash
- access the page with that has
- re-start the miner _(so that you don't break your dev envinronment)_

- To stop the miner call this rpc method via `curl`:
```bash
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"miner_stop","params":[]}' localhost:8545
```

- Create a transaction _(eg: create a new domain)_
- Get it's hash using Redux Devtools, by checking the payload of the `TRANSACTION_HASH_RECEIVED` action

Screenshot from 2020-11-30 21-15-44

- Access the page with that hash: you should see the page similar to the above screenshot _(the one with the golden dot)_

- Restart your miner:
```
curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"miner_start","params":[1]}' localhost:8545
```

Resolves DEV-42
Resolves DEV-52
Resolves DEV-51


src/data/resolvers/transactions.ts
```diff
@@ -211,3 +213,41 @@ export const getColonyUnclaimedTransfers = async (
...
+  Query: {
+    async transaction(_, { transactionHash, colonyAddress }) {
+      const { provider } = networkClient;
+      const colonyClient = await networkClient.getColonyClient(colonyAddress);
+      /*
+       * @TODO Optimze! Optimize! Optimize!
+       * We need to find a way to fetch specific events related to transaction
+       * As the way it's currently implement is not scalable at all...
+       */
+      const colonyEvents = await getColonyAllEvents(colonyClient);
```
rdig, line 229: I think I can improve this currently. Just try to fetch the event & log for the current transaction hash.

If that's is one of our contract tx's then it's an "action", otherwise it's just a normal transaction.

Then we can deal later with fetching subsequent events for this one particular transaction. But at any rate, it's better than fetching the whole events list and filtering it.

rdig: Nope. I was wrong.

Turns out that currently we don't have a way to fetch this for a single transaction hash, so for now we still have to rely on fetching all then filtering.

This is something I have to figure out, the most likely have to build it in `colonyJS`

But as it is, it can stay like this for now.

rdig: Nope again. I was wrong about being wrong.

Turns out Alex saved me once again. :tada: 

I'll put it in.

src/data/graphql/typeDefs.ts
```diff
@@ -18,6 +18,20 @@ export default gql`
     networkId: Int
   }
 
+  type ParsedEvent {
+    name: String
+    topic: String
+    values: String
```
rdig, line 24: I think I need help typing this properly, since `values` will be an object, of an _mostly -_ unknown shape

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+/*
+ * @TODO We need a decent way of detecting pending transaction
+ *
+ * Currently we only get these two states back from the chanin, and even so
+ * we only get it for post-byzantium forks
+ *
+ * I think there's a way of looking at the value of `blockNumber` but I need
+ * to look into it further
+ */
```
rdig, line 76: I think I have a way to do this. 

Damn, this PR just keeps dragging me back ...

src/data/graphql/typeDefs.ts
ceolson01: Hmmm... since graphql doesn't really support _unknown schemas_ (to my knowledge, at least), maybe something like an array of tuples (as key-value pairs) would suffice here? Obviously there are some drawbacks here, such as type safety anywhere these values would be used but... better than nothing 🤷‍♂️

src/img/naked-mole-404.svg
```diff
@@ -0,0 +1,11 @@
+<svg viewBox="0 0 660 350" class="naked-mole">
+    <style type="text/css">
+        .st0{fill:#00284B;}
+        .st1{fill:#F0F5F7;}
+        .st2{fill:#F2ADBD;}
+        .st3{fill:#F8D0D8;}
+        .st4{fill:#F5416E;}
+        .st5{fill:#FFFFFF;}
+    </style>
```
rdig, line 9: This doesn't belong here, I need to remove it

src/data/graphql/typeDefs.ts
alicjakujawa: Are we sure that we don't know shape of this object? All possible values of this object aren't define by ColonyNetwork? We can probably type all of those key-values pairs and just make them optional as each event return different values?

rdig: They're unknown in the sense that to know them all, it would be a lot of work to find out. True they all come from the contracts, but basically, we would need to get through all the events and find out what they're values are, ... which are a lot

ceolson01: True, but might be worth it, as that would provide support for discriminated unions via fragments which I think would be really helpful. Similar to the `eventContext` from the server events

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
rdig: I've implemented the pending state via dfdd3292ba4cde86f2750ea200fc03a7ec0a28d7

src/img/naked-mole-404.svg
rdig: Cleaned up via 4ac182545fb18d4acd98465c15ba1727995bac67

src/data/resolvers/transactions.ts
```diff
@@ -211,3 +220,67 @@ export const getColonyUnclaimedTransfers = async (
...
+        } = transactionReceipt;
+        const events = logs
+          ?.map((log) => colonyClient.interface.parseLog(log))
+          /*
+           * If the above parser find events that are not part of the colony client
+           * it will return them as `null` so we filter them out
+           */
+          .filter((log) => !!log)
+          .map(({ name, values, topic }) => ({ name, values, topic }));
+        return {
```
rdig, line 256: A crap. We also need to return blockTime in here so we can get an accurate date.

I'll put this in tommorow

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -0,0 +1,283 @@
...
+           * for message descriptors, and that doesn't support our complex text values
+           */}
+          <h1 className={styles.heading}>
+            <FormattedMessage
+              {...MSG.actionTitle}
+              values={titleDynamicValues}
+            />
+          </h1>
+          {!events && hash && (
+            <TransactionHash
```
rdig, line 251: Remember to pass down the `createdAt` value from the block time here

src/modules/dashboard/components/ActionsPage/TransactionHash/TransactionHash.tsx
```diff
@@ -0,0 +1,46 @@
...
+      <div className={styles.status}>
+        <TransactionStatus status={status} />
+      </div>
+    )}
+    <div className={styles.transaction}>
+      <Hash transactionHash={transactionHash} />
+      {showMeta && (
+        <TransactionMeta
+          transactionHash={transactionHash}
+          createdAt={new Date()}
```
rdig, line 36: This needs to pass down the `createdAt` value from the block Time

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -0,0 +1,289 @@
...
+           */
+          return events[0].name;
+        }
+        return false;
+      })(),
+    }),
+    [transactionData, fallbackUserData, userData],
+  );
+
+  if (!isTransactionFormat(transactionHash) || transactionDataError) {
```
alicjakujawa, line 190: sooo... if transactionDataError then render?

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -0,0 +1,289 @@
...
+               * @NOTE Otherwise it interprets 0 as false, rather then a index
+               * Typecasting it doesn't work as well
+               */
+              status={typeof status === 'number' && STATUS_MAP[status]}
+            />
+          )}
+          <ul>
+            <b>Events for the tx:</b>
+            {events.map(({ name, topic }) => (
+              <li key={topic || ''}>
```
alicjakujawa: empty string as a key? i think it will throw console warning if we won't have topic

src/modules/dashboard/components/ActionsPage/TransactionHash/Hash.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+interface Props {
+  transactionHash: string;
+  title?: MessageDescriptor;
+}
+
+const Hash = ({ transactionHash, title = MSG.fallbackTitle }: Props) => (
+  <>
+    <p className={styles.title}>
+      <FormattedMessage {...title} />
```
alicjakujawa, line 27: did you test how it looks and works when title is undefined?

src/modules/dashboard/components/ActionsPage/TransactionHash/TransactionHash.tsx
```diff
@@ -0,0 +1,46 @@
...
+          status={status === STATUS.Failed ? status : undefined}
```
alicjakujawa: Not sure if i got it, so we want show status only if Status is Failed? 
How TransactionMeta component would look like with status = undefined? 
Wondering if we could just pass status here, and in TransactionMeta handle it
```if status===STATUS.FAILED then display something```

File: src/modules/dashboard/components/ColonyNavigation/ColonyNavigation.tsx [hunk 12]
```diff
@@ -46,9 +46,8 @@ const ColonyNavigation = () => {
         text: MSG.linkTextEvents,
       },
       {
-        disabled: true,
         extra: MSG.linkExtraExtensions,
-        linkTo: `/colony/${colonyName}/extensions`,
+        linkTo: `/colony/${colonyName}/admin`,
```
alicjakujawa: is it mistake :D?

src/modules/dashboard/components/FourOFour/FourOFour.tsx
```diff
@@ -3,7 +3,7 @@ import { defineMessages, FormattedMessage } from 'react-intl';
 import { Link } from 'react-router-dom';
 
 import Logo from '../../../../img/logo.svg';
-import NakedMole from '../../../../img/naked-mole.svg';
+import NakedMole from '../../../../img/naked-mole-404.svg';
```
alicjakujawa, line 6: 😆

src/modules/dashboard/components/ColonyNavigation/ColonyNavigation.tsx
rdig: Yes, but it was since fixed :)

src/modules/dashboard/components/FourOFour/FourOFour.tsx
rdig: Heh

rdig: Originally I think it was called "naked mole rat"

src/modules/dashboard/components/ActionsPage/TransactionHash/TransactionHash.tsx
rdig: This was fixed as well, I wonder why you're seeing it, since it's an outdated version.

Maybe refresh ?

src/modules/dashboard/components/ActionsPage/TransactionHash/Hash.tsx
rdig: When the `title` is undefined it falls back to `MSG.fallbackTitle`

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
rdig: I don't care about this, this mostly debug log. 

This will be replaced by the components you are building in DEV-47

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
rdig: This is:
- if the hash is not the correct format
- Or there was an error fetching the local resolver (this could happen if we can't get either the transaction or the transaction receipt)

alicjakujawa (approved):
🔥

PR #2311, opened 2020-12-02
Add DetailsWidget component

alicjakujawa:
## Description

- This PR adds in DetailsWidget component. Most of component props are optional, as we not always shows every row in widget. From/To props are as ReacNode as we can send from/to team OR person. DetailsWidgetUser and DetailsWidgetTeam handle from/to items design. 

**New stuff** ✨

* DetailsWidget component
* DetailsWidgetUser component
* DetailsWidgetTeam component

Screenshot 2020-12-03 at 18 01 37

## TODO

- Find out how to get "active team" domainID on ActionPage
- Get all possible action types and icons for them

Resolves DEV-45

PR #2309, opened 2020-12-01
Add Create Payment modal

ArmandoGraterol:
Add "Create Payment" modal structure and styles without it being connected to the network

**New stuff** ✨

* New Create Payment modal
* New PermissionRequiredInfo component to show the required permissions to the users.
* New `background` appearance property on DialogSection to display the background on `grey`

## How to test it

You could use `ColonyFunding` to test it this way:

- [ ] Go to `ColonyFunding` on `dashboard/components/ColonyHome`, and `import CreatePaymentDialog from '~dashboard/CreatePaymentDialog';`

- [ ] Go to line 44 and instead of using the `TransferFundsDialog`, use `CreatePaymentDialog`

- [ ] Finally, on line 53, just pass `colonyAddress` to `openDialog`, like this:

```
useCallback(
    () =>
      openDialog({
        colonyAddress,
      }),
    [openDialog, colonyAddress],
  );
```

With that, you can access the modal through the `Move Funds` link on the home

FireShot Capture 018 - Colony - localhost
FireShot Capture 021 - Colony - localhost

Resolves DEV-61

PR #2313, opened 2020-12-04
Expenditure Flow Wiring

rdig:
## Description

This PR wires in the first flow of UAC, the `Payment` flow, which after going through a bunch of modal steps, will result in creating an `OneTxPayment` transaction.

The implementation of this was halted half-way, due to us switching to fetching data using The Graph.

**Changes**

- [x] Wired `ExpenditureDialog`
- [x] Wired `CreatePaymentDialog`
- [x] Added `createPaymentAction` saga
- [x] Added `withMeta` redux action util
- [x] Added `MetaWithHistory` redux action type
- [x] Added expenditure payment action types: `COLONY_ACTION_EXPENDITURE_PAYMENT`
- [x] Added core `FriendlyUserName`
- [x] Added `events` utils
- [x] Updated TS library parsing target
- [x] Refactored: `ActionsPage`
- [x] Refactored: `ActionsPageFeed`
- [x] Refactored: `ActionsPageFeedEvent`
- [x] Refactored: `ActionsPageFeedItem`
- [x] Refactored: `ActionsPageComment`
- [x] Refactored: `TransactionStatus`
- [x] Refactored: `DetailsWidgetTeam`
- [x] Refactored: `DetailsWidget`
- [x] Refactored: `DetailsWidgetUser`
- [x] Refactored: `PermissionsLabel`

**Not doing anymore**
- Wiring the ColonyActions list
- Adding more events copy
- Adding more actions copy

**Reviewers Notes**
- Sorry for it being so big
- Make sure to fetch the latest changes!
- Both `colonyNetwork` and `colony-js` changed, so make sure to reinstall modules and run provision:
```shell
npm i
npm run provision
```

**Demos**

cs-demo-expenditure-flow-final

Resolves DEV-62
Resolves DEV-85
Contributes to DEV-54

Blocked by #2309 
Blocked by #2311 
Blocked by DEV-60


src/data/graphql/typeDefs.ts
```diff
@@ -20,19 +20,22 @@ export default gql`
...
-    from: String!
+    emmitedBy: String!
   }
 
-  type Transaction {
+  type ColonyAction {
     hash: String!
-    from: String!
-    to: String!
+    transactionInitiator: String!
```
alicjakujawa, line 30: if we're refactoring everything from transaction to action, we probably should name it actionInititor

src/data/resolvers/colonyActions.ts
```diff
@@ -0,0 +1,180 @@
...
+              .map((clientType) => {
+                const type = clientType?.clientType;
+                const potentialParsedLog = clientType?.interface.parseLog(log);
+                if (potentialParsedLog) {
+                  const { name, values } = potentialParsedLog;
+                  return {
+                    name,
+                    values,
+                    createdAt,
+                    emmitedBy: type,
```
alicjakujawa, line 89: so if type is null then we pass here null, and from what i saw above emitedBy is required, so we probably should add to if potentiatlParsedLog && type

src/data/resolvers/colonyActions.ts
```diff
@@ -0,0 +1,180 @@
...
+           */
+          const paymentAddedEvent = reverseSortedEvents?.find(
+            (event) => event?.name === ColonyAndExtensionsEvents.PaymentAdded,
+          );
+          const payoutClaimedEvent = reverseSortedEvents?.find(
+            (event) => event?.name === ColonyAndExtensionsEvents.PayoutClaimed,
+          );
+          const { paymentId } = paymentAddedEvent?.values;
+          paymentDetails = await getPaymentDetails(
+            paymentId,
```
alicjakujawa, line 123: what if this will be undefined?

src/data/resolvers/colonyActions.ts
```diff
@@ -0,0 +1,180 @@
...
+       * addressses have a user profile created. But that might be error prone.
+       */
+      const { hash, to } = await provider.getTransaction(transactionHash);
+
+      return {
+        hash,
+        fromDomain: 1,
+        to,
+        status: 2,
+        events: null,
```
alicjakujawa, line 168: Would this actually work? 🤔  if we defined type as events: [ParsedEvent!]!

src/data/graphql/typeDefs.ts
rdig: Agree

src/data/resolvers/colonyActions.ts
rdig: Type cannot be `null`. There is always a `clientType` on the client instance

src/data/resolvers/colonyActions.ts
rdig: It actually cannot be `undefined`. If that happens _somehow_ it means the contracts are actually broken, and the app crashes long before that :smile: 

Also, let's just consider that it is `undefined`, and the app didn't crash so far. In that case, `getPaymentDetails` will return `null`. So not that big of a deal.

I will however put a check for if `paymentDetails` exists, before trying to assign values from it.

src/data/resolvers/colonyActions.ts
rdig: Yes it will. This is how `pending` transactions work.

You can test this locally if you want by stopping the miner, creating a transaction, then going to that transaction hash's route.

alicjakujawa:
Please disable form in case when there is now available funds 
Screenshot 2020-12-11 at 11 26 45

alicjakujawa:
Another bug i noticed, when i created expenditure and i was redirected to ActionPage i saw blank page and a lot of errors in console:
Screenshot 2020-12-11 at 11 36 05
Same when i refresh the page

rdig:
> Another bug i noticed, when i created expenditure and i was redirected to ActionPage i saw blank page and a lot of errors in console:
> <Image ... />
> Same when i refresh the page

I cannot reproduce this. Please pull the latest version of the code, then try again.

From your screenshot, it points at not being able to map out the roles based on the event name. But that can't actually happen, for at least two reasons:
- The roles map is manually defined, locally: https://github.com/JoinColony/colonyDapp/blob/feature/dev62-expenditure-flow-wiring/src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx#L53-L56
- The `eventName` has a default in case it has a falsy value: https://github.com/JoinColony/colonyDapp/blob/feature/dev62-expenditure-flow-wiring/src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx#L61

rdig:
> Please disable form in case when there is now available funds
> <Image ...>

The fix for this was not pleasant at all, as we ran into an issue with `Formik` where it can't do both `validationSchema` and setting the error state manually _(calling `setError)_, which we needed when checking for domain available funds.

I had to add a forced error state to the core `Input` component, as well as track a parallel error state, in order for this work.

But in the end I've made it do what we wanted it to:

uac-demo-expenditure-domain-funds-validation

alicjakujawa:
Ive got same error, so for example if eventName is "ColonyRoleSet" 
ROLES_MAP[eventName] was returning - undefined for me, then we was trying to map over it so it was causing error
ROLES_MAP[eventName] && ROLES_MAP[eventName].map - solving problem for me

So either make sure that ROLES_MAP handle all events, or just add this additional check, to not iterate over undefined

ArmandoGraterol (approved):
A PR to end all PRs :1st_place_medal:

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
```diff
@@ -33,102 +37,134 @@ const MSG = defineMessages({
...
-    variables: { colonyAddress: colonyAddress || '' },
-  });
+  const showFullDetails = actionType !== ColonyActions.Generic;
 
-  useEffect(() => {
-    if (data) {
-      const domain = data.colony.domains.find(
-        ({ ethDomainId }) => Number(domainId) === ethDomainId,
+  const shortenedHash = useMemo(() => {
+    if (!showFullDetails && transactionHash) {
```
alicjakujawa, line 89: I would consider to not use useMemo here, as normal operation on string seems to be not super expensive

alicjakujawa (approved):
Tested, works great 🔥

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
rdig: Sure

rdig: Fixed 03af527! :tada:

PR #2314, opened 2020-12-05
Expenditures Index Modal 

rdig:
## Description

This PR adds in the _Expenditures_ index modal, and, while I was at it, I've created a branching wizard hook, to help wire them all together when the time comes.

In the process, I took the liberty to fix some styles I found as not being in spec, in the core `IndexModal` component.

**Changes**

- [x] Added index `ExpendituresDialog`
- [x] Some fixes to the logic, types and styles of `core` `IndexModal`
- [x] Fix `ColonyTokenEditDialog` display name
- [x] Added `naiveBranchingDialogWizard` hook
- [x] Added a bunch of icons: `triangle-warning`, `emoji-dollar-stack`, emoji-superman`, `emoji-calendar`

**Notes on using the `naiveBranchingDialogWizard` hook**

To use this simple hook you have to first pass it an array of objects containing a dialog component, and optional props to be passed to said component.

This will inject the dialog components with an extra prop, `callStep`, which can be called at any time, using the component's display name _(eg: `"dashboard.ComponentName"`)_

To start the whole thing up, it will return an initial `callStep` method, which, as above, can be called using a component's name.

This is a branching wizard logic, so it doesn't care about a previous step or a next step. It can, and will, jump around in whatever order you tell it to.

Example:

Start Component:
```jsx
import { useNaiveBranchingDialogWizard } from '~utils/hooks';
import ComponentA from '../ComponentA';
import ComponentB from '../ComponentB';

const StartComponent = () => {
  const startWizardFlow = useNaiveBranchingDialogWizard([
    { component: ComponentA },
    { component: ComponentB, props: { buttonName: 'Go Back' }},
  ]);

  return (
    <button onClick={() => startWizardFlow('ComponentA')}>
      Start wizard
    </button>
  );
};
```

Component A:
```jsx
const ComponentA = ({ callStep }) => (
  <button onClick={() => callStep('ComponentB')}>
    Go to B
  </button>
);

ComponentA.displayName = 'ComponentA';
```

Component B:
```jsx
const ComponentB = ({ callStep, buttonName }) => (
  <button onClick={() => callStep('ComponentA')}>
    {buttonName}
  </button>
);

ComponentB.displayName = 'ComponentB';
```

**Demo**

cs-demo-expenditure-flow-1

With permissions check:
cs-demo-expenditure-flow-2-permissions

Resolves DEV-56
Resolves DEV-60


alicjakujawa:
One question to your notes about wizard hook. So from my understanding we can only pass/define props at the beginning - when we start using hook, like you show for ComponentB. What if we would need props from component B in component D? So we won't be able to define them at the beginning but we need to have them somehow..

rdig:
> What if we would need props from component B in component D

In that case my hook isn't adquate. This is why I canlled it naive :)

It's built with the current flow in mind: Index modal -> Index modal -> Final modal, so no props can be passed around currently.

As I've mentioned:
https://github.com/JoinColony/colonyDapp/blob/3d954cf47754ab4a843b1ed17a4a4b7a9faa63cd/src/utils/hooks/naiveBranchingDialogWizardHook.ts#L52-L53

If we really need state between modals, I can build that in pretty easy, but it's not going to be a "clean" solution, as you would need to pass around values manually from each modal.

So because of that, I won't wire that in, unless we _really_ need it :)

src/modules/core/components/IndexModal/IndexModalItem.tsx
```diff
@@ -1,31 +1,44 @@
...
-  icon,
-  onClick,
-  comingSoon,
+  item: {
+    title,
+    description,
+    icon,
+    onClick,
+    comingSoon = false,
+    permissionRequired = false,
```
alicjakujawa, line 38: not sure if those are needed, if something wasn't passed, it will be undefined so !comingSoon will give false either way

rdig: Sure, you're right. 

I just have this habit of initializing all boolean values manually :)

src/utils/hooks/naiveBranchingDialogWizardHook.ts
```diff
@@ -0,0 +1,90 @@
...
+interface WizardComponent {
+  component: <P extends object>(potentialProps?: P) => WizardDialogType<P>;
+  props?: PotentialProps;
+}
+
+type WizardMap = Record<string, WizardComponent>;
+
+const createWizardMap = (potentialWizardSteps: Steps): WizardMap => {
+  const wizardMap = {};
+  potentialWizardSteps.map((potentialWizardStep, index) => {
```
alicjakujawa, line 30: why those steps are potential ? :P

src/utils/hooks/naiveBranchingDialogWizardHook.ts
```diff
@@ -0,0 +1,90 @@
...
+ * @NOTE Initial implementation of a flow-less dialog wizard
+ *
+ * The idea behind this, is that it injects a `callStep` method in each
+ * initial declared step, with which, you can reference and call each other
+ * step in the registered wizard.
+ *
+ * This way it doesn't matter if you want to skip one step, or two, or to a
+ * different flow entirely. The wizard doesn't care, since it's not linear anymore.
+ *
+ * Currently it doesn't support passing state between dialog calls, but that can
```
alicjakujawa, line 52: founder answer :D

alicjakujawa (approved):
Works nice 🔥

alicjakujawa:
> > What if we would need props from component B in component D
> 
> In that case my hook isn't adquate. This is why I canlled it naive :)
> 
> It's built with the current flow in mind: Index modal -> Index modal -> Final modal, so no props can be passed around currently.
> 
> As I've mentioned:
> https://github.com/JoinColony/colonyDapp/blob/3d954cf47754ab4a843b1ed17a4a4b7a9faa63cd/src/utils/hooks/naiveBranchingDialogWizardHook.ts#L52-L53
> 
> If we really need state between modals, I can build that in pretty easy, but it's not going to be a "clean" solution, as you would need to pass around values manually from each modal.
> 
> So because of that, I won't wire that in, unless we _really_ need it :)

I mean, you wouldn't need to pass values manually from each modal. We could use context provider and just saving props in shared state. To avoid props drilling. But either way its not important here :D

src/utils/hooks/naiveBranchingDialogWizardHook.ts
rdig: I can't remember :)

I think it has something to do with the fact that you can initialize, in theory, the hook with an empty array `[]`, in which case there isn't any step.

ArmandoGraterol (approved):
Looks good to me :ok_hand: First time seeing a modal stepper implemented from scratch, looks interesting

PR #2317, opened 2020-12-07
Add and update Dialog component themes

ArmandoGraterol:
Adds new theme for DialogSection and updates some existing styles of Dialog according to design specs

**New stuff** ✨

* Added "footer" theme for DialogSection component
* Added "heading" and "footer" theme to TransferFundsDialog

**Changes** 🏗

* Updated border-radius of Dialog
* Updated padding of Dialog heading and main sections

FireShot Capture 049 - Colony - localhost

Resolves DEV-41

### DIFF
--- a/src/modules/core/components/Dialog/Dialog.css
+++ b/src/modules/core/components/Dialog/Dialog.css
@@ -8,7 +8,7 @@
 
 .main {
   width: 560px;
-  border-radius: var(--radius-large);
+  border-radius: var(--radius-normal);
   background-color: var(--colony-white);
 }
 
--- a/src/modules/core/components/Dialog/Dialog.css
+++ b/src/modules/core/components/Dialog/Dialog.css
@@ -15,7 +15,7 @@
 .dialogOuterActions {
   display: flex;
   justify-content: flex-end;
-  margin-bottom: 5px;
+  margin-bottom: 12px;
 }
 
 .closeIconButton {
--- a/src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialogForm.tsx
+++ b/src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialogForm.tsx
@@ -356,6 +356,7 @@ const TransferFundsDialogForm = ({
           text={{ id: 'button.confirm' }}
           loading={isSubmitting}
           disabled={!isValid}
+          style={{ width: styles.wideButton }}
         />
       </DialogSection>
     </>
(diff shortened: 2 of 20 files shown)


src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialogForm.tsx
```diff
@@ -344,7 +344,7 @@ const TransferFundsDialogForm = ({
           />
         </div>
       </DialogSection>
-      <DialogSection appearance={{ align: 'right' }}>
+      <DialogSection appearance={{ align: 'right', theme: 'footer' }}>
```
rdig, line 347: The confirm button needs to be wider. Currently yours is `120px`, and it needs to be `160px`.

Look up how we implemented this in `OrganizationAddDomains`: 
https://github.com/JoinColony/colonyDapp/blob/05131f8ce4c9e23793d2ada77ba7026a318f4ff2/src/modules/admin/components/Domains/OrganizationAddDomains.tsx#L70

By overwriting the component styles directly, with a value set in the css modules, using `@value`:
https://github.com/JoinColony/colonyDapp/blob/05131f8ce4c9e23793d2ada77ba7026a318f4ff2/src/modules/admin/components/Domains/OrganizationAddDomains.css#L3

rdig (changes requested):
I'm good with this, however, with the exception of one, you didn't update all the modals.

Modals still needed to be updated:
- ColonyPermissionsAddDialog
- ColonyTokenEditDialog
- CreatePaymentDialog
- PermissionManagementDialog
- TokenMintDialog
- TransferFundsDialog

Make sure to update the button width on all of them

:+1:

src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialogForm.tsx
ArmandoGraterol: Mmmm I think it's gonna be better if we update the size of large buttons. According to the style guide, the large button is 160x44 so by updating the padding applied by the appearance, we should be fine

rdig: Not necesarily. 

The `large` button theme is in use elsewhere as well. So if you change that, you'll modify a lot of places in the app we might not want to change.

That's why I suggested you make the change locally just for the dialog.

src/modules/core/components/Dialog/Dialog.css
```diff
@@ -8,7 +8,7 @@
 
 .main {
```
rdig, line 9: Please increase the distance between the actual modal and the `circle-close` icon

Screenshot from 2020-12-10 21-00-21

rdig (approved):
I'm all good with this.

But before you merge it, please increase the distance between the modal and the close button.

Nice work :+1:

PR #2318, opened 2020-12-08
Add ColorSelect core component

alicjakujawa:
## Description

- This PR adds in ColorSelect core component

**New stuff** ✨

* ColorSelect core component

**Changes** 🏗

* Added new theme type to Select - Grid

Screenshot 2020-12-09 at 16 06 41

## TODO

- Remove code from ColonyHome which is responsible for displaying ColorSelect. Its provided just for easier testing.

Resolves DEV-124

### DIFF
--- a/src/modules/core/components/ColorSelect/ColorSelect.tsx
+++ b/src/modules/core/components/ColorSelect/ColorSelect.tsx
@@ -65,7 +65,7 @@ const ColorSelect = ({ disabled, activeOption, onColorChange }: Props) => {
   return (
     <Form<FormValues>
       initialValues={{
-        activeColor: activeOption || Color.Yellow,
+        activeColor: `${activeOption || Color.Yellow}`,
       }}
       onSubmit={() => {}}
     >
--- a/src/modules/core/components/ColorSelect/ColorSelect.tsx
+++ b/src/modules/core/components/ColorSelect/ColorSelect.tsx
@@ -69,7 +69,7 @@ const ColorSelect = ({ disabled, activeOption, onColorChange }: Props) => {
       }}
       onSubmit={() => {}}
     >
-      <div className={styles.wrapper}>
+      <div className={styles.main}>
         <Select
           appearance={{
             theme: 'grid',
--- a/src/modules/core/components/ColorSelect/ColorSelect.tsx
+++ b/src/modules/core/components/ColorSelect/ColorSelect.tsx
@@ -44,7 +44,7 @@ const ColorSelect = ({ disabled, activeOption, onColorChange }: Props) => {
   const renderActiveOption = useCallback<
     (option: SelectOption | undefined, label: string) => ReactNode
   >(() => {
-    return <ColorTag color={activeOption || Color.Yellow} />;
+    return <ColorTag color={activeOption} />;
   }, [activeOption]);
 
   const options = useMemo<ComponentProps<typeof Select>['options']>(() => {
@@ -65,7 +65,7 @@ const ColorSelect = ({ disabled, activeOption, onColorChange }: Props) => {
   return (
     <Form<FormValues>
       initialValues={{
-        activeColor: `${activeOption || Color.Yellow}`,
+        activeColor: activeOption,
       }}
       onSubmit={() => {}}
     >
--- a/src/modules/core/components/ColorSelect/ColorSelect.tsx
+++ b/src/modules/core/components/ColorSelect/ColorSelect.tsx
@@ -1,0 +1,13 @@
+import React from 'react';
+
+const displayName = 'ColorSelect';
+
+interface Props {}
+
+const ColorSelect = ({}: Props) => {
+  return <div />;
+};
+
+ColorSelect.displayName = displayName;
+
+export default ColorSelect;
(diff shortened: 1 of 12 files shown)


rdig:
Please assign yourself to a PR you're working on, even if it's still draft. Thanks :+1:

src/modules/core/components/ColorSelect/ColorSelect.tsx
```diff
@@ -0,0 +1,94 @@
...
+  }, []);
+
+  return (
+    <Form<FormValues>
+      initialValues={{
+        activeColor: activeOption || Color.Yellow,
+      }}
+      onSubmit={() => {}}
+    >
+      <div className={styles.wrapper}>
```
rdig, line 72: Usually, we have the first "wrapper" class in the component named `main`.

Please try to do this from now on, for consistency's sake.

src/modules/core/components/ColorSelect/ColorSelect.tsx
```diff
@@ -0,0 +1,94 @@
...
+          appearance={{
+            theme: 'grid',
+          }}
+          elementOnly
+          label={MSG.labelColorSelect}
+          name="activeColor"
+          onChange={(val) => {
+            onChange(Number(val));
+          }}
+          options={options}
```
rdig, line 83: When re-opening the dropdown, the active color should have a blue outline, when shown along the other colors in the grid:

Screenshot from 2020-12-09 20-27-08

vs. the current one:

Screenshot from 2020-12-09 20-19-13

_(notice that no color is "selected" in the dropdown)_

rdig: Also, please ask _product_ if the "selected" color in the dropdown should have that white dot in the middle:

Screenshot from 2020-12-09 20-18-41

I personally think it's an error in the figma file, but please check.

src/modules/core/components/ColorSelect/ColorSelect.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  return (
+    <Form<FormValues>
+      initialValues={{
+        activeColor: activeOption || Color.Yellow,
+      }}
+      onSubmit={() => {}}
+    >
+      <div className={styles.wrapper}>
+        <Select
```
rdig, line 73: There's a bug where after either closing (by clicking outside) or closing by selecting a new color, the select component retains focus:

bug-color-select-focust-state

Notice how I need to click again, outside of the select component in order to get rid of the focus.

The focus on the element should be lost when closing it.

src/modules/core/components/ColorSelect/ColorSelect.tsx
```diff
@@ -0,0 +1,94 @@
...
+      }
+      return null;
+    },
+    [onColorChange],
+  );
+
+  const renderActiveOption = useCallback<
+    (option: SelectOption | undefined, label: string) => ReactNode
+  >(() => {
+    return <ColorTag color={activeOption || Color.Yellow} />;
```
rdig, line 47: Please make sure the active color is vertically centered with the caret icon. 

Currently it's not:

Screenshot from 2020-12-09 20-22-43

rdig (changes requested):
I like how it looks. 

It still has a couple of style issues that need to be fixed, as well as that bug where it retains focus after closing it.

After you fix the above it should be good to merge.

:taco:

src/modules/core/components/ColorSelect/ColorSelect.tsx
alicjakujawa: I don't think its a bug. It current implementation of Select.ts component:

  const close = useCallback(() => {
    setIsOpen(false);
    setSelectedOption(-1);
    // if (comboboxRef.current) {
    //   comboboxRef.current.focus();
    // }
  }, []);
I can remove if and it will fix problem - if we want

ArmandoGraterol (commented):
I noticed something strange... When I select `Pink` (The first color in the selector), the color showed in the input is `Yellow`, this doesn't happen to me with any other color, just `Pink`.

Other than that, and the thing that Raul mentioned about the carot not being centered, it looks good to me!

alicjakujawa:
> I noticed something strange... When I select `Pink` (The first color in the selector), the color showed in the input is `Yellow`, this doesn't happen to me with any other color, just `Pink`.
> 
> Other than that, and the thing that Raul mentioned about the carot not being centered, it looks good to me!

Great catch!

src/modules/core/components/Fields/Select/Select.tsx
```diff
@@ -124,9 +124,6 @@ const Select = ({
   const close = useCallback(() => {
     setIsOpen(false);
     setSelectedOption(-1);
-    if (comboboxRef.current) {
-      comboboxRef.current.focus();
-    }
```
rdig, line 129: Since you removed this piece of logic, the `ref` itself doesn't have a point anymore.

Please remove:
- the `useRef` call an import
- the `ref` prop

rdig (approved):
This is good.

But before you merge:
- remove the unused `ref` reference from `Select`
- remove the test from `ColonyHome`

src/modules/core/components/Fields/Select/Select.tsx
alicjakujawa: didint remove useRef import as useRef is also used by wrapper for handling outside click. Removed comboboxRef

PR #2319, opened 2020-12-09
Make tooltip copy message customisable

alicjakujawa:
## Description

- This PR adds in possibility to customise tooltip message. Message which is showing when we hover over address

**Changes** 🏗

* Added optional copyMessage prop to InvisibleCopyableAddress component. If message is not provided we are using default one
Screenshot 2020-12-09 at 10 26 11

Resolves DEV-125


rdig (approved):
Good to go!

PR #2321, opened 2020-12-10
Refactor MemberInfoPopover and TokenInfoPopover

ArmandoGraterol:
Refactor MemberInfoPopover and TokenInfoPopover so they handle cases where the user data is not available

**Changes** 🏗

* Moved MemberInfoPopover/TokenInfoPopover to MemberInfo/TokenInfo, and use the Popover to handle the case where the member/token data is not available.
* Changed UserInfoNotAvailable to NotAvailableMessage and added additional prop to make it more reusable.

FireShot Capture 058 - Colony - localhost
FireShot Capture 055 - Colony - localhost

Resolves DEV-96


src/modules/core/components/InfoPopover/MemberInfo.tsx
```diff
@@ -0,0 +1,163 @@
...
+                    title={formatMessage(MSG.achievementTitleText, {
+                      title,
+                      programTitle,
+                    })}
+                  />
+                ),
+            )}
+          </div>
+        </div>
+      )}
```
rdig, line 156: You can take this out as it's not needed anymore

rdig (approved):
I'm good with this too. Good turnaround on this! :taco: 

I've marked something that you should remove from `MemberInfo`, but after you've done that you can merge it.

PR #2322, opened 2020-12-10
Refactor TokenIcon component

alicjakujawa:
## Description

- This PR refactor current HookedTokenIcon component

**New stuff** ✨

* Fetching logos from trustwallet api
* Added iconName prop to HookedTokenIcon component

**Changes** 🏗

* If iconName provided then displaying Icon component instead of Avatar
*. Clearing not needed code

**Deletions** ⚰️

* eth-metadata-lib

Fetched ETH logo and blockie fallback for "A" token:
Screenshot 2020-12-10 at 18 18 03

Icon name provided:
Screenshot 2020-12-10 at 18 24 33

**TODO**
Find out if we can safely remove "name" prop from HookedTokenIcon component

Resolves DEV-122

### DIFF
--- a/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
+++ b/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
@@ -69,7 +69,7 @@ const HookedTokenIcon = ({
       }
     };
     loadTokenLogo();
-  }, [address, iconName]);
+  }, [address, iconName, dontFetch]);
 
   return (
     <>
--- a/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
+++ b/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
@@ -1,5 +1,7 @@
 import React, { useState, useEffect } from 'react';
+
 import { AddressZero } from 'ethers/constants';
+import { TOKEN_LOGOS_REPO_URL } from '~constants';
 
 import Avatar from '~core/Avatar';
 import { useDataFetcher } from '~utils/hooks';
@@ -29,11 +31,10 @@ interface Props {
 
   /** If provided than icon is display instead of Avatar */
   iconName?: string;
-}
 
-const TOKEN_LOGOS_REPO_URL =
-  'https://raw.githubusercontent.com/trustwallet' +
-  '/assets/master/blockchains/ethereum/';
+  /** If true logo fetching wont be fire */
+  dontFetch?: boolean;
+}
 
 const loadTokenImages = async (address: Address): Promise<Response> => {
   let tokenImageUrl = `${TOKEN_LOGOS_REPO_URL}${address}/logo.png`;
@@ -47,6 +48,7 @@ const HookedTokenIcon = ({
   name,
   token: { iconHash, address },
   iconName,
+  dontFetch,
   ...props
 }: Props) => {
   const [tokenImage, setTokenImage] = useState<string | undefined>();
@@ -58,7 +60,7 @@ const HookedTokenIcon = ({
 
   useEffect(() => {
     const loadTokenLogo = async () => {
-      if (address && !iconName) {
+      if (!dontFetch && address && !iconName) {
         const response = await loadTokenImages(address);
         if (!response.ok) {
           return;
@@ -72,11 +74,7 @@ const HookedTokenIcon = ({
   return (
     <>
       {iconName ? (
-        <Icon
-          appearance={{ size: 'medium' }}
-          name={iconName}
-          title={iconName}
-        />
+        <Icon name={iconName} title={name || address} {...props} />
       ) : (
         <Avatar
           avatarURL={tokenImage || ipfsIcon || undefined}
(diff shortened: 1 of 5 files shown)


src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -24,76 +26,67 @@ interface Props {
...
   name?: string;
+
+  /** If provided than icon is display instead of Avatar */
+  iconName?: string;
 }
 
-const checkSVG = (fileName: string) =>
-  fileName &&
-  fileName.substring(fileName.length - 3, fileName.length) === 'svg';
+const TOKEN_LOGOS_REPO_URL =
```
rdig, line 34: Please move this in `src/modules/constants.ts`

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -24,76 +26,67 @@ interface Props {
...
-      {...props}
-    >
-      {tokenSVG}
-    </Avatar>
+    <>
+      {iconName ? (
+        <Icon
+          appearance={{ size: 'medium' }}
+          name={iconName}
+          title={iconName}
```
rdig, line 78: title should be `{name || address}`

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -24,76 +26,67 @@ interface Props {
...
+        />
```
rdig, line 79: This overwrites the size of the icons, and it will always display the icon as `medium`, even if we don't want it to:

Screenshot from 2020-12-10 21-40-54

The component already receives a `size` prop, just that you're not passing it down. This should be, as the `Avatar` just passing down all the remaining `props` that the component receives.

Please change this to:
```suggestion
        <Icon name={iconName} title={name || address} {...props} />
```

This way, it will look proper in all instances:

Screenshot from 2020-12-10 21-41-54

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -24,76 +26,67 @@ interface Props {
...
-  import(
-    /* webpackMode: "eager" */ `../../../../../node_modules/eth-contract-metadata/images/${logo}`
-  );
+const loadTokenImages = async (address: Address): Promise<Response> => {
+  let tokenImageUrl = `${TOKEN_LOGOS_REPO_URL}${address}/logo.png`;
+  if (address === AddressZero) {
+    tokenImageUrl = `${TOKEN_LOGOS_REPO_URL}info/logo.png`;
+  }
+  return fetch(tokenImageUrl);
+};
```
rdig, line 44: I think there should be an extra prop to be passed down to this component: `dontFetch` that should prevent fetching of the token image. _(it should default to `false`)_

This way we can conditionally prevent fetches that we know are going to fail _(eg: for native tokens that won't have an icon)_

Screenshot from 2020-12-10 21-50-01

rdig (changes requested):
This is very good. It does everything I imagined this originally to do.

However, I did find some places that either need to be fixed or improved.

After you've done those, I'll re-review it.

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -24,76 +28,63 @@ interface Props {
...
+  dontFetch?: boolean;
+}
+
+const loadTokenImages = async (address: Address): Promise<Response> => {
+  let tokenImageUrl = `${TOKEN_LOGOS_REPO_URL}${address}/logo.png`;
+  if (address === AddressZero) {
+    tokenImageUrl = `${TOKEN_LOGOS_REPO_URL}info/logo.png`;
+  }
+  return fetch(tokenImageUrl);
+};
```
rdig, line 45: I'm all good with this, but I want to add more functionality. Sorry :)

We need an internal state that tracks if the icon `isFetching` and based on that show a loading spinner over the avatar. _(but only the avatar, there's no point in having it, if we just show a manual icon)_

This is because there's a flash of content while fetching, and when replacing the token avatar _(plus the user doesn't actually know that we're fetching in the background)_

demo-slow-token-loading

We have something similar for the `AvatarUploader` core component, _if you need_ to take inspiration from somewhere: https://github.com/JoinColony/colonyDapp/blob/master/src/modules/core/components/AvatarUploader/AvatarUploadItem.tsx#L79-L81

Otherwise, just implement this as you see fit :smile:

rdig: The way I'm seeing this look is an overlay over the existing blockie (just show the blockie initially), and that overlay has a small spinner loader inside of it.

When the file has finished loading, replace the blockie in the background (with the overlay still active), then remove the overlay entirely.

This way we both show a loading state to the user and prevent having the content flash white in the interim state after the file has been fetched and is being replaced.

rdig (changes requested):
So I'm all good with this. Very nice work.

The only reason I'm rejecting it is that I've realized we need more functionality when fetching the token icon, and that is to show a loading state.

_See the comments below_

Thanks! :100:

ArmandoGraterol (approved):
Looks good to me! :100:

rdig (approved):
Good to go! :+1:

PR #2328, opened 2020-12-14
Rewrite Icon core component

ArmandoGraterol:
Rewrite Icon component so it displays icons better

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/img/icons/circle-close-error.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="17" height="17" viewBox="0 0 17 17" fill="none" xmlns="http://www.w3.org/2000/svg">
```
rdig, line 1: height and width are not required tags in a svg file.

Also, why is the viewbox 17x17 ?

src/img/icons.json
```diff
@@ -19,6 +19,7 @@
     "circle-back",
     "circle-check",
     "circle-close",
+    "circle-close-error",
```
rdig, line 22: So this is just like `circle-close`, only it's red ? Couldn't that be achieved from css with `fill` ?

src/modules/core/components/Icon/Icon.tsx
```diff
@@ -75,6 +75,27 @@ const Icon = ({
...
+      case 'medium':
+        return '0 0 28 28';
+      case 'large':
+        return '0 0 42 42';
+      case 'huge':
+        return '0 0 64 64';
+      default:
+        return '0 0 18 18';
+    }
+  }, [appearance]);
```
rdig, line 97: Why you "can" programatically change the viewbox of svg elements, it doesn't make the actual `path` render better at different sizes -- that's why I suggested we need **different version** of file sizes

Also, the same effect can be achieved by setting a width and height for the icon component in css, wihtout involving js code _(but again, it will not look pixel-perfect for all sizes)_

rdig (changes requested):
>  They mentioned that we aren't ready for something like that,

Product doesn't decide what our code/app is ready or not ready for! Why didn't you get in touch with me about this? since this a pretty major change from the original scope of the issue.

Next time please come to me if something like this arises during implementation.

> Resolves DEV-120

This nowhere near resolves this.

This PR is not acceptable at all!
- it does not implement the scope it was originally supposed to
- it does not implement any actual improvements

src/img/icons.json
ArmandoGraterol: The lines of this icon are bolder, or wider than `circle-close`

PR #2329, opened 2020-12-15
TokenIcon component updates

alicjakujawa:
## Description

- This PR adds in saving icons to local storage to not fetch them every time

**New stuff** ✨

* Fetching images is not firing if on we're on dev environment
* After first image fetch, image is converted to base64image and save in localStorage

**Changes** 🏗

* Updated useEffect of HookedTokenIcon to check if image exist in local storage and if so, then take it. If not - we're normally fetching image

Resolves DEV-131

### DIFF
--- a/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
+++ b/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
@@ -48,7 +48,7 @@ const loadTokenImages = async (address: Address): Promise<Response> => {
 const getBase64image = (blob): Promise<any> => {
   return new Promise((resolve) => {
     const fileReader = new FileReader();
-    fileReader.onload = function () {
+    fileReader.onload = () => {
       return resolve(fileReader.result);
     };
     fileReader.readAsDataURL(blob);
--- a/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
+++ b/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
@@ -1,7 +1,8 @@
 import React, { useState, useEffect } from 'react';
+import { Network } from '@colony/colony-js';
 
 import { AddressZero } from 'ethers/constants';
-import { TOKEN_LOGOS_REPO_URL } from '~constants';
+import { TOKEN_LOGOS_REPO_URL, DEFAULT_NETWORK } from '~constants';
 
 import Avatar from '~core/Avatar';
 import { useDataFetcher } from '~utils/hooks';
@@ -50,7 +51,7 @@ const HookedTokenIcon = ({
   name,
   token: { iconHash, address },
   iconName,
-  dontFetch = process.env.NODE_ENV === 'development',
+  dontFetch = DEFAULT_NETWORK !== Network.Mainnet,
   ...props
 }: Props) => {
   const [tokenImage, setTokenImage] = useState<string | undefined>();
@@ -68,7 +69,7 @@ const HookedTokenIcon = ({
         return;
       }
 
-      if (address && !iconName) {
+      if (!dontFetch && address && !iconName) {
         const response = await loadTokenImages(address);
         if (response.ok) {
           const blob = await response.blob();
--- a/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
+++ b/src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
@@ -39,6 +39,8 @@ interface Props {
   dontFetch?: boolean;
 }
 
+const ICON_STORAGE = 'tokenImages';
+
 const loadTokenImages = async (address: Address): Promise<Response> => {
   let tokenImageUrl = `${TOKEN_LOGOS_REPO_URL}${address}/logo.png`;
   if (address === AddressZero) {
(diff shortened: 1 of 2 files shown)


src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -44,11 +45,21 @@ const loadTokenImages = async (address: Address): Promise<Response> => {
   return fetch(tokenImageUrl);
 };
 
+const getBase64image = (blob): Promise<any> => {
+  return new Promise((resolve) => {
+    const fileReader = new FileReader();
+    fileReader.onload = function () {
```
rdig, line 51: Why is this not an arrow function ?

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -44,11 +45,21 @@ const loadTokenImages = async (address: Address): Promise<Response> => {
   return fetch(tokenImageUrl);
 };
 
+const getBase64image = (blob): Promise<any> => {
```
rdig, line 48: I think this should be an util function, as we might use it for the user / colony avatar as well

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -44,11 +45,21 @@ const loadTokenImages = async (address: Address): Promise<Response> => {
...
+    fileReader.readAsDataURL(blob);
+  });
+};
+
 const HookedTokenIcon = ({
   name,
   token: { iconHash, address },
   iconName,
-  dontFetch,
+  dontFetch = process.env.NODE_ENV === 'development',
```
rdig, line 62: This should actually go by network, since that repo only has icon for mainnet ethereum

So something like:
```
DEFAULT_NETWORK !== Network.Mainnet
```

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
rdig, line 48: Why is this an `async` function, the only async call is done outside of this, when fetching the blob.

Also, this can be simplified as you don't need to set up the `onload` listener.

```js
const fileReader = new FileReader();
fileReader.readAsDataURL(blob);
return fileReader.result;
```

No async at all.

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -60,12 +71,22 @@ const HookedTokenIcon = ({
...
+        return;
+      }
+
       if (!dontFetch && address && !iconName) {
         const response = await loadTokenImages(address);
-        if (!response.ok) {
-          return;
+        if (response.ok) {
+          const blob = await response.blob();
+          const base64image = await getBase64image(blob);
```
rdig, line 84: As I've said above the `getBase64image` function doesn't need to be async

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -60,12 +71,22 @@ const HookedTokenIcon = ({
...
+          if (base64image) {
+            localStorage.setItem(address, base64image);
```
rdig, line 86: This sets the base64 string an address as a key value pair in the local storage. This has the potential of polluting the local storage way too much.

Please refactor this to store in under an object.

Something like:
```suggestion
const tokenImages = localStorage.getItem('tokenIcons`);
localStorage.setItem('tokenIcons', { ...tokenImages, [address]: base64image });
```

Then access it like that.

rdig (changes requested):
This is mostly good.

However, it still needs some work before it will be ready to merge.

I've outlined the required changes below.

rdig:
Also, why is not @ArmandoGraterol set as a reviewer to this PR?

Screenshot from 2020-12-16 10-22-51

ArmandoGraterol (commented):
Apart from what Raul mentioned already, this looks good to me! I'll review it again when the changes to Raul's comments are added in

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
alicjakujawa: With this, you will get only empty string

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
alicjakujawa: I was testing no async way, and i was getting empty string

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
rdig: hmmm... I remember testing locally and working

But I'm not saying you're not right :) Maybe I missed something.

I'll check again.

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
rdig: I'll have  a look again

rdig (approved):
Good to go now. This is a very nice improvment. :+1: 

_(Now if we can only find such a list of token images for `xdai`...)_

src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx
```diff
@@ -60,12 +65,37 @@ const HookedTokenIcon = ({
 
   useEffect(() => {
     const loadTokenLogo = async () => {
+      const imagesStorage = localStorage.getItem(ICON_STORAGE);
+      const parsedImagesStorage = imagesStorage && JSON.parse(imagesStorage);
+      if (parsedImagesStorage) {
+        const image = parsedImagesStorage[address];
+        if (image) {
+          setTokenImage(image);
+          return;
```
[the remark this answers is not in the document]
alicjakujawa: Lint don't like it 
```  67:36  error  Expected to return a value at the end of async arrow function  consistent-return```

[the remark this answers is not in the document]
rdig: Boo `eslint`!

Leave it in then;

PR #2331, opened 2020-12-15
Rework auth local storage token

ArmandoGraterol:
Reworked auth local storage tokens so they occupy 1 storage slot in the local storage instead of using 1 slot for each token

**Changes** 🏗

* Reworked setToken, getToken, clearToken so that they store, get, and clear tokens from a single localStorage slot
*  Added function to check if the localStorage slot is already in use

Resolves DEV-132

### DIFF
--- a/src/api/auth.ts
+++ b/src/api/auth.ts
@@ -16,16 +16,56 @@ const postRequest = async (path: string, data: object) => {
   return response.json();
 };
 
-export const setToken = (walletAddress: string, token: string) =>
-  localStorage.setItem(`${TOKEN_STORAGE}-${walletAddress}`, token);
-export const getToken = (walletAddress: string) =>
-  localStorage.getItem(`${TOKEN_STORAGE}-${walletAddress}`);
-export const clearToken = (walletAddress: string) =>
+export const setToken = (walletAddress: string, token: string) => {
+  const storedTokens = localStorage.getItem(TOKEN_STORAGE);
+
+  if (storedTokens) {
+    const parsedStoredTokens = JSON.parse(storedTokens);
+
+    parsedStoredTokens[`${TOKEN_STORAGE}-${walletAddress}`] = token;
+
+    localStorage.setItem(TOKEN_STORAGE, JSON.stringify(parsedStoredTokens));
+  }
+};
+
+export const getToken = (walletAddress: string) => {
+  const storedTokens = localStorage.getItem(TOKEN_STORAGE);
+
+  if (storedTokens) {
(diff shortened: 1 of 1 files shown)


src/api/auth.ts
```diff
@@ -16,16 +16,56 @@ const postRequest = async (path: string, data: object) => {
...
+    const parsedStoredTokens = JSON.parse(storedTokens);
+
+    return parsedStoredTokens[`${TOKEN_STORAGE}-${walletAddress}`];
+  }
+
+  return null;
+};
+
+export const clearToken = (walletAddress: string) => {
   localStorage.removeItem(`${TOKEN_STORAGE}-${walletAddress}`);
```
rdig, line 44: Why is this still here, this still tries to remove a token from the `${TOKEN_STORAGE}-${walletAddress}` key

src/api/auth.ts
```diff
@@ -16,16 +16,56 @@ const postRequest = async (path: string, data: object) => {
...
+export const setToken = (walletAddress: string, token: string) => {
+  const storedTokens = localStorage.getItem(TOKEN_STORAGE);
+
+  if (storedTokens) {
+    const parsedStoredTokens = JSON.parse(storedTokens);
+
+    parsedStoredTokens[`${TOKEN_STORAGE}-${walletAddress}`] = token;
+
+    localStorage.setItem(TOKEN_STORAGE, JSON.stringify(parsedStoredTokens));
+  }
```
rdig, line 28: What if we don't already have a seeded object in the local storage?

Please rewrite this to cover that case as well.

src/api/auth.ts
```diff
@@ -16,16 +16,56 @@ const postRequest = async (path: string, data: object) => {
...
+export const setTokenStorage = () =>
+  localStorage.setItem(TOKEN_STORAGE, JSON.stringify({}));
 
 export const authenticate = async (wallet) => {
   try {
+    const hasTokenStorageSet = !!localStorage.getItem(TOKEN_STORAGE);
+
+    if (!hasTokenStorageSet) {
+      setTokenStorage();
+    }
```
rdig, line 65: This should be moved in the `setToken()` method

rdig (changes requested):
This is mostly good. 

Just a quick refactor then it should be ready to go.

src/api/auth.ts
rdig: Haha :)

rdig (approved):
All good now. 

Nice work :100:

PR #2332, opened 2020-12-15
Removed ActivityFeed

ArmandoGraterol:
Removed ActivityFeed core component because it's no longer necessary

Resolves DEV-24


alicjakujawa (changes requested):
Looks good to me, the only thing i feel we're missing is removing ActivityFeedItem.ts where types are define, but i can be wrong. @rdig

rdig (changes requested):
Mostly good, but there are some leftovers.

Please also remove the `ActivityFeed` types and mentions from `immutable`
- https://github.com/JoinColony/colonyDapp/blob/9c75ef27f2f561034591547e0d86302f66e1b0d0/src/immutable/ActivityFeedItem.ts
- https://github.com/JoinColony/colonyDapp/blob/9c75ef27f2f561034591547e0d86302f66e1b0d0/src/immutable/index.ts#L3
- https://github.com/JoinColony/colonyDapp/blob/9c75ef27f2f561034591547e0d86302f66e1b0d0/src/immutable/Using-Immutable.md#L34-74

alicjakujawa (approved):

rdig (approved):
LGTM: :+1:

PR #2334, opened 2020-12-17
Actions Page Recovery Mode Tag - Dev-50

chinins:
## Description

This is a simple issue that adds a "Recovery" tag element at the top of the actions page, when the colony is put into recovery mode.

**New stuff** ✨

* Added `Recovery` string via `defineMessages` & `useIntl` hook
* Added styles for `Recovery` tag

Resolves Dev-50

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.css
@@ -5,6 +5,11 @@
   background-color: white;
 }
 
+.recoveryTag {
+  background-color: var(--pink);
+  color: rgb(249, 250, 250);
+}
+
 .container {
   display: flex;
   align-items: flex-start;
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.css
@@ -6,13 +6,14 @@
 }
 
 .recoveryTag {
-  padding: 4.5px 8px;
+  padding: 4px 8px;
   height: 20px;
   width: 69px;
   position: absolute;
   top: 129px;
   left: 24.3%;
-  border-radius: 5px;
+  vertical-align: middle;
+  border-radius: var(--radius-large);
   background-color: var(--pink);
   font-size: var(--size-tiny);
   font-weight: var(--weight-bold);
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.css
@@ -6,8 +6,20 @@
 }
 
 .recoveryTag {
+  padding: 4.5px 8px;
+  height: 20px;
+  width: 69px;
+  position: absolute;
+  top: 129px;
+  left: 24.3%;
+  border-radius: 5px;
   background-color: var(--pink);
-  color: rgb(249, 250, 250);
+  font-size: var(--size-tiny);
+  font-weight: var(--weight-bold);
+  text-align: center;
+  line-height: var(--size-tiny);
+  color: var(--text-invert);
+  letter-spacing: var(--spacing-medium);
 }
 
 .container {
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.css
@@ -9,7 +9,6 @@
   margin-bottom: 10px;
   padding: 4px 8px;
   height: 20px;
-  width: 69px;
   vertical-align: middle;
   border-radius: var(--radius-large);
   background-color: var(--pink);
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.css
@@ -10,11 +10,6 @@
   padding: 4px 8px;
   height: 20px;
   width: 69px;
-
-  /* position: absolute; */
-
-  /* top: 129px;
-  left: 24.3%; */
   vertical-align: middle;
   border-radius: var(--radius-large);
   background-color: var(--pink);
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.css
@@ -6,12 +6,15 @@
 }
 
 .recoveryTag {
+  margin-bottom: 10px;
   padding: 4px 8px;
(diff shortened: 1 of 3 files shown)


rdig:
Few notes:
- You don't need to put the issue id in the title of your PR
- You _do_ need to add it at the end of your PR, with one of [linear's link keywors](https://docs.linear.app/GitHub-and-GitLab-fa4b88df484343e4989538f066c729f3), even if it's a draft PR _(eg: "Resoves DEV-50")_
- Assign yourself to the PR
- Add a label to your PR _(in your case that would be `feature`)_
- Don't need to update the PR's description while in draft mode

src/modules/dashboard/components/ActionsPage/ActionsPage.css
```diff
@@ -5,6 +5,23 @@
   background-color: white;
 }
 
+.recoveryTag {
+  padding: 4.5px 8px;
```
rdig, line 9: Please don't use half pixel values

rdig, line 9: The text would look better "centered" vertically if the top and bottom padding would just be `4px`

```suggestion
  padding: 4px 8px;
```

Screenshot from 2020-12-18 16-03-36

src/modules/dashboard/components/ActionsPage/ActionsPage.css
```diff
@@ -5,6 +5,23 @@
...
+  height: 20px;
+  width: 69px;
+  position: absolute;
```
rdig, line 12: This has no reason to be an absolute element.

- Refactor `.container` to not include a top border
- create a `divider` element that has the solely purpose of displaying that diving line _(the border you just removed)_
- place your tag component inside the `.container` element, above the divider

src/modules/dashboard/components/ActionsPage/ActionsPage.css
```diff
@@ -5,6 +5,23 @@
...
+  top: 129px;
+  left: 24.3%;
+  border-radius: 5px;
```
rdig, line 15: You can safely use `--radius-large` for this one, to keep things consistent

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -264,6 +270,9 @@ const ActionsPage = () => {
 
   return (
     <div className={styles.main}>
+      {actionType === ColonyActions.Recovery && (
+        <p className={styles.recoveryTag}>{formatMessage(MSG.recoveryTag)}</p>
```
rdig, line 274: Please use the  `FormattedMessage` component where possible, rather than the programatic `formatMessage`.

```suggestion
        <p className={styles.recoveryTag}>
          <FormattedMessage {...MSG.recoveryTag} />
        </p>
```

rdig: The programmatic hook is a legacy leftover in `react-intl` and is mostly used for stuff like giving a component the ability to support both strings and message descriptors as values.

See this as an example:
https://github.com/JoinColony/colonyDapp/blob/6a8ca7f612960c8c7788833202eceb3f5a709216/src/modules/core/components/Heading/Heading.tsx#L55-L65

rdig (changes requested):
This still needs a bunch of work needed until it can be merged, but you're getting there :)

Most stuff that I've pointed out is related to things you must not have known, so this is mostly on me for not telling them to you, but hey... it's a learning process :)

Also, I noticed to basically copy-pasted the css specs from the Figma file. We don't really do that. We use Figma for visual guidance, but we decide how to actually implement the css declarations based on our component's needs, maintenability or otherwise "what makes sense". Figma specs are auto-generated, and they seem to be kind of crappy :P

That's all I have for now, let me know if you need help with this.

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -264,6 +268,17 @@ const ActionsPage = () => {
 
   return (
     <div className={styles.main}>
+      <div className={styles.container}>
+        <p className={styles.recoveryTag}>
+          <FormattedMessage {...MSG.recoveryTag} />
+        </p>
+      </div>
+      {actionType === ColonyActions.Recovery && (
```
ArmandoGraterol, line 276: I think this is old code right? It looks like the old tag before the improvement you did, the conditional should be wrapped around the div above this.

ArmandoGraterol (changes requested):
It's looking good besides the small thing I pointed out :ok_hand:

src/modules/dashboard/components/ActionsPage/ActionsPage.css
```diff
@@ -5,11 +5,23 @@
...
 
-.container {
-  display: flex;
-  align-items: flex-start;
-  justify-content: space-between;
+.recoveryTag {
+  margin-bottom: 10px;
+  padding: 4px 8px;
+  height: 20px;
+  width: 69px;
```
rdig, line 12: I don't think we need a fixed width, as we have no strict requirement for it's width. 

The padding will take care of that anyway.

Imagine this some other way. At some point, internationalization will kick in and we'll have multiple languages. Longer works will break this.

Just remove this and it's all fine.

src/modules/dashboard/components/ActionsPage/ActionsPage.css
```diff
@@ -21,7 +33,19 @@
...
   margin: 0 13.19% 0 24.3%;
   width: 900px;
-  border-top: 1px solid color-mod(var(--temp-grey-blue-7) alpha(15%));
+}
+
+.dividerTop {
+  composes: setMarginWidth;
+  height: 1px;
+  background-color: color-mod(var(--temp-grey-blue-7) alpha(15%));
+}
```
rdig, line 42: Usually, when I do dividers, I just go without a border, and just set a `background-color`, and a `height` of `1px`.

But this is just a side note, you can do it however you like it :)

rdig (approved):
Left a note that I think you should still address _(basically remove the `width`)_.

But I'm pre-approving this, so after you make that change, you can merge it.

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
rdig: Yes it is old code. Pull the new version, and then re-review it, as It's all good now.

PR #2338, opened 2020-12-21
Version upgrade flow wiring

alicjakujawa:
## Description

This PR wires up the "Version upgrade" flow

**New stuff** ✨

* Redux types for version upgrade action
* Saga for version upgrade action
* Values fetcher
* Event and action copies for version. upgrade action
* VersionUpgrade colony actions type and ColonyUpgraded event
* Annotation message

**Changes** 🏗

* NetworkContractUpgradeDialog refactoring as it was broken for upgrade banner
* Disable some major actions for colony v3 and v4

**Deletions** ⚰️

* Removed UpgradeNetworkDialog from ProfileAdvanced
* NetworkContractUpgradeDialog refactoring
* Remove old COLONY_VERSION_UPGRADE types and sagas

Resolves DEV-76
Resolves DEV-160


rdig:
As I mentioned on your other PR, this needs to also handle wiring up subgraph query and populating the subgraph with the action required to populate the actions list.

There's issue DEV-141 to track this.

package.json
```diff
@@ -132,7 +132,7 @@
   },
   "dependencies": {
     "@apollo/client": "^3.0.2",
-    "@colony/colony-js": "^3.0.0-alpha.15",
+    "@colony/colony-js": "^3.0.0-alpha.17",
```
rdig, line 135: So I think we broke this when we removed the filters from colonyJS (all those `null` arguments)

Since now we broke the roles system in the app. 

I'm logged in with the creator of the colony, but I just get the recovery role for that user, and because of this, I get an error saying I can't upgrade my colony

Screenshot from 2021-01-25 15-01-25

And these are the roles

Screenshot from 2021-01-25 15-02-27

rdig: But don't worry about this, I'll look into colonyJS and try to fix it

rdig (changes requested):
I've found some stuff that you need to change, especially with the way we show disabled elements.

But the biggest problem is coming from colonyJS which I think we broke when we "fixed" the logic in the getRoles methods. I'll look into this to try to fix it.

Also, the extensions route needs to be disabled for versions 3 and 4. The problem is we don't have access to that yet, as it's not finished.

Please open a new issue to track that, so we don't forget about it.

src/data/resolvers/colonyActions.ts
```diff
@@ -135,12 +137,15 @@ export const colonyActionsResolvers = ({
...
         );
-
-        const annotation = await getAnnotation(
-          from as string,
-          hash as string,
-          colonyClient as ColonyClientV5,
-        );
+        const clientVersion = await colonyClient?.version();
+        let annotation;
+        if (clientVersion.toString() === '5') {
```
rdig, line 142: Please don't hard-code the version number.

Instead use the version provided by colonyJS via `ColonyVersion`

```suggestion
        if (clientVersion === ColonyVersion.CeruleanLightweightSpaceship) {
```

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -143,32 +144,44 @@ const ColonyHome = ({ match, location }: Props) => {
...
    * which versions *must* be upgraded, and which can function as-is, even with
    * an older version
    */
   const mustUpgradeColony = canBeUpgraded(
-    data.colony,
+    colony,
     parseInt(networkVersion || '0', 10),
   );
+
+  const isSupportedColonyVersion = colony.version >= 5;
```
rdig, line 183: Same here, please use the version constant from colonyJS

```suggestion
        const isSupportedColonyVersion = colony.version >= ColonyVersion.CeruleanLightweightSpaceship);
```

src/modules/dashboard/components/ColonyTotalFunds/ColonyTotalFunds.tsx
```diff
@@ -45,12 +45,13 @@ const ColonyTotalFunds = ({
...
     tokens: colonyTokens,
     nativeTokenAddress,
+    version,
   },
 }: Props) => {
   const [currentTokenAddress, setCurrentTokenAddress] = useState<Address>(
     nativeTokenAddress,
   );
-
+  const isSupportedColonyVersion = version >= 5;
```
rdig, line 54: Same here, please use the version constant from colonyJS

```suggestion
  const isSupportedColonyVersion = colony.version >= ColonyVersion.CeruleanLightweightSpaceship);
```

src/modules/dashboard/sagas/actions.ts
```diff
@@ -569,6 +573,126 @@ function* createMintTokensAction({
...
+    const colonyManager = TEMP_getContext(ContextModule.ColonyManager);
+
+    const { version: newestVersion } = yield getNetworkContracts();
+    const currentVersion = parseInt(version, 10);
+    const nextVersion = currentVersion + 1;
+    if (nextVersion > parseInt(newestVersion, 10)) {
+      throw new Error('Colony has the newest version');
+    }
+
+    const supportAnnotation = currentVersion >= 5 && annotationMessage;
```
rdig, line 593: Same here, please use the version constant from colonyJS

```suggestion
  const supportAnnotation = currentVersion >= ColonyVersion.CeruleanLightweightSpaceship && annotationMessage;
```

src/modules/dashboard/components/AdvancedDialog/AdvancedDialog.tsx
```diff
@@ -104,6 +105,8 @@ const AdvancedDialog = ({
   const canEnterRecovery =
     hasRegisteredProfile && canEnterRecoveryMode(allUserRoles);
 
+  const canUpgradeVersion = hasRegisteredProfile && hasRoot(allUserRoles);
```
chinins, line 108: you don't need to add this variable, you can just you can just use `hasRootPermission` from line 103 - it's the same value (same root permission that is being used for several parts)

chinins (changes requested):
Everything seems ok to me except one UX thing. I think it is related to Raul's comment about roles.

The UX is misleading. I have all of the permissions, including the root needed for the upgrade. I can't upgrade because the colony has the latest version. BUT i still see the `Root` permission error.

I think that in both cases the submission should be disabled BUT the causes and hence the messages should be separated: `No permission` error when a user doesn't have a permission and `Can't upgrade because it's the latest version`.

package.json
chinins: I have all the roles when I login but I still see the permission error because the colony can't be upgraded as it has the latest version. I think the UX has to be changed slightly (please, see my comment about it in my review summary and let me know what you both think).

alicjakujawa:
@JoinColony/frontend its ready for re-review

chinins (approved):
All good! Great job :)

ArmandoGraterol (approved):
Working well on my side of things :fire:

src/modules/dashboard/components/DomainDropdown/CreateDomainButton.tsx
```diff
@@ -43,9 +44,15 @@ const CreateDomainButton = ({ colony }: Props) => {
...
+  const isSupportedColonyVersion =
+    colony.version >= ColonyVersion.CeruleanLightweightSpaceship;
   return (
-    <button className={styles.main} onClick={handleClick} type="button">
+    <button
+      className={styles.main}
+      onClick={handleClick}
+      type="button"
+      disabled={!isSupportedColonyVersion}
+    >
```
rdig, line 55: The styles for this, when disabled, are awful :)

They don't tell you in any way that this is actually disabled.

Leave it like this for now, but open a issue to track fixing this, as this is not a deal-breaker :)

rdig (approved):
This works exactly as I imagined it would.

Very nice job Alicja. You showed great determination in pulling through the looooooooooog dev process for this PR.

:1st_place_medal:

src/modules/dashboard/components/DomainDropdown/CreateDomainButton.tsx
alicjakujawa: https://linear.app/colony/issue/DEV-170/ui-improvements-in-disabled-state-for-colony-version-3-and-4 added to scope of this

PR #2340, opened 2020-12-22
Add subgraph events list

rdig:
This PR adds in subgraph event handlers for all the colony events and wires them in to be displayed in the `ColonyEventsList` component

**Changes** 

- [x] Updated `subgraph` submodule
- [x] Added subgraph events entities queries
- [x] Refactored core `ActionsList` to support events
- [x] Added `getEventsData` transformer
- [x] Added all support events as enum types
- [x] Added copy for the events list
- [x] `ColonyEvents` now uses subgraph events data

Note that this uses JoinColony/subgraph#13 to update the subgraph colony metadata history part

**Demo**

demo-events-list

Resolves DEV-133


src/modules/core/components/ActionsList/ActionsList.tsx
```diff
@@ -15,19 +14,26 @@ export interface ClickHandlerProps {
 }
 
 interface Props {
-  items: FormattedAction[];
+  items;
```
alicjakujawa, line 17: what happen here?

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -45,28 +47,36 @@ export enum Status {
 }
 
 interface Props {
-  item: FormattedAction;
+  item;
```
alicjakujawa, line 50: I feel it should be typed, i know that there will be a lot of "optonal" fields. But now we can puts here any object, and easily break the app

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -75,8 +85,16 @@ const ActionsListItem = ({
...
+  const initiatorUserProfile = useUser(
+    initiator ? createAddress(initiator) : '',
+  );
+  const agentUserProfile = useUser(agent ? createAddress(agent) : '');
+  const eventRecipient = values?.recipient
+    ? createAddress(values.recipient)
+    : null;
+  const recipientAddress = recipient
+    ? createAddress(recipient)
+    : null || eventRecipient || colonyAddress;
```
alicjakujawa, line 97: what is happening here? if eventRecipient and colonyAddress will be undefined we will get here undefined either way.

So why not just 
: eventRecipient || colonyAddress

src/modules/dashboard/transformers.ts
```diff
@@ -116,3 +122,36 @@ export const getActionsListData = (
...
+          transactionHash: hash,
+          colonyAddress: createAddress(colonyAddress),
+          createdAt: new Date(parseInt(`${timestamp}000`, 10)),
+          values,
+          displayValues: args,
+          fromDomain: values?.domainId || null,
+        };
+      }
+      return undefined;
+    })
```
alicjakujawa, line 156: if you would use reduce, you wouldn't need to return undefined and then filter undefined

src/data/graphql/queries.graphql
```diff
@@ -637,3 +637,23 @@ query SubgraphSingleDomain($colonyAddress: String!, $domainId: Int!) {
...
+    id
+    address
+    associatedColony {
+      colonyAddress: id
+      id: colonyChainId
+    }
+    transaction {
+      hash: id
+      block {
+        id
```
alicjakujawa, line 652: i dont think we need to query it

alicjakujawa (changes requested):

src/modules/core/components/ActionsList/ActionsListItem.tsx
rdig: I know. 

I just couldn't make it work properly the first time around so at one point I just gave up and strait removed it.

You're right that it should typed so I guess I'll give it one more  try then.

src/modules/core/components/ActionsList/ActionsList.tsx
rdig: See below

src/modules/dashboard/transformers.ts
rdig: Actually that's a fair point

src/data/graphql/queries.graphql
rdig: It doesn't hurt either

src/modules/core/components/ActionsList/ActionsListItem.tsx
rdig: Huh?

It recipient _or_ eventRecipient _or_ colonyaddress.

Just that the first part needs to check that the recipient actually exists, otherwise it can't call createAddress since it will error out: 
```
recipient
    ? createAddress(recipient)
    : null
```
So if that is null, than go to the the `eventRecipient || colonyAddress` check

Yeah I know it's kinda cumbersome, and this is because we're trying to make the ActionsList component handle events as well. 

I think a better idea would be for me to just create a separate one just for events. That way there would be no need for such "awkward" logic checks everywhere.

src/data/graphql/queries.graphql
alicjakujawa: I remember you was ask me to remove it in my PR, so thought that you actually dont like it 😏

rdig: Haha, I honestly don't remember that.

But fine. I'll remove it then :)

src/i18n/en-events.ts
```diff
@@ -10,6 +10,22 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.DomainAdded} {{initiator} added Team: {fromDomain}}
       other {{eventNameDecorated} emmited by {clientOrExtensionType}}
     }`,
+  'eventList.event': `{eventName, select,
+      ${ColonyAndExtensionsEvents.DomainAdded} {{agent} added Team: {fromDomain}}
+      ${ColonyAndExtensionsEvents.DomainMetadata} {{agent} changed Team {fromDomain} metdata to {metadata}}
```
ArmandoGraterol, line 15: Typo here :-) `metdata`

ArmandoGraterol (commented):
I noticed something strange when I try to access the `Events` page, I get this error

Screenshot from 2021-01-25 12-32-49

 The `undefined` on the consoles is how the `events` array is arriving at `immutableSort(events, sort)`

Now when I make a change, like adding a `console.log`, and the app recompiles, then the `Events` page loads correctly! With the 3 initial events from when you create the colony... But if I manually refresh, or try to access the events page through the navbar, it crashes again

rdig:
Nice catch! I'll look into this

chinins (commented):
I have exactly the same error as Armando. It seems that it's coming from `immutableSort` in `ColonyEvents`.

I agree with some of Alicja's comments (like types for Props). Other than that seems all good.

rdig:
@ArmandoGraterol @chinins 

Put in a fallback if we don't yet have events data. Also fixed typo's

@alicjakujawa 

Refactored the core `ActionList` to accept a rendering `Item` component, that way we can de-couple the actions and events render items and provide them with different data.

For this, I had to create a custom item component for the events list: `ColonyEventsListItem`

This allows us to scale both actions and events list independently, without worrying about data getting mixed up or hard to read, each will only handle its own set of values coming in.

Also, refactored the `getEventsListData` transformer to use reduce and to return more than _(that I needed in the new custom item component)_

This is ready for a new round of reviews when you have the time! :+1:

src/modules/dashboard/transformers.ts
```diff
@@ -157,3 +163,57 @@ export const getActionsListData = (
...
+        domainId,
+        recipient,
+        fundingPotId,
+        metadata,
+        token,
+        paymentId,
+        amount,
+        payoutRemainder,
+        decimals = '18',
+      } = JSON.parse(args);
```
alicjakujawa, line 193: it can throw error

src/modules/dashboard/transformers.ts
```diff
@@ -157,3 +163,57 @@ export const getActionsListData = (
...
+          fundingPot: fundingPotId,
+          metadata,
+          tokenAddress: token ? createAddress(token) : null,
+          paymentId,
+          decimals: parseInt(decimals, 10),
+          amount: amount || payoutRemainder || '0',
+        },
+      ];
+    }
+    return undefined;
```
alicjakujawa, line 218: Just to be sure... so if we have array [event1, event2, null, event3, event4], then  getEventsListData will return undefined? is it intended?

alicjakujawa (approved):
Small comments but pre-approving this as works great 🔥

src/modules/dashboard/transformers.ts
rdig: No. Actually good catch. I think this was an oversight on my part.

This should actually return the current value of the `processedEvents` array.

src/modules/dashboard/transformers.ts
rdig: Yes and no.

`args` will always exist if we have an event. And even if it's an empty string: `''`, it will still be parsed by `JSON.parse`.

But yeah, I'll put in a provision to fallback to a valid value if it does not exist

PR #2341, opened 2020-12-29
Update default state when nativeToken is fetched

alicjakujawa:
## Description

- This PR is fixing bug with displaying token information on home page

**Changes** 🏗

* Added useEffect to ColonyTotalFunds component to update default state whenever nativeToken is changing. The reason behind this is that while switching between colonies, nativeToken from previous company was set as default state

Resolves DEV-153


chinins:
Looks good to me

ArmandoGraterol (approved):
Works great for me :+1:

rdig (approved):
Works as expected. Very nice Alicja!

Screenshot from 2021-01-05 14-36-51

PR #2336, opened 2020-12-17
Fixes, and changes for Dapp styling

ArmandoGraterol:
This PR it's to improve the styling in general of the Dapp

**New stuff** ✨

* Added IBMPlexMono (techie font) SemiBold
* Added a different version of the caret-right icon for the right caret in the UAC
* Added correct colors and weight to InfoPopover heading, error, and etherscan link
* Added separator between "New Action" button and domain dropdown, and the event list in the Colony Home

**Changes** 🏗
* Improved alignment of copyable address
* Reduced width of Dialog core component
* Increased margin between Dialog outer actions and Dialog main content
* Updated color of Select core component to correct one
* Fixed extraTiny className not being applied to Icon
* Added correct font weight and color to MaskedAddress
* Updated Modal overlay color
* Reduced width of SinglePickerUser
* Updated Navbar text color
* Updated active item color of domain dropdown
* Updated font weight of Permission Management checkbox
* Increased width of TokenAmount second column in Transfer Funds Dialog
* Decreased padding of User Notification button

**Deletions** ⚰️
* Removed incorrect font (techie) from TokenSelector

Contributes to DEV-134

PR #2346, opened 2021-01-11
Mint tokens subgraph queries

alicjakujawa:
## Description

- This PR adds in subgraph queries for mintTokens action 

**New stuff** ✨

* Subgraph types for events query
* Events filter
* getValuesForActionType util method

**Changes** 🏗

* Updated recipient address in ActionListItem to support colony recipient 
* Updated getActionListData to correctly format values for events
* Added missing first arg to events filters

Screenshot 2021-01-13 at 11 19 49

Subgraph branch: https://github.com/JoinColony/subgraph/pull/7

Resolves DEV-136

### DIFF
--- a/src/data/graphql/queries.graphql
+++ b/src/data/graphql/queries.graphql
@@ -655,7 +655,6 @@ query SubgraphActions($colonyAddress: String!) {
     transaction {
       hash: id
       block {
-        id
         timestamp
       }
     }
@@ -666,7 +665,7 @@ query SubgraphActions($colonyAddress: String!) {
       }
     }
     args
-    address
+    address # Address of the contract that emitted the event
     name
   }
 }
--- a/src/data/graphql/queries.graphql
+++ b/src/data/graphql/queries.graphql
@@ -650,7 +650,7 @@ query SubgraphActions($colonyAddress: String!) {
       }
     }
   }
-  events(where: { associatedColony_contains: $colonyAddress }) {
+  events(where: { associatedColony_contains: $colonyAddress, name_in: ["TokensMinted(address,address,uint256)"] }) {
     id
     transaction {
       hash: id
--- a/src/data/graphql/queries.graphql
+++ b/src/data/graphql/queries.graphql
@@ -650,4 +650,23 @@ query SubgraphActions($colonyAddress: String!) {
       }
     }
   }
+  events(where: { associatedColony_contains: $colonyAddress }) {
+    id
+    transaction {
+      hash: id
+      block {
+        id
+        timestamp
+      }
+    }
+    associatedColony {
+      token {
+        decimals
+        symbol
+      }
+    }
+    args
+    address
+    name
+  }
 }
--- a/src/data/resolvers/task.ts
+++ b/src/data/resolvers/task.ts
@@ -42,6 +42,7 @@ const finalizeTaskWithTxHash = async (
     null,
     null,
     null,
+    null,
   );
   const payoutClaimedLogs = await getLogs(colonyClient, payoutClaimedFilter);
   const claimedLogForTask = payoutClaimedLogs.find(
@@ -105,6 +106,7 @@ export const taskResolvers = ({
       }
       if (existingEthPotId) {
         const payoutClaimedFilter = colonyClient.filters.PayoutClaimed(
+          null,
           existingEthPotId,
           null,
           null,
--- a/src/data/resolvers/task.ts
+++ b/src/data/resolvers/task.ts
@@ -107,6 +107,7 @@ export const taskResolvers = ({
       if (existingEthPotId) {
         const payoutClaimedFilter = colonyClient.filters.PayoutClaimed(
           null,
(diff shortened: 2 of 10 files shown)


src/data/graphql/queries.graphql
```diff
@@ -650,4 +650,23 @@ query SubgraphActions($colonyAddress: String!) {
       }
     }
   }
+  events(where: { associatedColony_contains: $colonyAddress, name_in: ["TokensMinted(address,address,uint256)"] }) {
```
rdig, line 653: I think it will work with just `TokensMinted`, but I'm fine with this anyway.

It will have to change anyway when we wire in the rest of the events and actions

src/data/graphql/queries.graphql
```diff
@@ -650,4 +650,23 @@ query SubgraphActions($colonyAddress: String!) {
...
+    id
+    transaction {
+      hash: id
+      block {
+        id
```
rdig, line 658: Do we need the block id ?

src/data/graphql/queries.graphql
```diff
@@ -650,4 +650,23 @@ query SubgraphActions($colonyAddress: String!) {
...
+      }
+    }
+    associatedColony {
+      token {
+        decimals
+        symbol
+      }
+    }
+    args
+    address
```
rdig, line 669: This is the `agent`, right ?

Maybe rename it so it's a bit more clear

src/modules/dashboard/transformers.ts
```diff
@@ -79,6 +83,30 @@ export const getActionsListData = (
...
+            const {
+              args,
+              associatedColony: {
+                token: { address: tokenAddress, symbol, decimals },
+              },
+              name,
+            } = unformattedAction;
+            const actionEvent = Object.entries(ACTIONS_EVENTS).find((el) =>
+              el[1]?.includes(name.split('(')[0]),
+            );
```
rdig, line 96: Maybe have a util for this at some point, since I think we're using this in two places now.

But you can leave it like this for now, maybe I'll fix it when I'll wire in the rest of the events

src/utils/colonyActions.ts
```diff
@@ -27,3 +33,25 @@ export const getDetailsForAction = (
...
+        initiator: argsObj.agent,
+        recipient: argsObj.who,
+        amount: argsObj.amount,
+      };
+    }
+    default: {
+      return {};
+    }
+  }
+};
```
rdig, line 57: I like this idea!

However:
- this should also handle the other actions which are not coming from events directly (eg: OneTxPayment)
- this should use a static map same as we do with the other utils

src/data/resolvers/task.ts
```diff
@@ -105,6 +106,7 @@ export const taskResolvers = ({
       }
       if (existingEthPotId) {
         const payoutClaimedFilter = colonyClient.filters.PayoutClaimed(
+          null,
           existingEthPotId,
```
rdig, line 110: You can safely add a `@ts-ignore` here to deal with that typescript error.

I _think_ is didn't generate something properly in `colony-js`, but until I get to investigate this, it can stay ignore

rdig (approved):
I'm good with this. Very nice work! :1st_place_medal: 

You're the first to wire a subgraph action that's only an event! Way to go!

**However, before you merge:**
- Merge the subgraph branch into master (in the subgraph repo): JoinColony/subgraph#7
- After you've merged the above, come back to this repo and update the `subgraph` submodule to the latest commit in it's `master` (if you're not sure about this, ping me after and I'll check if it's good before you merge)

Also, as a side note, @ArmandoGraterol after this PR is merged, please don't rebase your #2336 on `master` anymore. If you do that, we won't be able to deploy it to QA anymore, as the required contracts are not yet deployed to `goerli`

You can still work on it, just don't rebase it on `master`

This just goes for branches that need to be tested on QA (currently there's only yours and the Extension Manager), all other PRs / branches can be safely rebased on `master`

src/data/graphql/queries.graphql
alicjakujawa: It's the address of the contract that emitted the event, agent is sitting in args object. i will add comment there

src/utils/colonyActions.ts
alicjakujawa: For the first dot -> im disagree, i would keep this only for events, when we parsing "args" object / String
For other edge cases like OneTxPayment we should think about something scalable, but it will be easier  to create it, when we will have at least second case like OneTxPayment. Not here. 
For the second dot - i will create static map

rdig: > For the first dot -> im disagree, i would keep this only for events, when we parsing "args" object / String

In that case find a better name for that utis, as it's current name implies that it gets values for actions, including ones like OneTx, which will not be the case here.

I don't know, maybe something like `getValuesForSingleEventAction` ?

ArmandoGraterol (approved):
It's working great for me! Solid work :fire:

PR #2350, opened 2021-01-15
Wire "Colony Actions" filter by domain

chinins:
## Description

This PR adds filtering by domain to Colony Actions list

**New stuff** ✨

- add filtering for actions based on domain id

**Changes** 🏗

- updated copy when there are no actions (made it more general)

**Deletions** ⚰️

- commented out sorting by `activity` as this is not available yet

## TODO

- ideally to test all of the filtering cases but all of the actions need to be wired up and shown in the list to do this.

Resolves DEV-152

filter-gif

### DIFF
--- a/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
+++ b/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
@@ -19,7 +19,10 @@ import {
 } from '../shared/actionsSort';
 import { getActionsListData } from '../../transformers';
 import { useTransformer } from '~utils/hooks';
-import { FormattedAction } from '~types/index';
+import {
+  ColonyActions as ColonyActionTypes,
+  FormattedAction,
+} from '~types/index';
 
 import styles from './ColonyActions.css';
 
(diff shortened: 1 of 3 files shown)


src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -38,7 +41,7 @@ const MSG = defineMessages({
   },
   noActionsFound: {
     id: 'dashboard.ColonyActions.noActionsFound',
-    defaultMessage: `The colony did not create any actions yet.
+    defaultMessage: `There are no actions yet.
```
chinins, line 44: Thought it might be good to change it to a generic phrase that goes for both colony actions and domain actions

rdig: I mean we can even remove it.

Since after everything is wired up, all colonies will have at least one action (the first domain created), so you'll see this reflected in my #2348 

But yeah, leave the change in, it doesn hurt :)

rdig (approved):
Good to go!

PR #2352, opened 2021-01-18
Wire `Transfer Funds` annotation

chinins:
## Description

This wires up annotation in `Transfer Funds` dialog

**New stuff** ✨

- add annotation wiring in TransferFundsForm dialog
- add annotation & types in payload in createTransferFundsAction

**Changes** 🏗

 - change batching in `createPaymentAction` - for consistency purpose

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-158

Screenshot 2021-01-18 at 21 32 28

### DIFF
--- a/src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialog.tsx
+++ b/src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialog.tsx
@@ -34,6 +34,7 @@ export interface FormValues {
   toDomain?: string;
   amount: string;
   tokenAddress?: Address;
+  annotation: string;
 }
 
 interface CustomWizardDialogProps {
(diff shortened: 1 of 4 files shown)


src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialog.tsx
```diff
@@ -67,6 +68,7 @@ const TransferFundsDialog = ({
       .required()
       .moreThan(0, () => MSG.amountZero),
     tokenAddress: yup.string().address().required(),
+    annotation: yup.string().max(4000),
```
rdig, line 71: I would have renamed this in the component as well (to `annotationMessage`), rather than have to rename it in the transformer, but it doesn't really matter either way

So I'll chuck this to being a personal preference, so you can leave it like this and merge

rdig (approved):
Very nice. 

I appreciate that you refactored the payment action saga as well to keep the code consistent.

Way to go :tada:

PR #2353, opened 2021-01-19
Wire Edit Colony Flow

rdig:
## Description

This PR wires in the colony edit action flow:
- the actual colony action creation (via the modal)
- refactor to the way we upload avatars
- wire in the actual action for the action page
- wire in the subgraph action for this
- refactor colony data to come from subgraph events
- refactor the create colony saga to send metadata

**Changes**
- [x] Added `EDIT_COLONY` redux action types
- [x] Added `editColonyAction` actions saga
- [x] Update `EditColonyDialog` to send in correct values
- [x] Added colony edit action and event copy
- [x] Refactored colony action resolver to return "Wrong Colony" action type -- _it's a check to prevent opening actions from colonies that they not belong to_
- [x] `ActionsPage` now properly displays colony edit action
- [x] `DisplayWidget` now properly displays colony edit action
- [x] Fetch IPFS metadata in the actions resolver
- [x] Added colony metadata event from subgraph -- to populate the actions list
- [x] Fetch the colony's data from the subgraph
- [x] Update the cache after joining / leaving a colony
- [x] Subscribe to colony after creating it
- [x] Fix user token minting permissions refresh
- [x] Detect type of colony metadata change
- [x] Wire up `TokenManagement` dialog
- [x] Refactored `FriedlyName`
- [x] Cleanup old admin domains / profile logic
- [x] Better token validation and fail safes
- [x] **Many** improvements to local cache updates

Also note that:
- this uses JoinColony/subgraph#11 to update the subgraph colony metadata history part
- this uses JoinColony/colonyServer#132 where it removes the domains and colonies

So you'll need to run `npm i` and `npm provision` in order to make this work

**Demo**

Edit colony title

demo-colony-edit-action

Edit colony avatar

demo-colony-edit-logo

Edit tokens
demo-manage-tokens

Subscribe / unsubscribe
demo-subscriptions

Resolves DEV-77
Resolves DEV-140
Resolves DEV-144
Resolves DEV-147
Resolves DEV-156

And a bunch of in-passing fixes / cleanup

Resolves DEV-34
Resolves DEV-35
Resolves DEV-57
Resolves DEV-108
Resolves DEV-111


rdig:
Sorry about the size of this :(

But on the plus side, it ties everything up in a neat little bow. :bowtie:

src/modules/dashboard/components/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
```diff
@@ -197,7 +258,12 @@ const EditColonyDetailsDialogForm = ({
           text={{ id: 'button.confirm' }}
           onClick={() => handleSubmit()}
           loading={isSubmitting}
-          disabled={!userHasPermission}
+          disabled={
+            !userHasPermission ||
+            !isValid ||
+            avatarFileError ||
+            !canValuesBeUpdate()
```
alicjakujawa, line 265: why this is function? 🤔

src/modules/dashboard/sagas/actions.ts
```diff
@@ -735,6 +739,210 @@ function* createDomainAction({
...
+}: Action<ActionTypes.COLONY_ACTION_EDIT_COLONY>) {
+  let txChannel;
+  try {
+    const apolloClient = TEMP_getContext(ContextModule.ApolloClient);
+
+    /*
+     * Validate the required values for the payment
+     */
+    if (!colonyDisplayName && colonyDisplayName !== null) {
+      throw new Error('A colony name is required in order to edit the colony');
```
alicjakujawa, line 762: why !colonyDisplayName && colonyDisplayName !== null instead of just !colonyDisplayName ??

src/modules/dashboard/sagas/utils/updateColonyDisplayCache.ts
```diff
@@ -0,0 +1,59 @@
...
+      /*
+       * @NOTE Only return the value that you want to be set in the cache
+       * Any value you return will be set immediatly
+       *
+       * That's why we can't have a consistent return
+       */
+      // eslint-disable-next-line consistent-return
+      displayName: () => {
+        if (displayName || displayName === null) {
+          return displayName;
```
alicjakujawa, line 42: im lost here a bit. So if we want return displayName always (when its null or not) then why we just not do
displayName: () => displayName ?

src/utils/events/index.ts
```diff
@@ -270,6 +284,44 @@ const getCreateDomainActionValues = async (
   return domainAction;
 };
 
+const getColonyEditActionValues = async (
+  processedEvents: ProcessedEvent[],
+): Promise<Partial<ActionValues>> => {
+  const mintTokensEvent = processedEvents.find(
```
alicjakujawa, line 290: why this name? 🤔

alicjakujawa (changes requested):

src/data/graphql/typeDefs.ts
```diff
@@ -38,6 +38,9 @@ export default gql`
     amount: String!
     tokenAddress: String!
     annotationHash: String
+    colonyDisplayName: String!
+    colonyAvatarHash: String!
+    colonyTokens: [String]!
```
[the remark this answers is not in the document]
rdig: It might be an empty array

src/data/resolvers/colony.ts
```diff
@@ -138,60 +300,66 @@ export const colonyResolvers = ({
...
+          query: SubgraphColonyDocument,
+          variables: {
+            /*
+             * First convert it a string since in cases where the network name
+             * cannot be found via ENS it will throw an error
+             *
+             * Converting this to string basically converts the error object into
+             * a string form
+             */
+            address: address.toString().toLowerCase(),
```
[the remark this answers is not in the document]
rdig: It won't be.

This comes from colonyName query which takes the colony name and converts it into an address.

If that fails this doesn't get called.

src/data/resolvers/user.ts
```diff
@@ -155,5 +163,45 @@ export const userResolvers = ({
...
+            const subscribedColony = ((data?.colonies as unknown) as Array<{
+              id: string;
+              colonyChainId: string;
+              ensName: string;
+              metadata: string;
+            }>).find(({ id }) => id === colonyAddress.toLowerCase());
+            if (subscribedColony) {
+              userColonies.push(subscribedColony);
+            }
+            return null;
```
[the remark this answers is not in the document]
rdig: I don't understand how that would be more beneficial here ?

Sure it's a different approach to this sure, and it would remove the need for returns, but I don't see the benefit of using this over my current implementation.

src/modules/core/components/TokenEditDialog/TokenEditDialog.tsx
```diff
@@ -90,37 +105,71 @@ const TokenEditDialog = ({
...
       }
       addresses = [
         ...new Set(
           addresses
             .map((address) => createAddress(address))
-            .filter((address) => address !== AddressZero),
+            .filter((address) => {
+              if (address === AddressZero) {
+                return false;
+              }
```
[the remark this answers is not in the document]
rdig: Sure. Initially I thought my two filters would look "cleaner" :)

src/modules/dashboard/components/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
rdig: Custom validation, as we don't have a proper way (using `yup`) of checking if the name _or_ logo changed, but didn't change to the same values they were initially.

It's not ideal. but It's the only way we currently can validate this form properly

src/modules/dashboard/sagas/actions.ts
rdig: Because if it's undefined I don't want to send the date, but if it's null, I want to (this means the colony display name was removed by the user)

Same goes for the avatar

src/modules/dashboard/sagas/utils/updateColonyDisplayCache.ts
rdig: No. We don't want to return the `undefined` case, we just want either a string value, or null.

So that's why we have to do this little dance of filtering out null :(

src/utils/events/index.ts
rdig: I guess I copy-pasted it from somewhere :)

I'll change it.

src/data/resolvers/user.ts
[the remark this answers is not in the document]
alicjakujawa: So "map" is used to "decorate" values in the array, not to loop over array and create different array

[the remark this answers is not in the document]
rdig: I mean, I've been using `map` to modify an array my whole life :)

But I guess what you're saying makes sense, so I'll make the change.

src/modules/dashboard/components/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
alicjakujawa: Hmmm, so why not make it just a variable?

rdig: Actually a good point.

So initially, this had more "complex" logic and I needed a function to achieve that. But looking at the final result, it is indeed a good candidate to just change into a variable

src/modules/core/components/Fields/Input/InputComponent.tsx
```diff
@@ -111,6 +112,8 @@ const InputComponent = ({
         placeholder={placeholder}
         ref={innerRef}
         maxLength={maxLength}
+        // value={value}
```
chinins, line 115: should the comment be deleted or uncommented?

src/modules/core/components/FriendlyName/FriendlyName.tsx
```diff
@@ -30,20 +31,42 @@ interface Props {
...
+      displayName: colonyDisplayName,
+      colonyAddress,
+    } = colony;
+    colonyDisplay = colonyDisplayName || colonyName;
+    if (colonyAddress !== AddressZero) {
+      colonyDisplayAddress = colonyAddress;
+    }
+  }
+
+  // const isColonyAddress = colony && walletAddress === colony.colonyAddress;
```
chinins, line 67: should the commented lines be deleted?

src/modules/core/components/TokenEditDialog/TokenEditDialog.tsx
```diff
@@ -1,19 +1,26 @@
 import React, { useCallback, useState, useMemo } from 'react';
```
chinins, line 1: Couple of points regarding token management dialog:

1. When there is an error in a new address (in the input) the color doesn't change to red. I think to be consistent with other dialogs/forms it should change color.

2. When deleting token there are 2 warning messages for Appolo cache (i saw it only when deleting):

```
warning 1:
Cache data may be lost when replacing the tokenAddresses field of a ProcessedColony object.

To address this problem (which is not a bug in Apollo Client), define a custom merge function for the ProcessedColony.tokenAddresses field, so InMemoryCache can safely merge these objects:

  existing: ["0x8972e86549bb8E350673e0562fba9a4889d01637","0x7F176F534C31294fDbA77D80bB9d01279820C063"]
  incoming: ["0x7F176F534C31294fDbA77D80bB9d01279820C063"]

For more information about these options, please refer to the documentation:

  * Ensuring entity objects have IDs: https://go.apollo.dev/c/generating-unique-identifiers
  * Defining custom merge functions: https://go.apollo.dev/c/merging-non-normalized-objects

  
  warning 2:
  Cache data may be lost when replacing the tokens field of a ProcessedColony object.

To address this problem (which is not a bug in Apollo Client), define a custom merge function for the ProcessedColony.tokens field, so InMemoryCache can safely merge these objects:

  existing: [{"__ref":"Token:0x0000000000000000000000000000000000000000"},{"__ref":"Token:0x8972e86549bb8E350673e0562fba9a4889d01637"},{"__ref":"Token:0x7F176F534C31294fDbA77D80bB9d01279820C063"}]
  incoming: [{"__ref":"Token:0x0000000000000000000000000000000000000000"},{"__ref":"Token:0x7F176F534C31294fDbA77D80bB9d01279820C063"}]

For more information about these options, please refer to the documentation
...
```

ArmandoGraterol (approved):
It's working great for me :ok_hand: This a beast of PR

src/data/graphql/typeDefs.ts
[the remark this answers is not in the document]
alicjakujawa: From my understanding, by adding String! inside of array, you just forcing, that if array has item, then it need to be string. so with  [String!]! ->
[] - is ok
['string'] - is ok
[null] - is not ok

alicjakujawa (changes requested):
Screenshot 2021-01-27 at 12 13 21
* We need to handle error when user actually will choose file >1MB
* Also i think, im not able to choose svg format avatar
Apart of this bug just update graphql schema type and this HUGE PR will be ready to merge 🔥

src/modules/dashboard/components/EditColonyDetailsDialog/EditColonyDetailsDialogForm.tsx
rdig: I thought I did? Did I not commit the change? What's going on here????

rdig:
@alicjakujawa 

> * We need to handle error when user actually will choose file >1MB * Also i think, im not able to choose svg format avatar Apart of this bug just update graphql schema type and this HUGE PR will be ready to merge fire

re-install node modules and try this again please. We handle upload errors. I think that's coming from a library not existing on your envrionment (I've had this before when testing Olga's PR)

alicjakujawa:
@rdig removed and installed node_modules, and same error. Also im not able to upload svg file

chinins (approved):
Looks and works fine for me.

rdig:
@alicjakujawa 

1. Cannot replicate your first issue, with not handling errors. Works fine locally and even on QA. Here's a demo:
demo-cannot-upload-large-files

2. You were right about the svg files, the mime type was not set correctly. I've fixed that now:
demo-can-upload-svg

3. I've changed the `canValuesBeUpdate` function into a simple variable now

alicjakujawa (approved):
🔥

PR #2355, opened 2021-01-20
Wire up Permission Management flow and subgraph queries

ArmandoGraterol:
This PR wires the permission management flow in the UAC, and adds ColonyRoleSet events to the event list

**New stuff** ✨

* Connected PermissionManagementDialog with the rest of the UAC flow
* Added warnings and validation for having no permission to modify roles
* Added SetUserRoles related action, events, saga, mapping, types, copies, event values, and queries
* Added DetailsWidgetRoles (section used in ActionPage to display modified permission labels)

**Changes** :fire:
* Updated subgraph submodule with new SetUserRoles event handlers

**Deletions** ⚰️

* Removed old roles sagas
* Removed old SetUserRoles role types
* Removed ColonyPermissionAddDialog

FireShot Capture 115 - Colony - localhost
FireShot Capture 112 - Colony - localhost

Resolves DEV-74
Resolves DEV-143


src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -299,6 +303,55 @@ const ActionsPage = () => {
...
+        unassignedRoles,
+        assignedRoles,
+      )}`;
+    }
+
+    roleTitle += roles.length > 1 ? ' permissions' : ' permission';
+
+    return roleTitle;
+  };
+
```
alicjakujawa, line 354: i would extract those methods to some utils/helper file, or utils/roles

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -54,6 +55,7 @@ interface Props {
 }
 
 const ActionsPageEvent = ({
+  index: eventIndex,
```
alicjakujawa, line 58: why we just not name this prop eventIndex?

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx
```diff
@@ -41,6 +54,13 @@ const MSG = defineMessages({
     id: 'dashboard.PermissionManagementDialog.selectUser',
     defaultMessage: 'Member',
   },
+  noPermissionFrom: {
+    id:
+      'dashboard.CreatePaymentDialog.CreatePaymentDialogForm.noPermissionFrom',
```
alicjakujawa, line 59: I don't think id is correct

src/modules/dashboard/sagas/actions.ts
```diff
@@ -1086,6 +1086,180 @@ function* editColonyAction({
...
+
+    let annotationMessageIpfsHash = null;
+    if (annotationMessage) {
+      annotationMessageIpfsHash = yield call(
+        ipfsUpload,
+        JSON.stringify({
+          annotationMessage,
+        }),
+      );
+    }
```
alicjakujawa, line 1129: Those sagas was refactored in @rdig https://github.com/JoinColony/colonyDapp/pull/2359
We need to adjust to those change. You would need to cherry-pick this commit->
https://github.com/JoinColony/colonyDapp/pull/2359/commits/d32a4a67748d09203cef281539718bf03bdc5313

src/types/colonyActions.ts
```diff
@@ -121,6 +124,8 @@ export interface FormattedAction {
   transactionHash: string;
   createdAt: Date;
   commentCount: number;
+  role: ColonyRole;
+  setTo: string;
```
alicjakujawa, line 128: why string? 🤔

alicjakujawa (dismissed):

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
ArmandoGraterol: Sure :+1:

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx
ArmandoGraterol: You are right, must be the result of my copy-pasting :(

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -84,6 +85,13 @@ const ActionsListItem = ({
     ({ ethDomainId }) => ethDomainId === parseInt(toDomainId, 10),
   );
 
+  const getFormattedRole = () => {
+    const roleNameMessage = { id: `role.${role}` };
+    return `${setTo === 'true' ? 'assigned' : 'removed'} ${formatMessage(
```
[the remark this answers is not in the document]
ArmandoGraterol: It comes out as a string due to [this](https://github.com/JoinColony/subgraph/pull/15/files#diff-24656a8181a67d58132644c18a1f3ccdf0085bfac7eb07d9912d8725a887925cR29). Perhaps I could convert to a boolean in the util function that gets the action values

src/types/colonyActions.ts
ArmandoGraterol: See [this](https://github.com/JoinColony/colonyDapp/pull/2355#discussion_r570970278) comment

src/data/graphql/queries.graphql
```diff
@@ -388,7 +392,8 @@ query SubgraphActions($skip: Int!, $first: Int!, $colonyAddress: String!) {
       "TokensMinted(address,address,uint256)",
       "DomainAdded(address,uint256)",
       "ColonyMetadata(address,string)",
-      "ColonyFundsMovedBetweenFundingPots(address,uint256,uint256,uint256,address)"
+      "ColonyFundsMovedBetweenFundingPots(address,uint256,uint256,uint256,address)",
+      "ColonyRoleSet(address, address, uint256, uint8, bool)"
```
rdig, line 396: Please remove the spaces between the args, to keep this consistent with the others

src/i18n/en.json
```diff
@@ -115,6 +115,10 @@
     "transaction.ColonyClient.moveFundsBetweenPotsWithProofs.title": "Move Funds Between Pots",
     "transaction.ColonyClient.moveFundsBetweenPotsWithProofs.description": "Move Token from one domain pot to another",
 
+    "transaction.group.setUserRoles.title": "Set user roles",
+    "transaction.group.setUserRoles.description": "Set user roles",
+    "transaction.ColonyClient.setUserRolesWithProofs.title": "Set user roles",
+    "transaction.ColonyClient.setUserRolesWithProofs.description": "Set user roles",
     "transaction.group.setRoles.title": "Set roles",
     "transaction.group.setRoles.description": "Sign the transactions required to set the selected roles.",
```
rdig, line 123: If you've added new message descriptors for the new roles actions, please make sure to remove the old ones

src/i18n/en-events.ts
```diff
@@ -9,6 +9,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.TokensMinted} {{initiator} minted {amount} {tokenSymbol} to {recipient}}
       ${ColonyAndExtensionsEvents.DomainAdded} {{initiator} added Team: {fromDomain}}
       ${ColonyAndExtensionsEvents.ColonyUpgraded} {This colony has upgraded to {newVersion}}
+      ${ColonyAndExtensionsEvents.ColonyRoleSet} {{initiator} {role} permission in the {toDomain} team to {recipient}}
       other {{eventNameDecorated} emmited by {clientOrExtensionType}}
     }`,
   /*
```
rdig, line 15: Please also add copy for the events:

Screenshot from 2021-02-05 15-42-09

Let me know if you need help coming up with the correct copy

src/modules/core/components/ActionsList/ActionsListItem.tsx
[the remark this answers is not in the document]
rdig: Alicja is partly right. 

I know it's coming from the subgraph as a string, that we can't do anything about since it's a limitation, be can convert it client side.

So you should type it as a string (since it's coming in as a string), then convert it to a true boolean in the app.

This was we won't have any surprises down the line.

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
rdig: I second this. They have no palce inside a React component

src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -299,6 +303,55 @@ const ActionsPage = () => {
...
+      roleTitle += ` and removed the${getFormmatedRoleList(
+        unassignedRoles,
+        assignedRoles,
+      )}`;
+    }
+
+    roleTitle += roles.length > 1 ? ' permissions' : ' permission';
+
+    return roleTitle;
+  };
```
rdig, line 353: The role title is not generated correctly. It always shows as the permission being set in the Root domain, even though it was set in a sub-domain.

Here's an example: I've given permissions to user "b" in domain "Second"

Screenshot from 2021-02-05 15-52-16

But the title still shows that permissions were added in domain "Root"

Note that the sidebar widget displays the correct domain

rdig: Also, I think the permissions should be displayed in lower case when showing them in the titles:
- both the action page title
- and the even copy

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
```diff
@@ -168,6 +173,16 @@ const DetailsWidget = ({
...
+      {detailsForAction.Permissions && (
+        <div className={styles.item}>
+          <div className={styles.label}>
+            <FormattedMessage {...MSG.roles} />
+          </div>
+          <div className={styles.value}>
+            {values?.roles && <DetailsWidgetRoles roles={values.roles} />}
+          </div>
+        </div>
+      )}
```
rdig, line 185: Screenshot from 2021-02-05 15-52-16

Please add more padding to the section that displays multiple roles. In cases like this, when there's two or more roles being set, it breaks the UI

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidgetRoles/DetailsWidgetRolesItem.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { useIntl } from 'react-intl';
+
+import { ActionUserRoles } from '~types/index';
+import PermissionsLabel from '~core/PermissionsLabel';
+
+import styles from './DetailsWidgetRoles.css';
+
+const displayName =
+  // eslint-disable-next-line max-len
+  'dashboard.ActionsPage.DetailsWidget.DetailsWidgetRoles.DetailsWidgetRolesItem';
```
rdig, line 11: If you add backticks ` you won't need to disable line length, as those are ignored by our eslint rules, as we assume they are used for long strings

```suggestion
  `dashboard.ActionsPage.DetailsWidget.DetailsWidgetRoles.DetailsWidgetRolesItem`;
```

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx
```diff
@@ -228,8 +271,32 @@ const PermissionManagementDialog = ({
...
+                          <PermissionsLabel
+                            permission={ColonyRole.Administration}
+                            name={{ id: `role.${ColonyRole.Administration}` }}
+                          />
+                        ),
+                      }}
+                    />
+                  </div>
+                </DialogSection>
+              )}
```
rdig, line 299: This requirement display is **wrong!**

Just above, in this file, you are checking for the `Architecture` permission:

https://github.com/JoinColony/colonyDapp/blob/6e965069641df996d29f3105799bf5c519f8dc50/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx#L212-L215

But when showing the user the error message, you are telling them it requires both `Funding` and `Adminstration` ?

Screenshot from 2021-02-05 15-52-49

Please fix this

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx
```diff
@@ -242,6 +309,7 @@ const PermissionManagementDialog = ({
                   text={{ id: 'button.confirm' }}
                   type="submit"
                   style={{ width: styles.wideButton }}
+                  disabled={!userHasPermission || !isValid}
```
rdig, line 312: This form is not validating! At all!

Confirm should not be enabled if I didn't changed the user's permissions.

bug-permissions-form-not-validating

Not how if I send the transaction, it won't actually create one since the contracts validate this as not being an actual change, so when you get redirected, you run into an error.

We should prevent this by validating the form properly.

src/modules/dashboard/sagas/actions.ts
rdig: :point_up: What @alicjakujawa said :)

src/modules/dashboard/sagas/actions.ts
```diff
@@ -1086,6 +1086,180 @@ function* editColonyAction({
...
+    yield apolloClient.query<
+      ProcessedColonyQuery,
+      ProcessedColonyQueryVariables
+    >({
+      query: ProcessedColonyDocument,
+      variables: {
+        address: colonyAddress,
+      },
+      fetchPolicy: 'network-only',
+    });
```
rdig, line 1228: You don't need to run this at the end since no "actual" colony data needs to be refreshed

rdig (changes requested):
This needs a lot of work still:
- action title is not working properly
- form validation was not added all
- some style fixes still need to be added
- copy for the events list was not added

Also, there's something I've noticed, that was not mentioned anywhere for this:

Screenshot from 2021-02-05 15-41-56

The above is how the colony front page looks after creating the colony first time

Yes, I know that by default the network contract adds all the permissions to the user that created the colony, but I'm wondering if we need to filter them out _(same way that Alicja did with domains and Olga with transfer funds)_

So please go ask someone in Product how we should handle them, if the original permission sets need to be filtered out, or can we leave them in?

Let me know if you need help with any of this.

src/data/graphql/queries.graphql
ArmandoGraterol: If I do that, then the query won't catch the ColonyRoleSet events. The name of the query includes the white spaces

ArmandoGraterol: If I do that, then the query won't catch the ColonyRoleSet events. The name of the query includes the white spaces

src/i18n/en-events.ts
```diff
@@ -43,6 +44,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.PayoutClaimed} {{agent} claimed a payout of {amount} {tokenSymbol} from funding pot {fundingPot}}
       ${ColonyAndExtensionsEvents.ColonyMetadata} {{agent} changed Colony metadata to {metadata}}
       ${ColonyAndExtensionsEvents.ColonyFundsMovedBetweenFundingPots} {{agent} transferred {amount} {tokenSymbol} between pots}
+      ${ColonyAndExtensionsEvents.ColonyRoleSet} {{agent} {roles} in {domain} to {recipient}}
```
rdig, line 47: This does not work well for "removed", since it says "to", which only applies to adding. In the case of removing it should say "from"

```suggestion
      ${ColonyAndExtensionsEvents.ColonyRoleSet} {{agent} {roles} in {domain} from {recipient}}
```

Here's an example:
Screenshot from 2021-02-10 17-37-07

Consider creating two messages descriptors Id for this instance

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -93,6 +95,13 @@ const ActionsListItem = ({
     ({ ethDomainId }) => ethDomainId === parseInt(toDomainId, 10),
   );
 
+  const getFormattedRole = () => {
+    const roleNameMessage = { id: `role.${role}` };
+    return `${setTo ? 'Assign' : 'Remove'} ${formatMessage(
+      roleNameMessage,
+    ).toLowerCase()} permission`;
+  };
+
```
rdig, line 104: As I've stated above, consider using two message descriptors for this event, and in that case, you can use this same function, to just return the correct message descriptor id

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
```diff
@@ -168,6 +173,16 @@ const DetailsWidget = ({
           </div>
         </div>
       )}
+      {detailsForAction.Permissions && (
+        <div className={styles.item}>
+          <div className={styles.label}>
+            <FormattedMessage {...MSG.roles} />
```
rdig, line 179: The "Roles" label should be aligned to the top of the section, not the middle _(see the same way we do it for domains descriptions)_

Screenshot from 2021-02-10 17-37-31

Everything else here is fine _(padding between the roles lables)_

src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.tsx
```diff
@@ -77,6 +79,13 @@ const ColonyEventsListItem = ({
     ({ address }) => address === nativeTokenAddress,
   );
 
+  const getFormattedRole = () => {
+    const roleNameMessage = { id: `role.${role}` };
+    return `${setTo ? 'assigned' : 'removed'} the ${formatMessage(
+      roleNameMessage,
+    ).toLowerCase()} permission`;
+  };
+
```
rdig, line 88: This is the same as the one in the `ActionsListItem`.

Since it's duplicated code, please refactor this out and move it into utils/events so that it can be shared

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -196,6 +205,7 @@ const ActionsListItem = ({
                 decimals: getTokenDecimalsWithFallback(decimals),
                 fromDomain: domainName || fromDomain?.name || '',
                 toDomain: toDomain?.name || '',
+                roles: role ? getFormattedRole() : '',
               }}
             />
           </div>
```
rdig, line 211: The actions list should filter out the initial permissions, which it still does not:

Screenshot from 2021-02-10 17-34-02

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx
```diff
@@ -242,6 +306,11 @@ const PermissionManagementDialog = ({
                   text={{ id: 'button.confirm' }}
                   type="submit"
                   style={{ width: styles.wideButton }}
+                  disabled={
+                    !userHasPermission ||
+                    !isValid ||
+                    isEqual(sortBy(values.roles), sortBy(initialValues.roles))
+                  }
```
rdig, line 313: This does not do proper validation yet.

Here's two cases I found:

1. I can remove **my own** permissions in "Root" making the colony basically unusuable.

Screenshot from 2021-02-10 17-19-17
Screenshot from 2021-02-10 17-19-54
Screenshot from 2021-02-10 17-19-57

Ask product about this, but I don't think this is what they want.

2. It seems I cannot remove some user permissions: If the user has a single permission, de-selecting the check box, does not light up the "Confirm" button:

Screenshot from 2021-02-10 17-35-22
Screenshot from 2021-02-10 17-35-26

Please be more thorough when doing form validations

src/modules/dashboard/sagas/actions.ts
rdig: This wasn't changed yet...

src/modules/dashboard/sagas/actions.ts
rdig: I think you addressed this privately, as this still needs to be in.

src/utils/hooks/useFormatRolesTitle.ts
```diff
@@ -0,0 +1,58 @@
...
+  const { formatMessage } = useIntl();
+
+  if (!roles) {
+    return { roleTitle };
+  }
+
+  const assignedRoles = roles.filter((role) => role.setTo);
+  const unassignedRoles = roles.filter((role) => !role.setTo);
+
+  const getFormattedRoleList = (roleGroupA, roleGroupB) => {
```
rdig, line 15: :+1:

rdig (changes requested):
This still needs a bunch of work done, as validations still don't work properly.

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementDialog.tsx
ArmandoGraterol: Consulted with product about the first case. They said we keep it that way, if someone wants to remove their permissions then we shouldn't prevent that

rdig: Ok then.

In that case just take care of the second case

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -164,6 +168,7 @@ const ColonyActions = ({
   const actions = useTransformer(getActionsListData, [
     (data && { ...data, events: uniqueEvents }) || data,
     commentCount?.transactionMessagesCount,
+    oneTxPaymentExtensionData?.oneTxPaymentExtensionAddress,
   ]);
```
rdig, line 172: There's one thing I haven't caught until now.

If we add two (or three, or whatever number) permissions to a user, we show two actions entries in the actions list, but since we only have one action/transaction hash, we direct the user to the same action page, no matter what action they click:

Screenshot from 2021-02-12 13-48-28

The above directs the user to the same page, this:
Screenshot from 2021-02-12 14-04-47

So I think we should only show one action in the action list, but with same merged title that you do for the event title.

So in this case, we would only have one entry, that would have the title:
```
Assign the architure, administration permission in second to b
```

I think the easiest way to achieve this, is to hook into the transformer and if the action type is a permission, and the transaction hash is the same, just leave one in.

rdig (changes requested):
This is all good to merge, mostly.

I still want you to do a change to only show one action instances where we change multiple permissions in one single transaction.

I've left notes on how to do that below.

But all in all, this is very nice work. Kudos!

rdig (approved):
I found some yup validations that _can_ be done better, which apparently I missed in previous review rounds.

But in the interest of time, we can merge this as-is

PR #2356, opened 2021-01-22
Wire up `Move Funds` action query

chinins:
## Description

Create and wire up "Move Funds" actions Subgraph queries. Corresponds to the subgraph#14 PR `Add `funds moved` to subgraph`.

Checked that filtering by domain for move funds action works (2nd screenshot as an example)

**New stuff** ✨

- add `MoveFunds` case to `getValuesForActionType`
- add `ColonyFundsMovedBetweenFundingPots` event to graphql actions query
- add `getDomainsforMoveFundsActions` helper function in `utils/events` to get correct domains from funding pots
- format and filter events in `ColonyActions` component
- add hard limit on the token symbol input when creating Colony (Dev-164)

**Changes** 🏗

- add missing parts in move funds action

**Deletions** ⚰️

none

## TODO

- potentially move `TEMP_getContext(ContextModule.ColonyManager)` initialisation from `ColonyAction` component to `getDomainsforMoveFundsActions` helper function

Resolves DEV-135
Resolves DEV-164

Screenshot 2021-01-27 at 14 35 27
Screenshot 2021-01-27 at 14 35 51

### DIFF
--- a/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
+++ b/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
@@ -1,7 +1,8 @@
-import React, { useCallback, useState, useMemo } from 'react';
+import React, { useCallback, useEffect, useState, useMemo } from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { useHistory } from 'react-router-dom';
 
+import { ContextModule, TEMP_getContext } from '~context/index';
 import ActionsList, {
   ClickHandlerProps as RedirectHandlerProps,
 } from '~core/ActionsList';
@@ -19,6 +20,7 @@ import {
   ActionsSortSelectOptions,
 } from '../shared/actionsSort';
 import { getActionsListData } from '../../transformers';
+import { getDomainsforMoveFundsActions } from '~utils/events';
 import { useTransformer } from '~utils/hooks';
 import {
   ColonyActions as ColonyActionTypes,
@@ -75,9 +77,18 @@ const ColonyActions = ({
 
   const [dataPage, setDataPage] = useState<number>(1);
 
+  const [formattedActions, setFormattedActions] = useState<FormattedAction[]>(
+    [],
+  );
+
   const ITEMS_PER_PAGE = 10;
 
   const history = useHistory();
+
+  const colonyManager = useMemo(
+    () => TEMP_getContext(ContextModule.ColonyManager),
+    [],
+  );
 
   const { data, loading: paymentActionsLoading } = useSubgraphActionsQuery({
     variables: {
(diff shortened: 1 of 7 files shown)


src/utils/colonyActions.ts
```diff
@@ -56,6 +56,13 @@ export const getValuesForActionType = (
         fromDomain: argsObj.domainId,
       };
     }
+    case ColonyActions.MoveFunds: {
+      return {
+        amount: argsObj.amount,
+        fromDomain: argsObj.fromPot,
+        toDomain: argsObj.toPot,
+      };
```
rdig, line 64: Please also return the initiator from the agent:

```suggestion
      return {
        amount: argsObj.amount,
        fromDomain: argsObj.fromPot,
        toDomain: argsObj.toPot,
        initiator: argsObj.agent,
      };
```

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -91,6 +102,23 @@ const ColonyActions = ({
...
+      data === undefined
+        ? []
+        : data.events.filter(
+            (action) =>
+              !data?.oneTxPayments.some(
+                (paymentAction) =>
+                  paymentAction.transaction?.hash === action.transaction.hash,
+              ),
+          ),
+    [data],
```
rdig, line 119: This whole thing needs a try/catch block.

We have colonies (broken ones) right now on QA that don't have `action.transaction`

Yes they are broken, but just because that it doesn't mean we can't provision and provide some level of fallback for that.

rdig: I really don't like how this is so hard to read, but due to time constraints it will have to do

src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.tsx
```diff
@@ -178,7 +178,7 @@ const StepCreateToken = ({
                 appearance={{ theme: 'fat' }}
                 maxLength={5}
                 data-test="defineTokenSymbol"
-                formattingOptions={{ uppercase: true }}
+                formattingOptions={{ uppercase: true, blocks: [5] }}
```
rdig, line 181: :+1:

rdig (approved):
A few minor changes and a try/catch block and this is ready to go

ArmandoGraterol (commented):
I noticed something while testing

At least for me, it's not showing the events of the transfer of funds unless I reload the page. It's not like for example, when you mint a token, that you don't need to reload the page to see it on the list when you go back from the action page. It feels like it's not refetching or updating the list.

And also something strange... When I made the fund transfer and reloaded to see the event, I opened the UAC again, and now whenever I select an option in the UAC the list strangely seems to load again each time I selected an option or opened a new dialog. See the gif for example on how it's behaving:

Colony

rdig:
@ArmandoGraterol 

> At least for me, it's not showing the events of the transfer of funds unless I reload the page. It's not like for example, when you mint a token, that you don't need to reload the page to see it on the list when you go back from the action page. It feels like it's not refetching or updating the list.

This is not a problem introduced in here, this was always the case. Hopefully it will be solved by DEV-165

@chinins 
> And also something strange... When I made the fund transfer and reloaded to see the event, I opened the UAC again, and now whenever I select an option in the UAC the list strangely seems to load again each time I selected an option or opened a new dialog. See the gif for example on how it's behaving:

This however is accurate. I think it's because the filtering in the component gets reset every time. Please look into this.

PR #2357, opened 2021-01-25
Edit domain flow wiring

alicjakujawa:
## Description

This PRs adds in domain edit wiring flow

**New stuff** ✨

* Added EditDomain action, static maps, action copies
* Added transaction copies
* Added action types
* Added saga for editing domain
* Added EditDomainDialog and EditDomainDialogForm
* Added method for getting action values from DomainMetadata object
* Added redux types
* Added SubgraphDomainMetadata query
* Added helpers to determine what values was change
* Added helper to get correct message descriptors

**Changes** 🏗

* Refactored CreateEditDomainDialog

Screenshot 2021-01-29 at 15 05 31
Screenshot 2021-01-29 at 15 10 23

Resolves DEV-69
Resolves DEV-139


ArmandoGraterol (approved):
Looking good for me :sunglasses:

src/i18n/en-events.ts
```diff
@@ -8,6 +8,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.ColonyFundsMovedBetweenFundingPots} {{initiator} transferred {amount} {tokenSymbol} from the {fromDomain} to the {toDomain}}
       ${ColonyAndExtensionsEvents.TokensMinted} {{initiator} minted {amount} {tokenSymbol} to {recipient}}
       ${ColonyAndExtensionsEvents.DomainAdded} {{initiator} added Team: {fromDomain}}
+      ${ColonyAndExtensionsEvents.DomainMetadata} {{initiator} changed {fromDomain} teams's from}
```
chinins, line 11: Is it used anywhere? It doesn't seem to be complete but I couldn't find it in this form anywhere in the app.

chinins (changes requested):
It works fine for me. Great job! Just a couple of small things:

- when I create domain I have 2 actions in the actions list corresponding to it. I think the "details edited" action should not be shown when creating a domain. Please, see below my screenshot - that's only after creating domain, didn't edit it.
Screenshot 2021-02-02 at 17 22 28

- The permission copy at the bottom of the edit domain doesn't seem to be correct. `domainName` doesn't seem right
Screenshot 2021-02-02 at 17 33 29

src/i18n/en-events.ts
alicjakujawa: nee, need to remove it, old forgot thing 😆  good catch

alicjakujawa:
> It works fine for me. Great job! Just a couple of small things:
> 
> * when I create domain I have 2 actions in the actions list corresponding to it. I think the "details edited" action should not be shown when creating a domain. Please, see below my screenshot - that's only after creating domain, didn't edit it.
> 
> Screenshot 2021-02-02 at 17 22 28
> 
> * The permission copy at the bottom of the edit domain doesn't seem to be correct. `domainName` doesn't seem right
> 
> Screenshot 2021-02-02 at 17 33 29

haha another forgot thing, will update it

src/modules/core/components/Fields/Input/InputComponent.tsx
```diff
@@ -97,7 +97,7 @@ const InputComponent = ({
       onChange={onChange}
       placeholder={placeholder}
       ref={innerRef}
-      defaultValue={value}
+      value={value || ''}
```
rdig, line 100: Why was this changed? 

Note that using `value` as opposed to `defaultValue` will make the input go in and out of React controlled / uncontrolled state.

Also, why does this default to an empty string? This is why we have `initialValues`. Please remove this.

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -183,18 +259,28 @@ const ActionsPageEvent = ({
...
+              ),
+              domainColor: (
+                <ColorTag color={Number(domainColor) || Color.LightPink} />
+              ),
+              // Fallback to current name/description when there is no previous metadata obj
+              oldDescription:
+                previousDomainMetadata?.domainPurpose || domainPurpose,
+              oldName: previousDomainMetadata?.domainName || domainName,
+              domainPurpose,
+              domainName,
```
rdig, line 283: The logic here is all good, but the display is not.

Currently, you show the most current domain name, which is not always correct, especially if you go back through multiple edits.

It should just show the old and new domain names relevant to the action that was taken

Here's an example:

Screenshot from 2021-02-03 00-10-07

I've changed the name from three to four, but this shows the domain name as "SAME", which is the most recent name I've set to the domain.

This is not correct. This should just say "Changed team's name from three to four"

While I understand this was the copy spec that product handed over, I think the spec/copy is wrong and should be changed. 

Please raise this problem with Product.

src/i18n/en-events.ts
```diff
@@ -36,6 +37,14 @@ const eventsMessageDescriptors = {
...
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.nameDescription`]: `{initiator} changed {fromDomain} teams's name and description from {oldName}, {oldDescription} to {domainName}, {domainPurpose}`,
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.nameColor`]: `{initiator} changed {fromDomain} teams's name and color from {oldName}, {oldColor} to {domainName}, {domainColor}`,
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.descriptionColor`]: `{initiator} changed {fromDomain} teams's description and color from {oldDescription}, {oldColor} to {domainPurpose}, {domainColor}`,
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.description`]: `{initiator} changed {fromDomain} teams's description from {oldDescription} to {domainPurpose}`,
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.name`]: `{initiator} changed {fromDomain} teams's name from {oldName} to {domainName}`,
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.color`]: `{initiator} changed {fromDomain} teams's color from {oldColor} to {domainColor}`,
+  [`event.${ColonyAndExtensionsEvents.DomainMetadata}.fallback`]: `{initiator} changed this domain, but values are the same`,
```
rdig, line 47: I think these should come before the `eventList.event` entry in this list.

`eventList.event` will most likely grow by a  lot as time goes on

src/modules/dashboard/components/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -0,0 +1,217 @@
...
+      </DialogSection>
+      <DialogSection>
+        <div className={styles.nameAndColorContainer}>
+          <div className={styles.domainName}>
+            <Input
+              label={MSG.name}
+              name="domainName"
+              appearance={{ colorSchema: 'grey', theme: 'fat' }}
+              disabled={!canEditDomain}
+              maxLength={20}
```
rdig, line 158: The team name input should the same length as the others

Screenshot from 2021-02-03 00-15-51

Currently it's a bit shorter

src/modules/dashboard/sagas/actions.ts
```diff
@@ -873,6 +873,170 @@ function* createDomainAction({
...
+     */
+    let annotationMessageIpfsHash = null;
+    if (annotationMessage) {
+      annotationMessageIpfsHash = yield call(
+        ipfsUpload,
+        JSON.stringify({
+          annotationMessage,
+        }),
+      );
+    }
```
rdig, line 921: Please have a look at how I refactored these in #2359 

https://github.com/JoinColony/colonyDapp/blob/fixes/qa-initial-stabilization/src/modules/dashboard/sagas/actions.ts#L171-L184

The idea here is basically:
- you move them below, next to the actual transaction in which they are being called
- you put the transaction in a pending state
- then you send the upload action
- then you update the transaction's params with the ipfs hash
- then you send the actual transaction

This makes it so that the transaction will show a loading indicator as soon as it starts to upload the ipfs data.

This is not noticeable on local, but it's a problem on QA, where, the UI will be blocked until IPFS finishes the upload. If you refactor this as I've proposed, you'll already inform the user that something going on in the background.

If you have more questions let me know and I'll come help :)

src/modules/dashboard/components/EditDomainDialog/EditDomainDialogForm.tsx
```diff
@@ -0,0 +1,217 @@
...
+      setFieldValue('domainName', selectedDomain.name || undefined);
+      setFieldValue('domainPurpose', selectedDomain.description || undefined);
+      setDomainColor(selectedDomain.color);
+    }
+  };
+  useEffect(() => {
+    if (domainId) {
+      handleDomainChange(domainId);
+    }
+  }, []);
```
rdig, line 113: This has missing dependencies: `domainId` and `handleDomainChange`

rdig (changes requested):
A lot of good work was made here! But this is still not done, as there are a bunch of stuff we need to "fix", especially regarding the actions list

1. The action list should not display a "domain edited" action if we create a new domain. Only the edit action should show one:

Screenshot from 2021-02-02 23-51-41

I know this is a weird issue, but we've run into this before. See @chinins 's #2356 where she has the exact same problem, and she filters out the MoveFundsBetweenPots action

2. The list should show the domain name that was changed at the time of change, not the most recent name that the domain has. For this we'll need to parse the metadata when displaying domain actions. Otherwise we run into issues like this, where the we show the same name for the whole list:
Screenshot from 2021-02-02 23-54-59

(Note that I have the same problem on the events page, that I'll have to fix after you merge this PR)

3. The above also applies to the action page (i've left a note below) and the action details, widget, as that needs to show the name **at the time of change** not, the most recent one

Here's a demo:
demo-domains-have-same-name

I think this is all I could find so far. Let me know if I can help you with getting this done faster.

src/modules/core/components/Fields/Input/InputComponent.tsx
alicjakujawa: it wont go in and out of React controlled/uncontrolled, because you have default value passed. This is the only way we can fill out form with current domain data. Unless you know about better solution to fix this problem

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
alicjakujawa: Im not sure what exact problem you see with this? It will be hard for me to raise this problem with Product without explaining what is the problem. So can you give me more details?

src/modules/dashboard/components/EditDomainDialog/EditDomainDialogForm.tsx
alicjakujawa: Nope, i want fire this only once, on first render

alicjakujawa: When you pass empty array as second parameter of useEffect it will behave like componentDidMount.

src/modules/core/components/Fields/Input/InputComponent.tsx
rdig: >  This is the only way we can fill out form with current domain data. 

But then how did it work for the colony edit modal then? When I implemented that, it worked if I just passed in `defaultValue={value}`

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
alicjakujawa: Ahh i think a got it, so you want to update copies to show just "from to" values?

src/modules/dashboard/components/EditDomainDialog/EditDomainDialogForm.tsx
rdig: I know that, I just don't understand why do you only need this to run at first render.

But it really doesn't matter this is a minor nit-pick, also I trust that you did the right thing :)

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
rdig: So the problem here is that we show the most recent domain name, even for actions that happened in the past.

In my example "SAME NAME" is the latest domain name I've changed. This should only be shown on the _latest_ domain edit action.

The previous one, should have the old name, and the one before that the even-older name, ... and so on.

If you want we can jump on a call and I can show you the problem live

rdig: Exactly. Since it makes more sense that way

ArmandoGraterol (commented):
I noticed something strange, along the lines of what Raul already mentioned...

FireShot Capture 126 - Colony - localhost

I initially created a team called Wonker 1, changed the color, then changed the name to Wonker 1.2, and finally changed the name again to Wonker 1.3.

As you can see, almost every action displays the correct name for the team at the time of the change, except for the initial team creation. 

The initial creation although it displays the newest name of the team in the action list, when you enter the ActionsPage, displays the name at the time of the creation, so it just that action in the action list which is considering the newest name.

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -191,8 +207,8 @@ const ActionsListItem = ({
             </FormattedDateParts>
             {fromDomain && (
               <span className={styles.domain}>
-                {fromDomain.name ? (
-                  fromDomain.name
+                {domainName || fromDomain.name ? (
+                  domainName || fromDomain.name
                 ) : (
                   <FormattedMessage
```
rdig, line 213: This message descriptor should say team, not domain

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -94,6 +104,12 @@ const ActionsListItem = ({
     [handleOnClick, id, transactionHash],
   );
 
+  let domainName;
+  if (metadataJSON && actionType === ColonyActions.EditDomain) {
+    const domainObject = parseDomainMetadata(metadataJSON);
+    domainName = domainObject.domainName;
+  }
+
```
rdig, line 112: I think this is the only thing I could find "wrong" with this PR, the domain name in the actions list shows the "new" name, not the old one that was set when it was created.

Here's how to replicate:
1. create domain -> "domain second"
2. edit domain, from "domain second" -> "domain second plus"

The first entry in the list (the create domain one) should show "domain second", but currently it shows "domain second plus"

However when you click on the action, and go to action page it shows the correct values.

Here's a demo of the bug in action:

bug-domain-rename-action-list-current-domain

src/modules/dashboard/components/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+  cancel,
+  close,
+  colony,
+  colony: { colonyAddress, colonyName, domains },
+  selectedDomainId,
+}: Props) => {
+  const history = useHistory();
+
+  const validationSchema = yup.object().shape({
+    domainName: yup.string(),
```
rdig, line 50: I think we should restrict these to the value we set in the input's `maxLength` both for `domainName` and `domainPurpose`

src/modules/dashboard/components/EditDomainDialog/EditDomainDialog.tsx
```diff
@@ -0,0 +1,104 @@
...
+    domainId: yup.number().required(),
+    domainColor: yup.string(),
+    domainPurpose: yup.string(),
+    annotationMessage: yup.string().max(4000),
+  });
```
rdig, line 55: So that will be `20` and `90`: https://github.com/JoinColony/colonyDapp/blob/5ce5c8d735f69f5b1caf7851ee4ebfa05fa9976e/src/modules/dashboard/components/EditDomainDialog/EditDomainDialogForm.tsx#L156-L170

So something like this:
```suggestion
  const validationSchema = yup.object().shape({
    domainName: yup.string().max(20),
    domainId: yup.number().required(),
    domainColor: yup.string(),
    domainPurpose: yup.string().max(90),
    annotationMessage: yup.string().max(4000),
  });
```

Yes, I know that could be circumvented in other places, but that doesn't mean we can't properly validate it :)

PS: please add the same validation to the create domain modal

rdig (changes requested):
There's one thing that still doesn't work quite right:
- action list shows the new domain name instead of the old one

Beyond that, this is good to go, so once you've fixed that I'll approve and you can merge it.

rdig (approved):
All good now. Ready to merge!

chinins (approved):
Reviewed and tested - all good. Great job!

PR #2358, opened 2021-01-29
Fix the user's wallet route

chinins:
## Description

Fix the user's wallet route & edit tokens dialog & form

**New stuff** ✨

- add `UserTokenEditDialogForm` 
- add tokens to `InMemoryCache` user
- add balance fetching in `getTokenData` for `token` resolver

**Changes** 🏗

- change balance  from resolver function to field in `Token`
- fix several small things in `token` part of `user` resolver 
- update `UserTokens` query & typedef

**Deletions** ⚰️

- delete `balance` function in `token` resolver

## TODO

- none

Resolves DEV-168

Screenshot 2021-02-02 at 15 41 52
Screenshot 2021-02-02 at 15 41 10

### DIFF
--- a/src/data/graphql/typeDefs.ts
+++ b/src/data/graphql/typeDefs.ts
@@ -69,7 +69,7 @@ export default gql`
     symbol: String!
     iconHash: String
     verified: Boolean!
-    balance(walletAddress: String!): String!
+    balance: String
     balances(colonyAddress: String!, domainIds: [Int!]): [DomainBalance!]!
   }
 
@@ -124,7 +124,7 @@ export default gql`
 
   extend type User {
     reputation(colonyAddress: String!, domainId: Int): String!
-    tokens: [Token!]!
+    tokens(walletAddress: String): [Token!]!
     tokenTransfers: [Transfer!]!
     processedColonies: [ProcessedColony!]!
   }
--- a/src/data/graphql/typeDefs.ts
+++ b/src/data/graphql/typeDefs.ts
@@ -69,7 +69,7 @@ export default gql`
     symbol: String!
     iconHash: String
     verified: Boolean!
-    balance: String
+    balance: String!
     balances(colonyAddress: String!, domainIds: [Int!]): [DomainBalance!]!
   }
 
@@ -124,7 +124,7 @@ export default gql`
 
   extend type User {
     reputation(colonyAddress: String!, domainId: Int): String!
-    tokens(walletAddress: String): [Token!]!
+    tokens(walletAddress: String!): [Token!]!
     tokenTransfers: [Transfer!]!
     processedColonies: [ProcessedColony!]!
   }
--- a/src/modules/dashboard/components/Wallet/UserTokenEditDialog.tsx
+++ b/src/modules/dashboard/components/Wallet/UserTokenEditDialog.tsx
@@ -16,6 +16,8 @@ interface Props {
   walletAddress: Address;
 }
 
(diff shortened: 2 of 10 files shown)


src/data/graphql/typeDefs.ts
```diff
@@ -69,7 +69,7 @@ export default gql`
     symbol: String!
     iconHash: String
     verified: Boolean!
-    balance(walletAddress: String!): String!
+    balance: String
```
alicjakujawa, line 72: Are we sure that balance CAN be a null? I feel that here we should have String!

src/data/graphql/typeDefs.ts
```diff
@@ -124,7 +124,7 @@ export default gql`
 
   extend type User {
     reputation(colonyAddress: String!, domainId: Int): String!
-    tokens: [Token!]!
+    tokens(walletAddress: String): [Token!]!
```
alicjakujawa, line 127: Can we query tokens without walletAddress?

src/modules/dashboard/components/Wallet/UserTokenEditDialogForm.tsx
```diff
@@ -0,0 +1,222 @@
...
+
+import { UserTokens, OneToken, useLoggedInUser } from '~data/index';
+import { Address } from '~types/index';
+import { createAddress } from '~utils/web3';
+
+import styles from './UserTokenEditDialogForm.css';
+
+const MSG = defineMessages({
+  title: {
+    id: 'dashboard.UserTokenEditDialog.UserTokenEditDialogForm.title',
```
alicjakujawa, line 24: I think those ids are to change, UserTokenEditDialog is not needed here

chinins: Why? I thought that this was a convention that we are using, e.g.: 
`dashboard.TransferFundsDialog.TransferFundsDialogForm.title` (and also other dialog forms)
I am happy to change, just want to understand what would be the difference in this case.

src/data/graphql/typeDefs.ts
chinins: for the user no, you are right - good catch.

src/data/graphql/typeDefs.ts
chinins: yeah, probably

src/modules/dashboard/components/Wallet/UserTokenEditDialog.tsx
```diff
@@ -50,7 +51,7 @@ const UserTokenEditDialog = ({ cancel, close, walletAddress }: Props) => {
      * the user and wire it up for launch
      */
     // @ts-ignore
-    <TokenEditDialog
+    <DialogForm
```
rdig, line 54: Haha. You just fixed it!

So you can safely remove the above comment :)

src/modules/dashboard/components/Wallet/UserTokenEditDialogForm.tsx
rdig: Alicja is right here. this should be:
```
dashboard.Wallet.UserTokenEditDialogForm.title
```

Also the component should have a `displayName`, which should be `dashboard.Wallet.UserTokenEditDialogForm`

rdig (approved):
I'm good with this.

Pre-approving, but please take care of the stuff Alicja mentioned before merging!

ArmandoGraterol (approved):
It's looking dandy for me :fire: Great work!

PR #2361, opened 2021-02-04
ColonyEvents: sorting, filtering, load more, fix `TransferFunds` event

chinins:
### Description

This PR adds sorting & filtering of events list, load more button & logic & fixes `TransferFunds` event

**New stuff** ✨

- add filtering of events list
- add load more logic & button to the events list

**Changes** 🏗

 - refactored & updated names for sorting
 - fix how `TransferFunds` shows up

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-148
Resolves DEV-149
Resolves DEV-151

filter-domain
load-more


src/modules/dashboard/components/ColonyEvents/ColonyEvents.tsx
```diff
@@ -55,29 +69,62 @@ const ColonyEvents = ({ colony: { colonyAddress }, colony }: Props) => {
...
+                Number(displayValues.toPot) === ethDomainId) ||
+              /* when no specific domain in the event it is displayed in Root */
+              (ethDomainId === 1 &&
+                event.domainId === null &&
+                event.fundingPot === undefined &&
+                displayValues.toPot === undefined &&
+                displayValues.fromPot === undefined)
+            );
+          }),
+    [ethDomainId, events],
```
rdig, line 107: This is not needed here.

Consider the events list as a raw feed from the contract events.

That means no parsing of data what-so-ever, except for those that already come in the event `args`.

Please remove it

src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.tsx
```diff
@@ -108,6 +112,8 @@ const ColonyEventsListItem = ({
     metadata,
     tokenAddress,
     paymentId,
+    fromPot: displayValuesObject.fromPot,
+    toPot: displayValuesObject.toPot,
```
rdig, line 116: This shouldn't be needed anymore once you remove the `filtereEvents` logic, as it should just display data that comes in from the event only

rdig (changes requested):
This is good, except:
- We don't need to parse data for the events list, just display it as-is

I've left notes below on what you need to remove to make this list simpler.

rdig (approved):
All good now!

PR #2363, opened 2021-02-05
Add beta caution banner

chinins:
## Description

Add a little alert that cautions that it's BETA version. It should be displayed on each page and stay in the same place when scrolled.

I've added the component to the `App.tsx` file as it seems the most appropriate place to me.

**New stuff** ✨

- add `BetaCautionAlert` component & css

**Changes** 🏗

-  none

**Deletions** ⚰️

- delete the old alert that was on the transaction details page when not loggedin

## TODO

none

Resolves DEV-192

Screenshot 2021-02-09 at 10 05 33

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,7 +7,6 @@ import { ApolloProvider } from '@apollo/client';
 
 import layout from '~styles/layout.css';
 import { DialogProvider } from '~core/Dialog';
-import BetaCautionAlert from '~core/BetaCautionAlert';
 
 import messages from './i18n/en.json';
 import actionMessages from './i18n/en-actions';
@@ -40,7 +39,6 @@ const App = ({ store }: Props) => (
             <DialogProvider>
               <div className={layout.stretch}>
                 <Routes />
-                <BetaCautionAlert />
               </div>
             </DialogProvider>
           </BrowserRouter>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -7,6 +7,7 @@ import { ApolloProvider } from '@apollo/client';
 
 import layout from '~styles/layout.css';
 import { DialogProvider } from '~core/Dialog';
+import BetaCautionAlert from '~core/BetaCautionAlert';
 
 import messages from './i18n/en.json';
 import actionMessages from './i18n/en-actions';
@@ -39,6 +40,7 @@ const App = ({ store }: Props) => (
             <DialogProvider>
               <div className={layout.stretch}>
                 <Routes />
+                <BetaCautionAlert />
               </div>
             </DialogProvider>
           </BrowserRouter>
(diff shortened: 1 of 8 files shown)


src/App.tsx
```diff
@@ -39,6 +40,7 @@ const App = ({ store }: Props) => (
             <DialogProvider>
               <div className={layout.stretch}>
                 <Routes />
+                <BetaCautionAlert />
```
rdig, line 43: I mean, this works as well, but I'd like to have this just to set up the providers, and not display any actual components in it.

Can you please move it to the `pages` module, in one of the templates _(See: `RouteLayouts/Default`)_

rdig (approved):
All good here, I'd just wish you would render it somewhere else rather than the `App` root component. 

I've left a comment about the above.

ArmandoGraterol (approved):
Good enough for me :sunglasses:

src/modules/core/components/BetaCautionAlert/BetaCautionAlert.tsx
```diff
@@ -0,0 +1,57 @@
...
+              <FormattedMessage {...MSG.cautionText} />
+            </div>
+          </div>
+        </>
+      )}
+    </div>
+  );
+};
+
+export default BetaCautionAlert;
```
rdig, line 57: This shouldn't be hardcoded! _(like this)_

Still hardcoded, but passed through a `const` so it can be more easily spotted when scanning the file visually.

```suggestion
import React, { useState } from 'react';
import { FormattedMessage } from 'react-intl';

import ExternalLink from '~core/ExternalLink';

import styles from './BetaCautionAlert.css';

const MSG = {
  beta: {
    id: 'BetaCautionAlert.beta',
    defaultMessage: 'BETA 🤓',
  },
  cautionText: {
    id: 'BetaCautionAlert.cautionText',
    defaultMessage: 'Use with caution!',
  },
  learnMore: {
    id: 'BetaCautionAlert.learnMore',
    defaultMessage: 'Learn more',
  },
};

const LEARN_MORE_LINK = 'https://colony.io';

const BetaCautionAlert = () => {
  const [isHovered, setIsHovered] = useState(false);

  const toggleHover = (hasHover) => setIsHovered(hasHover);

  return (
    <div
      className={styles.container}
      onMouseEnter={() => toggleHover(true)}
      onMouseLeave={() => toggleHover(false)}
    >
      {isHovered ? (
        <ExternalLink
          text={MSG.learnMore}
          className={styles.link}
          href={LEARN_MORE_LINK}
        />
      ) : (
        <>
          <div className={styles.pinkStripe} />
          <div>
            <div className={styles.betaText}>
              <FormattedMessage {...MSG.beta} />
            </div>
            <div className={styles.cautionText}>
              <FormattedMessage {...MSG.cautionText} />
            </div>
          </div>
        </>
      )}
    </div>
  );
};

export default BetaCautionAlert;
```

rdig (approved):
All good, except I'd like the link to be passed through a constant _(like I showed you below in the suggestion)_

Also, not that you _should_ anything about it _(this is by design)_, but on monitors like mine, it's pretty hard to spot ...

Screenshot from 2021-02-09 20-36-56

_(and this is not even the full width... :slightly_smiling_face: )_

PR #2364, opened 2021-02-07
Use subscriptions to fetch Subgraph Actions and Events

rdig:
## Description

This PR refactors (initially) the actions and events queries and replaces them with subscriptions through WebSockets so that data for the will be reflected instantly once it gets ingested by the subgraph

**Changes**

- [x] Added `subscriptions-transport-ws` dependency
- [x] Added `SUBGRAPH_WS_ENDPOINT` env variable
- [x] Wired up WebSockets connection to the subgraph via the Apollo Client
- [x] Added `subscription` queries for both Actions and Events _(they can no longer be combined)_
- [x] Replaced queries with subscriptions in `ColonyEvents`
- [x] Replaced queries with subscriptions in `ColonyActions`
- [x] Refactored `getActionsListData` to support data coming from subscriptions
- [x] Remove actions queries in sagas since they are not longer needed
- [x] Added local events subscriptions resolvers
- [x] Add `SUBGRAPH_WS_ENDPOINT` to the docker image from [JoinColony/colony-docker-images](https://github.com/JoinColony/colony-docker-images) via [3d5fe96](https://github.com/JoinColony/colony-docker-images/commit/3d5fe96aef1159264fd06a036c8435e4c7f7764e)

**Demo**

demo-actions-subscriptions

Resolves DEV-165

I also think I inadvertently fixed `DEV-194` when removing the ColonyManager dependency in the component

Resolves DEV-194


ArmandoGraterol (approved):
Tested all the actions and events available as of now, and it's working well! The actions appear in the list instantly :+1: 

Also, tested signing out and it appears that it's also fixed

chinins (approved):
Looks and works great! I really liked how you moved some of the logic for filtering and formatting of actions to resolvers and transformers.

PR #2366, opened 2021-02-08
Prevent ColonyHome children constant re-rendering

rdig:
## Description

This PR fixes the problem of the actions and events lists constantly re-rendering when having the UAC modals opened and clicking around them.

This happened, apparently, because the `useDialog` hook instantiated in `ColonyHome` and which would change every time a modal would change (it's `id` would change actually), which will make the whole component re-render, including its children.

So since we don't actually need the upgrade colony modal to be declared in the `ColonyHome` component, we created a subcomponent to hold the logic that instantiates the `useDialog` hook.

**Changes**

- [x] Added `ColonyUpgrade` subcomponent
- [x] Added `ColonyFinishDeployment` subcomponent

Resolves DEV-195


ArmandoGraterol (approved):
Looks good to me! Tested it, and it's not reloading anymore :ok_hand:

chinins (approved):
Looks good to me. Works well as well.

PR #2368, opened 2021-02-09
Create Colony IPFS Metadata Upload Fallback/Retry

rdig:
This basically adds a try/catch provision in the colony create saga that:
- tries to upload metadata to ipfs
- if it fails, it tries again (we assume the first failure was a connection error)
- if both the above fail, set it to an empty string so that the contract call will go through

**Changes**

- [x] Add try/catch block in `colonyCreate` saga to retry IPFS upload

Resolves DEV-197

### DIFF
--- a/src/modules/dashboard/sagas/colonyCreate.ts
+++ b/src/modules/dashboard/sagas/colonyCreate.ts
@@ -38,6 +38,7 @@ import {
 } from '../../core/actionCreators';
 import { createTransaction, createTransactionChannels } from '../../core/sagas';
 import { ipfsUpload } from '../../core/sagas/ipfs';
+import { log } from '~utils/debug';
 
 interface ChannelDefinition {
   channel: Channel<any>;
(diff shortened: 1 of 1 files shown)


chinins (approved):
It's bit tricky to test the failure of ipfs upload only, especially only on the second try. But the code makes sense.

src/modules/dashboard/sagas/colonyCreate.ts
```diff
@@ -260,27 +261,61 @@ function* colonyCreate({
...
+        colonyMetadataIpfsHash = yield call(
+          ipfsUpload,
+          JSON.stringify({
+            colonyName,
+            colonyDisplayName: displayName,
+            colonyAvatarHash: null,
+            colonyTokens: [],
+          }),
+        );
+      }
```
alicjakujawa, line 300: If the second upload call will throw an error, where we're catching this error?

rdig: the try / catch inside the saga is not about catching the errors, we're just using it as a mechanism for retry-ing.

Both IPFS libs that we use handle their own error catching, we just use the try/catch block for setting the hash value _(what's why we check if it's a string, because if it throws an Error we don't actually know)_

Anyway, this is just a stop-gap solution, until we implement a more proper way of hadling IPFS uploads and their failure states _(most likely via a queue and using Service Workers)_

alicjakujawa (approved):

PR #2369, opened 2021-02-09
Don't upload Colony ENS name to IPFS metadata

rdig:
This PR removes `colonyName` (the colony's ens name) from being uploaded to the ipfs metadata object. This is done because of two things:
- while we're not using it right now, someone, down the line will and that will be a problem since it's unreliable _(it can be modified outside of the dapp)_
- shaving a few bytes off the upload data

Obiously one is more important than the other :)

**Changes** 

- [x] Remove `colonyName` from the metadata ipfs upload in the `editColonyAction` saga

Resolves DEV-196

### DIFF
--- a/src/modules/core/components/Fields/Checkbox/Checkbox.css
+++ b/src/modules/core/components/Fields/Checkbox/Checkbox.css
@@ -17,7 +17,7 @@
   height: 18px;
   width: 18px;
   position: relative;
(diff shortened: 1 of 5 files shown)


ArmandoGraterol (approved):
I'll miss uploading the Colony ENS name, but oh well! It's working fine without it

alicjakujawa (approved):

PR #2370, opened 2021-02-09
Fixed domain selector not changing domains on ColonyFunding

ArmandoGraterol:
Should fix the domain selector not changing the domain of the tokens displayed

Resolves DEV-188


chinins (approved):
It works for the root domain. I will fix the bug that you found while doing this PR (DEV-199)

rdig (approved):
:+1:

PR #2372, opened 2021-02-10
Add feedback link & integrate feedback.fish

chinins:
## Description

This is to add a link to an external tracking service called feedback.fish

This is integrated by having a button in the bottom right corner, and, upon clicking it, the  feedback.fish widget will show.

**New stuff** ✨

- add `@feedback-fish/react` package
- add `FeedbackWidget` component that integrates the service
- create `FEEDBACK_FISH_PROJECT_ID` env variable (so far empty)

**Changes** 🏗

none

**Deletions** ⚰️

- delete `height: auto` for `iframe` in `main.css` so that the external component can be displayed properly

## TODO

- To create a corporate account with Feedback.fish and add the project ID to the env variable

Resolves DEV-198

Screenshot 2021-02-11 at 11 51 52

### DIFF
--- a/src/modules/core/components/FeedbackWidget/FeedbackWidget.css
+++ b/src/modules/core/components/FeedbackWidget/FeedbackWidget.css
@@ -12,4 +12,8 @@
   font-weight: var(--weight-bold);
   color: color-mod(var(--temp-grey-blue-7) alpha(85%));
   cursor: pointer;
+
+  &:focus {
+    outline: none;
+  }
 }
--- a/src/modules/core/components/FeedbackWidget/FeedbackWidget.css
+++ b/src/modules/core/components/FeedbackWidget/FeedbackWidget.css
@@ -1,0 +1,15 @@
+.container {
+  position: absolute;
+  right: 34px;
+  bottom: 16px;
+}
+
+.button {
+  padding: 0;
+  border: none;
+  background-color: transparent;
+  font-size: var(--size-tiny);
+  font-weight: var(--weight-bold);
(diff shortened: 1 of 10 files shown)


src/modules/core/components/FeedbackWidget/FeedbackWidget.css
```diff
@@ -0,0 +1,16 @@
+.button {
+  padding: 0;
+  position: absolute;
+  right: 34px;
+  bottom: 16px;
+  border: none;
```
rdig, line 6: So the feedback fish widget is kinda garbage in the way it works.

I'm sure you've noticed this but upon first render it renders outside of the screen:

demo-feedback-widget-outside-screen

This is because it uses popper under the hood and that needs a center point reference to align properly to it.

Because of this we can hack around it:
- set the width equal to the feedback fish popup
- align your text to the right
- reduce the right positioning (to better align with the feedback fish popup)
- compensate with padding, the lost right position space

So something like this:
```suggestion
  padding: 0;
  position: absolute;
  right: 25px;
  bottom: 16px;
  border: none;
  width: 320px;
  padding-right: 9px;
  text-align: right;
```

This will still keep the feedback link in your original position, but allow the feeback fish popoup to render in the same place every time:

demo-feedback-widget-proper

rdig (changes requested):
I want you to make two changes to this before we can merge it:
- color the heart emoji red _(see Figma for the proper color)_
- make the feedback fish popover render in the same place always _(I've left examples and instructions on how to do this)_

rdig (approved):
Good to go!

PR #2376, opened 2021-02-15
Fixed FriendlyName showing incorrect name/address

ArmandoGraterol:
Fixes FriendlyName always showing the name of the colony when the initiator/recipient is not a member of the colony

Resolves DEV-183

### DIFF
--- a/src/modules/core/components/FriendlyName/FriendlyName.tsx
+++ b/src/modules/core/components/FriendlyName/FriendlyName.tsx
@@ -86,9 +86,11 @@ const FriendlyName = ({
       addressRef.current.style.fontSize = `${inheritedFontSize - 1}px`;
     }
   }, [addressRef, autoShrinkAddress]);
+  const isColony = user?.profile.walletAddress === colony?.colonyAddress;
+
   return (
     <div className={styles.main}>
-      {userDisplay || colonyDisplay || (
+      {userDisplay || (isColony && colonyDisplay) || (
         <MaskedAddress
           address={userDisplayAddress || colonyDisplayAddress}
           full={!maskedAddress}


src/modules/core/components/FriendlyName/FriendlyName.tsx
```diff
@@ -86,9 +86,11 @@ const FriendlyName = ({
       addressRef.current.style.fontSize = `${inheritedFontSize - 1}px`;
     }
   }, [addressRef, autoShrinkAddress]);
+  const isColony = user?.profile.walletAddress === colony?.colonyAddress;
+
   return (
     <div className={styles.main}>
-      {userDisplay || colonyDisplay || (
+      {userDisplay || (isColony && colonyDisplay) || (
```
alicjakujawa, line 93: Are we sure that if userDisplayAddress is undefined/null, it will be always colony?

ArmandoGraterol: Before, it was treated that way. The `userDisplay` is always null/undefined when the address is not really a part of the colony (So, an address that is not a member of the colony for example). The `colonyDisplay`, from what I've seen, is always a truthy value (the name of the colony where the transaction originated from).

If it's not a member of the colony, and it's not the colony itself, then it just displays the masked address.  Before, from what I tested, the conditional never went past the `colonyDisplay` because if `userDisplay` was false, then the `colonyDisplay` was **always** true

alicjakujawa: but if its not colony itself, then there is a possibility that we display colonyDisplayAddress if userDisplayAddress is undefined. So the question is, if userDisplayAddress is undefined, then are we 100% sure that its colonyAddress?

ArmandoGraterol: Nope, it's not guaranteed that it will be a colonyAddress. From my understanding, if the `userDisplay` is undefined, then that only means 2 things:

A. The address to display is the colony's address
B. The address to display is from something or someone outside the colony (Another colony, a user which hasn't joined in the colony, etc...)

alicjakujawa: So, in this case we will still have bug as we display colonyDisplayAddress not "user which hasn't joined the colony" 🤔

ArmandoGraterol: Are you saying that because you tested it and it's still showing the colonyDisplay? Because case B should be covered by the:

`|| (
        <MaskedAddress ... /> )` 

If the address it's from a user which hasn't joined then it should show a masked address, and the colonyDisplay should only be appearing when the address is exactly the colony's address

ArmandoGraterol: With the `isColony &&` I added, an unjoined user should not be displayed as the colony name

alicjakujawa: Nope im thinking about scenario when:
* User not joined colony, so userDisplayAddress is null
* colonyDisplayAddress probably won't be null
In that case, we will put to MaskedAddress - colonyDisplayAddress
but we probably want to show address of user which is not a member do we?

ArmandoGraterol: Oh, I think I see what you mean! No, in that case, the userDisplayAddress won't be null because we would obtain the address from the processed action/event, most likely from the `receptor` prop. 

Yes, the user is not a known user to the colony, so it can't obtain the username(userDisplay), but it can still obtain the wallet address through the data that we receive from the action/event/others so it's not null.

alicjakujawa (approved):
🔥

src/modules/core/components/FriendlyName/FriendlyName.tsx
rdig: I like this back and forth! Nice going @alicjakujawa and @ArmandoGraterol

rdig (approved):
Yep, I'm good with this.

PR #2377, opened 2021-02-16
Allow users to create payments in subdomains

rdig:
This PR fixes the issue where a user, even with the correct permissions, could not create a payment in a subdomain.

This was because the domain dropdown was disabled unless you had permissions in "Root".

To fix this, I've enabled the dropdown for all access levels so that you can get to the domain you have permissions in. While I was at it I've fixed a bug where the custom error validation was not resetting properly.

fix-subdomain-payment

As a future improvement, we could remove all domains from the list, in which the user doesn't have permissions.

Resolves DEV-211


chinins (approved):
Works and looks good to me.

PR #2379, opened 2021-02-16
Fix 404 redirection in ColonyHome page

ArmandoGraterol:
* Fixed `data?.processedColony?.colonyName !== colonyName` always being true(and thus, always showing the loader) when `processedColony` arrives as `undefined`
* Added extra conditional for 404 redirection that checks for errors on the `data.colonyAddress` prop

Resolves DEV-210

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
@@ -97,7 +97,10 @@ const ColonyHome = ({ match, location }: Props) => {
    * This problem is made even worse in an production environment where loading
    * times are slow.
    */
-  if (loading || data?.processedColony?.colonyName !== colonyName) {
+  if (
+    loading ||
+    (data?.processedColony && data.processedColony.colonyName !== colonyName)
+  ) {
     return (
       <div className={styles.loadingWrapper}>
         <LoadingTemplate loadingText={MSG.loadingText} />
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
```diff
@@ -97,15 +97,23 @@ const ColonyHome = ({ match, location }: Props) => {
...
       </div>
     );
   }
 
-  if (!colonyName || error || !data?.processedColony) {
+  if (
+    !colonyName ||
+    error ||
+    !data?.processedColony ||
+    (data?.colonyAddress as any) instanceof Error
```
chinins, line 115: Why do you need this check for instanceof Error?

It works without it and i think that the error case is covered by `error` check

chinins (commented):
It works good. Just need an answer to my question about instanceof Error check before approving

src/modules/dashboard/components/ColonyHome/ColonyHome.tsx
rdig: This is correct, it's how we can detect a error in the first resolver we are calling.

See: https://github.com/JoinColony/colonyDapp/blob/a364c74f0d3f7fa356807ad242eea29984e8737b/src/data/resolvers/colony.ts#L226-L243

rdig (approved):
Good to go

PR #2380, opened 2021-02-17
Add validation for AddressZero while creating colony

alicjakujawa:
## Description

This PR adds in validation in SelectTokenStep to prevent user from entering ETH/Xdai tokens while creating colony

Screenshot 2021-02-17 at 10 50 14
 

Resolves DEV-206

### DIFF
--- a/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
+++ b/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
@@ -1,11 +1,11 @@
 import { $PropertyType } from 'utility-types';
 import { FormikBag } from 'formik';
 import React, { useCallback, useState } from 'react';
-import { defineMessages, FormattedMessage } from 'react-intl';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
 import * as yup from 'yup';
 import { AddressZero } from 'ethers/constants';
 
-import { DEFAULT_NETWORK_TOKEN, XDAI_TOKEN } from '~constants';
+import { DEFAULT_NETWORK_TOKEN } from '~constants';
 import { WizardProps } from '~core/Wizard';
 import { Form, Input } from '~core/Fields';
 import Heading from '~core/Heading';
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
```diff
@@ -57,6 +59,14 @@ const MSG = defineMessages({
...
       'Not a valid token. Only ERC20 tokens with 18 decimals are supported.',
   },
+  xdaiAddress: {
+    id: 'dashboard.CreateColonyWizard.StepSelectToken.xdaiAddress',
+    defaultMessage: 'You cannot use XDAI token as a native token for colony.',
+  },
+  ethAddress: {
+    id: 'dashboard.CreateColonyWizard.StepSelectToken.xdaiAddress',
+    defaultMessage: 'You cannot use ETH token as a native token for colony.',
+  },
```
rdig, line 69: This works as well, but when I suggested you use the `DEFAULT_NETWORK_TOKEN` constant, I was expecting you to pass it along as a value.

```suggestion
  ethAddress: {
    id: 'dashboard.CreateColonyWizard.StepSelectToken.xdaiAddress',
    defaultMessage: 'You cannot use {tokenSymbol} token as a native token for colony.',
  },
```

Then call it passing along a values object:
```
values={{ tokenSymbol: DEFAULT_NETWORK_TOKEN.symbol }}
```

That way you don't really care what network we are on, you just care that the token address is not `0x00000`

src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
```diff
@@ -71,8 +81,17 @@ const MSG = defineMessages({
...
+const isXdai = DEFAULT_NETWORK_TOKEN.symbol === XDAI_TOKEN.symbol;
+
 export const validationSchema = yup.object({
-  tokenAddress: yup.string().address(() => MSG.invalidAddress),
+  tokenAddress: yup
+    .string()
+    .address(() => MSG.invalidAddress)
+    .test(
+      'is-not-addressZero',
+      () => (isXdai && MSG.xdaiAddress) || MSG.ethAddress,
```
rdig, line 92: So what happens when we add another network to the mix, one that is not ethereum and not xdai?

See my above suggestion about passing in the `DEFAULT_NETWORK_TOKEN` as a value to the message descriptor.

I think you'll need to use `formattedMessage()` from `react-intl` for this

src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
rdig: See an usage example here:

https://github.com/JoinColony/colonyDapp/blob/c3299c8db8a1f2a40bd18deeb4a51495893d81c7/src/modules/dashboard/components/CreateColonyWizard/StepUserName.tsx#L155-L161

rdig (changes requested):
I like your fix, but this will not work, if we add a third network to the mix.

It just assumes that we are either on eth or xdai. But this might not be the case in the future.

src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
alicjakujawa: i also wanted to do like that, but like you see, yup object schema is define outside of the component. And to use formatMessage method we need to use hook useUntil(). I was also trying to return component <FormattedMessage /> but it also wasn't work

rdig (approved):
Good to go!

PR #2381, opened 2021-02-17
Add token activation button to UserNavigation

alicjakujawa:
## Description

This PR adds in button for token activation to user navigation bar

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves DEV-212


src/data/graphql/queries.graphql
```diff
@@ -204,6 +204,21 @@ query UserTokens($address: String!) {
   }
 }
 
+query UserBalanceWithLock($address: String!, $tokenAddress: String!) {
+  user(address: $address) {
```
chinins, line 208: Is `withLock` and further in the code `userLock`  inactive tokens? If yes, can we please rename it to `inactive`. In my popover I will actually have total locked tokens which you don't refer to here (as far as I understand), so I think it will be a bit confusing. 

I'm not trying to nitpick, I just think it will be easier to have the names corresponding to the UI: active, inactive, total, total locked (I am not sure about the last one as I still need to understand what it refers to).

src/modules/pages/components/RouteLayouts/UserNavigation.tsx
```diff
@@ -56,6 +78,16 @@ const UserNavigation = () => {
...
+  const userLock = userData?.user.userLock;
+  const nativeToken = userLock?.nativeToken;
+  const userActiveBalance = nativeToken?.balance;
+
+  const lockedBalance =
+    (userLock && bigNumberify(userLock.balance)) || bigNumberify(0);
+  const walletBalance =
+    (userActiveBalance && bigNumberify(userActiveBalance)) || bigNumberify(0);
+  const totalBalance = lockedBalance.add(walletBalance);
+
```
chinins, line 90: as per my previous comment, could we rename to active/inactive balance, please?

chinins (changes requested):
Several points:

- as per my comments, could we rename the balances to active/inactive to avoid confusion and have it corresponding to UI? It took me a while to process the `lock` names  and wallet balances :)

- focus styles are not correct: i think you need to remove outline & add border color same blue as the font color. Also there is a little stipe of the neighboring color that gets in the outline.

- from Jack's discord message, I thought we should show decimals when 5 figure numbers or smaller. Right now, no decimals are shown with small numbers.

- there is no tooltip on hover

- when I make a payment that increases the token number it doesn't update the dot color & the number of tokens until i refresh the page

How should I test when tokens are inactive? Or is it only possible when we have actions wired?

the current focus styles:
Screenshot 2021-02-22 at 14 17 27

chinins:
This helps me for my branch. I will rebase and can actually pass proper props :)

chinins:
If some of my comments are not clear or I misunderstood something, please, let me know and we can also have a quick call to discuss.

src/data/graphql/queries.graphql
alicjakujawa: so this is the contract:
```  struct Lock {
    // User's lock count
    uint256 lockCount;
    // Deposited balance
    uint256 balance;
    // Weighted average of deposit timestamps (no longer used)
    uint256 DEPRECATED_timestamp; // solhint-disable-line var-name-mixedcase
    // Pending balance (from failed transfers), can claim with a forced deposit
    uint256 pendingBalance;
  }```
and this is the network method im calling: getUserLock

I would honestly prefer to be consistent in naming with network.
Balance - from Lock contract is the balance which is currently "stake or locked" for this user
balance from nativeToken - Is balance which this user have in this token
TOTAL balance = userLock.balance + userLock.nativeToken.balance
I wouldn't name userLock.balance = inactive - because from different perspective is not necessarily "inactive" is just "staked" or "locked"

**Comment by alicjakujawa** [comment]
> Several points:
> 
> * as per my comments, could we rename the balances to active/inactive to avoid confusion and have it corresponding to UI? It took me a while to process the `lock` names  and wallet balances :)
> * focus styles are not correct: i think you need to remove outline & add border color same blue as the font color. Also there is a little stipe of the neighboring color that gets in the outline.
> * from Jack's discord message, I thought we should show decimals when 5 figure numbers or smaller. Right now, no decimals are shown with small numbers.
> * there is no tooltip on hover
> * when I make a payment that increases the token number it doesn't update the dot color & the number of tokens until i refresh the page
> 
> How should I test when tokens are inactive? Or is it only possible when we have actions wired?
> 
> the current focus styles:
> Screenshot 2021-02-22 at 14 17 27

* About decimals:
Screenshot 2021-02-22 at 16 32 59
* about styles "nitpicks" its for sure not. yet done, it will be in following steps, as this task either way will be a feature branch for this feature, so i will go with them after we will have "base" from which we can work. To not block you and Armando in your tasks :)
* tooltips - I will also add it, but lets first agree about base. 
* for the last point, i will debug create expenditure saga, probably something there is missing :)

File: src/data/graphql/queries.graphql (hunk 1 above)
Comment (chinins) [reply src/data/graphql/queries.graphql:208]: I see your point... All of these balances are not clear to me... Especially I know that we should show another one as per Karol's message in discord:
"these are still active tokens, however they are already in use somewhere. for example, being part of expenditure budget, at pending state, and so on. so as a result, they cannot be withdrawn. let me know if that sounds clear"

Can we maybe have a call where you explain the contract balances to me, please? Should we include Alex as well?

**Comment by chinins** [comment]
Of course, the base is good. Let's just connect regarding these balances. I'll ping you in discord.

File: src/modules/pages/components/RouteLayouts/UserNavigation.css [hunk 3]
```diff
@@ -106,17 +106,19 @@
...
   border-radius: var(--radius-normal);
   background-color: color-mod(var(--temp-grey-blue-7) alpha(10%));
-  font-size: 13px;
-  font-weight: var(--weight-medium);
   color: var(--temp-grey-blue-7);
   cursor: pointer;
 }
 
+.walletAddress span {
+  font-size: 13px;
```
ArmandoGraterol, line 118: You could use `--size-smallish` here :dog:

src/modules/pages/components/RouteLayouts/UserNavigation.css
```diff
@@ -142,3 +150,33 @@
   color: var(--colony-white);
   box-shadow: 0px 2px 11px var(--drop-shadow);
 }
+
+.tokens {
+  composes: walletAddress;
+  margin-right: -5px;
+  border: 0;
+  background: transparent;
+  font-size: var(--size-smallis);
```
ArmandoGraterol, line 159: Typo here. Should be `--size-smallish`

src/modules/pages/components/RouteLayouts/UserNavigation.css
```diff
@@ -106,17 +106,19 @@
...
   background-color: color-mod(var(--temp-grey-blue-7) alpha(10%));
-  font-size: 13px;
-  font-weight: var(--weight-medium);
   color: var(--temp-grey-blue-7);
   cursor: pointer;
 }
 
+.walletAddress span {
+  font-size: 13px;
+  font-weight: 600;
```
ArmandoGraterol, line 119: Do we have a font that has 600 weight available? I think we only have 400, 500, and 700

ArmandoGraterol (changes requested):
Nice! Apart from some things Olga already mentioned, and these small things I noticed, this looks good

src/modules/pages/components/RouteLayouts/UserNavigation.css
alicjakujawa: Thats true 🤔  Designs want font-weight: 600, i think i will switch it to 500 for now and we will see feedback on QA

rdig: If there's any feedback on QA about this, ignore it for now.

We're not adding a new font-weight just for this. That involves adding the required font files, which are about 300k

src/data/resolvers/user.ts
```diff
@@ -41,6 +44,35 @@ const getUserReputation = async (
...
 };
 
+const getUserLock = async (
+  client: ApolloClient<object>,
+  colonyManager: ColonyManager,
+  walletAddress: Address,
+  tokenAddress: Address,
+): Promise<UserLock> => {
+  const { networkClient } = colonyManager;
+  const tokenLockingClient = await networkClient.getTokenLockingClient();
```
rdig, line 54: This is not the correct way of getting the token locking client. 

I'm actually surprised this worked.

Apply this patch:
```diff
diff --git a/src/lib/ColonyManager/ColonyManager.ts b/src/lib/ColonyManager/ColonyManager.ts
index 0b41f7682..20b1e995a 100644
--- a/src/lib/ColonyManager/ColonyManager.ts
+++ b/src/lib/ColonyManager/ColonyManager.ts
@@ -9,8 +9,10 @@ import {
   Extension,
   OneTxPaymentClient,
   TokenClient,
+  TokenLockingClient,
   ExtensionClient,
 } from '@colony/colony-js';
+import getTokenLockingClient from '@colony/colony-js/lib/clients/TokenLockingClient';
 
 import ENS from '~lib/ENS';
 import { isAddress } from '~utils/web3';
@@ -27,6 +29,8 @@ export default class ColonyManager {
 
   tokenClients: Map<Address, Promise<TokenClient>>;
 
+  tokenLockingClients: Map<Address, TokenLockingClient>;
+
   networkClient: ColonyNetworkClient;
 
   provider: Provider;
@@ -37,6 +41,7 @@ export default class ColonyManager {
     this.colonyClients = new Map();
     this.extensionClients = new Map();
     this.tokenClients = new Map();
+    this.tokenLockingClients = new Map();
     this.networkClient = networkClient;
     this.provider = networkClient.provider;
     this.signer = networkClient.signer;
@@ -117,6 +122,11 @@ export default class ColonyManager {
     identifier?: AddressOrENSName,
   ): Promise<TokenClient>;
 
+  async getClient(
+    type: ClientType.TokenLockingClient,
+    identifier?: AddressOrENSName,
+  ): Promise<TokenLockingClient>;
+
   async getClient(
     type: ClientType.OneTxPaymentClient,
     identifier?: AddressOrENSName,
@@ -146,6 +156,18 @@ export default class ColonyManager {
         const colonyClient = await this.getColonyClient(identifier);
         return colonyClient.tokenClient;
       }
+      case ClientType.TokenLockingClient: {
+        if (!identifier)
+          throw new Error('Need colony identifier to get TokenLockingClient');
+        const colonyClient = await this.getColonyClient(identifier);
+        const tokenLockingClient = this.getTokenLockingClient(identifier);
+        /*
+         * @TODO This needs a proper fix in colonyJS
+         */
+        // @ts-ignore
+        colonyClient.tokenLockingClient = tokenLockingClient;
+        return tokenLockingClient;
+      }
       case ClientType.OneTxPaymentClient: {
         if (!identifier)
           throw new Error('Need colony identifier to get OneTxPaymentClient');
@@ -173,4 +195,13 @@ export default class ColonyManager {
     }
     return clientPromise;
   }
+
+  async getTokenLockingClient(address: Address): Promise<TokenLockingClient> {
+    let clientPromise = this.tokenLockingClients.get(address);
+    if (!clientPromise) {
+      clientPromise = getTokenLockingClient(address, this.signer);
+      this.tokenLockingClients.set(address, clientPromise);
+    }
+    return clientPromise;
+  }
 }
```

Then get it from the colonyManager like this:

```
await colonyManager.getClient(
  ClientType.TokenLockingClient,
  colonyAddress,
);

File: src/data/resolvers/user.ts [hunk 7]
```diff
@@ -41,6 +44,35 @@ const getUserReputation = async (
   return reputationAmount;
 };
 
+const getUserLock = async (
+  client: ApolloClient<object>,
```
rdig, line 48: please name this `apolloClient` so it's more descriptive

src/modules/core/components/Numeral/Numeral.tsx
```diff
@@ -59,7 +59,7 @@ const Numeral = ({
     prefix,
     suffix,
     integerSeparator,
-    truncate,
+    truncate: convertedNum > 10 ** 4 ? 0 : truncate,
```
rdig, line 62: I'd still want us to be able to control `truncate` manually.

So the "large number" logic should go in some other prop we just toggle true/false. (which should be true by default)

You can name it something like: `reducedOutput`

src/modules/pages/components/RouteLayouts/UserNavigation.tsx
```diff
@@ -89,31 +120,51 @@ const UserNavigation = () => {
...
+              />
+              <Numeral
+                suffix={` ${nativeToken?.symbol} `}
+                unit={getTokenDecimalsWithFallback(nativeToken?.decimals)}
+                value={totalBalance}
+                truncate={3}
+              />
+            </button>
+          </>
+        )}
```
rdig, line 141: This should be moved into a sub-component, along with all the lock / balance checks

rdig (commented):
I'm fine for the most part with this, however, I left some comments that should be improved upon.

As for the thing Olga mentioned, that this is not complete _(missing styles, tooltips, etc)_: please open another issue to track those, and add them in a serparate PR

PR #2383, opened 2021-02-17
Fix/dev 202 colony selectors styles

chinins:
## Description

This update colony selectors to the latest design as per the link:
https://www.figma.com/file/DjVCPq7LpkjEMEfQZtKgBdC1/Style-Guide?node-id=3473%3A238

The Figma specs show height and width of 37 px. However, I took a standard size `s` of 34px for the avatar due to 3 reasons:
- we don't have 37px size and I think it would be a bit much to add it 
- didn't look that good
- the user avatar in the top right corner and the avatars in actions list are 34 px and i think it's better to keep it consistent

**New stuff** ✨

- none

**Changes** 🏗

- increased the size of the colony avatars
- increased the border of selected avatar to 3px
- added a transparent border of 1 px between the image and the blue selected border
- changed color of the plus icon to lighter grey as per specs 
- updated the padding-margin of the list items & the calculation of the max height of the list container

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-202

Screenshot 2021-02-17 at 17 34 23


src/img/icons/circle-plus.svg
```diff
@@ -1,4 +1,4 @@
 <svg width="28" height="28" viewBox="0 0 28 28" fill="none" xmlns="http://www.w3.org/2000/svg">
-	<circle cx="14" cy="14" r="13.5" stroke="#3C444D"/>
-	<path fill-rule="evenodd" clip-rule="evenodd" d="M14.25 6C14.1119 6 14 6.11193 14 6.25V13.5H6.75C6.61193 13.5 6.5 13.6119 6.5 13.75V14.25C6.5 14.3881 6.61193 14.5 6.75 14.5H14V21.75C14 21.8881 14.1119 22 14.25 22H14.75C14.8881 22 15 21.8881 15 21.75V14.5H22.25C22.3881 14.5 22.5 14.3881 22.5 14.25V13.75C22.5 13.6119 22.3881 13.5 22.25 13.5H15V6.25C15 6.11193 14.8881 6 14.75 6H14.25Z" fill="#3C444D"/>
+	<circle cx="14" cy="14" r="13.5" stroke="#76748B"/>
```
alicjakujawa, line 2: Why this change? I think circle-plus is used in some other places? We can change color of this with css if we need

src/modules/dashboard/components/SubscribedColoniesList/SubscribedColoniesList.tsx
```diff
@@ -57,7 +57,7 @@ const SubscribedColoniesList = () => {
                     <ColonyAvatar
                       colony={colony}
                       colonyAddress={colonyAddress}
-                      size="xs"
+                      size="s"
```
alicjakujawa, line 60: I think we should stick to sizes define in Avatar component so in case of "s" it will be 34x34 instead of adding new case in css 36x36, maybe lets ask Karol if he will agree with that?

alicjakujawa (changes requested):
Looks good, just small comments 🔥

src/img/icons/circle-plus.svg
chinins: Yeah, I see your point. I was not 100% sure. I decided for this option because ofconsistency. So far this emoji is used  in 2 places to add colonies so they should be the same and in create domain from actions page where i think it would be good to have the same color.

But again I am not sure. I am happy to change back and apply it with css.

ArmandoGraterol (approved):
New styles are looking good to me :+1:

alicjakujawa (approved):

PR #2384, opened 2021-02-18
Display correct block explorer name in Transaction Meta

rdig:
## Description

This PR fixes an issue reported on the xdai mainnet where the block explorer, on the action page, would still show up as Etherescan, despite xdai using Blockscout as it's explorer.

The link to the specific transaction was created correctly, it was just the name that was at fault here.

This was due to a message descriptor, to which we would pass a correct block explorer value, but since it was hard-coded it didn't actually display it.

**Changes** 

- [x] Fixed message descriptor in `TransactionMeta` to actually display values

Resolves DEV-217


chinins (approved):
All good!

PR #2385, opened 2021-02-18
Implement token activation popover

chinins:
## Description

Implement Token Activation popover.

It should:
- Show the total amount of tokens.
- Show the active amount of the previous total amount of tokens.
- Show the number of unclaimed tokens (doesn't count to the total of tokens).
- Show the inactive amount of the previous total amount of tokens.
-  Have 2 tabs to change between deposit and withdraw
- Have 1 input for each tab to write the amount of token the user wishes to deposit/withdraw. With a Max button to quickly put the max amount of available tokens. 
- Have tooltips explaining what are active tokens, inactive tokens, and unclaimed transactions

**The claims tab for now should not be there as this feature is not yet implemented. I realised it after I have already added it. So I commented out some code that we can use when we add `ClaimsTab`.**

**New stuff** ✨

- add `emoji-padlock-closed` svg
- add `white` button variant
- add `TokenActivationPopover`
- add `TokenActivationContent`
- add `ClaimsTab`
- add `TokensTab`
- add `ChangeTokenStateForm`
- cherry picked Armando's commit to add action types for the feature

**Changes** 🏗

- add `UserToken` to `AnyToken` type
- make label optional prop in `Input` core component
- add `formatTokenValue` utils helper & use it in `Numeral` component. I did that so that number of tokens & token symbol can be separate when using `formatTokenValue` and we have the same code that does it.
- add `className` prop to `Tab` & `TabList` so that it's not mandatory to use their default styles

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-213


src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+import { UserToken } from '~data/generated';
+import { ActionTypes } from '~redux/index';
+import { pipe, mapPayload } from '~utils/actions';
+
+import styles from './TokenActivationContent.css';
+
+const MSG = defineMessages({
+  changeState: {
+    id: 'users.TokenActivation.TokenActivationContent.TokensTab.changeState',
```
ArmandoGraterol, line 20: This should be changed from `.TokensTab.` from to`ChangeTokenStateForm` right?

ArmandoGraterol: Same with the other message ids

src/modules/users/components/TokenActivation/TokenActivationContent/TokensTab.tsx
```diff
@@ -0,0 +1,149 @@
...
+    defaultMessage: 'Active',
+  },
+  inactive: {
+    id: 'users.TokenActivation.TokenActivationContent.TokensTab.inactive',
+    defaultMessage: 'Inactive',
+  },
+  activeTokensTooltip: {
+    id:
+      // eslint-disable-next-line max-len
+      'users.TokenActivation.TokenActivationContent.TokensTab.activeTokensTooltip',
```
ArmandoGraterol, line 29: You could use back quotes( \` ) here to avoid having to use the `eslint-disable-next-line max-len`

src/modules/users/components/TokenActivation/TokenActivationContent/TokensTab.tsx
```diff
@@ -0,0 +1,149 @@
...
+    defaultMessage: `Tokens are “Active” when they’ve been deposited to a
+     contract which lets them get ‘locked’ when you need to stake,
+     or claim a share of Rewards. You can withdraw tokens back
+     to your wallet any time, you just need to clear any locks first.`,
+  },
+  inactiveTokensTooltip: {
+    id:
+      // eslint-disable-next-line max-len
+      'users.TokenActivation.TokenActivationContent.TokensTab.inactiveTokensTooltip',
```
ArmandoGraterol, line 38: Same here, could use back quotes

src/modules/core/components/Button/Button.css
```diff
@@ -322,3 +322,19 @@
 .themeDottedArea:hover {
   opacity: 0.7;
 }
+
+/*
+ * White
+ * buttons with white background, --temp-grey-blue-7 border & font color
+ */
+
+.themeWhite {
```
rdig, line 331: Are you sure `themeGhost` doesn't do exactly this ?

src/modules/core/components/Numeral/Numeral.tsx
```diff
@@ -54,19 +53,15 @@ const Numeral = ({
...
+    unit,
+    value,
     prefix,
     suffix,
     integerSeparator,
-    truncate: reducedOutput ? largeNumberTruncate : truncate,
-  })(parseFloat(convertedNum));
+    truncate,
+    reducedOutput,
+  });
```
rdig, line 64: Can you explain this change please?

Also this needs very rigorous testing, as we use `Numeral` in a hell of alot of places. I just want to make sure we didn't borked some other place of the dapp

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+        onSelect={(newIndex) => {
+          setTabIndex(newIndex);
+        }}
+      >
+        <TabList
+          className={styles.tabsList}
+          containerClassName={styles.tabsListContainer}
+        >
+          <Tab
+            //  selectedClassName={styles.tabSelected}
```
rdig, line 36: ?

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+            className={styles.tab}
+          >
+            <FormattedMessage {...MSG.yourTokens} />
+          </Tab>
+          {/* <Tab selectedClassName={styles.tabSelected} className={styles.tab}>
+            <FormattedMessage {...MSG.claims} />
+          </Tab> */}
```
rdig, line 43: Is this disabled for a reason, or leftover ?

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
```diff
@@ -0,0 +1,56 @@
...
+        </TabList>
+        <TabPanel className={styles.tabContainer}>
+          <TokensTab {...props} />
+        </TabPanel>
+        {/* <TabPanel className={styles.tabContainer}>
+          <ClaimsTab />
+        </TabPanel> */}
```
rdig, line 50: Same here

src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
```diff
@@ -24,22 +26,37 @@ const UserTokenActivationButton = ({ nativeToken, userLock }: Props) => {
...
+      inactiveTokens={inactiveBalance}
+      totalTokens={totalBalance}
+      lockedTokens={lockedBalance}
+      token={nativeToken}
+    >
+      {({ toggle, ref }) => (
+        <>
+          <button
+            type="button"
+            className={styles.tokens}
```
rdig, line 40: I know this was not added here, but while you're at it...

Hovering the tokens should show a `cursor: pointer;`

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+            text={MSG.activate}
+          />
+        </div>
+        <div className={isActivate ? styles.withdrawInactive : styles.withdraw}>
+          <Button
+            appearance={{ theme: !isActivate ? 'primary' : 'white' }}
+            onClick={() => setIsActivate(false)}
+            text={MSG.withdraw}
+          />
+        </div>
```
rdig, line 101: Colors for the "inactive" button are way **stronger** than the spec.

They need to be a softer shade of grey _(unless this was changed with product after the fact?)_

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+const ChangeTokenStateForm = ({
+  token,
+  tokenDecimals,
+  activeTokens,
+  inactiveTokens,
+}: ChangeTokenStateFormProps) => {
+  const [isActivate, setIsActivate] = useState(true);
+
+  const formAction = useCallback(
+    (actionType: '' | '_ERROR' | '_SUCCESS') =>
```
rdig, line 61: Might not need to use the underscore `_` in the name, but you call :)

However please don't use a empty string as a value or a type. Give it some name, doesn't really matter is it's only being used locally.

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  const transform = useCallback(
+    pipe(
+      mapPayload(({ amount }) => {
+        // Convert amount string with decimals to BigInt (eth to wei)
+        const formtattedAmount = bigNumberify(
+          moveDecimal(amount, tokenDecimals),
+        );
+
+        return { amount: formtattedAmount };
```
rdig, line 76: don't you need to send (at least) the colony address as well ?

Without it, the redux transaction channel doesn't know for what colony to instantiate the client for.

Now that I think about it more, should this also have the user address? Since the transaction will be made for that user.

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
```diff
@@ -0,0 +1,162 @@
...
+  return (
+    <div className={styles.changeTokensState}>
+      <div className={styles.changeStateTitle}>
+        <FormattedMessage {...MSG.changeState} />
+      </div>
+      <div className={styles.changeStateButtonsContainer}>
+        <div className={isActivate ? styles.activate : styles.activateInactive}>
+          <Button
+            appearance={{ theme: isActivate ? 'primary' : 'white' }}
+            onClick={() => setIsActivate(true)}
```
rdig, line 91: Maybe?
```suggestion
            onClick={() => setIsActive(true)}
```

src/modules/users/components/TokenActivation/TokenActivationContent/TokensTab.tsx
```diff
@@ -0,0 +1,149 @@
...
+  }).split(' ')[0];
+
+  return (
+    <>
+      <div className={styles.totalTokensContainer}>
+        <TokenIcon token={token || {}} size="xs" />
+        <p className={styles.totalTokens}>
+          {formattedTotalAmount} <span>{token?.symbol}</span>
+        </p>
+      </div>
```
rdig, line 85: Large amounts break the design.

Screenshot from 2021-03-01 17-43-29

I think that if you change the popover to have a min-width, it will resize accordingly

rdig: (note that the 1px vertical line is actually the token icon being squished)

src/modules/users/components/TokenActivation/TokenActivationContent/TokensTab.tsx
```diff
@@ -0,0 +1,149 @@
...
+    id:
+      // eslint-disable-next-line max-len
+      'users.TokenActivation.TokenActivationContent.TokensTab.inactiveTokensTooltip',
+    defaultMessage: `Inactive tokens are contained in your own wallet.
+     You need to “Activate” them to stake, or be eligible to receive Rewards.`,
+  },
+  lockedTokensTooltip: {
+    id:
+      // eslint-disable-next-line max-len
+      'users.TokenActivation.TokenActivationContent.TokensTab.lockedTokensTooltip',
```
rdig, line 45: Same here about the backticks

src/utils/numbers/index.ts
```diff
@@ -5,3 +5,5 @@ import { BigNumberish, bigNumberify } from 'ethers/utils';
  */
 export const bnLessThan = (a: BigNumberish, b: BigNumberish) =>
   bigNumberify(a).lt(bigNumberify(b));
+
+export { formatTokenValue } from './formatTokenValue';
```
rdig, line 9: I would have put this at the beginning of the file (after the imports)

rdig (changes requested):
This is mostly good functionally speaking, however some more UI polish is required.

Also, did we already built states for locked active tokens and staked tokens?

Ie:
Screenshot from 2021-03-01 17-48-23

Or are we implementing that only when we get to claims ?

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
chinins: good catch! :)

src/modules/core/components/Button/Button.css
chinins: Yeah, I am sure. You can actually check if you change 'white' for 'ghost' in the file. The differences are:
- transparent background vs. white background
- font, border color
- hover background

src/modules/core/components/Numeral/Numeral.tsx
chinins: For the total tokens in  the Figma specs token amount & token symbol are of different font-sizes. I have no way to separate them with using Numeral component as it uses a span element.
So I created a helper that just extracts the logic for the amount & symbol from the Numeral. It's the same logic, just encapsulated in a helper. I then use this helper in my `TokensTab` component & can split the amount & symbol & style them differently.

I've already checked if it's working fine but can do some more testing.

Screenshot 2021-03-02 at 11 37 10

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
chinins: As discussed yesterday on a call, it's to be uncommented when the claims tab is added

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
chinins: As discussed yesterday on a call, it's to be uncommented when the claims tab is added

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
chinins: As discussed yesterday on a call, it's to be uncommented when the claims tab is added

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
chinins: yes, I was missing opacity!

src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
chinins: I was going more for activate/withdraw distinction, but I do agree that isActive is better

ArmandoGraterol (approved):
Looks good to me :fire:

ArmandoGraterol (approved):
Still looking good! The changes seem to be working fine

chinins:
This PR is ready for the re-review.

The max button on the input component gave me a bit of trouble because of `Cleave`. The `handleCleaveChange` code seems a bit clunky but this is the only way that I found that works in all cases and doesn't break other forms.

rdig:
@olga can you please rebase this? And also it's base branch so that I can properly review this

Thank you

src/modules/users/components/TokenActivation/TokenActivationContent/TokensTab.tsx
```diff
@@ -0,0 +1,213 @@
...
+              <FormattedMessage {...MSG.staked} />
+            </TokenTooltip>
+            <div className={styles.tokenNumbersLocked}>
+              <Numeral
+                value={lockedTokens}
+                suffix={` ${token?.symbol}`}
+                unit={tokenDecimals}
+                truncate={3}
+              />
+            </div>
```
rdig, line 165: This should be a separate list entry, why is it bundled with the "active" list entry ?

rdig, line 165: Personally, I would only show the "staked" section if we have staked tokens, otherwise it's kinda pointeless.

Can you check with product if that is the intended functionality, or that we are supposed to show it at all times.

If so, please make the change, otherwise leave it as it is.

rdig (approved):
I'm good with this, I just have a question about the "staked" section (see below)

Screenshot from 2021-03-12 11-18-19

src/modules/users/components/TokenActivation/TokenActivationContent/TokensTab.tsx
chinins: Checked with Karol and he confirmed that this should be shown even when 0.

chinins: It was under the active in my mind, but you are right. I've corrected it.

PR #2386, opened 2021-02-18
Fix ERC20 unclaimed transfers to show up properly

rdig:
## Description

This PR fixes a urgent bug found in production where a colony could not claim their non-native token transfers.

This was due to the fact that our tokens differ from ERC20 ones in values named in the abi (check my comment), so that means that for ERC20 it could not fetch the correct values

**Changes** 

- [x] transfer resolvers handle ERC20 token values
- [x] `UnclaimedTransfer` add loading state

Resolves DEV-218


chinins (approved):
Seems good to me!

PR #2387, opened 2021-02-18
Fix FriendlyName not using the colony display

ArmandoGraterol:
Fixed FriendlyName not using the colony display when a user is not provided

Resolves DEV-219


rdig (approved):
Good to go.

But please pay more attention in the future, as I really don't like merging broken stuff into production.

PR #2388, opened 2021-02-21
Remove what's left of the admin module

rdig:
This PR removes what was left of the `admin` module as it's not being used anymore.

Also, while I was at it, I removed the leftover edit domain redux sagas that we forgot to remove when implementing the edit domain action.

Resolves DEV-33
Resolves DEV-37


alicjakujawa (approved):

chinins (approved):
Looks good

ArmandoGraterol (approved):

PR #2390, opened 2021-02-22
Wire up Token Locking feature actions

ArmandoGraterol:
Creates token locking feature related actions, and wires up deposit and withdraw action in the token locking popover

Resolves DEV-214


rdig:
What's the status of this?

Is it done or are you still working on it ?

ArmandoGraterol:
> What's the status of this?
> 
> Is it done or are you still working on it ?

It's mostly done. By that, I mean that I could put it on review because it's working as it is, but I would prefer for the token popover(something Olga is working on) to be merged so that you guys can test this easily and with the complete flow

rdig:
Ok, we'll wait for that to merge then

chinins (approved):
All good! 🔥

rdig:
@ArmandoGraterol can you please rebase this before I start reviewing it ?

ArmandoGraterol:
> @ArmandoGraterol can you please rebase this before I start reviewing it ?

Done!

src/i18n/en.json
```diff
@@ -52,6 +52,14 @@
...
     "transaction.debug.description": "DEBUG: context: {context} methodName: {methodName}",
 
+    "transaction.group.deposit.title": "Activate tokens",
+    "transaction.group.deposit.description": "Activate tokens",
+    "transaction.TokenLockingClient.deposit.title": "Activate tokens",
+    "transaction.group.withdraw.title": "Withdraw tokens",
+    "transaction.group.withdraw.description": "Withdraw tokens",
+    "transaction.TokenLockingClient.wthdraw.description": "Withdraw tokens",
+    "transaction.TokenLockingClient.withdraw.title": "Withdraw tokens",
+    "transaction.TokenClient.approve.title": "Approve amount of tokens",
```
rdig, line 62: As above, this needs a description message descriptor counterpart

```suggestion
    "transaction.TokenClient.approve.title": "Approve amount of tokens",
    "transaction.TokenClient.approve.description": "Approve amount of tokens",
```

src/modules/users/sagas/user.ts
```diff
@@ -209,10 +219,157 @@ function* userLogout() {
...
+  meta,
+  payload: { tokenAddress, amount, colonyAddress },
+}: Action<ActionTypes.USER_DEPOSIT_TOKEN>) {
+  const txChannel = yield call(getTxChannel, meta.id);
+  try {
+    const apolloClient = TEMP_getContext(ContextModule.ApolloClient);
+    const colonyManager = TEMP_getContext(ContextModule.ColonyManager);
+    const { walletAddress } = yield getLoggedInUser();
+
+    // eslint-disable-next-line max-len
```
rdig, line 232: Please leave a note when it's not obvious why you disabled the `max-len` rule.

_(I'm assuming, in this case, it's because `prettier` doesn't suggest a good solution)_

src/modules/users/sagas/user.ts
```diff
@@ -209,10 +219,157 @@ function* userLogout() {
...
+      ready: false,
+    });
+
+    yield createGroupTransaction(deposit, {
+      context: ClientType.TokenLockingClient,
+      methodName: 'deposit',
+      identifier: colonyAddress,
+      params: [tokenAddress, amount, false],
+      ready: false,
+    });
```
rdig, line 269: The amount value needs to be converted to a big number.

Otherwise, for values with more than 15 zeros. this will give an error.

So something like:
```
params: [tokenAddress, new BigNumber(amount), false],
```

This applies to both approve and deposit

src/modules/users/sagas/user.ts
```diff
@@ -209,10 +219,157 @@ function* userLogout() {
...
+    const apolloClient = TEMP_getContext(ContextModule.ApolloClient);
+    const { walletAddress } = yield getLoggedInUser();
+
+    const { withdraw } = yield createTransactionChannels(meta.id, ['withdraw']);
+
+    yield fork(createTransaction, withdraw.id, {
+      context: ClientType.TokenLockingClient,
+      methodName: 'withdraw',
+      identifier: colonyAddress,
+      params: [tokenAddress, amount, false],
```
rdig, line 323: Same as above, about the amount value needing to be converted into a big number

rdig (changes requested):
This is mostly good and it works almost as expected.

I've left a few notes about minor changes that need to be done.

But the biggest change (the one for which I've rejected this) is that the token amount form needs to be cleared after the transactions are succesfull.

Here's a demo:

bug-activate-tokens-form-not-cleared

The field (that has the max button on it) needs to be reset after the transaction fires the success action.

Yes, I know this is not something you introduced in this PR, but since this is the last one for this feature, might as well do it here.

Other than that, this is some very nice work.

src/modules/users/sagas/user.ts
ArmandoGraterol: That's correct, I'll leave a comment then

rdig (approved):
Very nice work! This is good to go!

PR #2393, opened 2021-02-25
Remove task feature

alicjakujawa:
## Description

This PR  removing Task feature which is not used anymore
 
**Changes** 🏗

* Updated cacheUpdates file
* Updated InboxItem component
* Moved tasks array to utils/tasks

**Deletions** ⚰️

* ItemsList component
* Task related mutations
* TaskSkills component
* Task related fragments queries and type defs
* Task related checks 
* Tasks from immutablejs
* Task resolver
* Translations
* Assignment component
* Filters
* Task saga
* Redux types
* Task action types
* Payout fragment

Resolves DEV-23


src/modules/core/components/PayoutsList/PayoutsList.tsx
```diff
@@ -28,7 +27,7 @@ interface Props {
   maxLines?: number;
 
   /** Payouts list containing all the payouts */
-  payouts: Payouts;
+  payouts: any[];
```
chinins, line 30: Why is it switched to any?

src/modules/users/components/Inbox/InboxItem/InboxItem.tsx
```diff
@@ -203,7 +198,7 @@ const InboxItem = ({
                     <Numeral
                       suffix={` ${token ? token.symbol : ''}`}
                       integerSeparator=""
-                      value={value}
+                      value={value || '-'}
```
chinins, line 201: Why do we need to pass '-' and not just empty string? It seems that they both work but empty string is more of a convention in our code, unless there is a reason?

src/utils/tasks/taskSkillsTree.goerli.ts
```diff
@@ -2,9 +2,7 @@
  * Skills Tree For the Goerli Network
  */
 
-import { ConsumableItem } from '~core/ItemsList';
-
-const taskSkills: ConsumableItem[] = [
+const taskSkills = [
```
chinins, line 5: Are these task skills still used anywhere or will be used in the future? I did a simple variable search of `taskSkill` & `getTaskSkillsTreeInUse` and didn't see any except the declaration ones. But maybe I am missing something?

chinins (commented):
The app is working correctly and all the changes seem legit. I just have a couple questions (please, see my code comments)

src/modules/core/components/PayoutsList/PayoutsList.tsx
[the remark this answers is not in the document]
alicjakujawa: Because i removed type  Payouts, as it was used by Task feature. But component stays for now (because maybe will be use in future 🤦‍♀️ . Im actually thinking of moving this type back

src/modules/users/components/Inbox/InboxItem/InboxItem.tsx
alicjakujawa: The reason is, i checked how we was using it everywhere, and we was indeed putting '-' as fallback

src/utils/tasks/taskSkillsTree.goerli.ts
alicjakujawa: Not for now, but in task requirements was written to keep skills for future :)

chinins (approved):
Thank you for the answers! Maybe just keep the Payout then but other than that - good for me.

ArmandoGraterol (approved):
Nice work :+1:

src/data/graphql/fragments.graphql
```diff
@@ -104,7 +62,7 @@ fragment EventFields on Event {
     profile {
       avatarHash
       displayName
-      username
+     username
```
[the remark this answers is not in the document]
rdig: Yeah, I think this is a mistake. Please put the space back in :P

src/modules/core/components/ItemsList/ItemsList.tsx
```diff
@@ -1,317 +0,0 @@
...
-  /** Callback function, called after the field value has been set */
-  onChange?: (value: number | null | undefined) => void;
-
-  /** Wheather or not to show the Popover's arrow */
-  showArrow?: boolean;
-}
-
-const displayName = 'ItemsList';
-
-const ItemsList = ({
```
rdig, line 70: While this was used strictly on the task, it's built for being generalized.

I has a very nice piece of logic that allows you to have nested items in the list.

Please put this back in, as we still want to keep it.

src/modules/core/components/PayoutsList/PayoutsList.tsx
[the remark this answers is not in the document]
rdig: No, don't. Leave it as `any`.

We can add proper types for this, once we actually need to use it.

src/utils/tasks/taskSkillsTree.goerli.ts
```diff
@@ -2,9 +2,7 @@
...
   {
```
rdig, line 6: You can remove the task skills lists for now. We can bring them back when we'll need them.

rdig (changes requested):
This is all good, except:
- please add back the core `ItemsList` component
- you can remove the task skills lists for all the networks.

Nice cleanup!

src/utils/tasks/taskSkillsTree.goerli.ts
rdig: We won't. When I first wrote that I had other plans :)

Those skills are generated from the contracts, so they don't really matter.

PR #2394, opened 2021-02-26
Fixed UserPermissions headRole conditional

ArmandoGraterol:
Fixed Recovery role not showing because the previous conditional checked if the value of the headRole is a falsey value (like `0`, which is the role id of Recovery). Now it only checks if it's `null` or `undefined`

FireShot Capture 155 - Colony - localhost

Resolves DEV-226

### DIFF
--- a/src/modules/dashboard/components/UserPermissions/UserPermissions.tsx
+++ b/src/modules/dashboard/components/UserPermissions/UserPermissions.tsx
@@ -1,5 +1,7 @@
 import React from 'react';
 import { ColonyRole } from '@colony/colony-js';
+import { isNil } from 'lodash';
+
 import PermissionsLabel from '~core/PermissionsLabel';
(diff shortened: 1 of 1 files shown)


chinins (approved):
Works as it should!

src/modules/dashboard/components/UserPermissions/UserPermissions.tsx
```diff
@@ -34,7 +36,7 @@ const UserPermissions = ({ roles, directRoles, appearance }: Props) => {
   const [headRole, ...restRoles] = sortedRoles;
   return (
     <div className={getMainClasses(appearance, styles)}>
-      {headRole && (
+      {!isNil(headRole) && (
```
rdig, line 39: I don't really like us relying on `lodash`, so please try to avoid it as much as possible

I'm fine with this for now and you can leave it in, just try to avoid it in the future.

rdig (approved):
Good to go

PR #2395, opened 2021-02-26
Added loading message to colony extension section

ArmandoGraterol:
* Fixed loading state of colony extension section defaulting to the old spinner and hiding the section title.
* Added loading message to loading state of the colony extension section
* Fixed section title colony members having a different color when on loading state

FireShot Capture 158 - Colony - localhost

Resolves DEV-222

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.css
+++ b/src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.css
@@ -24,3 +24,11 @@
 .invisibleLink:hover {
   color: color-mod(var(--dark) alpha(50%));
 }
+
+.loadingText {
+  display: inline-block;
+  margin: -8px 0 0 5px;
+  vertical-align: middle;
+  font-size: var(--size-small);
+  letter-spacing: initial;
+}
--- a/src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
@@ -46,41 +46,37 @@ const ColonyExtensions = ({ colony: { colonyName, colonyAddress } }: Props) => {
     );
   }
 
-  return (
+  return data?.processedColony ? (
     <div className={styles.main}>
-      {data?.processedColony && (
-        <>
-          <Heading appearance={{ size: 'normal', weight: 'bold' }}>
-            <FormattedMessage {...MSG.title} />
-          </Heading>
-          <ul>
-            {data.processedColony.installedExtensions
-              .filter(
-                (extension) =>
-                  extension.details.initialized &&
-                  !extension.details.missingPermissions.length,
-              )
-              .map((extension) => {
-                const { address, extensionId } = extension;
-                return (
-                  <li key={address} className={styles.extension}>
-                    <NavLink
-                      className={styles.invisibleLink}
-                      to={`/colony/${colonyName}/extensions/${extensionId}`}
-                      text={extensionData[extensionId].name}
-                    />
-                    <ExtensionStatus
-                      installedExtension={extension}
-                      deprecatedOnly
-                    />
-                  </li>
-                );
-              })}
-          </ul>
-        </>
-      )}
+      <Heading appearance={{ size: 'normal', weight: 'bold' }}>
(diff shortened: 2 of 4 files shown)


src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
```diff
@@ -15,6 +15,10 @@ const MSG = defineMessages({
     id: 'dashboard.ColonyHome.ColonyExtensions.title',
     defaultMessage: 'Enabled extensions',
   },
+  loadingData: {
+    id: 'dashboard.ColonyHome.ColonyMembers.loadingData',
+    defaultMessage: 'Loading enabled extensions information...',
```
chinins, line 20: I don't want to be picky but it just seems a bit too much for loading info. I saw that you just replicated what members loading shows but I would drop information from both. It's not that important.

src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.css
```diff
@@ -24,3 +24,11 @@
 .invisibleLink:hover {
   color: color-mod(var(--dark) alpha(50%));
 }
+
+.loadingText {
+  display: inline-block;
+  margin: -8px 0 0 5px;
```
chinins, line 30: why -8px here? I think it looks better without these margins. The text is way to close to the spinner. Or is it the specs?

chinins (commented):
I just have one question about the text margins (please, see me code comment). I think it needs a bit of space between the spinner & text.

src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
```diff
@@ -63,7 +63,12 @@ const ColonyExtensions = ({ colony: { colonyName, colonyAddress } }: Props) => {
             })}
         </ul>
       ) : (
-        <SpinnerLoader />
+        <>
+          <SpinnerLoader />
+          <span className={styles.loadingText}>
+            <FormattedMessage {...MSG.loadingData} />
+          </span>
+        </>
```
rdig, line 71: This is not correct.

Yes I know you just expanded on previous functionality, but now is a chance to fix it.

So this is not correct because if it doesn't find any extensions (and it might happen for non-fully deployed colonies), you will always show a spinner. This is not correct. We only show a spinner while actually loading, but if we don't find data, than we hide the spinner, and just don't show anything else.

Please change it to something like this:
```
if (loading) {
  return <Spinner ... >
}

return data && <Component ... >
```

Note that if we don't have any data, **don't** show the title at all, hide it.

src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
rdig: Agree, this should just be `Loading extensions data ...`

src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.css
rdig: No. Armando is actually right here.

The `-8px` correctly aligns the text with the spinner

rdig (changes requested):
So your changes are good and what I have expected from this issue.

I just want you to "fix" the logic of the spinner / data showing in this widget. 

(It's not something you added, it wasn't implemented properly the first time around)

src/modules/dashboard/components/ColonyHome/ColonyExtensions/ColonyExtensions.tsx
```diff
@@ -29,41 +33,51 @@ const ColonyExtensions = ({ colony: { colonyName, colonyAddress } }: Props) => {
...
-                    installedExtension={extension}
-                    deprecatedOnly
-                  />
-                </li>
-              );
-            })}
-        </ul>
-      ) : (
-        <SpinnerLoader />
+      {data?.processedColony && (
```
rdig, line 51: Please include the `main` div as well in your logic check.

So something like;

`return data?.processedColony ? <div className="main">...</div> : null`

rdig (approved):
All good here, just include the main div in the display logic check

PR #2396, opened 2021-02-26
Add Tooltip to colony event list titles

ArmandoGraterol:
This should stop the tooltip of the titles in the event list to display `[object object]` whenever a dynamic value is used

Resolves DEV-193

### DIFF
--- a/src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.css
+++ b/src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.css
@@ -108,3 +108,8 @@
 .titleDecoration {
   color: var(--pink);
 }
+
+.tooltip span {
+  vertical-align: text-top;
+  font-size: var(--size-tiny);
+}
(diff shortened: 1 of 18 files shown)


chinins (approved):
All good!

src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.css
```diff
@@ -108,3 +108,8 @@
 .titleDecoration {
   color: var(--pink);
 }
+
+.tooltip span {
+  vertical-align: text-top;
+  font-size: var(--size-tiny);
+}
```
rdig, line 115: Just to keep it in line with the other tooltip styles (meaning they being bigger) please add:
```
  padding: 8px 4px;
```

To the styles

rdig: If you are really feeling adventurous :) you can add this as a "theme" for the tooltip

rdig (approved):
Very nice!!!! Big kudos for this.

I didn't tough of this solution.I was thinking of something waaaay more complicated. I like the route you went with.

PR #2397, opened 2021-03-01
Fix Token Management Modal Back Button

chinins:
## Description

Due to some oversight, the back button for the Token Management modal is actually a "Cancel" button, and breaks out of the wizard flow.

I think the reason is that we have the same dialog open from the funding page and there is no back there, only cancel. So I just did back button depending on the passed props.

**New stuff** ✨

none

**Changes** 🏗

- add option back prop to the dialog & invoke it if it's passed

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-229

token-mngt-dialog2

token-mngt-dialog

### DIFF
--- a/src/modules/core/components/TokenEditDialog/TokenEditDialog.tsx
+++ b/src/modules/core/components/TokenEditDialog/TokenEditDialog.tsx
@@ -68,6 +68,7 @@ interface Props {
   // Token list from json file. Not supported on local env
   tokensList?: AnyToken[];
   colony: Colony;
+  back?: () => void;
 }
 
 interface FormValues {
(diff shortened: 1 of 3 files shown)


rdig (approved):
Good to go

ArmandoGraterol (approved):
Working as expected :ok_hand:

rdig:
Hey @chinins 

I've just discovered that we have the same problem with the permissions modal.

Screenshot from 2021-03-01 21-59-08

Can you please change it for this one too ?

Please note that we call this modal from two places:
- colony home actions button
- members page _(`/members` route)_

PR #2400, opened 2021-03-04
Add themes and threshold to ProgressBar

ArmandoGraterol:
* Added `appearance` prop with `theme` and `size` to ProgressBar
* Added threshold to ProgressBar

FireShot Capture 163 - Colony - localhost

To test it:

1. Copy and paste this whenever you want(ProgressBar needs to be imported of course): 
`<ProgressBar value={50} appearance={{ theme: 'dark', size: 'small' }} threshold={70} />`.
2. ?????
3. Profit

Resolves DEV-238


alicjakujawa (approved):
Almost there 🔥 
Small nitpicks:
1. Add border-radius: 1px to threshold
2. Change bg color to 254,94,124,1

Pre-approving this

rdig (changes requested):
This is all good, except for the threshold line not having the proper size when `size` is set to `normal`

src/modules/core/components/ProgressBar/ProgressBar.tsx
```diff
@@ -1,8 +1,15 @@
 import React from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 
+import { getMainClasses } from '~utils/css';
+
 import styles from './ProgressBar.css';
 
+interface Appearance {
+  theme?: 'default' | 'dark';
```
rdig, line 9: So the 'default' theme, when in 'normal' size, has a background wrapper:

Screenshot from 2021-03-12 11-41-19

While, in the 'small' size, it does not

Screenshot from 2021-03-12 11-41-14

It think we should remove it for the 'normal' size as well to keep it consistent (an just have the grey "fill" portion of the bar)

rdig (approved):
I'm good with this now, however, please address the comment that I left before about the default small vs. normal styles.

Screenshot from 2021-03-12 11-41-24

src/modules/core/components/ProgressBar/ProgressBar.tsx
ArmandoGraterol: Sounds good to me! It also looks better without the wrapper

PR #2402, opened 2021-03-05
Create Slider component

alicjakujawa:
## Description

This PR adds in core slider component

**New stuff** ✨

* Slider core component

**Changes** 🏗

Primary theme of slider with limit
Screenshot 2021-03-08 at 10 35 23
Danger theme with limit
Screenshot 2021-03-08 at 10 36 41
Danger theme without limit
Screenshot 2021-03-08 at 10 38 12

Resolves DEV-234


ArmandoGraterol (approved):
Looks good to me! I just noticed something small and it's that the circle of the `danger` theme is not as center-aligned as the one in the `primary` theme, prob because the height of the slider is different?

Perhaps I'm being nitpicky

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,107 @@
...
+  theme?: 'primary' | 'danger';
+};
+
+interface Props {
+  value: number;
+  max?: number;
+  limit?: number;
+  appearance?: Appearance;
+  onChange: (val: any) => void;
+}
```
rdig, line 15: This component needs to be a form element, without that it won't work for our use case.

For that we need to have a `name` for it.

Note that this is a required prop.

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,107 @@
...
+      markPositionTop: -2,
+    },
+    danger: {
+      backgroundColor: '#FE5E7C',
+      borderColor: '#FE5E7C',
+      height: 3,
+      markHeight: 8,
+      markWidth: 2,
+      markPositionTop: -3,
+    },
```
rdig, line 62: There's a difference in styles between the two themes of the slider. The danger theme is "thicker" than the primary one:

Screenshot from 2021-03-09 14-06-33

Personally I like how the danger one looks, so I would say to make the primary one the same thinckness at the danger theme.

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,107 @@
...
+  limit?: number;
+  appearance?: Appearance;
+  onChange: (val: any) => void;
+}
+
+const displayName = 'Slider';
+
+const Slider = ({ value, max = 100, onChange, limit, appearance }: Props) => {
+  const [sliderValue, setSliderValue] = useState<number>(value);
+
```
rdig, line 21: It's pretty straight forward to set up the field, you just need to bring in the `useField` component   from formik

```
  const [, , { setValue }] = useField(name);
```

Note that since we are using a slider, we don't actually have to handle any errors :)

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,107 @@
...
+        setSliderValue(val);
+        onChange(val);
+      }
+      if (limit && sliderValue > limit) {
+        setSliderValue(limit);
+        onChange(limit);
+      }
+    },
+    [setSliderValue, onChange, limit, sliderValue],
+  );
```
rdig, line 38: Using the `setValue` method imported from formik, you'll need to call it every time the value changes:

```suggestion
  const onSliderChange = useCallback(
    (val): void => {
      if ((limit && sliderValue < limit) || val < sliderValue || !limit) {
        setSliderValue(val);
        setValue(val);
        onChange(val);
      }
      if (limit && sliderValue > limit) {
        setSliderValue(limit);
        setValue(limit);
        onChange(limit);
      }
    },
    [setSliderValue, onChange, limit, sliderValue],
  );
```

And that's it, now this will function as form element.

If you really want to go a step further, you can also add a `isConnected` prop (that defaults to `true`) and use that only make this a connected form element if that is set. 

But that is optional, as most likely we'll only use this in a form configuration.

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -0,0 +1,107 @@
...
+        border: 0,
+        borderRadius: 0,
+        top: styles.markPositionTop,
+        marginLeft: 0,
+      }}
+      railStyle={{
+        backgroundColor: '#C2CCCC',
+        height: styles.height,
+        backgroundImage: `linear-gradient(90deg, #76748B 0% ${gradientStopPercentage}%, transparent ${gradientStopPercentage}%)`,
+      }}
```
rdig, line 100: I think these styles can be set in css modules, and just imported.

(I think, I'm not sure, and haven't tested if css modules knows to convery `background-image` to `backgroundImage`)

rdig (changes requested):
This is mostly good, however it still needs work in order to make it work as form connected component.

src/modules/core/components/Slider/Slider.tsx
alicjakujawa: haha i actually more like primary more :D But if you wont make them same thickness ask Karol if he is ok with that

src/modules/core/components/Slider/Slider.tsx
alicjakujawa: And what is wrong with current implementation?

rdig (approved):
Please add a `sizing` prop to `Appearance`

PR #2408, opened 2021-03-09
Bug/dev 230 token activ

alicjakujawa:
## Description

This PR fixes buttons alignment as well as updating balance after payment action 

**Changes** 🏗

* Updated buttons alignment
* Added refetch query to payment saga

Resolves DEV-230


rdig (approved):
Good to go

PR #2409, opened 2021-03-10
Create radio buttons structure

alicjakujawa:
## Description

This PR adds in core radio buttons

**New stuff** ✨

* RadioButtons component
* RadioButton component

Default state
default

True selected
true-selected

False selected
false-selected

Fully staked
fully-staked

Resolves DEV-237


src/modules/core/components/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { SimpleMessageValues } from '~types/index';
+
+import styles from './RadioButton.css';
+
+export interface Appearance {
+  theme?: 'primary' | 'danger';
+  direction?: 'horizontal' | 'vertical';
+}
+
+export interface RadioButtonTypes {
```
chinins, line 17: why `Types` & not `Props`? I thought props was a default.

ArmandoGraterol (approved):
Looking good to me :+1:

src/modules/core/components/RadioButton/RadioButton.tsx
```diff
@@ -0,0 +1,102 @@
...
+  /** Description text values for intl interpolation */
+  labelValues?: SimpleMessageValues;
+  /** Button description */
+  description?: string | MessageDescriptor;
+  /** Description text values for intl interpolation */
+  descriptionValues?: SimpleMessageValues;
+  /** Button icon */
+  icon?: string;
+  /** If the input is checked */
+  checked: boolean;
```
chinins, line 35: I think that `checked` prop should be excluded or made optional when you pass the `RadioButtonTypes` interface as options in `RadioButton`, otherwise it will give a type error when you instantiate it. You are not using it in `RadioButton` as prop but create a value instead.

chinins (commented):
All good, just a couple of types questions/remarks.

src/modules/core/components/RadioButton/RadioButton.tsx
alicjakujawa: Default of what? This is just name of interface

rdig (changes requested):
I'll good with most of this, however it still needs some changes:
- needs to be moved in `core/components/Fields`
- needs to be renamed (we already have `Radio`)
- description shouldn't be visible when `direction` is `vertical`

Other than that, it seems to function as we need it to:

Screenshot from 2021-03-12 11-00-33

rdig (approved):
All good now! :+1:

PR #2411, opened 2021-03-11
Add MemberReputation component

alicjakujawa:
## Description

This PR adds in MemberReputation core component

**New stuff** ✨

* MemberReputation component
* Displaying reputation in UserNavigation

**Changes** 🏗

* Updated MembersListItem to use new MembersReputation component
* Updated Members component to not fetch total reputation [handled by member component itself]
* Members reputation would always have star before number [important for consistency, approve by product] 

Screenshot 2021-03-15 at 08 52 07
Screenshot 2021-03-15 at 08 47 31

Resolves DEV-232


chinins (changes requested):
All looks good except for 2 UI points:

- when the reputation is undefined the UI is a bit broken:
Screenshot 2021-03-15 at 18 33 16

- i see the star in figma specs smaller than the font and it's currently larger. Maybe there are updated specs somewhere but I didn't manage to find them - can you please clarify this point.
Screenshot 2021-03-15 at 18 29 06

src/modules/core/components/MemberReputation/MemberReputation.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { Address } from '~types/index';
+import Numeral from '~core/Numeral';
+import Icon from '~core/Icon';
+
+import styles from './MemberReputation.css';
+
+const MSG = defineMessages({
+  starReputationTitle: {
+    id: 'MemberReputation.starReputationTitle',
+    defaultMessage: `User reputation value: {reputation}`,
```
rdig, line 18:
```suggestion
    defaultMessage: `User reputation value: {reputation}%`,
```

src/modules/pages/components/RouteLayouts/UserNavigation.css
```diff
@@ -142,3 +142,20 @@
...
+  height: 26px;
+  border-radius: var(--radius-normal);
+  background-color: color-mod(var(--temp-grey-blue-7) alpha(10%));
+}
+
+.reputation span {
+  display: inline-block;
+  vertical-align: middle;
+  font-size: 9px; /* it seems we start useing this font regulary, time for var? */
+}
```
rdig, line 161: So my only complaint about this is here the size of the text. 

It's too small:

Screenshot from 2021-03-15 21-41-07

I know this is not your fault, as the specs show it like that:

Screenshot from 2021-03-15 21-46-49

I think that the specs are wrong. I think the font should be a bigger size and the side padding should be bigger. 

If you really want to, you can ask Karol, but these change should be done for consistency, otherwise the user navigation will look bad.

Here's how I changed them locally:

Screenshot from 2021-03-15 21-50-19

Increased the font size to 13px and removed the vertical-align _(it aligns perfectly with a font of 13px)_, and increased the side padding to 7px

```js
.reputation {
  ...
  padding: 0 7px;
  ...
}

.reputation span {
  display: inline-block;
  font-size: var(--size-smallish);
}
```

rdig (approved):
I'm all good with this, just that I think the user navigation specs are wrong!

I've left a comment explaining why and how to change them

I've also pre-approved this, but only merge them after you update the user navigation reputation.

alicjakujawa:
> All looks good except for 2 UI points:
> 
> * when the reputation is undefined the UI is a bit broken:
> 
> Screenshot 2021-03-15 at 18 33 16
> 
> * i see the star in figma specs smaller than the font and it's currently larger. Maybe there are updated specs somewhere but I didn't manage to find them - can you please clarify this point.
> 
> Screenshot 2021-03-15 at 18 29 06

Hmm designs says font-size: 9px and start ~10x10 and i applied those styles, but i can agree that it looks different xD

src/modules/pages/components/RouteLayouts/UserNavigation.css
alicjakujawa: i love it, i will make this change, and i also think that is something wrong with specs, because it looks like things are similar size, but when you apply those styles it looks a lot different

ArmandoGraterol (approved):
This is looking quite good! Nice work :ok_hand:

chinins (approved):
Looks so much better with updated styles!

PR #2412, opened 2021-03-11
Create tag components variants

chinins:
## Description

Create variants for all tags used in Motion flow.

To test - add the code below wherever you want to see the tag and change the prop values:
`<Tag text="Text" appearance={{ theme: 'blue', colorSchema: 'plain', fontSize: 'tiny' }} />`

**New stuff** ✨

- added new themes to `Tag` component
- added `fontSize` tiny prop value and css
- add `colorSchema` plain & inverted

**Changes** 🏗

- updated golden color variable (also done in token activation popover but it's not yet merged into master)
- change padding left & right from `10px` to `8px` as agreed with Karol

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-235

### DIFF
--- a/src/modules/core/components/Tag/Tag.css
+++ b/src/modules/core/components/Tag/Tag.css
@@ -3,6 +3,7 @@
   border-radius: var(--radius-large);
   font-size: 0.8rem;
   font-weight: var(--weight-bold);
+  color: var(--colony-white);
 }
 
 .baseStyles + .baseStyles {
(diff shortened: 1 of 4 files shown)


alicjakujawa (approved):
Looks ok 👍

src/modules/core/components/Tag/Tag.tsx
```diff
@@ -6,7 +6,9 @@ import { useMainClasses } from '~utils/hooks';
 import styles from './Tag.css';
 
 interface Appearance {
-  theme: 'primary' | 'light' | 'golden' | 'danger';
+  theme: 'primary' | 'light' | 'golden' | 'danger' | 'pink' | 'blue';
+  fontSize?: 'tiny';
+  colorSchema?: 'inverted' | 'plain';
```
rdig, line 11: we should have a default color schema, for the "full color" version of the tag:

Screenshot from 2021-03-12 11-28-51

You can name it however you'd like, just that it needs to be there so that we can explicitly set it _(right now, the only way to "set it" it to leave the `colorSchema` prop empty)_

src/modules/core/components/Tag/Tag.css
```diff
@@ -1,8 +1,9 @@
 .baseStyles {
-  padding: 1px 10px 2px;
+  padding: 1px 8px 2px;
   border-radius: var(--radius-large);
   font-size: 0.8rem;
   font-weight: var(--weight-bold);
+  color: var(--colony-white);
 }
```
rdig, line 7: This is not something that you added, but something I've noticed:

Screenshot from 2021-03-12 11-28-12

Tags are broken up at the end of the line. I don't that this should be the case.

Please add either something like `display: inline-block` or `white-space: nowrap` to prevent that from happening.

rdig (approved):
I'm good with this, but please address the comments I left

rdig:
This can be merged, right ?

PR #2414, opened 2021-03-11
Add motion and objection displays

ArmandoGraterol:
Add motion and objection displays core component

Some examples to play with:

```
<ArgumentDisplay appearance={{ theme: 'motion' }} argument={{ author: '0xb77D57F4959eAfA0339424b83FcFaf9c15407461', comment: "Design team funds are drained. I would like to suggest we move some coins from Dev team.", timestamp: Date.now()}} />

<ArgumentDisplay appearance={{ theme: 'objection' }} argument={{ author: '0x9dF24e73f40b2a911Eb254A8825103723E13209C', comment: "I strongly object to this! We’re going to need a vote.", timestamp: Date.now()}} />
```
FireShot Capture 180 - Colony - localhost

Resolves DEV-236


rdig:
@ArmandoGraterol please name PRs better, even if they are in "Draft" mode _(this one has the name of your first commit)_

src/modules/core/components/ArgumentDisplay/ArgumentDisplay.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const displayName = 'ArgumentDisplay';
+
+const ArgumentDisplay = ({
+  appearance = { theme: 'motion' },
+  argument: { author, timestamp, comment },
+}: Props) => {
+  const user = useUser(author);
+  const tagTheme = appearance.theme === 'motion' ? 'primary' : 'pink';
+  const tagText = appearance.theme === 'motion' ? 'Motion' : 'Objection';
```
alicjakujawa, line 34: This is a bit too trivial (think of cases where we could have more themes)
We could for example create some MAP object which would return correct arguments for theme

src/modules/core/components/ArgumentDisplay/ArgumentDisplay.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const UserAvatar = HookedUserAvatar({ fetchUser: false });
+
+  return (
+    <div
+      className={`${styles.container} ${getMainClasses(appearance, styles)}`}
```
alicjakujawa, line 39: i think if you create .main class in styles file and apply container styles, then here you could just do: 
```suggestion
      className={getMainClasses(appearance, styles)}
```

alicjakujawa (approved):
2 suggestions, apart of this, everything looks good to me :)

src/modules/core/components/ArgumentDisplay/ArgumentDisplay.tsx
ArmandoGraterol: Not working, from what I saw inside `getMainClasses` I think it only works like that when no `theme` is provided

src/modules/core/components/ArgumentDisplay/ArgumentDisplay.tsx
ArmandoGraterol: Decided to make it simple because we would only have 2 themes for now. But yeah I could make a more dynamic way of choosing the themes for the Tag

chinins: I agree with Alicja. I would suggest to have the same theme names as in the `Tag` and just pass it directly.
Also I would suggest to pass a prop with the tag text (by the way it should be done with react intl I think).
This way you decouple appearance and contents and it will give you more flexibility for potential future components.

chinins (commented):
All good except for theme and props as per my commnet

chinins (approved):

src/modules/core/components/ArgumentDisplay/ArgumentDisplay.tsx
ArmandoGraterol: I like that idea! I'll implement it

PR #2415, opened 2021-03-13
Create `Toggle` form core component

chinins:
## Description

Create a core form `Toggle` component that will be used for motions & disputes feature.

style guide link: https://www.figma.com/file/DjVCPq7LpkjEMEfQZtKgBdC1/Style-Guide?node-id=2475%3A1828

To test, add anywhere where you already have a form:
```
 <Toggle label="Label" name="toggle" />
```

**New stuff** ✨

- add `Toggle` core component

**Changes** 🏗

- none

**Deletions** ⚰️

- none

## TODO

- [ ] Add TODOs

Resolves DEV-233


src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+    modifiers: [
+      {
+        name: 'offset',
+        options: {
+          offset: [-5, 7],
+        },
+      },
+    ],
+  },
+};
```
rdig, line 88: I mean, I can't remember the last time we set `defaultProps` on a react component since it's much easier to just initialize props with what values you need.

Eg:
```jsx
const Toggle = ({
  name,
  label,
  disabled: false,
  elementOnly: false,
  tooltipText,
  tooltipPopperProps = {
    placement: 'top-end',
    modifiers: [
      {
        name: 'offset',
        options: {
          offset: [-5, 7],
        },
      },
    ],
  },
}: Props) => {
  ...
}
```

But in the end it's your choice how you want to implement this, so I'm good with either approach.

src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,90 @@
...
+        </Tooltip>
+      )}
+    </div>
+  );
+};
+
+Toggle.displayName = displayName;
+
+Toggle.defaultProps = {
+  disable: false,
```
rdig, line 75: This should say `disabled`, **not** `disable`

```suggestion
  disabled: false,
```

src/modules/core/components/Fields/Toggle/Toggle.tsx
chinins: i followed `Checkbox` example :) but i guess it was an old one. Updated now.

src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+import InputLabel from '~core/Fields/InputLabel';
+import Icon from '~core/Icon';
+import { Tooltip } from '~core/Popover';
+
+import styles from './Toggle.css';
+
+const displayName = 'Toggle';
+
+interface Props {
```
rdig, line 13: This needs `appearance` themes.

Off the top of my head:
- primary (switch is green)
- danger (switch is pink)
- minimal (doesn't have an outline, just a grey backround like the progress bar does)

But these can be all added in time, when they are needed. 

For now, please just implement the first two: primary and danger

src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,84 @@
...
+        >
+          <div className={styles.icon}>
+            <Icon
+              name="question-mark"
+              appearance={{ size: 'small' }}
+              title=""
+            />
+          </div>
+        </Tooltip>
+      )}
```
rdig, line 77: I would argue that this doesn't fit in here since it should be added externally on a case by case basis.

But since you've added it already, we might as well take advantage of it.

So just leave it in :)

src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div>
+        <input
+          name={name}
+          type="checkbox"
+          disabled={disabled}
+          aria-checked={value}
+          aria-disabled={disabled}
+          className={styles.delegate}
+          onChange={onChange}
```
rdig, line 56: You might need to create a custom method for this.

I've noticed a bug.

If the initial value is `true` and you try to click it, it takes to clicks to change the state:
- 1st click does nothing
- 2nd click changes the state to `false`

This only happens the first time around. All subsequent clicks work as expected.

I wasn't able to pinpoint the cause of this so you'll need to do some investigating. _(it works as expected if the initial value is `false`)_

Might be because the `toggleSwitch` is positioned absolute, but again, I'm not sure.

Here's the setup I used to test:
```js
      <Form onSubmit={console.log} initialValues={{ toggle: true }}>
        <Toggle name="toggle" />
        <Button type="submit" text={{ id: 'button.confirm' }} />
      </Form>
```

rdig (changes requested):
Except for the bug I found when clicking with the initial state set to `true`, this is all good. I've only rejected it because of that.

Screenshot from 2021-03-15 21-28-36

Try to investigate the root cause and see what's going on there.

ArmandoGraterol (approved):
It seems like the bug Raul found is fixed! All good from my side :+1:

rdig (approved):
Very nice! :tada:

PR #2417, opened 2021-03-16
Create logic for displaying toggle in modals

alicjakujawa:
## Description

This PR adds in logic for displaying toggle in modals

**New stuff** ✨

* Added hook which return information if extension is enabled
* Fetching and passing this flag in ColonyHomeActions and ColonyFundingMenu

**Changes** 🏗

* Updated heading to lets us add toggle inline with title
* Updated core Toggle component to support Message descriptors

Resolves DEV-250

### DIFF
--- a/src/modules/core/components/Dialog/DialogSection.css
+++ b/src/modules/core/components/Dialog/DialogSection.css
@@ -5,10 +5,10 @@
 }
 
 .themeHeading {
-  padding: 32px 20px 20px;
   display: flex;
-  align-items: center;
   justify-content: space-between;
+  align-items: center;
+  padding: 32px 20px 20px;
 }
 
 .themeSidePadding {
--- a/src/modules/core/components/Dialog/DialogSection.css
+++ b/src/modules/core/components/Dialog/DialogSection.css
@@ -6,6 +6,9 @@
 
 .themeHeading {
   padding: 32px 20px 20px;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
 }
 
 .themeSidePadding {
--- a/src/modules/core/components/Fields/Toggle/Toggle.tsx
+++ b/src/modules/core/components/Fields/Toggle/Toggle.tsx
@@ -58,7 +58,11 @@ const Toggle = ({
   return (
     <div className={styles.container}>
       {!elementOnly && label && (
-        <InputLabel label={label} labelValues={labelValues} appearance={{ colorSchema: 'grey' }} />
+        <InputLabel
+          label={label}
+          labelValues={labelValues}
+          appearance={{ colorSchema: 'grey' }}
+        />
       )}
       <div>
         <input
@@ -78,11 +82,13 @@ const Toggle = ({
       {tooltipText && (
         <Tooltip
           appearance={{ theme: 'dark' }}
-          content={tooltipText && tooltipText === 'string' ? (
-            tooltipText
-          ) : (
-            <FormattedMessage {...tooltipText} values={tooltipTextValues} />
-          )}
+          content={
+            tooltipText && tooltipText === 'string' ? (
+              tooltipText
+            ) : (
+              <FormattedMessage {...tooltipText} values={tooltipTextValues} />
+            )
+          }
           trigger="hover"
           popperProps={tooltipPopperProps}
         >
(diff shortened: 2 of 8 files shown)


src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -18,7 +19,7 @@ interface Appearance {
 interface Props {
   appearance?: Appearance;
   name: string;
-  label?: string;
+  label?: string | MessageDescriptor;
```
rdig, line 22: This should also have a `labelValues` which will bring in the (potentially) message descriptor variables

src/modules/core/components/Fields/Toggle/Toggle.tsx
```diff
@@ -18,7 +19,7 @@ interface Appearance {
...
   disabled?: boolean;
   tooltipText?: string;
```
rdig, line 24: Same for tooltip text, should be a message descriptor and there should also be a `tooltipTextValues` prop.

_(Yes, I know you didn't add them like this, but while you are at it, might as well make the change)_

src/modules/dashboard/components/ColonyFundingMenu/ColonyFundingMenu.tsx
```diff
@@ -65,6 +67,7 @@ const ColonyFundingMenu = ({
   const handleMintTokens = useCallback(() => {
     openTokenMintDialog({
       colony,
+      isVotingExtensionEnabled,
     });
   }, [colony, openTokenMintDialog]);
```
rdig, line 72: This needs `isVotingExtensionEnabled` as a dependency, no ?

src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
```diff
@@ -106,14 +107,17 @@ const ManageFundsDialog = ({
...
   const { walletAddress, username, ethereal } = useLoggedInUser();
 
   const allUserRoles = useTransformer(getAllUserRoles, [colony, walletAddress]);
 
   const hasRegisteredProfile = !!username && !ethereal;
   const canMoveFunds = hasRegisteredProfile && canFund(allUserRoles);
-  const canMintNativeToken = colony.canMintNativeToken && hasRoot(allUserRoles);
+  const canMintNativeToken =
+    (colony.canMintNativeToken && hasRoot(allUserRoles)) ||
+    isVotingExtensionEnabled;
```
rdig, line 120: While the logic is good, I think this is hard to read.

I would have left the `canMintNativeToken` logic as it is, and just change the check in the object:
```
    {
      title: MSG.mintTokensTitle,
      description: MSG.mintTokensDescription,
      icon: 'emoji-seed-sprout',
      permissionRequired: !isVotingExtensionEnabled || !canMintNativeToken,
      permissionInfoText: MSG.permissionsListText,
      permissionInfoTextValues: {
        permissionsList: (
          <FormattedMessage {...MSG.mintTokensPermissionsList} />
        ),
      },
      onClick: () => callStep(nextStepMintTokens),
    },
```

src/modules/dashboard/components/TokenMintDialog/TokenMintForm.tsx
```diff
@@ -70,8 +77,11 @@ const TokenMintForm = ({
           appearance={{ size: 'medium', margin: 'none', theme: 'dark' }}
           text={MSG.title}
         />
+        {userHasPermissions && isVotingExtensionEnabled && (
+          <Toggle label={MSG.force} name="toggle" />
```
rdig, line 81: We need a proper name for this. How about `forceMotion`

src/modules/core/components/Dialog/DialogSection.css
```diff
@@ -5,6 +5,9 @@
 }
 
 .themeHeading {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
```
rdig, line 10: Please check if this change doesn't break the styles of the modals that don't a toggle.

rdig (approved):
Looks good to me. I have left a few comments that need to be addressed, but after that, you can merge it.

ArmandoGraterol (approved):
Nice! So nice I'll show this PR to my mom :+1:

src/modules/core/components/Dialog/DialogSection.css
alicjakujawa: i was checking :)

PR #2421, opened 2021-03-22
Fix Permission Modal bugs

alicjakujawa:
## Description

- This PR adds fixes bugs related to setting permissions in management modal

**Changes** 🏗

* Updated architecture description to be more general
* Fixed advanced dialog to let user with root (and only root) permission to enter Permission Management dialog
* Fixed userHasPermissions flag in PermissionManagementDialog
* Fixed canRoleBeSet callback in PermissionManagementForm
* Hide root and recovery permission checkboxes from subdomains

**Notes** 🏗
* Run npm install before QA
* To change root domain permissions you need to have ROOT
* To change any permission in subdomain, you need to have architecture
* If you have architecture permission in subdomain, you CANNOT change anything in this domain -> only in future sub domains

Resolves DEV-259

### DIFF
--- a/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
+++ b/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback } from 'react';
+import React, { useCallback, useMemo } from 'react';
 import { defineMessages } from 'react-intl';
 import { ColonyRole, ROOT_DOMAIN_ID } from '@colony/colony-js';
 import sortBy from 'lodash/sortBy';
@@ -114,6 +114,16 @@ const PermissionManagementForm = ({
     [onDomainSelected],
   );
 
+  const filteredRoles = useMemo(
+    () =>
+      domainId !== ROOT_DOMAIN_ID
+        ? availableRoles.filter(
+            (role) => role !== ColonyRole.Root && role !== ColonyRole.Recovery,
+          )
+        : availableRoles,
+    [availableRoles, domainId],
+  );
+
   return (
     <>
       <div className={styles.domainSelectContainer}>
@@ -130,7 +140,7 @@ const PermissionManagementForm = ({
         appearance={{ colorSchema: 'grey' }}
       />
       <div className={styles.permissionChoiceContainer}>
-        {availableRoles.map((role) => {
+        {filteredRoles.map((role) => {
           const roleIsInherited =
             !userDirectRoles.includes(role) &&
             userInheritedRoles.includes(role);
--- a/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
+++ b/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
@@ -90,15 +90,7 @@ const PermissionManagementForm = ({
           return false;
       }
     },
-    [
-      currentUserRoles,
-      domainId,
-      rootAccounts,
-      userDirectRoles,
-      canSetPermissionsInRoot,
-      hasArchitectureInRoot,
-      hasRoot,
-    ],
+    [domainId, canSetPermissionsInRoot, hasArchitectureInRoot, hasRoot],
   );
 
   const domainSelectOptions = sortBy(
@@ -121,7 +113,7 @@ const PermissionManagementForm = ({
             (role) => role !== ColonyRole.Root && role !== ColonyRole.Recovery,
           )
         : availableRoles,
-    [availableRoles, domainId],
+    [domainId],
   );
 
   return (
--- a/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
+++ b/src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
@@ -35,6 +35,7 @@ interface Props {
   domainId: number;
   rootAccounts: Address[];
   userDirectRoles: ColonyRole[];
+  currentUserRolesInRoot: ColonyRole[];
   userInheritedRoles: ColonyRole[];
   colonyDomains: DomainFieldsFragment[];
   userHasPermission: boolean;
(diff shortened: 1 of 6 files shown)


src/data/graphql/queries.graphql
```diff
@@ -303,6 +303,18 @@ query RecoveryRolesUsers($colonyAddress: String!, $endBlockNumber: Int) {
   }
 }
 
+query ArchitectureRoleUsers($colonyAddress: String!) {
+  architectureRoleUsers(colonyAddress: $colonyAddress) @client {
+    id
+    profile {
+      avatarHash
+      displayName
+      username
```
rdig, line 312: This is so f****** stupid!!!!!

This doesn't work because of `username` and I have no idea why!!!!!!

(there are a couple of more issues, but the main problem is `username`)

If you change this to anything other than, `username` it will work.

This is how I've changed it:
```suggestion
      userName
```

src/data/resolvers/role.ts
```diff
@@ -0,0 +1,78 @@
...
+          }
+          return userWithProfile;
+        });
+        return users;
+      } catch (error) {
+        console.error(error);
+        return [];
+      }
+    },
+  },
```
rdig, line 77: There's a couple of issues that you need to tackle here:
- change `username` to something else (see above)
- you need to return `avatarHash`, `displayName` and `userName` for the minimal profile
- you need to check if the `subscribedUsers` data and `colonyAddress` exists before starting to use it

Here's my implementation:
```js
    async architectureRoleUsers(_, { colonyAddress }) {
      try {
        const subscribedUsers = await apolloClient.query<
          ColonySubscribedUsersQuery,
          ColonySubscribedUsersQueryVariables
        >({
          query: ColonySubscribedUsersDocument,
          variables: {
            colonyAddress,
          },
        });

        if (subscribedUsers?.data?.subscribedUsers && colonyAddress) {
          const colonyClient = (await colonyManager.getClient(
            ClientType.ColonyClient,
            colonyAddress,
          )) as ColonyClientV6;

          const rolesSet = await getMultipleEvents(colonyClient, [
            colonyClient.filters.ColonyRoleSet(null, null, null, null, null),
          ]);

          const allUsers = subscribedUsers?.data?.subscribedUsers || [];
          const userWithArchitectureRoles = getUsersArchitectureRoles(rolesSet);

          return userWithArchitectureRoles.map((userAddress) => {
            const userWithProfile = allUsers.find(
              ({ profile: { walletAddress } }) =>
                walletAddress.toLowerCase() === userAddress.toLowerCase(),
            );
            if (!userWithProfile) {
              const minimalUser = getMinimalUser(createAddress(userAddress));
              return {
                ...minimalUser,
                profile: {
                  userName: null,
                  displayName: null,
                  avatarHash: null,
                  walletAddress: minimalUser.id,
                },
              };
            }
            return {
              ...userWithProfile,
              profile: {
                ...userWithProfile.profile,
                userName: userWithProfile?.profile?.username || null,
              },
            };
          });
        }
        return [];
      } catch (error) {
        console.error(error);
        return [];
      }
    },
```

Combined with the the `username` prop rename, this works perfectly:

Screenshot from 2021-03-22 20-48-46

rdig (commented):
Besides the things I've pointed out, you also need to filter out the extensions addresses 

_(if you look at my screenshot you'll see that the address that has the minimal profile, is actually the `OneTxPayment` extension address to which we've added, by default, the architecture permission)_

To get the installed extensions addresses you need to call this query: `SubgraphColony` and that you can get all the installed extensions addresses from the `extensions` prop _(or you can use the processed colony query as well)_

Also. God f***ing damn apollo. I have no idea why it doesn't like the `username` name :(

src/data/resolvers/role.ts
alicjakujawa: Omg, how you found out about this naming problem? Thanks!!

chinins (approved):
Just one question about dependencies, other than that looks good.

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
```diff
@@ -62,24 +73,32 @@ const PermissionManagementForm = ({
...
+            (domainId === ROOT_DOMAIN_ID && hasRoot) || hasArchitectureInRoot
+          );
 
         default:
           return false;
       }
     },
-    [currentUserRoles, domainId, rootAccounts, userDirectRoles],
+    [
+      currentUserRoles,
```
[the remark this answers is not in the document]
alicjakujawa: nope, i will clean this up

ArmandoGraterol (approved):
Seems to be working fine :+1:

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
```diff
@@ -95,6 +114,16 @@ const PermissionManagementForm = ({
...
   );
 
+  const filteredRoles = useMemo(
+    () =>
+      domainId !== ROOT_DOMAIN_ID
+        ? availableRoles.filter(
+            (role) => role !== ColonyRole.Root && role !== ColonyRole.Recovery,
+          )
+        : availableRoles,
+    [availableRoles, domainId],
```
rdig, line 124: Same as what @chinins mentioned above, `availableRoles` is not needed here.

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
```diff
@@ -50,7 +51,17 @@ const PermissionManagementForm = ({
   colonyDomains,
   onDomainSelected,
   userHasPermission,
+  currentUserRolesInRoot,
 }: Props) => {
+  const canSetPermissionsInRoot =
+    domainId === ROOT_DOMAIN_ID &&
+    currentUserRoles.includes(ColonyRole.Root) &&
+    (!userDirectRoles.includes(ColonyRole.Root) || rootAccounts.length > 1);
```
rdig, line 59: Shouldn't this be `>= 1` ?

rdig (approved):
Very nice work! :1st_place_medal: 

Thank you untangling and making sense of all these edge cases.

src/modules/dashboard/components/PermissionManagementDialog/PermissionManagementForm.tsx
alicjakujawa: Im actually not sure about this part of logic 🤔 This logic exist before, but now im not sure if its correct honestly...\
 1. Should we block possibility of  removing root permission by person which also holding root? 🤔  (probably question to product)
 2. Should we block possibility of removing root from last root? (also probably question to product)

PR #2427, opened 2021-03-23
Filter out Reputation Mining Cycle Incoming Transfers

rdig:
This PR filters out _spammy_ incoming transfers that are generated by the Reputation Miner on each Cycle. These serve no purpose and will be fixed (at some point) on the network side as well, so that they don't get emitted.

But until then, we filter them out client side as well.

The way to tell them apart is:
- they are sent from the colony network address (ether router)
- they have a `0` amount

**Changes** 

- Refactored `getColonyUnclaimedTransfers` to use `Array.reduce` instead of `Array.map`
- Filter out mining reputation cycle transfers inside of `getColonyUnclaimedTransfers`

**Screenshots**

Before:
Screenshot from 2021-03-23 14-36-19

After:
Screenshot from 2021-03-23 14-36-23

Resolves DEV-267


chinins (approved):
All good.

PR #2428, opened 2021-03-24
Add `TokenSymbolSelector` & replace in `TransferFunds` & `CreatePayment`

chinins:
## Description

This PR adds  new core component TokenSymbolSelector that will work similarly to the Select, just that it will also need to display the token's avatar.

**New stuff** ✨

- add `TokenSymbolSelector`

**Changes** 🏗

- replace `Select` with `TokenSymbolSelector` in `TransferFundsDialogForm` & `CreatePaymentDialogForm`
- update css in `TransferFundsDialogForm` so that the avatar is not squashed

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-265

Screenshot 2021-03-24 at 18 57 54
Screenshot 2021-03-24 at 18 58 20
Screenshot 2021-03-24 at 18 58 47

### DIFF
--- a/src/modules/core/components/Fields/Select/Select.tsx
+++ b/src/modules/core/components/Fields/Select/Select.tsx
@@ -29,7 +29,7 @@ const MSG = defineMessages({
   },
 });
 
-interface Props {
+export interface Props {
   /** Appearance object */
   appearance?: Appearance;
 
@@ -255,6 +255,8 @@ const Select = ({
           activeOption.label,
           activeOption.labelValues,
         );
+      } else if (activeOption.labelElement) {
+        activeOptionLabel = activeOption.labelElement;
       } else {
         activeOptionLabel = activeOption.label;
       }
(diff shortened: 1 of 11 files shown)


src/modules/core/components/Fields/Select/Select.tsx
```diff
@@ -255,6 +255,8 @@ const Select = ({
           activeOption.label,
           activeOption.labelValues,
         );
+      } else if (activeOption.labelElement) {
+        activeOptionLabel = activeOption.labelElement;
       } else {
         activeOptionLabel = activeOption.label;
       }
```
rdig, line 262: I mean, I would have just extended `activeOption.label` to be able to support react components as well as strings / message descriptors.

But it's your decision ultimately how you want to implement this.

rdig (approved):
Very nice work here, yet again :100: 

Screenshot from 2021-03-24 21-18-53

Left a comment, but it's more of a nitpick than anything else.

src/modules/core/components/Fields/Select/Select.tsx
chinins: @rdig, I actually did go this route initially. But the way label is used in `SelectOption` & also because of the types, there would need to be more of redoing Select & supporting components and types. So I changed to the `labelElement` version.

PR #2429, opened 2021-03-25
Motion and Disputes StakingWidget and Staking Saga

ArmandoGraterol:
This PR adds in the Staking widget (for the YAY side), adds the staking as well as improving a lot of subtleties related to sagas, the Slider and the motions feed.

Changes:
- Added `StakingWidget` component
- Added `stakeMotion` saga and redux types
- Added `motionStakes` motions resolver
- Refactor core `Slider` to be able to reset values, be in a disabled state

**Note that most of the below fixes are not strictly related to this PR, but I've done the here anyway**

Fixes:
- Fix `colonyActions` saga to be able to load action pages (after you install the Voting Reputation extensions)
- Actions list fixed 10% stake filtering on motions

Demo:

User is logged out:
Screenshot from 2021-04-22 22-20-53

Motion is not fully staked on the YAY side:
Screenshot from 2021-04-22 22-21-04

Motion is fully staked on the YAY side:
Screenshot from 2021-04-22 22-20-45

demo-vr-staking-widget

Resolves DEV-256


src/modules/dashboard/sagas/actions/stakeMotion.ts
```diff
@@ -0,0 +1,116 @@
+import { call, fork, put, takeEvery } from 'redux-saga/effects';
+import { ClientType, VotingReputationClient } from '@colony/colony-js';
```
rdig, line 2: There's not such export in colony-js, the export you're looking for `ExtensionClient`

src/modules/core/components/Button/Button.tsx
```diff
@@ -22,7 +22,8 @@ export interface Appearance {
     | 'blue'
     | 'no-style'
     | 'white'
-    | 'dottedArea';
+    | 'dottedArea'
+    | 'pink';
```
rdig, line 26: Why was this added? This already exists!!! It's called `danger`

Please remove the "pink" theme

src/data/graphql/queries.graphql
```diff
@@ -242,6 +242,9 @@ query ColonyAction($transactionHash: String!, $colonyAddress: String!) {
     domainName
     domainPurpose
     domainColor
+    motionId
+    motionDomainId
+    rootHash
```
rdig, line 247: Why added them explicitly here?

You already have to `MotionCreated` event in the `events` object. Why not take your values from there ?

This is just pointless duplication!!!

Have a look at @alicjakujawa 's PR on how to do this: https://github.com/JoinColony/colonyDapp/blob/b04d8e6df8f93bf4ce07d621cb3c68b4f1b15c36/src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx#L65-L68

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+
+import { Context } from '~context/index';
+
+export const stakesResolvers = ({
+  colonyManager,
+}: Required<Context>): Resolvers => ({
+  Query: {
+    async stakeMotionLimits(
+      _,
+      { colonyAddress, userAddress, motionId, tokenDecimals },
```
rdig, line 13: Why do we need the tokenDecimals here passed in?

You can easily get them from the colony client!

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+    ) {
+      try {
+        const colonyClient = (await colonyManager.getClient(
+          ClientType.ColonyClient,
+          colonyAddress,
+        )) as ColonyClientV6;
+        const votingReputationClient = await colonyClient.getExtensionClient(
+          Extension.VotingReputation,
+        );
```
rdig, line 22:
```suggestion
        const votingReputationClient = await colonyManager.getClient(
          Extension.VotingReputation,
          colonyAddress,
        );
```

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+        const {
+          skillRep,
+          stakes,
+          domainId,
+        } = await votingReputationClient.getMotion(motionId);
+        const { skillId } = await colonyClient.getDomain(domainId);
+        const { reputationAmount } = await colonyClient.getReputation(
+          skillId,
+          userAddress,
+        );
```
rdig, line 32: Don't you need the custom `rootHash` in here ?

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+        // @NOTE There's no prettier compatible solution to this :(
+        // eslint-disable-next-line max-len
+        const totalStakeFraction = await votingReputationClient.getTotalStakeFraction();
+        // eslint-disable-next-line max-len
+        const userMinStakeFraction = await votingReputationClient.getUserMinStakeFraction();
+
+        const totalStaked = bigNumberify(stakes[1])
+          .div(bigNumberify(10).pow(tokenDecimals))
+          .toNumber();
```
rdig, line 41: How is this total staked, when you are only calculating the stakes from the YAY side ?

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+          .div(bigNumberify(10).pow(tokenDecimals * 2))
+          .toNumber();
+        const userMinStakeAmount = skillRep
+          .mul(totalStakeFraction)
+          .mul(userMinStakeFraction)
+          .div(bigNumberify(10).pow(tokenDecimals * 3))
+          .toNumber();
+        const formattedReputationAmount = reputationAmount
+          .div(bigNumberify(10).pow(tokenDecimals))
+          .toNumber();
```
rdig, line 53: Again, as I mentioned in the other PR review:
- **don't** convert values toNumber, convert them toString as it's the only way to handle big values
- **don't** convert values to ETH before export, leave them as WEI, and use `Numeral` inside the component to display them

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+        const totalStakeFraction = await votingReputationClient.getTotalStakeFraction();
+        // eslint-disable-next-line max-len
+        const userMinStakeFraction = await votingReputationClient.getUserMinStakeFraction();
+
+        const totalStaked = bigNumberify(stakes[1])
+          .div(bigNumberify(10).pow(tokenDecimals))
+          .toNumber();
+        const totalStakeAmount = skillRep
+          .mul(totalStakeFraction)
+          .div(bigNumberify(10).pow(tokenDecimals * 2))
```
rdig, line 44: For the native token, this 10 to the power of 36, why ?

This is super wrong

src/data/resolvers/stakes.ts
```diff
@@ -0,0 +1,81 @@
...
+          .toNumber();
+        const userMinStakeAmount = skillRep
+          .mul(totalStakeFraction)
+          .mul(userMinStakeFraction)
+          .div(bigNumberify(10).pow(tokenDecimals * 3))
```
rdig, line 49: For the native token, this 10 to the power of 54!!!!!!!, why ?

This is super wrong

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
```diff
@@ -137,6 +141,15 @@ const MintTokenMotion = ({
...
         <div className={styles.details}>
+          {motionState === MotionState.StakeRequired && (
+            <StakingWidget
+              motionId={motionId}
+              motionDomainId={motionDomainId}
+              rootHash={rootHash}
+              colonyAddress={colony.colonyAddress}
+              tokenDecimals={decimals}
+            />
+          )}
```
rdig, line 152: I don't think this is correct, since if there is "some", but not 100% stake, you basically won't show the widget

src/modules/dashboard/sagas/actions/stakeMotion.ts
```diff
@@ -0,0 +1,116 @@
...
+import { putError, takeFrom } from '~utils/saga/effects';
+
+import {
+  createTransaction,
+  createTransactionChannels,
+  getTxChannel,
+} from '../../../core/sagas';
+import { transactionReady } from '../../../core/actionCreators';
+
+function* stakeMotionAction({
```
rdig, line 15: This is not an action, it should not be named `stakeMotionAction`, and it should not live in the `actions` folder

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,142 @@
...
+            limit={maxStake}
+          />
+          <div className={styles.buttonGroup}>
+            <Button
+              type="submit"
+              disabled={!isValid || !isSubscribed}
+              text={MSG.stakeButton}
+            />
+            <Button appearance={{ theme: 'pink' }} text={MSG.objectButton} />
+          </div>
```
rdig, line 133: The slider and buttons are not aligned properly on the same vertical line

Screenshot from 2021-04-16 15-30-20

rdig (changes requested):
This still needs a lot of work: the logic in the resolver is bad and it has a lot of code repetition

That's why the slider always shows 0, since it can't actually calculate the values properly _(you divide them by a too high of a number)_

Screenshot from 2021-04-16 15-30-20

1. Please test your PRs thoroughly before submitting them for review
2. This is **again**, a very low quality PR, and is frankly not acceptable.

src/modules/dashboard/sagas/actions/stakeMotion.ts
ArmandoGraterol: I'll change it to `ExtensionClient`, but how come there's a `VotingReputationClient` export available? I'm using `4.0.0-rc.0`

rdig: There isn't. That's why your CI doesn't pass, and that's why you get this error:

Screenshot from 2021-04-16 16-01-15

ArmandoGraterol: Mmmm, I just saw that the VotingReputationClient it's in the `.d.ts` of the index of colonyJS. That's why it was letting me.

Either way, fixed.

src/modules/core/components/Button/Button.tsx
ArmandoGraterol: It doesn't exist

`themeDanger` uses:   --danger: rgb(248, 43, 101);
`themePink` uses:   --pink: rgb(254, 94, 124);

The pink looks kind of red, but it's a noticeable different shade of red (pinkish red :D) than the one that uses the danger theme. Should I still remove it?

src/data/resolvers/stakes.ts
ArmandoGraterol: Thank you, must have been erased in the rebase I did

src/data/resolvers/stakes.ts
ArmandoGraterol: Sure, will change the name to `YAYSideTotal` or something like that

rdig: No, but I meant don't we need to total both stakes for this to work? Both yay and nay

ArmandoGraterol: Mmmm I'm not so sure about it. The reason as to why I'm calculating the total staked over the YAY side it's because of these 2 edge cases Alex mentioned to me:

> 1. When the motion is nearly totally staked, such that the remaining amount of stake is smaller than minstake, we always allow the user to stake that much, otherwise the motion would get stuck
> 2. Maxstake can only be up to the amount of stake required to get to 100% (note that this and the one above can be different depending on which side the user is staking on)

I use totalStaked to get what's the amount left to achieve the 100% from that side of the vote and limit the Slider based on that

src/data/resolvers/stakes.ts
ArmandoGraterol: Is that so? I asked Alex how could I get the user-readable/ ETH version of the amounts and he explained it to me like this:

> both totalStakeFraction and userMinStakeFraction are stored as fixed-point decimals on chain and wherever you're using either you have to divide by 10 ** 18 to get the 'real' number
> if you're using both, as in the case of minStake, you have to divide by 10 ** 18 twice, or by 10 ** 36

This is my interpretation of that explanation.  It's 10 ^ 54 due to `totakStakeFraction`, `userMinStakeFraction`, and `skillRep`, from what I understand, the 3 of them are stored on 10 ^ 18

src/data/resolvers/stakes.ts
ArmandoGraterol: See my reasoning below

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
ArmandoGraterol: What would be a better check? In the design, the widget to `Stake` only seems to appear when the Motion is in a `Stake required` status

src/data/resolvers/stakes.ts
ArmandoGraterol: The `ReactSlider` needs `number` values to set a min, max, and limit. 

How could not use them and provide what the Slider needs? Providing percentages instead of the number themself?

For example, if minStake is 250, maxStake is 500, and the requiredStake is 1000, then:

The min of the Slider is 25%(250), the max is 50%(500), and the max is 100%(1000) ? 

And to get the amount the user input with the slider we use a a rule of 3

What do you think?

src/modules/core/components/Button/Button.tsx
rdig: Yes, remove it. We can deal with polishing the app at a later date.

src/data/resolvers/stakes.ts
rdig: Ok, but convert it to a string rather then a number _(since it will be quite a large number, and JS can't handle it)_

src/data/resolvers/stakes.ts
rdig: So you're 90% right :)

Talked to Alex, and this should actually be:
```
        const totalStakeAmount = skillRep
          .mul(totalStakeFraction)
          .div(bigNumberify(10).pow(tokenDecimals + 18))
```

src/data/resolvers/stakes.ts
rdig: Same as above, 90% right. Talked to Alex, it should be:
```
        const userMinStakeAmount = skillRep
          .mul(totalStakeFraction)
          .mul(userMinStakeFraction)
          .div(bigNumberify(10).pow(tokenDecimals + 2 * 18))
```

src/data/resolvers/stakes.ts
rdig: What about if you have a low value? 0.00001 

JS won't handle that, much less pass it to the React slider.

I think it's fine for now, but you have edge cases here which you are not covering

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
rdig: This should show for these three states:
- MotionState.StakeRequired
- MotionState.Motion
- MotionState.Objection

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -89,7 +96,7 @@ const Slider = ({
 
   return (
     <ReactSlider
-      min={0}
+      min={min}
       step={1}
```
rdig, line 100: Step should be configurable as well, what if we want to stake 1.25 tokens ?

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.css
```diff
@@ -0,0 +1,35 @@
+.wrapper {
+  display: flex;
+  flex-direction: column;
+  margin-top: 15px;
```
rdig, line 4: Margin to should be ~30

Screenshot from 2021-04-20 13-55-06

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.css
```diff
@@ -0,0 +1,35 @@
...
+  font-size: var(--size-smallish);
+  font-weight: var(--weight-bold);
+  color: var(--dark);
+}
+
+.description {
+  margin-bottom: 18px;
+  font-size: var(--size-tiny);
+  line-height: 18px;
+  color: color-mod(var(--temp-grey-blue-7) alpha(85%));
```
rdig, line 18: Weight for description is bold

Spec
Screenshot from 2021-04-20 13-55-40

Yours
Screenshot from 2021-04-20 13-55-42

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.css
```diff
@@ -0,0 +1,35 @@
...
+  margin-top: 32px;
+}
+
+.amount {
+  align-self: flex-end;
+  margin-bottom: 2px;
+  font-size: var(--size-tiny);
+  line-height: 18px;
+  color: color-mod(var(--temp-grey-blue-7) alpha(85%));
+}
```
rdig, line 31: Amount should be bold

Spec
Screenshot from 2021-04-20 13-55-52

Yours
Screenshot from 2021-04-20 13-55-50

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,144 @@
...
+              limit={maxStake}
+            />
+          </div>
+          <div className={styles.buttonGroup}>
+            <Button
+              type="submit"
+              disabled={!isValid || !isSubscribed}
+              text={MSG.stakeButton}
+            />
+            <Button appearance={{ theme: 'pink' }} text={MSG.objectButton} />
```
rdig, line 134: Button text is not aligned properly

Screenshot from 2021-04-20 13-55-23

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,144 @@
...
+          </p>
+          <span className={styles.amount}>{values.amount}</span>
+          <div className={styles.sliderContainer}>
+            <Slider
+              name="amount"
+              value={values.amount}
+              min={minStake}
+              max={requiredStake}
+              limit={maxStake}
+            />
```
rdig, line 126: The slider needs to show the token symbol next to the amount

Screenshot from 2021-04-20 13-55-52

rdig: Step for the slider should be 0.25 not 1

rdig: The slider should be:
- limit is the the totalstake - currentStake (only that difference)
- if the user doesn't have enough active tokens to cover the above difffence, then the new limit is the current users active tokens
- current value is the current staked so far _(the saga should update this, after you stake some more)_


Here's the example

Saga is already fully staked:
Screenshot from 2021-04-20 14-29-12

But the slider still shows that I've staked 0, and allows me to stake the full requiredStake value
Screenshot from 2021-04-20 14-28-32

rdig (approved):
There's still a lot of work to do here.

Resolver logic seems to be getting better, but the UI is not.

Slider still doesn't work as expected _(I've left comments detailing this)_, plus there are inconsistencies with the spec in terms of spacing, font weights, etc.

I don't like the quality of your recent submitted PRs and I hope you can improve this soon.

rdig (changes requested):
I meant to reject this with the previous review. :(

src/data/resolvers/stakes.ts
ArmandoGraterol: Yeah, that actually makes more sense :)

rdig:
Note: Took over from @ArmandoGraterol

alicjakujawa (changes requested):
Screenshot 2021-04-23 at 14 15 39
Missing validation, when i dont have enough activated tokens.

For the second problem i mentioned on discord i asked Alex and waiting for response

alicjakujawa:
Also if we cannot stake twice for the same motion we should disable stake button. I cannot move slider after i staked - thats why im assuming that we cannot stake twice.

chinins (changes requested):
I have an error of insufficient stake that happens if I stake without touching the slider (so the default starting amount).
Screenshot 2021-04-23 at 12 20 08

Plus, the whole maxStake (the one that I mention in one of my comments comment) is not very clear to me. I can stake twice. But no matter what the first stake was (unless it's fully staked) the second stake is the minimal that is required but the slider minimum is still shown like the first time (more than the actual stake is done).

Please, see below. The minimum stake is 2983 in this case
Screenshot 2021-04-23 at 15 50 47

ArmandoGraterol:
> Also if we cannot stake twice for the same motion we should disable stake button. I cannot move slider after i staked - thats why im assuming that we cannot stake twice.

@alicjakujawa I think it's not letting you move the slider because if you are the only user, and you stake once, the next maximum amount you can stake is going to be the amount the motion requires to be fully staked, but also, each user has a minimum that they can stake

So, if the required stake to fully stake the motion is **500**, and you stake **300,** then the next time the max should be **200**, but at the same time, if you are the only user in the colony, then the minimum a user has to stake is **250**. And so the max becomes **200**, but the min is **250**

That's my guess as to why it doesn't let you move the slider after you stake once and you don't stake it fully the first time around.

alicjakujawa:
> > Also if we cannot stake twice for the same motion we should disable stake button. I cannot move slider after i staked - thats why im assuming that we cannot stake twice.
> 
> @alicjakujawa I think it's not letting you move the slider because if you are the only user, and you stake once, the next maximum amount you can stake is going to be the amount the motion requires to be fully staked, but also, each user has a minimum that they can stake
> 
> So, if the required stake to fully stake the motion is **500**, and you stake **300,** then the next time the max should be **200**, but at the same time, if you are the only user in the colony, then the minimum a user has to stake is **250**. And so the max becomes **200**, but the min is **250**
> 
> That's my guess as to why it doesn't let you move the slider after you stake once and you don't stake it fully the first time around.

so we need to disable stake button then, as currently we can click it and it cause error :)

rdig:
We need to talk about this as I think it's some confusion going on.

But at any rate, once we find a solution we all agree on, I am happy to fix it :)

chinins:
@rdig , I've added `UserBalanceWithLockQuery` to `cahceUpdates` in `motionSaga` to update the token balance - in my object saga branch.
Just FYI, so that you don't add it as well.

Screenshot 2021-04-23 at 17 09 08

rdig:
> @rdig , I've added `UserBalanceWithLockQuery` to `cahceUpdates` in `motionSaga` to update the token balance - in my object saga branch.
> Just FYI, so that you don't add it as well.
> 
> Screenshot 2021-04-23 at 17 09 08

Very nice!

We're also going to need a resolver to fetch the number of activated tokens, since if you don't have enough, you won't be able to stake.

This affects both you an me.

I'll add it in, and you can just copy from mine.

src/modules/core/components/Slider/Slider.css
```diff
@@ -0,0 +1,7 @@
+.main :global(.rc-slider-disabled) {
+  padding: 6px 0 7px;
+}
+
+.main :global(.rc-slider-disabled .rc-slider-handle) {
+  border-color: var(--text-disabled) !important;
```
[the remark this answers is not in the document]
rdig: We already have a parent class _(It's `main`)_. 

I just couldn't find another way of over-writing that value without using `!important` as "something" inside `rc-slider` prevents normal overwrites

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,255 @@
...
+        setFieldValue('amount', parseFloat(userStakeBottomLimit));
+        scrollToRef?.current?.scrollIntoView({ behavior: 'smooth' });
+      }
+    },
+    [data, nativeToken, scrollToRef],
+  );
+
+  /*
+   * @TODO Add proper loading state
+   */
```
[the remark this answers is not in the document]
rdig: No. I'll remove it.

"Proper" loading state means that initially I only had a empty `div` in there as a placeholder :)

But that's gone now, and I forgot to remove the comment

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,255 @@
...
+        resetForm({});
+        setFieldValue('amount', parseFloat(userStakeBottomLimit));
+        scrollToRef?.current?.scrollIntoView({ behavior: 'smooth' });
+      }
+    },
+    [data, nativeToken, scrollToRef],
+  );
+
+  /*
+   * @TODO Add proper loading state
```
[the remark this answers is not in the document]
rdig: Nope, this is a leftover

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,255 @@
...
+      >
+        {({ values }) => (
+          <div className={styles.wrapper}>
+            <div className={styles.title}>
+              <Heading
+                text={MSG.title}
+                className={styles.title}
+                appearance={{ size: 'normal', theme: 'dark', margin: 'none' }}
+              />
+              <QuestionMarkTooltip
```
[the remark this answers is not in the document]
rdig: Screenshot from 2021-04-26 12-41-22
Screenshot from 2021-04-26 12-41-34

Ca't replicate this

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,255 @@
...
+                limit={userStakeTopLimit}
+                step={0.01}
+                disabled={!canUserStake || remainingToStakeSafe === 0}
+              />
+            </div>
+            <div className={styles.buttonGroup}>
+              <Button
+                appearance={{ theme: 'primary', size: 'medium' }}
+                type="submit"
+                disabled={!canUserStake || remainingToStakeSafe === 0}
```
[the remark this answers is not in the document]
rdig: You cannot stake less than than the `minUserStake` since this is enforced by the contracts, so both the values and display are correct.

But what happens underneath the hood, is that the contract only takes enough to fully stake the motion and returns the rest to you (yes, you can over-stake), so in a sense you are correct, that you are "only" staking `remainingToStakeSafe`, while sending the transaction for the `minUserStake` amount

This is the correct implementation.

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
```diff
@@ -0,0 +1,299 @@
...
+              <Button
+                appearance={{ theme: 'primary', size: 'medium' }}
+                type="submit"
+                disabled={!canUserStake}
+                text={MSG.stakeButton}
+              />
+              <Button
+                appearance={{ theme: 'danger', size: 'medium' }}
+                text={MSG.objectButton}
+                disabled={!canUserStake}
```
chinins, line 287: I think that you can't use the same condition for object button as for the stake button because if the yay side is fully staked, we can still object and stake objection. I think it will work if we add a check for the Nay side.
But I can do it in my branch.

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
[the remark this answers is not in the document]
chinins: Strange. Then I'll just fix it in my branch...

chinins (approved):
Everything works well now. The only thing is that object button is disabled if the motion is fully staked on yay side which I think shouldn't be the case. But I can fix it in my branch.

alicjakujawa (approved):
Good to merge 🔥  Bug with slider will be fix in separate PR

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
rdig: Sure, you're in control of the Objections part anyway :)

PR #2430, opened 2021-03-26
Add mini side loading states

chinins:
## Description

This PR adds `MiniSpinnerLoader` core component to replace repetitive code in places where small spinners are used.

**New stuff** ✨

- add `MiniSpinnerLoader`

**Changes** 🏗

- replace `SpinnerLoader` with `MiniSpinnerLoader`
- replace repetitive css with standard css in `MiniSpinnerLoader`

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-227

### DIFF
--- a/src/modules/core/components/Preloaders/MiniSpinnerLoader.tsx
+++ b/src/modules/core/components/Preloaders/MiniSpinnerLoader.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { FormattedMessage, MessageDescriptor } from 'react-intl';
+
+import Heading, { Appearance as HeadingAppearance } from '~core/Heading';
+import { SimpleMessageValues } from '~types/index';
+
+import SpinnerLoader, { Appearance } from './SpinnerLoader';
+
+interface Props {
+  appearance?: Appearance;
+  titleAppearance?: HeadingAppearance;
+  title?: string | MessageDescriptor;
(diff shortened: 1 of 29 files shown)


src/modules/core/components/Preloaders/MiniSpinnerLoader.tsx
```diff
@@ -0,0 +1,49 @@
...
+      {title && (
+        <Heading
+          text={title}
+          textValues={titleTextValues}
+          appearance={titleAppearance}
+        />
+      )}
+      <SpinnerLoader appearance={appearance} />
+      {loadingText && (
+        <span className={loadingTextClassName || styles.loadingText}>
```
rdig, line 41: I don't think we should have this customizable, but it your choice in the end, so I'm fine if you decide to leave it in.

rdig (approved):
Very nice work on cleaning this up! :1st_place_medal:

ArmandoGraterol (approved):
Looks good to me :ok_hand:

src/modules/core/components/Preloaders/MiniSpinnerLoader.tsx
chinins: In some components the styles are a bit different - like when we have it in a dialog and a couple of other instances. So I prefer to leave it like that.

PR #2431, opened 2021-03-26
Extract IPFS logic from ActionsPageFeedItem

ArmandoGraterol:
Extracted IPFS and annotation related logic from ActionsPageFeedItem into ActionsPageFeedItemWithIPFS

**New stuff** ✨

* New ActionsPageFeedItemWithIPFS component which receives all the props of the normal version of the component plus an `hash` props which is used to retrieve IPFS data (Right now it's only used to retrieve annotations)

**Deletions** ⚰️

* IPFS and annotation logic in ActionsPageFeedItem

Resolves DEV-228


rdig (approved):
Not what I had originally in mind when I wrote this issue, but I like your approach.

I've tested it and it works great.

Nice job! :tada:

chinins (approved):
Great work! 🔥

PR #2432, opened 2021-03-29
Dynamic tag on Motion Page

alicjakujawa:
## Description

This PR adds in dynamic tag on motion page

**New stuff** ✨

* Handling motion created events
* Getting correct action type for motion
* Getting needed args from motion
* Added helper to generate correct motion states based on motion network states and additional conditions 
* Added motionTagMap which return correct translations/styling for tags based on motion state
* Added queries/types for motionState
* Fetching motion state and displaying tag in MintTokenMotion component
* colonyMotions utils

Resolves DEV-253


src/utils/colonyMotions.ts
```diff
@@ -0,0 +1,101 @@
...
+export enum MotionState {
+  Motion = 'Motion',
+  StakeRequired = 'StakeRequired',
+  Voting = 'Voting',
+  Reveal = 'Reveal',
+  Objection = 'Objection',
+  Failed = 'Failed',
+  Passed = 'Passed',
+  Invalid = 'Invalid',
+}
```
rdig, line 23: I think I'll integrate these into colonyJS as well

src/utils/colonyMotions.ts
```diff
@@ -0,0 +1,101 @@
...
+    theme: 'primary',
+    colorSchema: 'plain',
+    name: MSG.passedTag,
+  },
+  [MotionState.Invalid]: {
+    theme: 'pink',
+    colorSchema: 'plain',
+    name: MSG.invalidTag,
+  },
+};
```
rdig, line 101: Hmmm....

I'm good with having a tag map, but I don't think it should live inside `utils`

It should be somewhere next to the motion component, so inside `ActionPage`

src/data/graphql/typeDefs.ts
```diff
@@ -62,6 +62,7 @@ export default gql`
     domainPurpose: String!
     domainColor: String!
     blockNumber: Int!
+    motionState: String
```
rdig, line 65: Shouldn't this be `String!` ? I mean, the motion will always have some state

src/data/resolvers/colonyActions.ts
```diff
@@ -149,16 +161,28 @@ export const colonyActionsResolvers = ({
...
+            colonyClient as ColonyClient,
+            reverseSortedEvents[0],
+          );
+        } else {
+          actionType = getActionType(reverseSortedEvents);
+        }
         const actionValues = await getActionValues(
           reverseSortedEvents,
           colonyClient as ColonyClient,
+          votingClient as ExtensionClient,
```
rdig, line 181: This doesn't work when the `VotingReputation` extension is not installed. It will break loading a normal action, since the voting client won't be available

rdig (changes requested):
I'm fine with most of this. Very nice work! :100: 

The only thing I'm rejecting this for is that it won't work if the `VotingReputation` extension is not yet installed, because the `colonyActions` resolver always assumes it's installed, which might not always be the case.

You need to add some checks in there to only try to access it if it's installed, or if needed.

Also, while I know this was not added in your PR, I've noticed that, after I installed the VotingReputation extension, I can't actually create an extension through the saga:

Screenshot from 2021-04-01 16-40-43

But this might be more of a question for @chinins since she added the saga.

src/utils/colonyMotions.ts
alicjakujawa: Didint see any good place for it in there 🤔 any ideas?

src/data/graphql/typeDefs.ts
alicjakujawa: unless it wasn't a motion?

src/utils/colonyMotions.ts
rdig: `dashboard/ActionsPage/staticMaps` just so that all are in the same place.

And yes I realize that this isn't the ideal solution, since these are not really actions anymore, but motions.

rdig (approved):
Very nice! Working as expected! This some quality code you've added in :1st_place_medal: 

Screenshot from 2021-04-02 11-58-41

PS: About the error I've posted on my first review, about not being able to create a motion, disregard that, I was just being a moron, and not testing it properly _(my user didn't have reputation)_

PR #2434, opened 2021-03-31
Add objection modal and saga

chinins:
## Description

This PR adds objection part to the staking. The object button in `StakingWidget` triggers `RaiseObjectionDialog` if nobody has staked objection. If there is already some stake in objection it will go to objection widget.

I have added placeholder for Armando's component and a condition for it as well.

**New stuff** ✨

- add `RaiseObjectionDialog` with forms etc
- add annotation message to staking saga
- modify `StakingWidget` and staking logic to include objection case

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


rdig:
What's the status of this ?

chinins:
This is on hold until Armando merges staking widget and saga PR

PR #2435, opened 2021-03-31
Create motion event

alicjakujawa:
## Description

This PR displaying motion created event on Action Page

**New stuff** ✨

* Displayed create motion event on Action Page

**NOTE**
We decided to keep lighter font

Screenshot 2021-04-02 at 13 36 39

Resolves DEV-272

### DIFF
--- a/src/i18n/en-events.ts
+++ b/src/i18n/en-events.ts
@@ -13,6 +13,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.RecoveryStorageSlotSet} {{initiator} submitted new value for slot {storageSlot}}
       ${ColonyAndExtensionsEvents.RecoveryModeExitApproved} {{initiator} approved exiting}
       ${ColonyAndExtensionsEvents.RecoveryModeExited} {{initiator} exited Recovery Mode}
+      ${ColonyAndExtensionsEvents.MotionCreated} {{initiator} {reputation} created a {motionTag}}
       other {{eventNameDecorated} emmited by {clientOrExtensionType}}
     }`,
   /*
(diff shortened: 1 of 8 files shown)


src/i18n/en-events.ts
```diff
@@ -13,6 +13,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.RecoveryStorageSlotSet} {{initiator} submitted new value for slot {storageSlot}}
       ${ColonyAndExtensionsEvents.RecoveryModeExitApproved} {{initiator} approved exiting}
       ${ColonyAndExtensionsEvents.RecoveryModeExited} {{initiator} exited Recovery Mode}
+      ${ColonyAndExtensionsEvents.MotionCreated} {{initiator} {reputation} created a {motionTag}}
```
rdig, line 16: So there's two things here:
1. I think we need to add reputation to all items in the feed that have an initiator, including the normal actions
2. I see the reputation being more integrated with the initiator, rather than being a separate entity

So in my head, you just have the initiator as a motion descriptor value:
```js
      ${ColonyAndExtensionsEvents.MotionCreated} {{initiator} created a {motionTag}}
```

Then when passing along the values, you just do something like this:
```js
    initiator: (
      <>
        <span className={styles.titleDecoration}>
          <FriendlyName user={initiator} autoShrinkAddress />
        </span>
        <div className={motionSpecificStyles.reputation}>
          <MemberReputation
            walletAddress={actionInitiator}
            colonyAddress={colony.colonyAddress}
          />
        </div>
      </>
    ),
```

Meaning you pass both the initiator and reputation in the same component

rdig (approved):
I'm good with this, apart from the way you display the initiator's reputation.

I've left some comments on how I would approach this.

If you want to change it here, then go for it, if not, just open a new issue to keep track of it, so we can change it later down the line.

chinins (approved):
Works great, looks great - nice work!

rdig (approved):
Good to go again! Very nice!

PR #2436, opened 2021-04-02
Add `CountdownTimer` component

chinins:
## Description

This PR adds `CountdownTimer` component that will show one of these options:

- Time left to stake
- Motion will fail in
- Voting ends in
- Reveal ends in
- Motion will pass in

**Important to read!** I have tried to use `react-intl` option for the time left but it just shortens it to hours, days etc and doesn't show seconds when the time left is more than a minute etc. I have also looked for other appropriate packages but due to css, translation, the same shortening as in `react-intl` or the low popularity of packages I decided that it's cleaner & easier to add it on my own. The code is not complicated and it's quite concise. 

**New stuff** ✨

- add `VotingExtensionParams` type & query
- add `CountdownTimer` component
- add `extensionsResolvers`
- add `calculateTimeLeft` helper function
- add `CountdownTimer` to `MintTokenMotion`. For now with just the stake props.
- add some css to `DefaultAction.css` file to reflect the countdown timer correctly on the actions page

**Changes** 🏗

I have cherry-picked 2 commits from the master branch with `MiniSpinnerLoader` as I wanted to use it in the code. Hope it's ok? If yes, then you can just ignore all the loader files. If not I will remove.

**Deletions** ⚰️

None

## TODO

none

Resolves DEV-270

Screenshot 2021-04-05 at 11 39 16


src/modules/core/components/CountDownTimer/CountDownTimer.css
```diff
@@ -0,0 +1,10 @@
+.container {
+  font-size: var(--size-tiny);
+  font-weight: var(--weight-bold);
+  color: color-mod(var(--temp-grey-blue-7) alpha(85%));
+}
+
+.time {
+  color: var(--dark);
+  opacity: 100%;
```
ArmandoGraterol, line 9: I think the time would look better and closer to the design specs if the opacity was `85%`

ArmandoGraterol (approved):
Other than the opacity of the time, this looks good to me!

src/modules/core/components/CountDownTimer/CountDownTimer.tsx
```diff
@@ -0,0 +1,125 @@
...
+        setTimeLeft(calculateTimeLeft(createdAt, stakePeriod));
+      }, 1000);
+      return () => clearInterval(timer);
+    }
+    return undefined;
+  }, [createdAt, stakePeriod]);
+
+  if (loading) {
+    return <MiniSpinnerLoader loadingText={MSG.loadingText} />;
+  }
```
rdig, line 94: You need to include the `data === undefined` check in here as well.

Otherwise, you get a brief flash of the loading state, then it disappears until it loads all the data, then finally it shows the time.

Here's the demo:

bug-countdown-spinner

src/modules/core/components/CountDownTimer/calculateTimeLeft.ts
```diff
@@ -0,0 +1,18 @@
+export const calculateTimeLeft = (createdAt: number, actionPeriod?: number) => {
+  if (!actionPeriod) {
+    return undefined;
+  }
+  // 1617362399585
```
rdig, line 5: I assume this is not needed

src/modules/core/components/CountDownTimer/calculateTimeLeft.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (difference > 0) {
+    return {
+      days: Math.floor(difference / (1000 * 60 * 60 * 24)),
+      hours: Math.floor((difference / (1000 * 60 * 60)) % 24),
+      minutes: Math.floor((difference / 1000 / 60) % 60),
+      seconds: Math.floor((difference / 1000) % 60),
+    };
+  }
+
+  return undefined;
```
rdig, line 17: Why two `return undefined` ?

You can simplify this down to one, by checking if the `actionPeriod` exists

src/modules/core/components/CountDownTimer/calculateTimeLeft.ts
```diff
@@ -0,0 +1,18 @@
+export const calculateTimeLeft = (createdAt: number, actionPeriod?: number) => {
```
rdig, line 1: This method should go into `utils/time`

_(you'll need to create the file + folder structure for this)_

src/data/resolvers/extensions.ts
```diff
@@ -0,0 +1,42 @@
...
+        const stakePeriod = stakePeriodBigNum.toNumber();
+
+        const submitPeriodBigNum = await extensionClient.getSubmitPeriod();
+        const submitPeriod = submitPeriodBigNum.toNumber();
+
+        const revealPeriodBigNum = await extensionClient.getRevealPeriod();
+        const revealPeriod = revealPeriodBigNum.toNumber();
+
+        const escalationPerBigNum = await extensionClient.getEscalationPeriod();
+        const escalationPeriod = escalationPerBigNum.toNumber();
```
rdig, line 28: As a precaution, this should be converted to strings, as JS is weird when it comes to large numbers.

It works perfectly as-is, I just wanted to mention this so you could be aware

rdig (changes requested):
This is mostly good _(it functions correctly)_, however there's still changes required before we can merge.

I've left comments in the relevant places.

src/modules/core/components/CountDownTimer/CountDownTimer.css
chinins: Good catch!

chinins:
The PR is ready for re-review.

FYI, after a conversation with Alex - we need to use blockchain and not real time as current time, since there can be up to 15 min of discrepancy. It took me a while to get to the current implementation as I went down a different alley in the beginning and that didn't really work out.

So now the time left refers to the blockchain time + the action period (stake, reveal etc).

Please, double check my logic and let me know if you have any questions

src/modules/core/components/CountDownTimer/CountDownTimer.tsx
```diff
@@ -0,0 +1,125 @@
...
+            {...MSG.minutes}
+            values={{ minutes: timeLeft.minutes }}
+          />
+        )}
+        <FormattedMessage
+          {...MSG.seconds}
+          values={{ seconds: timeLeft.seconds }}
+        />
+      </span>
+    </div>
```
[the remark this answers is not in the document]
chinins: I actually worked from the screen that doesn't have  it so I didn't realise there are cases when there is an icon. I've now added an option to add the icon with the tooltip.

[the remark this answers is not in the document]
chinins: I am also waiting for the info from the product. But we have an option to have both variants: with and without icon.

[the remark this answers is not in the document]
chinins: Confirmed by product: sometimes will have the icon and sometimes we won't.

ArmandoGraterol (approved):
If I understood correctly the logic changes, it seems to be working fine for me :+1:

src/modules/core/components/CountDownTimer/CountDownTimer.tsx
```diff
@@ -0,0 +1,122 @@
...
+        );
+      }, 1000);
+      return () => clearInterval(timer);
+    }
+    return undefined;
+  }, [createdAt, stakePeriod, differenceVsBCTime]);
+
+  if (loading || data === undefined) {
+    return <MiniSpinnerLoader loadingText={MSG.loadingText} />;
+  }
```
rdig, line 85: The bug where the timer is not displayed after loading is still there:

bug-countdown-spinner-2

I think the only way to fix this properly is to store our own loading checks in react's state, and only setting that to false when `calculateTimeLeft` is done

rdig (approved):
I'm all good with this, but the problem with the countdown now showing _(for a brief period)_ is still there :(

PR #2437, opened 2021-04-06
Prevent motion triggering when reputation is  zero

chinins:
## Description

This PR prevents triggering a motion when colony reputation is zero and adds an explanation copy in the `TokenMint` dialog.

The figma link:
https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=4020%3A335

**New stuff** ✨

- add `colonyReputation` query, type & resolver
- add `useColonyReputation` hook that returns `isColonyReputationZero` check
- add `NotEnoughReputation` core component

**Changes** 🏗

- small update to `TokenMint` dialog to avoid repeated code

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-268

Screenshot 2021-04-06 at 15 36 57


src/modules/core/components/NotEnoughReputation/NotEnoughReputation.tsx
```diff
@@ -0,0 +1,34 @@
...
+    id: 'NotEnoughReputation.description',
+    defaultMessage: `If you would like to enforce the action you can use
+    permissions. To create the action toggle on Force option in top
+    right corner.`,
+  },
+});
+
+const displayName = 'NotEnoughReputation';
+
+const NotEnoughReputation = () => (
```
rdig, line 21: I don't think this should be a core component as it's not general enough.

So either move it out of core, or tell my why you think it should remain in here :)

src/utils/hooks/useColonyReputation.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export const useColonyReputation = (colonyAddress: Address) => {
+  const { data, error } = useColonyReputationQuery({
+    variables: {
+      address: colonyAddress,
+    },
+  });
+
+  return {
+    isColonyReputationZero: data?.colonyReputation === '0' || error,
```
rdig, line 12: I don't like the name of this prop.

I would have gone with `colonyHasReputation`, but it's your call I guess

src/modules/dashboard/components/TokenMintDialog/TokenMintForm.tsx
```diff
@@ -72,6 +76,12 @@ const TokenMintForm = ({
 
   const requiredRoles: ColonyRole[] = [ColonyRole.Root];
 
+  const { isColonyReputationZero } = useColonyReputation(colony.colonyAddress);
+
+  const onlyForceAction =
+    isVotingExtensionEnabled && isColonyReputationZero && !values.forceAction;
```
rdig, line 82: Shouldn't the force toggle also check if the person perming the forced action, actually has the required permissions ?

Or do we check that some place else ?

rdig (approved):
Minor nitpicks / code structure stuff, but other than those, I'm good with this.

Nice work!

src/modules/dashboard/components/TokenMintDialog/TokenMintForm.tsx
ArmandoGraterol: Aren't those permission checked by the line below this one? 

`const inputDisabled = !userHasPermission || onlyForceAction;`

ArmandoGraterol (approved):
Looks good to me :ok_hand:

src/modules/dashboard/components/TokenMintDialog/TokenMintForm.tsx
chinins: yeah, it's done separately

src/utils/hooks/useColonyReputation.ts
chinins: done

src/modules/core/components/NotEnoughReputation/NotEnoughReputation.tsx
chinins: moved to dashboard

src/modules/dashboard/components/TokenMintDialog/TokenMintForm.tsx
chinins: Actually, it is checked when displaying Toggle. But it made me think of something and i've added a little missing part - `colonyHasReputation` to this check:
```
const userHasPermission =
    canMintTokens || (isVotingExtensionEnabled && colonyHasReputation);
```

ArmandoGraterol (approved):
Still working fine for me! :ok_hand:

rdig (approved):
Good to go

PR #2438, opened 2021-04-12
Add Motion has passed event

alicjakujawa:
## Description

This PR adds in motion has passed event

**New stuff** ✨

* Added motionSystemMessages resolver and query
* Added motionSystemMessage query to MintTokenMotion component
* Added System messages to MintTokenMotion component

Screenshot 2021-04-15 at 13 35 22

## TODO

- Found small bug connected to motion state, it will be fixed in DEV-285

## How to test [sorry it wont be super easy]
In truffle console you need to create needed events.

```
ethers = require("ethers")

c = await IColony.at(colonyAddress)
cnaddress = await c.getColonyNetwork()
cn = await IColonyNetwork.at(cnaddress)
tladdress = await cn.getTokenLocking()
tl = await ITokenLocking.at(tladdress)

votingAddress = await cn.getExtensionInstallation(web3.utils.soliditySha3("VotingReputation"), c.address);

v = await VotingReputation.at(votingAddress);

tokenAddress = await c.getToken()

t = await Token.at(tokenAddress)

// Amount required to stake depends on how much reputation was
// in the colony when the motion was created, and the total_stake_fraction that the extension was deployed with.

t.approve(tl.address, AMOUNT);
tl.deposit(t.address, AMOUNT, true);
// 1 here is the domain the motion has been made in
c.approveStake(v.address, 1, AMOUNT)

// The 1 before AMOUNT here is to stake that the motion should take place. If you want to stake against it, use 0.

// If the motion is not in the root domain the second and third parameters might be different. They are domain proofs, that the `withProofs` helpers deal with in colonyJS

// key value branchmask siblings come from the reputation oracle.

// Query http://localhost:3001/reputation/local/${motion.roothash}/${c.address}/${skillid corresponding to the domain the hash is taking place in }/${user address}

voting.stakeMotion(motionid, 1, ethers.constants.MaxUint256, 1, AMOUNT, key, value, branchmask, siblings)```

To make it passed:
```curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[604800],"id":1}' localhost:8545
curl -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[],"id":1}' localhost:8545```
```

Resolves DEV-275

### DIFF
--- a/src/data/index.ts
+++ b/src/data/index.ts
@@ -38,6 +38,7 @@ import { metaColonyResolvers } from './resolvers/metacolony';
 import { eventsResolvers } from './resolvers/events';
 import { recoveryModeResolvers } from './resolvers/recovery';
 import { extensionsResolvers } from './resolvers/extensions';
+import { motionsResolvers } from './resolvers/motions';
 
 import { FixedToken } from '../types';
 
@@ -64,6 +65,7 @@ export const resolvers: ResolverFactory[] = [
   eventsResolvers,
   recoveryModeResolvers,
   extensionsResolvers,
+  motionsResolvers,
 ];
 
 // export all the generated types and helpers
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -5,6 +5,7 @@ import {
   getLogs,
   getBlockTime,
   ExtensionClient,
+  NetworkMotionState,
 } from '@colony/colony-js';
 
 import { Context } from '~context/index';
@@ -14,7 +15,6 @@ import {
   SystemMessage,
   SystemMessagesName,
 } from '~dashboard/ActionsPageFeed';
-import { NetworkMotionState } from '~utils/colonyMotions';
 
 import { ProcessedEvent } from './colonyActions';
 
@@ -35,6 +35,7 @@ export const motionsResolvers = ({
       );
       const systemMessages: SystemMessage[] = [];
 
+      // @TODO Add missing types to colonyjs
       // @ts-ignore
       const motionStakedFilter = votingReputationClient.filters.MotionStaked(
         motionId,
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -5,7 +5,7 @@ import {
   getLogs,
   getBlockTime,
   ExtensionClient,
-  NetworkMotionState,
+  MotionState as NetworkMotionState,
 } from '@colony/colony-js';
 
 import { Context } from '~context/index';
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -4,13 +4,11 @@ import {
   ClientType,
   getLogs,
   getBlockTime,
-  ExtensionClient
+  ExtensionClient,
 } from '@colony/colony-js';
 
 import { Context } from '~context/index';
-import {
-  ColonyAndExtensionsEvents,
-} from '~types/index';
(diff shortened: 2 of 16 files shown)


src/data/index.ts
```diff
@@ -38,6 +38,7 @@ import { metaColonyResolvers } from './resolvers/metacolony';
 import { eventsResolvers } from './resolvers/events';
 import { recoveryModeResolvers } from './resolvers/recovery';
 import { extensionsResolvers } from './resolvers/extensions';
+import { motionsResolvers } from './resolvers/motions';
```
rdig, line 41: I created the same file, just that mine is called `motion` :P

But don't worry, I'll merge them in my PR

src/data/resolvers/motions.ts
```diff
@@ -0,0 +1,110 @@
...
+        null,
+        null,
+        null,
+      );
+      const motionStakedLogs = await getLogs(
+        votingReputationClient,
+        motionStakedFilter,
+      );
+      // @ts-ignore
+      // eslint-disable-next-line max-len
```
rdig, line 50: Please leave some comments so that in the future we'll remember why these were ignored 

_(because of `colonyJS` obviously, but we might forget :P)_

src/data/resolvers/motions.ts
```diff
@@ -0,0 +1,110 @@
...
+} from '@colony/colony-js';
+
+import { Context } from '~context/index';
+import { ColonyAndExtensionsEvents } from '~types/index';
+import {
+  ActionsPageFeedType,
+  SystemMessage,
+  SystemMessagesName,
+} from '~dashboard/ActionsPageFeed';
+import { NetworkMotionState } from '~utils/colonyMotions';
```
rdig, line 17: This should be imported from `colonyJS` . It's available since version `4.0.0-rc.0`

```suggestion
import { MotionState as NetworkMotionState } from '@colony/colony-js';
```

src/data/resolvers/motions.ts
```diff
@@ -0,0 +1,110 @@
...
+
+      if (
+        motionNetworkState === NetworkMotionState.Finalizable ||
+        motionNetworkState === NetworkMotionState.Finalized
+      ) {
+        const newestStakeOrVoteEvent = sortedEvents.find(
+          (event) =>
+            event.name === ColonyAndExtensionsEvents.MotionStaked ||
+            event.name === ColonyAndExtensionsEvents.MotionVoteRevealed,
+        );
```
rdig, line 92: Why are you looking for the `MotionStaked` / `MotionReveled` events in here ?

Wouldn't this work without checking for them?

I mean the motion is already in the `Finalizable`/`Finalized` state, meaning either we staked or voted, but either way we got to the end of the flow.

So we just need to decide if it passed or failed _(which you do, for both stakes or votes)_

So I think this will work even without searching for those two events.

rdig (approved):
Very nice work in here @alicjakujawa! :tada: 

I think the logic in the motions resolver can be simplified a bit, but other than that this is good to go!

Screenshot from 2021-04-16 14-03-23

src/data/resolvers/motions.ts
alicjakujawa: because i need timestamp of the newest event

rdig: You are right. I haven't thought of that. 

Very nice! :+1:

chinins (approved):
All good - great work!

PR #2441, opened 2021-04-12
Add MotionStaked event to ActionPage

ArmandoGraterol:
Added MotionStaked event to feed of MintTokenMotion

**New stuff** ✨

* Added query and local resolver to get events for a specific motion
* Connected eventsForMotion query to MintTokenMotion page
* Added amountTag and staker event value to feed item

**How to test** (If #2429 isn't merged) :leaves:
1. Check out to this branch: `testing/backed-motion-event`. Which has the staking widget and event combined.
2. Start the stack
3. Get reputation
4. Create a motion
5. Deposit the tokens you are going to stake
6. Stake in the motion
7. Profit

FireShot Capture 206 - Colony - localhost

Resolves DEV-271


chinins (commented):
The code looks ok. But I failed to test. I followed the instructions but either the rebase went wrong or one of the branches was wrongly rebased.... 
As discussed on discord, I would suggest for you to create a separate branch where it's correctly rebased (I assume you already have it since you tested it yourself, so this means just pushing it to github) where everybody can test it...

src/data/resolvers/motion.ts
```diff
@@ -0,0 +1,87 @@
...
+      bigNumberify(motionId),
+      null,
+      null,
+      null,
+    ),
+  );
+
+  const parsedMotionEvents = await Promise.all(
+    [...motionStakedLogs].map(async (log) => {
+      const potentialParsedLog = votingReputationClient.interface.parseLog(log);
```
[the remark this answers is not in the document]
ArmandoGraterol: Nice catch! This name seems to be a remnant of the code in the `colonyActions` file where from my understanding the log can potentially be `undefined/null`. Here, I think that possibility doesn't exist so I'll change the name

chinins:
I've tested on the new testing branch and it works well. The only thing is that there is extra margin-left of 10px for the tag that shows the staked amount. It comes for basic tag styles (when there are 2 elements in a row). I think you should remove it. 

Current: 

Screenshot 2021-04-15 at 21 25 13

What it is without the margin-left:

Screenshot 2021-04-15 at 21 25 43

src/data/graphql/queries.graphql
```diff
@@ -242,6 +242,7 @@ query ColonyAction($transactionHash: String!, $colonyAddress: String!) {
     domainName
     domainPurpose
     domainColor
+    motionId
```
rdig, line 245: As I've mentioned in the previous PR review, you don't need this here, you already have in the `events` object!

src/modules/dashboard/components/ActionsPageFeed/ActionsPageFeed.tsx
```diff
@@ -75,6 +79,7 @@ interface Props {
   actionType?: string;
   actionData: ColonyAction;
   colony: Colony;
+  token?: TokenInfoQuery['tokenInfo'];
```
rdig, line 82: Why is this here ?

Why not take the native token value from the `colony` prop ?

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -60,6 +67,7 @@ interface Props {
   emmitedBy?: string;
   actionData: ColonyAction;
   colony: Colony;
+  token?: TokenInfoQuery['tokenInfo'];
```
rdig, line 70: Same here, why do you need an extra prop passed down ?

This is available to you in the `colony` object, but you should know this

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -334,6 +343,34 @@ const ActionsPageEvent = ({
...
+                  <Tag
+                    appearance={{
+                      theme: 'primary',
+                      colorSchema: 'inverted',
+                      fontSize: 'tiny',
+                    }}
+                  >
+                    <Numeral
+                      value={values?.stakeAmount || 0}
+                      unit={getTokenDecimalsWithFallback(token?.decimals || 0)}
```
rdig, line 356: don't put `0` as a fallback for decimals!!! That's super bad. 

You're basically trying to calculate _(with 0)_, 10 to the power of 0, which is all sorts of nasty.

**That's why you are using `getTokenDecimalsWithFallback` so you don't have to pass in a fallback!**

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -334,6 +343,34 @@ const ActionsPageEvent = ({
...
+                amountTag: (
+                  <Tag
+                    appearance={{
+                      theme: 'primary',
+                      colorSchema: 'inverted',
+                      fontSize: 'tiny',
+                    }}
+                  >
+                    <Numeral
+                      value={values?.stakeAmount || 0}
```
rdig, line 355: This does not properly calculate the value for the stakeAmount.

Screenshot from 2021-04-16 15-36-19

Please revise your logic in the resolver

For reference the max stake is 995 and the amount I staked is 800

The values are in WEI

rdig (changes requested):
This still needs work as it does not calculate the staked amounts properly.

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
ArmandoGraterol: Tested it again (Both with the staking widget and with your method) after rebasing and doing the other changes requested, and it seems to be working as expected!

src/data/resolvers/motions.ts
```diff
@@ -20,6 +20,52 @@ import {
...
+
+  const parsedMotionEvents = await Promise.all(
+    [...motionStakedLogs].map(async (log) => {
+      const parsedLog = votingReputationClient.interface.parseLog(log);
+      const { address, blockHash, blockNumber, transactionHash } = log;
+      const {
+        name,
+        values: { amount, ...rest },
+      } = parsedLog;
+      const stakeAmount = name === 'MotionStaked' ? amount : null;
```
rdig, line 42: Please use the type in here

```suggestion
      const stakeAmount = name === ColonyAndExtensionsEvents.MotionStaked ? amount : null;
```

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -334,6 +344,37 @@ const ActionsPageEvent = ({
...
+                      margin: 'none',
+                    }}
+                  >
+                    <Numeral
+                      value={values?.stakeAmount || 0}
+                      unit={getTokenDecimalsWithFallback(
+                        colonyNativeToken?.decimals,
+                      )}
+                    />{' '}
+                    {values?.tokenSymbol}
```
rdig, line 362: Please use the numeral suffix

```suggestion
                    <Numeral
                      value={values?.stakeAmount || 0}
                      unit={getTokenDecimalsWithFallback(
                        colonyNativeToken?.decimals,
                      )}
                      suffix={` ${values?.tokenSymbol}`}
                    />
```

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -334,6 +344,37 @@ const ActionsPageEvent = ({
...
                 storageSlot: values?.slot?.toHexString(),
+                amountTag: (
+                  <Tag
+                    appearance={{
+                      theme: 'primary',
+                      colorSchema: 'inverted',
+                      fontSize: 'tiny',
+                      margin: 'none',
+                    }}
+                  >
```
rdig, line 355: Please apply the following css to the inline tags for this message:
```css
    vertical-align: top;
    margin-top: 2px;
```
This will align the tags properly with the text.

Here's how it looks currently:
Screenshot from 2021-04-20 12-10-54

And here's how it looks with the tags aligned
Screenshot from 2021-04-20 12-04-12

If you decide to wrap it in another element, make sure to also apply `display: inline-block;` to that element as well

rdig (approved):
I still want you to do some changes, but after that you can merge it

PR #2443, opened 2021-04-16
Fix motion states

alicjakujawa:
## Description

IN cases when motion is in Finalizable or Finalized state, and there was no voting, check stakes array to find out if motion passed or failed

**Changes** 🏗

* Updated getMotionState logic

Resolves DEV-285

### DIFF
--- a/src/utils/events/index.ts
+++ b/src/utils/events/index.ts
@@ -524,7 +524,8 @@ const getMotionState = async (
         : MotionState.Motion;
     case NetworkMotionState.Finalizable:
     case NetworkMotionState.Finalized:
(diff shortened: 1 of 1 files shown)


rdig (commented):
So I already fixes this in my #2442 

See this: https://github.com/JoinColony/colonyDapp/blob/e3077c7f24d17db85a4527e8f0836a308a7a455e/src/utils/events/index.ts#L526-L557

It also takes into account:
- stakes
- votes
- the math bug that's inside the contracts currently

So you can either discard this PR and wait for me to merge mine, or just cherry pick my changes and merge yours with my changes

PR #2452, opened 2021-04-21
Add MotionStaked NAY side event variant

ArmandoGraterol:
- Updated ActionsPageFeedd to make MotionStaked event dynamic and change backed side tag according to vote
- Truncated stakeAmount to 2 decimals only

FireShot Capture 223 - Colony - localhost


rdig (approved):
Seems to be working as expected now.

Good to go

Screenshot from 2021-04-21 18-30-28

PR #2420, opened 2021-03-17
Add TotalStakeWidget to motion action page

ArmandoGraterol:
Added TotalStakeWidget with reducers included

**New stuff** ✨

* Added TotalStakeWidget to show the total staked for/against the motion
* Added TotalStakeWidget subcomponent to allow the user to select the side to stake on
* Added query with a local resolver to get the user stake, total stake, and required stake
* Added getMotionRequiredStake function

**Changes** 🏗

* Integrated StakingWidget with TotalStakeWidget

**When there's no stake on the NAY side**
FireShot Capture 239 - Colony - localhost

**When there's a stake on the NAY side but the user is not logged in**
FireShot Capture 242 - Colony - localhost

**When there's a stake on the NAY side and the user is logged in**
FireShot Capture 245 - Colony - localhost

**When there's a stake on the NAY side and the user selected a side to stake in**
FireShot Capture 248 - Colony - localhost

**When you choose a side to stake on, a StakingWidget will appear but with a `Back` button**
FireShot Capture 251 - Colony - localhost

**When you choose the NAY side. TODO: Add Objection StakingWidget**
FireShot Capture 254 - Colony - localhost

Resolves DEV-255

PR #2454, opened 2021-04-21
Add MotionFullyStaked system message

ArmandoGraterol:
* Added system message for when a motion is fully staked
* Fixed system message timestamp not being properly aligned with the message
* Added `getMotionRequiredStake` util function (This was also added in another PR of mine, will resolve the conflict/remove duplicates when the time comes).
* Fixed `motionDomain` missing a default value in `colonyAction`
* Removed left margin of ObjectionTag when it's on a message

FireShot Capture 233 - Colony - localhost

Resolves DEV-274

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -235,11 +235,11 @@ export const motionsResolvers = ({
         (event) => event.name === ColonyAndExtensionsEvents.MotionVoteSubmitted,
       );
 
-      const newestMotionStakedEvent = sortedEvents.find(
+      const latestMotionStakedEvent = sortedEvents.find(
         (event) => event.name === ColonyAndExtensionsEvents.MotionStaked,
       );
 
-      if (newestMotionStakedEvent) {
+      if (latestMotionStakedEvent) {
         // eslint-disable-next-line max-len
         const totalStakeFraction = await votingReputationClient.getTotalStakeFraction();
         const requiredStake = getMotionRequiredStake(
@@ -252,7 +252,7 @@ export const motionsResolvers = ({
           systemMessages.push({
             type: ActionsPageFeedType.SystemMessage,
             name: SystemMessagesName.MotionFullyStaked,
-            createdAt: newestMotionStakedEvent.createdAt,
+            createdAt: latestMotionStakedEvent.createdAt,
           });
         }
       }
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -248,7 +248,7 @@ export const motionsResolvers = ({
           18,
         );
 
-        if (motion.stakes[MotionVote.Yay].eq(requiredStake)) {
+        if (motion.stakes[MotionVote.Yay].gte(requiredStake)) {
           systemMessages.push({
             type: ActionsPageFeedType.SystemMessage,
             name: SystemMessagesName.MotionFullyStaked,
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -14,7 +14,7 @@ import { Resolvers } from '@apollo/client';
 import { Context } from '~context/index';
 import { createAddress } from '~utils/web3';
 import { getMotionActionType, getMotionState } from '~utils/events';
-import { MotionVote } from '~utils/colonyMotions';
+import { MotionVote, getMotionRequiredStake } from '~utils/colonyMotions';
 import { ColonyAndExtensionsEvents } from '~types/index';
 import {
   UserReputationQuery,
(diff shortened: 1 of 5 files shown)


src/i18n/en-system-messages.ts
```diff
@@ -7,6 +7,7 @@ const systemMessagesMessageDescriptors = {
       ${SystemMessagesName.EnoughExitRecoveryApprovals} {Enough permission holders have now signed to exit recovery mode. As long as no further storage slots are updated, a recovery permission holder may now sign a transaction to reactivate the colony.}
       ${SystemMessagesName.MotionHasPassed} {{motionTag} has {passedTag} and may be finalized.}
       ${SystemMessagesName.MotionRevealPhase} {It's time to {revealTag} votes to the world!}
+      ${SystemMessagesName.MotionFullyStaked} {{motionTag} is fully staked; Staking period has reset. As long as there is no {objectionTag}, the motion will pass.}
```
chinins, line 10: I noticed that in Figma there is little space between `objectionTag` and a coma that you don't have and I think it looks a bit better with little space. But maybe not too important.

src/data/resolvers/motions.ts
```diff
@@ -158,6 +158,28 @@ export const motionsResolvers = ({
       const timeToSubmitMS = motion.events[1].toNumber() * 1000;
       const timeToSubmitInPast = timeToSubmitMS < blocktime;
 
+      const newestMotionStakedEvent = sortedEvents.find(
```
chinins, line 161: why newest? do we have an earlier one? if yes maybe latest is a better prefix? newest doesn't sound grammatically correct. It's not a big deal though.

chinins (approved):
Looks good, just 2 little nitpicks.

src/data/resolvers/motions.ts
ArmandoGraterol: They are synonyms, grammatically they both are valid. 

However, I think it's more common to say "latest event" than "newest event" for example, actually never heard anyone refer to a recent event as "newest" in english :)

So yeah, I could change it just for the sake of it sounding right, or more "native-like"

src/data/resolvers/colonyActions.ts
```diff
@@ -242,6 +242,7 @@ export const colonyActionsResolvers = ({
           domainColor: null,
           blockNumber,
           motionState: null,
+          motionDomain: null,
```
rdig, line 245: I approve of this fix, but it's not complete. 

To be complete you need to add both `motionState` and `motionDomain` default values to the case where we don't have a `transactionReceipt`.

Like I've did here: https://github.com/JoinColony/colonyDapp/blob/704ce13611d8d2d73daea8645957723c48f6b2c4/src/data/resolvers/colonyActions.ts#L290-L291

Or you can just leave it like this, since when my PR will be merged it will be fixed.

src/data/resolvers/motions.ts
```diff
@@ -158,6 +158,28 @@ export const motionsResolvers = ({
...
+        (event) => event.name === ColonyAndExtensionsEvents.MotionStaked,
+      );
+
+      if (latestMotionStakedEvent) {
+        // eslint-disable-next-line max-len
+        const totalStakeFraction = await votingReputationClient.getTotalStakeFraction();
+        const requiredStake = getMotionRequiredStake(
+          motion.skillRep,
+          totalStakeFraction,
+          18,
```
rdig, line 171: These should be the token's actual decimals, why is this value hard coded ?

src/data/resolvers/motions.ts
```diff
@@ -158,6 +158,28 @@ export const motionsResolvers = ({
...
+        );
+
+        if (motion.stakes[MotionVote.Yay].eq(requiredStake)) {
```
rdig, line 174: It might also be bigger by 1 than the required stake _(due to a bug in the contracts)_ so the easiest way around that is to check if greater than or equal

```suggestion
        if (motion.stakes[MotionVote.Yay].gte(requiredStake)) {
```

src/utils/colonyMotions.ts
```diff
@@ -93,3 +94,15 @@ export const MOTION_TAG_MAP = {
     name: MSG.invalidTag,
   },
 };
+
+export const getMotionRequiredStake = (
```
rdig, line 98: Make sure to replace this in the other places in the dapp where we use a required stake

src/data/resolvers/motions.ts
rdig: As I've stated above, here's the test for this:

Motion is fully staked on the yay side:
Screenshot from 2021-04-26 10-29-55

But the actual BigNumber values you are comparing here are off by one:
Screenshot from 2021-04-26 10-31-17

rdig (changes requested):
This is mostly good, as the bulk of the logic for this is here.

But it needs a few adjustments.

Screenshot from 2021-04-26 10-32-35

src/data/resolvers/motions.ts
ArmandoGraterol: We addressed this in #2420, I think. It's ok to leave it like this

rdig (approved):
Good to go, just make sure to rebase before merging

PR #2459, opened 2021-04-23
Flesh out DetailsWidget of motion page

ArmandoGraterol:
## Description

- Add details to DetailsWidget of motion page

**New stuff** ✨

* Added MintTokenMotions emoji and details

**Changes** 🏗

* Added dynamic message-id to DetailsWidget to cover motion titles

## TODO

- Add the details for the other motions when they are ready

FireShot Capture 257 - Colony - localhost

Resolves DEV-300

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
@@ -86,6 +86,7 @@ const MintTokenMotion = ({
   colonyAction,
   token: { decimals, symbol },
   transactionHash,
+  recipient,
   initiator,
 }: Props) => {
   const bottomElementRef = useRef<HTMLInputElement>(null);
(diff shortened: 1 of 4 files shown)


alicjakujawa (approved):
Screenshot 2021-05-05 at 18 11 07
Works good, the only think i would double check if we want so big space between those 2 widgets (double check with product)

ArmandoGraterol:
> Screenshot 2021-05-05 at 18 11 07
> 
> Works good, the only think i would double check if we want so big space between those 2 widgets (double check with product)

Hmmm ... I feel like that white space is where the result of the voting phase should show

alicjakujawa:
> > Screenshot 2021-05-05 at 18 11 07
> > Works good, the only think i would double check if we want so big space between those 2 widgets (double check with product)
> 
> Hmmm ... I feel like that white space is where the result of the voting phase should show

of course, but not in all states :D So we need to find out if this white space should exist for other states

chinins (approved):
Works well!

PR #2460, opened 2021-04-26
Fix displaying staked tokens

alicjakujawa:
## Description

This PR fixes displaying staked tokens in TokenActivationWallet

**Changes** 🏗

* Different way of calculating staked tokens, as getTotalObligation doesn't include tokens staked on voting reputation extension
* Fetched MotionStaked events, grouped them by motion id and vote
* Filtered them out by MotionRewardClaimed events (mapped by vote and motionId)
* stakedTokens = sum(filtered motionStaked events) + totalObligation
* Fixed wrong calculation of active tokens and total user balance

Screenshot 2021-04-26 at 17 14 24

Screenshot 2021-04-28 at 10 49 19

Resolves DEV-308

### DIFF
--- a/src/data/resolvers/user.ts
+++ b/src/data/resolvers/user.ts
@@ -5,8 +5,9 @@ import {
   ROOT_DOMAIN_ID,
   getBlockTime,
   getLogs,
+  getEvents,
 } from '@colony/colony-js';
-import { BigNumber } from 'ethers/utils';
+import { BigNumber, bigNumberify } from 'ethers/utils';
 import { AddressZero, HashZero } from 'ethers/constants';
 
 import { Context } from '~context/index';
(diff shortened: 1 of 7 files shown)


ArmandoGraterol (commented):
The code looks good to me, however, I couldn't test this.

When I create a colony and try to do a payment I got this error: 

`VotingReputation extension is not installed for this colony`

The payment action appears on the action list, but after the transaction is completed it doesn't redirect me to the ActionsPage. Also, my token balance isn't appearing in the navbar.

After installing the voting extension then the token balance appears, and I'm redirected correctly to the ActionsPage after making a payment.

So perhaps it's missing a check somewhere that checks if the voting extension is not/is installed

src/data/resolvers/user.ts
```diff
@@ -50,6 +51,78 @@ const getUserReputation = async (
...
+  colonyManager: ColonyManager,
+  colonyAddress: Address,
+  walletAddress: Address,
+): Promise<BigNumber> => {
+  const votingReputationClient = await colonyManager.getClient(
+    ClientType.VotingReputationClient,
+    colonyAddress,
+  );
+  /**
+   * @NOTE If there will be more staking events
```
rdig, line 64: You mean different event than `MotionStaked`, right ?

rdig (approved):
Looks and works good.

I just have a couple of notes:
- we need to wire the `getUserStakedBalance` query refresh inside the `updateMotionValues` saga helper _(so that the values update when staking on a motion)_
- I noticed a bug with gas station transaction notification _(that was most likely introduced by the Token Activation feature)_. While you're at it, mind fixing it?

How it is now: _(when there's a transaction to be signed)_
Screenshot from 2021-04-27 23-04-25

How it should be: _(when there's a transaction to be signed)_
Screenshot from 2021-04-27 23-05-30

alicjakujawa:
> The code looks good to me, however, I couldn't test this.
> 
> When I created a colony and try to do a payment I got this error:
> 
> `VotingReputation extension is not installed for this colony`
> 
> The payment action appears on the action list, but after the transaction is completed it doesn't redirect me to the ActionsPage. Also, my token balance isn't appearing in the navbar.
> 
> After installing the voting extension then the token balance appears, and I'm redirected correctly to the ActionsPage after making a payment.
> 
> So perhaps it's missing a check somewhere that checks if the voting extension is not/is installed

This is actually really good catch, i again forgot to check if extension is installed :P

src/data/resolvers/user.ts
alicjakujawa: yes, more ways of token staking (new events)

rdig:
> The code looks good to me, however, I couldn't test this.

> When I created a colony and try to do a payment I got this error:

> VotingReputation extension is not installed for this colony

But when you do a payment, is the `OneTxPayment` extension, not `VotingReputation`, so maybe something got mixed up ?

alicjakujawa:
> > The code looks good to me, however, I couldn't test this.
> 
> > When I created a colony and try to do a payment I got this error:
> 
> > VotingReputation extension is not installed for this colony
> 
> But when you do a payment, is the `OneTxPayment` extension, not `VotingReputation`, so maybe something got mixed up ?

nope, it causes our wallet to "refetch" as we display balance, so it fire my method which use voting extension, i fixed it already :) Will push changes whenever i finish with rest of things

rdig:
Very nice

PR #2461, opened 2021-04-26
Motions and Disputes Claim Rewards Widget 

rdig:
## Description

This PR adds in the claim stakes widget which will be available at the end of a motion (win or loose), if both you've only staked, or if you staked and voted.

Note that this does not include the saga, that will be a separate PR

**Changes** 

- [x] Refactored `FinalizeMotionWidget` into `FinalizeMotionAndClaimWidget` -- _they share a lot of data_
- [x] Added `motionFinalized` client resolver
- [x] Added `motionStakerReward` client resolver

**Demo**

demo-vr-claim-tokens

Resolves DEV-294


alicjakujawa (approved):
Almost there, pre-approving, just please fix this bug (wrong number format is for Winnings)
Screenshot 2021-04-27 at 10 31 22

PR #2463, opened 2021-04-27
Motions and Disputes Reveal Vote Saga

rdig:
## Description

This wires up the reveal vote saga for the motions

**Note when testing** We need to sign a message in order to generate the voting secret hash. When testing with the ganache wallet this will happen so fast you won't see it _(see the second demo below)_, but just know that it's there and that you should be aware of it

**Changes**

- [x] Added `revealVoteMotion` saga
- [x] Updated `updateMotionValues` saga util
- [x] Wired up `RevealWidget`

**Demos**

- Signing with a **metamask wallet** -- manually have to sign both the message, and the transaction
- Since only one user in the colony, it will go to finalize directly after revealing
demo-vr-reveal-instant

- Signing with a **software wallet** -- automatically signing messages + transactions.
- Also multiple available votes, since there are multiple users in the colony, hence it will not finalize instantly
demo-vr-reveal-wait

Resolves DEV-309


alicjakujawa (changes requested):
While testing with ganache account 
Screenshot 2021-04-27 at 16 53 01

alicjakujawa (approved):

PR #2466, opened 2021-04-28
Display total number of votes progress bar

alicjakujawa:
## Description

This PR adds in progress bar which is displaying in voting state

**New stuff** ✨

* VotingState resolver
* Displayed Progress bar with correct values
* Refetching votingState query to updateMotionValues helper

Screenshot 2021-04-28 at 13 26 40
Screenshot 2021-04-28 at 13 16 44

Resolves DEV-299

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -481,10 +481,6 @@ export const motionsResolvers = ({
     },
     async votingState(_, { colonyAddress, motionId }) {
       try {
-        const colonyClient = (await colonyManager.getClient(
-          ClientType.ColonyClient,
-          colonyAddress,
-        )) as ColonyClientV6;
         const votingReputationClient = await colonyManager.getClient(
           ClientType.VotingReputationClient,
           colonyAddress,
@@ -493,7 +489,7 @@ export const motionsResolvers = ({
           skillRep,
           repSubmitted,
         } = await votingReputationClient.getMotion(motionId);
-        const motion = await votingReputationClient.getMotion(motionId);
+        // eslint-disable-next-line max-len
         const maxVoteFraction = await votingReputationClient.getMaxVoteFraction();
 
         const threasholdValue = skillRep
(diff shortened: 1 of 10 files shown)


src/data/resolvers/motions.ts
```diff
@@ -478,6 +478,33 @@ export const motionsResolvers = ({
...
+        const {
+          skillRep,
+          repSubmitted,
+        } = await votingReputationClient.getMotion(motionId);
+        // eslint-disable-next-line max-len
+        const maxVoteFraction = await votingReputationClient.getMaxVoteFraction();
+
+        const threasholdValue = skillRep
+          .mul(maxVoteFraction)
+          .div(bigNumberify(10).pow(18));
```
rdig, line 496: I think @ArmandoGraterol already added an until for this, so that we don't have to keep repeating this code.

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.css
```diff
@@ -6,3 +6,15 @@
...
+  margin: 0 6px;
+  font-size: var(--size-tiny);
+  font-weight: var(--weight-bold);
+  color: var(--temp-grey-blue-7);
+}
+
+.progressBarContainer {
+  margin-left: 4px;
+  width: 116px;
+}
```
rdig, line 20: I think the progress bar should be fixed, otherwise it jumps around based on the number of seconds.

See this example:
bug-vr-voting-treshold

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
```diff
@@ -186,9 +222,28 @@ const MintTokenMotion = ({
...
+              </span>
+              <div className={motionSpecificStyles.progressBarContainer}>
+                <ProgressBar
+                  value={currentReputationPercent}
+                  threshold={threasholdPercent}
+                  max={100}
+                  appearance={{
+                    size: 'small',
+                    backgroundTheme: 'dark',
+                    barTheme: 'danger',
```
rdig, line 241: So the way that I understood the specs, is that the progress bar is green _(primary)_, and the threshold bar red, not both red.

But I might be wrong, so check with Product

rdig (approved):
There's some minor fixes to be done, but other than that, it's good to go!

PR #2467, opened 2021-04-28
ObjectionRaised event for motion page

ArmandoGraterol:
**New stuff** ✨

* Added message for ObjectionRaised event
* Added handler to detect if the ObjectionRaised has ocurred

**Changes** 🏗

* Added vote prop to eventValue type
* Added blockNumber to ProcessedEvent type

Resolves DEV-273

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -81,6 +81,35 @@ const getMotionEvents = async (
     (firstEvent, secondEvent) => firstEvent.createdAt - secondEvent.createdAt,
   );
 
+  const firstMotionStakedNAYEvent = sortedMotionEvents.find(
+    (event) =>
+      event.name === ColonyAndExtensionsEvents.MotionStaked &&
+      event.values.vote.eq(0),
+  );
+
+  if (firstMotionStakedNAYEvent) {
+    const {
+      values,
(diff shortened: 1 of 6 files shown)


src/data/resolvers/motions.ts
```diff
@@ -81,6 +81,35 @@ const getMotionEvents = async (
...
+      transactionHash,
+    } = firstMotionStakedNAYEvent;
+    sortedMotionEvents.push({
+      type: ActionsPageFeedType.NetworkEvent,
+      name: ColonyAndExtensionsEvents.ObjectionRaised,
+      /*
+       * @NOTE: I substract 1 second out of the timestamp
+       * to make the event appear before the first NAY stake
+       */
+      createdAt: firstMotionStakedNAYEvent.createdAt - 1000,
```
alicjakujawa, line 104: We can move it to constant, as we will do exact same things for other events to always appear first

alicjakujawa (changes requested):
Screenshot 2021-04-29 at 11 23 00
Almost there, just please make sure that system message "motion is fully staked; Staking period has reset. As long as there is no objection" appear before objections

alicjakujawa (approved):
Awesome 🔥

PR #2472, opened 2021-04-30
Add MotionRewardClaimed event

ArmandoGraterol:
Add MotionRewardClaimed event to the motion page feed

FireShot Capture 263 - Colony - localhost

Resolves DEV-296


chinins (approved):
Works well!

alicjakujawa (approved):
Im almost sure but could be wrong. We can have 2 events per user (for yes and no) if they voted multiple times. But we're "claiming" once from the dapp (at the bottom it create 2 events). So we can endup

PR #2473, opened 2021-04-30
Fixed user outcome message not considering winner side

ArmandoGraterol:
Fixed outcome for the voter showing that the user's side lost if they voted for the NAY side, even if the NAY side actually won.

Resolves DEV-310


alicjakujawa (approved):

chinins (approved):
Good to go!

PR #2476, opened 2021-05-04
Fix invalid tag

alicjakujawa:
## Description

This PR fixes incorrectly displayed invalid tag when motion is. failing

**New stuff** ✨

* Added FailedNoFinalizable motion state

**Changes** 🏗

* Updated getMotionState helper to return FailedNoFinalizable when motionNetworkState is failed (when there was not enough votes for yes) 

## How to test

1. Create motion
2. Stake small amount for yes / or stake for no
3. run:
curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[604800],"id":1}' localhost:8545

The motion will fail due to not enough stakes. It wont be finalizable, so you should see "failed" tag, but you shouldn't see "finalize" widget

Resolves DEV-319

### DIFF
--- a/src/utils/colonyMotions.ts
+++ b/src/utils/colonyMotions.ts
@@ -32,6 +32,7 @@ export enum MotionState {
   Objection = 'Objection',
   Failed = 'Failed',
   Passed = 'Passed',
+  FailedNoFinalizable = 'FailedNoFinalizable',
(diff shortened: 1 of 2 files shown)


rdig (approved):
Good to go!

PR #2481, opened 2021-05-06
Filter out events from contract

alicjakujawa:
## Description

This PR fixes duplicated events in Action List. When motion is finalised, contract emit event for this action (for example "Mint tokens") Because of that, we're ending with to Mint tokens event in action list (for one of them, initiator is contract) Because of this issue, people was thinking that "someone which is not a member of colony, minted tokens with force" (reported on QA)

**Changes** 🏗

* Filtering out events where initiator is contract

## TODO

- Make sure if we're not "filtering too much"
- If so, narrow down filtering to = agent !== votingContractAddress

Resolves DEV-321

### DIFF
--- a/src/modules/dashboard/transformers.ts
+++ b/src/modules/dashboard/transformers.ts
@@ -89,28 +89,16 @@ export const getActionsListData = (
         if (isTransactionRepeated) return acc;
 
         /*
-         * Filter out events that have the recipient an extension's address
+         * Filter out events that have the recipient or initiator an extension's address
          *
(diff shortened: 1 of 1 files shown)


ArmandoGraterol (approved):
Nice! Seems like the whole "minting tokens without being a member and having permissions" is solved with this :P

chinins (approved):
Looks good to me! 🔥

rdig (approved):
No we're talking :)

Nice job @alicjakujawa

PR #2483, opened 2021-05-06
Fix ActionPage showing StakeRequired on Failed motion

ArmandoGraterol:
Fixed ActionPage showing the StakeRequired motion state when the motion actually failed **without** reloading the dapp. 

The ActionsPage wouldn't refetch the action data when you **re-enter** the ActionsPage, therefore if the motion failed it would still show that a stake is required, even if the system message in the feed already shows that the motion failed.

### How to test:
1. Create colony
2. Get reputation
3. Create motion
4. Stake for YAY and/or NAY, but not fully
5. Wait for the motion to fail, or use the time machine to travel to the future (AKA. Reputation monitor, make payments with the monitor ON)
6. Go back to the ActionPage and you should see that it failed, shouldn't be able to stake.

Resolves DEV-320

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
@@ -18,10 +18,16 @@ import {
   useColonyFromNameQuery,
   useUser,
   useTokenInfoLazyQuery,
+  useMotionStatusLazyQuery,
 } from '~data/index';
 import { NOT_FOUND_ROUTE } from '~routes/index';
-import { ColonyActions, ColonyMotions } from '~types/index';
+import {
+  ColonyActions,
+  ColonyMotions,
+  ColonyAndExtensionsEvents,
+} from '~types/index';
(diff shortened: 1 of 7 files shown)


src/modules/dashboard/components/ActionsPage/ActionsPage.tsx
```diff
@@ -98,7 +98,7 @@ const ActionsPage = () => {
       loading: colonyActionLoading,
       error: colonyActionError,
     },
-  ] = useColonyActionLazyQuery();
+  ] = useColonyActionLazyQuery({ fetchPolicy: 'network-only' });
```
alicjakujawa, line 101: We don't want this, as colonyAction data is super heavy. The one way of doing this is to create one more query, which return motion status, and adding it as dependency to this query. So whenever motion status will change, then this query will be updated. @rdig idea :) If you have some other ideas, feel free to use them!

alicjakujawa (changes requested):

alicjakujawa (approved):

PR #2484, opened 2021-05-10
Fix question mark tooltip stying

chinins:
## Description

This PR fixes the alignment of question mark in the voting widget.

In the notion document it is also said that the alignment in staking widget is correct but not in the objection modal. Both of this components use the same component and the alignment is exactly the same (i've also checked visually). But please tell me if you see it differently, I will then correct.

https://www.notion.so/colony/Question-marks-styling-19571a27bcd14959a6952659a3541033

**New stuff** ✨

none

**Changes** 🏗

- Update markup & css of the `VoteDetails` component

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-326


ArmandoGraterol (dismissed):
I'm seeing the tooltips like this :(

FireShot Capture 284 - Colony - localhost
FireShot Capture 285 - Colony - localhost
FireShot Capture 286 - Colony - localhost

Now they all look aligned to the bottom instead of the center

chinins:
:( i guess it depends on the OS & your browser. What are you using?

ArmandoGraterol:
> :( i guess it depends on the OS & your browser. What are you using?

Ubuntu 20.04.2 and Chrome 90.0.4430.93

rdig:
Updated all the places where we use `QuestionMarkTooltip` to look properly aligned vertically.

Please ignore previous reviews as they no longer reflect the code. See my screenshots for the latest version.

chinins (commented):
It's not aligned centrally on my machine 😭

Screenshot 2021-05-24 at 15 56 22

PR #2487, opened 2021-05-11
Default 1 Gwei Gas Price on Xdai

rdig:
## Description

This PR sets the default gas price value on Xdai and Local networks to 1 Gwei as that is a safe value for the transaction to go trough, while consuming the minimal amount of Xdai

**Changes**

- [x] Added immutable `Transaction` default `gasPrice` value
- [x] Updated `TRANSACTION_CREATED` reducer to include gas price and limit


chinins (approved):
All good!

PR #2488, opened 2021-05-11
Add mapping for all motions for `DetailsWidget`

chinins:
## Description

Add mapping for all motions to be used in `Details widget`

**New stuff** ✨

Add mappings, translation strings etc.

**Changes** 🏗

none

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-330


alicjakujawa (approved):

alicjakujawa (commented):
Hmm, i think we also need to map those values in component itself (actionAndEventValues), unless we already have it?

chinins:
They are done via static maps - if I understand you correctly. I will make a note  and test every motion once again once Armando merges his PR that decodes all other motions and will fix if anything missing. But I think everything should be already there.

PR #2489, opened 2021-05-11
Decode motion types and redirect them to default motion

ArmandoGraterol:
Add correct redirection to the motion page for the rest of the motions that we already have up and working

**New stuff** ✨

* Added a `getMotionValues`  function to get the common/default values that every motions  needs
* Added `MotionValues` type to cover the common motion action values
* Decoded motion type for `SetUserRoleMotion`, `CreateDomainMotion`, `EditColonyMotion`, `PaymentMotion`, `EditDomainMotion`, and `MoveFundsMotion`
* Added previously decoded motion types to ActionsPage switch statement

**Changes** 🏗

* Changed MintTokenMotion name to DefaultMotion

Resolves DEV-298


src/types/colonyMotions.ts
```diff
@@ -4,8 +4,20 @@ export enum ColonyMotions {
...
 }
 
 export const motionNameMapping = {
   mintTokens: ColonyMotions.MintTokensMotion,
+  addDomain: ColonyMotions.CreateDomainMotion,
+  editDomain: ColonyMotions.EditDomainMotion,
+  setUserRoles: ColonyMotions.SetUserRolesMotion,
+  moveFundsBetweenPots: ColonyMotions.MoveFundsMotion,
+  editColony: ColonyMotions.ColonyEditMotion,
+  makePaymentFundedFromDomain: ColonyMotions.PaymentMotion,
```
chinins, line 22: We need to coordinate better - i've already done it in my PR :)

chinins (changes requested):
When I do any other motions other than mint tokens I see the title as `Generic motion we don't have information about`. As far as I understand it should be properly mapped to the correct title like mint tokens.

You don't need to do anything for the details widget as it's done in my PR (like names & icons) but i think the titles should be correct. . Below are examples of 3 different motions.

Also, could you please add the dev issue number to your branch names in the future  if it's not a problem for you? Then it's easier to find the issue in linear, to check for details saves a bit of time.

Screenshot 2021-05-14 at 21 29 14

Screenshot 2021-05-14 at 21 28 44



Screenshot 2021-05-14 at 21 27 42

ArmandoGraterol:
> When I do any other motions other than mint tokens I see the title as `Generic motion we don't have information about`. As far as I understand it should be properly mapped to the correct title like mint tokens.
> 
> You don't need to do anything for the details widget as it's done in my PR (like names & icons) but i think the titles should be correct. . Below are examples of 3 different motions.
> 
> Also, could you please add the dev issue number to your branch names in the future if it's not a problem for you? Then it's easier to find the issue in linear, to check for details saves a bit of time.

@chinins From what I understand, that ain't the case. I also had this confusion at the beginning and I had to ask Raul. 

This task, from my understanding, it's about:

1. Renaming the MintToken action page to something generic
2. Decode the action type and determine the motion type
3. Based on if it's a motion or not, redirect to the generic corresponding page (DefaultAction or DefaultMotion)

To get the correct titles for the motion page, I would have to decode the motion's values but this task is only for the decoding of action type, and then the decoding of the motion type.

Now, about putting the issue number in the branch name ... Sure, no problem :)

src/types/colonyMotions.ts
ArmandoGraterol: Tbf, at least it's just this small piece of code (and well, the emojis) that should be repeating :P

chinins (approved):
Ok, I see - thank you for clarification! All good then.

rdig (approved):
I'm approving this, assuming the actual page values will be added in a different PR _(for the page title and details widget)_

PR #2492, opened 2021-05-12
Fix Staking, Voting, Revealing Countdown Timer

rdig:
## Description

This PR attempts to fix the countdown timer displayed on a motion for it's different states: staking, voting, revealing

**WARNING**
This both applies to testing and in production. If no block is being mined **after** the motion countdown finishes, the motion state won't be updated _(on chain)_ that means, that if no transaction go through _(simplest example)_ when the local page's countdown reaches `0` and the refresh triggers, the state will stay the same _(this is actually correct, since this is how the contracts work)_

Luckily, on QA and production, this won't be a issues, since a block will be mined every 15-30 seconds. **The problem** however is that when testing locally, **you must create a new transaction within the last 5 seconds of the countdown timer in order for the automatic update to trigger** otherwise the state of the motion won't change.

See the demo for how this is supposed to work, as well as check out this comment: https://github.com/JoinColony/colonyDapp/blob/fix/dev-345-countdown-timer/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx#L94-L110

**Changes** 

- [x] Refactor core `CountDownTimer`
- [x] Refactor `motionTimeoutPeriods` resolver
- [x] Removed `blockTime` resolver 
- [x] Refactor `splitTimePeriods` util
- [x] Reafactor `ActionPage` component to remove use of `motionStatus` resolver _(was un-needed)_
- [x] Refactor `MintTokenMotion` component to use `motionStatus` resolver to fetch motion status
- [x] Refactor `updateMotionValues` saga helper to use `motionStatus` resolver

**Additional Fixes**

- [x] Fix `FinalizeMotionAndClaimWidget` styles _(when no claim/finalize button is displayed)_

**Demo**

demo-vr-motion-countdown

Resolves DEV-345


src/utils/time/calculateTimeLeft.ts
```diff
@@ -1,22 +1,11 @@
...
-  actionPeriod?: number,
-) => {
-  const difference = actionPeriod
-    ? createdAt +
-      actionPeriod * 1000 -
-      (Date.now() + differenceVsBlockChainTime)
-    : 0;
-
-  if (difference > 0) {
+export const splitTimePeriod = (period: number) => {
```
chinins, line 1: Maybe the file should be rename to `splitTimeLeft` as well for consistency?

chinins (approved):
Great stuff - I really like how you solved issue and simplified some of the stuff. I can appreciate it as it did cause me a lot of pain :)

src/utils/time/calculateTimeLeft.ts
rdig: Sure :+1:

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
```diff
@@ -228,8 +233,9 @@ const MintTokenMotion = ({
       </div>
     ),
   };
-  const motionStyles = MOTION_TAG_MAP[motionState || MotionState.Invalid];
 
+  const motionState = motionStatusData?.motionStatus || MotionState.Invalid;
```
alicjakujawa, line 237: Because of this line, we're seeing for a short while invalid tag when data is loading. I think we need to add loading state to tag component

alicjakujawa (approved):
Two minor things, one in comment, and one:
Screenshot 2021-05-17 at 14 57 14
i know it wasn't scope of this task, but it should be small css update to make it fit in one line, with text-overflow (please also check when progress bar is appearing)

Approving either way

src/modules/dashboard/components/ActionsPage/ActionsComponents/MintTokenMotion.tsx
rdig: Maybe not show it at all until we a "proper" state for it ?

alicjakujawa: Yes, it make even more sense

PR #2493, opened 2021-05-13
Hide claim button if there is nothing to claim

alicjakujawa:
## Description

This PR hide claim button when there is nothing to claim

Resolves DEV-334


rdig (approved):
Good to go!

PR #2495, opened 2021-05-13
Create payment motion saga & update createPayment dialog accordingly

chinins:
## Description

This PR adds motions to `CreatePayment` flow & fixes the reputation check in `EditDomain` dialog.

It uses some new stuff from the network so you need to run provision and rebuild.

**New stuff** ✨

- `createPayment` saga
- toggle, force & other checks in `EditDomain` dialog

**Changes** 🏗

- add `domainId` as an optional argument to `ColonyReputation` query to get domain reputaiton
- add `domainId` to `useColonyReputation` & `useDialogActionPermissions` hooks
- refactor createDomain & createRoot motions with new createMotion
- update permissions checks in `ExpendituresDialog`

**Deletions** ⚰️

none

## TODO

- replace `getExtensionPermissionProofs` function declaration with import from 'colony-js' (if/after Raul exports it in colony-js)

Resolves DEV-302
Resolves DEV- 346
Resolves DEV- 342


src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -242,18 +244,28 @@ const CreatePaymentDialogForm = ({
...
 
-  const canMakePayment =
+  const [userHasPermission, onlyForceAction] = useDialogActionPermissions(
+    colony.colonyAddress,
+    hasRoles,
+    isVotingExtensionEnabled,
+    values.forceAction,
+  );
+
+  const isPaymentExtensionInstalled =
```
alicjakujawa, line 260: We can use useEnabledExtensions hook which have isOneTxPaymentExtensionEnabled flag. This one is not enough either way as we need to make sure that this extension is enabled

src/modules/dashboard/components/ExpendituresDialog/ExpendituresDialog.tsx
```diff
@@ -90,9 +91,11 @@ const ExpendituresDialog = ({
   });
 
   const canMakePayment =
-    colonyExtensionsData?.processedColony?.installedExtensions?.find(
+    (colonyExtensionsData?.processedColony?.installedExtensions?.find(
```
alicjakujawa, line 94: same here

alicjakujawa (changes requested):
Almost there :) Small comments. Also wasn't able to test the part with subdomains, dont know why (and i dont think is connected to you PR) but i couldn't create new domain at all (no errors) Will try to test again on Monday

src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.tsx
chinins: this was already the code in both instances but a very good catch!

ArmandoGraterol (changes requested):
I can't make payments without the voting extension enabled :(

Just made a fresh colony, minted some tokens, and tried to make some payments to get reputation. (I also have all the permissions)

FireShot Capture 308 - Colony - localhost

I think now it's checking that I also have the voting extension to be able to make a payment

src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -242,17 +240,31 @@ const CreatePaymentDialogForm = ({
...
-      ({ extensionId }) => extensionId === Extension.OneTxPayment,
-    ) || false;
+  const [userHasPermission, onlyForceAction] = useDialogActionPermissions(
+    colony.colonyAddress,
+    hasRoles,
+    isVotingExtensionEnabled,
+    values.forceAction,
+  );
+
+  // const isPaymentExtensionInstalled =
```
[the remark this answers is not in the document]
chinins: no, it's leftover code - good catch!

chinins:
Good catch about expenditures dialog! It shouldn't, just fixed it.

src/modules/dashboard/components/ExpendituresDialog/ExpendituresDialog.tsx
```diff
@@ -73,26 +73,25 @@ const ExpendituresDialog = ({
...
   colony,
+  isVotingExtensionEnabled,
   nextStep,
 }: Props) => {
   const { walletAddress, username, ethereal } = useLoggedInUser();
+  const { isOneTxPaymentExtensionEnabled } = useEnabledExtensions({
+    colonyAddress,
+  });
+
+  // const isVotingExtensionEnabled = false;
```
rdig, line 84: I'm assuming this is a leftover debug value

src/modules/dashboard/sagas/motions/createEditDomainMotion.ts
```diff
@@ -79,13 +79,13 @@ function* createEditDomainMotion({
     txChannel = yield call(getTxChannel, metaId);
 
     // setup batch ids and channels
-    const batchKey = 'createDomainMotion';
+    const batchKey = 'createMotion';
 
     const {
-      createDomainMotion,
+      createMotion,
```
rdig, line 85: I mean... the tx channel name didn't necessarily need renaming, just the `methodName` we pass to the transaction.

But with all that being said, I very much appreciate the extra effort :1st_place_medal:

src/modules/dashboard/sagas/motions/paymentMotion.ts
```diff
@@ -24,26 +19,22 @@ import {
...
     ColonyRole.Funding,
-    // address,
+    address,
   );
-
-  const [adminPDID, adminCSI] = yield call(
-    getPermissionProofs,
-    client,
+  const [adminPDID, adminCSI] = await getPermissionProofs(
+    colonyClient,
```
rdig, line 34: I'll put this on my list for tommorow to make a new build of colonyJS that exports this for you.

_(I keep putting it off, as more "pressing" stuff always seems to take it's place. Sorry about that)_

src/modules/dashboard/sagas/motions/paymentMotion.ts
```diff
@@ -138,33 +142,45 @@ function* createPaymentMotion({
...
+      params: [
+        domainId,
+        childSkillIndex,
+        client.address,
+        encodedAction,
+        key,
+        value,
+        branchMask,
+        siblings,
+      ],
```
rdig, line 183: When trying to create this payment in a child domain, this errors out:

Screenshot from 2021-05-18 00-04-43

I'm assuming it doesn't like the `childSkillIndex` value, but this is just a wild guess. I think you might need to bring in Alex to help you on this.

rdig (changes requested):
I rejected this because of the "wrong" value we're passing to the transaction when making a payment from a child domain, causing the transaction to fail.

Other then that, I'll all good with this PR

src/modules/dashboard/sagas/motions/paymentMotion.ts
chinins: Do you have reputation in the domain where you are creating the motion? I think this is the reason why.
I can't replicate. I've done several tests and it works every time except when I don't have reputations in that domain.

But it's a good catch because I forgot to add the reputation check for domain reputation.

chinins: i've fixed it the reputation check and am pushing the changes.

src/modules/dashboard/components/ExpendituresDialog/ExpendituresDialog.tsx
chinins: of course :)

ArmandoGraterol (approved):
Looks good to me :ok_hand:

src/modules/dashboard/components/ExpendituresDialog/ExpendituresDialog.tsx
```diff
@@ -73,26 +73,23 @@ const ExpendituresDialog = ({
...
-    variables: { address: colonyAddress },
-  });
+    ((canAdminister(allUserRoles) && canFund(allUserRoles)) ||
+      isVotingExtensionEnabled);
 
-  const canMakePayment =
-    colonyExtensionsData?.processedColony?.installedExtensions?.find(
-      ({ extensionId }) => extensionId === Extension.OneTxPayment,
-    ) || false;
+  const canMakePayment = isOneTxPaymentExtensionEnabled || false;
```
rdig, line 92: Hmmm..... I don't think this is correct

Either that `false` was added for debugging purposes, or it doesn't have a place in there

rdig, line 92: Also, this variable naming is very bad, since we already have `canCreatePayment`, and this being called `canMakePayment` is very confusing....

rdig (approved):
Seems to be working as expected now, however a better test will be when all the motion pages will be wired in.

Also, have a look into the `canMakePayment` logic check as I don't think that is correct, at least not entirely.

src/modules/dashboard/components/ExpendituresDialog/ExpendituresDialog.tsx
chinins: it was there before but I have removed it now.

chinins: agree, update that too

PR #2496, opened 2021-05-13
Fix slider limit

ArmandoGraterol:
Fixed slider limit, should now actually stop at the limit of staking

Resolves DEV-339

### DIFF
--- a/src/modules/core/components/Slider/Slider.tsx
+++ b/src/modules/core/components/Slider/Slider.tsx
@@ -63,9 +63,9 @@ const Slider = ({
           onChange(val);
         }
       }
-      if (limit && sliderValue > limit) {
+      if (limit && (sliderValue > limit || val > limit)) {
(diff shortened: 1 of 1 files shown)


alicjakujawa (changes requested):
I think something is still wrong, see screen:
Screenshot 2021-05-14 at 12 51 22
I didint stake, but from what i see, i cannot stake, but limit is showing that max stake for me is "something"

But actually i cannot stake, so the limit should be 0

alicjakujawa:
And addiitonal screen, it can be of course some issues with slider disabled logic

alicjakujawa (approved):

alicjakujawa:
Approving, because bugs i mentioned wasn't introduce here

chinins (approved):
I have the same problem as Alicja but if that should be dealt with in a separate PR I am approving

PR #2501, opened 2021-05-14
Improve Token Activation popover styles

ArmandoGraterol:
Improved some of the styles of the Token Activation popover

* Aligned token avatar to the center
* Updated padding and margin so that it looks more like the specs
* Updated max button font weight
* Improved alignment of Confirm button and wallet address button

FireShot Capture 305 - Colony - localhost

Resolves DEV-327

### DIFF
--- a/src/modules/core/components/Fields/Input/InputComponent.css
+++ b/src/modules/core/components/Fields/Input/InputComponent.css
@@ -184,6 +184,7 @@
   border: none;
   background: none;
   font-size: var(--size-tiny);
+  font-weight: var(--weight-bold);
   color: var(--action-secondary);
   cursor: pointer;
 }
--- a/src/modules/pages/components/RouteLayouts/UserNavigation.css
+++ b/src/modules/pages/components/RouteLayouts/UserNavigation.css
@@ -96,6 +96,8 @@
 }
 
 .walletAddress {
+  padding: 2px 9px 3px;
+
   /*
    * @NOTE We need to do border sizing shenanigans due to how the border color
    * is required by spec to have opacity.
@@ -106,7 +108,7 @@
    * active class (the one that has a different colored border), we remove the extra padding
    * and set the border
    */
-  padding: 2px 9px 3px;
+  height: 100%;
   border: none;
   border-radius: var(--radius-normal);
   background-color: color-mod(var(--temp-grey-blue-7) alpha(10%));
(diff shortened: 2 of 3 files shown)


alicjakujawa (commented):
What designs you was using for updating styles in here? https://www.figma.com/file/m96yJW1PPXezFl2mkuYha6/Token-Activation?node-id=1275%3A0 i see here, that popover looks a bit different (for example "Staked" aligment) and some other things

ArmandoGraterol:
> What designs you was using for updating styles in here? https://www.figma.com/file/m96yJW1PPXezFl2mkuYha6/Token-Activation?node-id=1275%3A0 i see here, that popover looks a bit different (for example "Staked" aligment) and some other things

I was using this one: https://www.figma.com/file/m96yJW1PPXezFl2mkuYha6/Token-Activation?node-id=1689%3A7694

As that is the one I see marked as a "Reference"

chinins (commented):
What about the 2 buttons: activate and deactivate? i think there is the same problem of vertical text align as was with the confirm button

ArmandoGraterol:
> What about the 2 buttons: activate and deactivate? i think there is the same problem of vertical text align as was with the confirm button

Hmmm, that's true. 

They only mentioned the Confirm button so I tunneled down on that. I will adjust the other 2 buttons tomorrow

rdig (changes requested):
I'm mostly good with this, but:

1. Confirm button looks correct:
Screenshot from 2021-05-19 00-14-36

2. However the Activate/Deactivate buttons do not look aligned proper:
Screenshot from 2021-05-19 00-14-39

3. Shouldn't the staked tokens indicator be aligned to the left on the same column as the other two (active / inactive)
Screenshot from 2021-05-19 00-14-46

ArmandoGraterol:
> I'm mostly good with this, but:
> 
> 1. Confirm button looks correct:
>    Screenshot from 2021-05-19 00-14-36
> 2. However the Activate/Deactivate buttons do not look aligned proper:
>    Screenshot from 2021-05-19 00-14-39
> 3. Shouldn't the staked tokens indicator be aligned to the left on the same column as the other two (active / inactive)
>    Screenshot from 2021-05-19 00-14-46

3.Well, in the reference screen the yellow dot is aligned with the "Activate" and "Deactivate" text ... I guess they want to convey that it's a subcategory of active tokens?

chinins (approved):
Looks good to me! Indeed, the yellow dot was supposed to be moved to the right from the very beginning.

alicjakujawa (approved):
Screenshot 2021-05-20 at 15 29 05
Small nitpick - just make sure that title is in line with avatar, but apart of this im approving 🔥

PR #2504, opened 2021-05-17
Fix token activation displays of locked tokens

ArmandoGraterol:
- Fixed locked token display using token balance(active tokens) as locked tokens
- Fixed not being able to deactivate ANY token when you have locked tokens

Resolves DEV-350

### DIFF
--- a/src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
+++ b/src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
@@ -94,10 +94,14 @@ const ChangeTokenStateForm = ({
     [isActivate],
   );
 
-  const maxAmount = useMemo(() => moveDecimal(tokenBalance, -tokenDecimals), [
-    tokenDecimals,
-    tokenBalance,
-  ]);
+  const maxAmount = useMemo(
+    () =>
+      moveDecimal(
+        tokenBalance.sub(isActivate ? 0 : lockedTokens),
+        -tokenDecimals,
+      ),
+    [tokenDecimals, tokenBalance],
+  );
 
   const transform = useCallback(
     pipe(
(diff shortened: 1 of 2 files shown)


alicjakujawa (changes requested):
Screenshot 2021-05-19 at 08 50 49
There is one more thing to fix in this PR
I want deactivate my "active tokens" but like you see in screen, confirm button is still disabled

ArmandoGraterol:
> Screenshot 2021-05-19 at 08 50 49
> 
> There is one more thing to fix in this PR I want deactivate my "active tokens" but like you see in screen, confirm button is still disabled

@alicjakujawa That's intentional according to Product, you shouldn't be able to deactivate tokens if you have locked tokens

rdig:
> @alicjakujawa That's intentional according to Product, you shouldn't be able to deactivate tokens if you have locked tokens

Shouldn't then max just fill the value for `active` minus `staked`, that would be the _proper_ max tokens that can be deactived

ArmandoGraterol:
> > @alicjakujawa That's intentional according to Product, you shouldn't be able to deactivate tokens if you have locked tokens
> 
> Shouldn't then max just fill the value for `active` minus `staked`, that would be the _proper_ max tokens that can be deactived

@rdig  Hmmm, I think that when they said that "you can't deactivate tokens if you have locked tokens" they meant that if for example you have 100 active tokens, and 20 of those 100 active tokens are locked then you can't even deactivate a single one

So, even if we subtracted the locked tokens from the max amount you still shouldn't be able to deactivate them. 

Perhaps @chinins can chime in into this, as I believe she was the one who implemented the conditional that disables the buttons in case of the user having locked tokens while on deactivate (`!isActivate && hasLockedTokens`). But I think that's how it should work

chinins (approved):
Looks good!

alicjakujawa:
@ArmandoGraterol why this is still open?

alicjakujawa (approved):
Good to merge

PR #2506, opened 2021-05-19
Adjust displaying rewards updated colonyNetwork

alicjakujawa:
## Description

This PR updates the way we display rewards. From now, in voting state we're not able to display exact value, we only displaying range. When we are in Reveal state - we display exact value. 

**New stuff** ✨

* Clny token icon

**Changes** 🏗

* Updated Numeral component with "darken" theme
* Updated VoteDetails Widget to display rewards value based on state
* Updated motionVoterReward resolver
* Updated colonyNetwork version
* Updated colonyjs version

reveal
vote

Resolves DEV-358


src/modules/core/components/Numeral/Numeral.css
```diff
@@ -50,10 +50,17 @@
   }
 }
 
+/* I see that in designs Dark is actually --dark not --grey-3 so maybe we
+should think about replacing this color here instead
+of adding one more theme option */
```
chinins, line 55: Is `dark` theme used for Numeral a lot? I think it's a good idea not  to add too many themes. Maybe ask product about that? Usually, it's just an oversight on their part and they are happy to have less shades.

src/modules/core/components/Numeral/Numeral.tsx
```diff
@@ -7,7 +7,7 @@ import { formatTokenValue } from '~utils/numbers';
 import styles from './Numeral.css';
 
 interface Appearance {
-  theme: 'primary' | 'grey' | 'dark' | 'blue';
+  theme: 'primary' | 'grey' | 'dark' | 'blue' | 'darken';
```
chinins, line 10: Why is the name of the theme `darken`? It's nitpicking but it sounds bizarre because I think it's a verb. Maybe we can use `darker` instead?

chinins (approved):
Looks good to me, just one little naming thing to look at.

src/data/resolvers/motions.ts
```diff
@@ -425,11 +426,37 @@ export const motionsResolvers = ({
...
         if (data?.userReputation) {
-          const reward = await votingReputationClient.getVoterReward(
-            bigNumberify(motionId),
-            bigNumberify(data.userReputation),
-          );
-          return reward.toString();
+          let reward = bigNumberify(0);
+          let minReward = bigNumberify(0);
+          let maxReward = bigNumberify(0);
+          if (state === NetworkMotionState.Submit) {
```
rdig, line 432: Why are you checking the state of the motion?

I don't believe the voter reward changes when the motion state changes? Does it?

src/data/resolvers/motions.ts
```diff
@@ -425,11 +426,37 @@ export const motionsResolvers = ({
...
+                bigNumberify(data.userReputation),
+                createAddress(userAddress),
+              );
+              minReward = minUserReward;
+              maxReward = maxUserReward;
+            } catch (error) {
+              /* Not ideal, but when we not logged in useLoggedInUser is returning wrong userAddress
+               When we pass wrong address to getVoterRewardRange method it returning error.
+              */
+            }
```
rdig, line 448: This should be handled on the contracts side. We should be able to call this with any valid wallet address and get some value back, not just a user that voted. Otherwise, this can be used to discover who voted and who didn't.

Please tell Alex about this.

src/modules/core/components/Numeral/Numeral.tsx
rdig: I agree with Olga here, `darken` is a bad name.

Maybe come up with something that is not necessarily related to the shade. Maybe something like: `NightGlow` or something equally ridiculous :)

src/modules/dashboard/components/ActionsPage/VoteWidget/VoteDetails.tsx
```diff
@@ -152,12 +160,54 @@ const VoteDetails = ({
               </div>
             </div>
             <div className={styles.value}>
-              {voterReward?.motionVoterReward && (
-                <Numeral
-                  unit={getTokenDecimalsWithFallback(nativeToken?.decimals)}
-                  value={voterReward.motionVoterReward}
-                  suffix={` ${nativeToken?.symbol}`}
-                />
+              {voterReward?.motionVoterReward && !loadingVoterReward && (
```
rdig, line 163: Please don't put the `loading` variable into this check.

Add a loading state to this component, using the `MiniSpinnerLoader` like we do in most other widgets.

Here's and example: https://github.com/JoinColony/colonyDapp/blob/feature/dev-358-rewards/src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx#L202-L216

rdig (approved):
Functionality wise this is good. I like most of the logic in the code as well. However, I've left some comments that I want you to address.

Pre-approving this.

Screenshot from 2021-05-21 16-46-21

Screenshot from 2021-05-21 16-46-30

ArmandoGraterol (approved):
Looks good to me!

I agree that "darken" should have another name tho as it's a verb, not an adjective/name.

Maybe "DarkAlicja"? :P

src/modules/core/components/Numeral/Numeral.tsx
alicjakujawa: because i didint have any other idea xD

src/data/resolvers/motions.ts
alicjakujawa: because i don't want to call function when its not needed?  performance wise

alicjakujawa:
@ArmandoGraterol @rdig @chinins love that you all hate this name 😆 i hate it too, but its because of too many "kind" of themes, instead of being consistent. I didint have any better idea so took stupid one. i will ask product to either stick to current "Dark" or update "dark" to be new dark.

alicjakujawa:
Updated dark theme to use --dark variable, as i didint see any usage of "dark" theme in codebase

PR #2508, opened 2021-05-20
Fix countdown timer not updating

ArmandoGraterol:
Fixed countdown timer not updating their periods after a change of state in the motion

Resolves DEV-357

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
+++ b/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
@@ -70,6 +70,7 @@ const CountDownTimer = ({
       colonyAddress,
       motionId,
     },
+    notifyOnNetworkStatusChange: true,
   });
 
   const currentStatePeriod = useCallback(() => {
@@ -147,7 +148,7 @@ const CountDownTimer = ({
     if (state && data) {
       refetch();
     }
-  }, [state]);
+  }, [state, data, refetch]);
 
   /*
    * Split the time into h/m/s for display purpouses
--- a/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
+++ b/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
@@ -166,7 +166,7 @@ const CountDownTimer = ({
     if (data && !isStakingPhaseState) {
       refetch();
     }
-  }, [state, data, refetch]);
+  }, [isStakingPhaseState, data, refetch]);
 
   /*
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
```diff
@@ -143,6 +143,12 @@ const CountDownTimer = ({
     walletAddress,
   ]);
 
+  useEffect(() => {
+    if (state && data) {
+      refetch();
+    }
+  }, [state]);
```
rdig, line 150: This needs to have both `data` and `refetch` as dependencies! Please don't ignore `eslint` rules

rdig, line 150: So this is technically working, however, when the state changes, there is a brief flash where the time is still shows wrong (adding staking + voting time)

Here's an example:
bug-vr-counter-wrong-time-flash

In my example, the timeout is set to 1h. As you can see when the state changes to Reveal, for a brief second it shows as 2h (voting time + reveal time), then, after the `refetch()` call finishes, it shows the proper time.

rdig (changes requested):
This is almost good. It shows the correct time, but when the motion state changes, there  is a brief flash where the time shown is wrong.

While locally this is not a big issue, on production, where everything is slow, this will be more visible.

chinins (approved):
Looks good to me

alicjakujawa:
@ArmandoGraterol @rdig what is the state of this?

rdig:
Why am I tagged in this ?

rdig:
@alicjakujawa why did you tag me for this ?

alicjakujawa:
> @alicjakujawa why did you tag me for this ?

Because you are as "pending reviewer"

rdig:
I mean... if it's ready I'll gladly review it, but I have no idea what the state of this is

ArmandoGraterol:
@alicjakujawa @rdig This is ready for review! :)

rdig (changes requested):
So after staking (but not so much that the state changes -- imagine just staking 3 out of 10), the counter gets reset.

This is incorrect, since the counter just counts down until the state changes.

Most likely to fix this you'll need to modify the resolver.

alicjakujawa (changes requested):
Unfortunately im still experiencing this bug. When im vote and it will go to reveal state, im seeing voting state time + reveal state time untill i will refresh page.

alicjakujawa (approved):

chinins (approved):
Seems to be working properly!

PR #2510, opened 2021-05-21
Slider staking limit visual, logic & explanation messages

chinins:
## Description

This refactors the logic for slider limit visuals, logic & not enough reputation/tokens messages.

**New stuff** ✨

nothing new

**Changes** 🏗

- refactor `Slider` limit logic & visuals
- simplify `StakingValidationError`
- add `StakingValidationError` for tokens to `StakingSlider`
- changed staking slider logic to calculate % so that the visual corresponds to the numbers shown
- simplified `StakingWidget`

**Deletions** ⚰️

none

## TODO

Potentially extract logic for % calculation to a separate function

Resolves DEV-317

Not sure whether to show no reputation/ no tokens messages when user is logged off (for now showing them):

Screenshot 2021-05-21 at 12 28 04

Demo
stake-limit

When a person is logged in, has active tokens but doesn't have reputation to stake
Screenshot 2021-05-21 at 12 30 28

### DIFF
--- a/src/modules/core/components/Slider/Slider.css
+++ b/src/modules/core/components/Slider/Slider.css
@@ -22,6 +22,6 @@
   Removes `left` property from the dot for the correct visual position.
   I tried many options but that's the only one I found working
 */
-.primary > div > div > span {
+.primary :global(.rc-slider-dot) {
   left: 0 !important;
 }
--- a/src/modules/core/components/Slider/Slider.css
+++ b/src/modules/core/components/Slider/Slider.css
@@ -18,7 +18,10 @@
   box-shadow: 0 0 5px var(--danger);
 }
 
-/* This is to remove `left` from the dot for it have correct visual position */
+/*
+  Removes `left` property from the dot for the correct visual position.
+  I tried many options but that's the only one I found working
+*/
 .primary > div > div > span {
   left: 0 !important;
 }
--- a/src/modules/core/components/Slider/Slider.css
+++ b/src/modules/core/components/Slider/Slider.css
@@ -17,3 +17,8 @@
 .danger :global(.rc-slider-handle:active) {
   box-shadow: 0 0 5px var(--danger);
 }
+
+/* This is to remove `left` from the dot for it have correct visual position */
+.primary > div > div > span {
+  left: 0 !important;
+}
--- a/src/modules/core/components/Slider/Slider.tsx
+++ b/src/modules/core/components/Slider/Slider.tsx
@@ -54,10 +54,15 @@ const Slider = ({
     return limit ? limit * 100 : 0;
   }, [limit]);
 
+  const gradientPercentage = useMemo(
+    () => (limitValue >= 100 ? 100 : limitValue),
+    [limitValue],
+  );
+
   const onSliderChange = useCallback(
     (val): void => {
       if (
-        (limitValue && sliderValue < limitValue) ||
+        (limit !== undefined && sliderValue < limitValue) ||
         val < sliderValue ||
         !limitValue
       ) {
@@ -67,7 +72,10 @@ const Slider = ({
           onChange(val);
         }
       }
-      if (limitValue && (sliderValue > limitValue || val > limitValue)) {
+      if (
+        limit !== undefined &&
+        (sliderValue > limitValue || val > limitValue)
+      ) {
         setSliderValue(limitValue);
         setValue(limitValue);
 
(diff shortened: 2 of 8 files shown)


src/modules/core/components/Slider/Slider.css
```diff
@@ -17,3 +17,11 @@
...
   box-shadow: 0 0 5px var(--danger);
 }
+
+/*
+  Removes `left` property from the dot for the correct visual position.
+  I tried many options but that's the only one I found working
+*/
+.primary > div > div > span {
+  left: 0 !important;
+}
```
rdig, line 27: I think this also works:
```suggestion
.primary :global(.rc-slider-dot) {
  left: 0 !important;
}
```

The advantages of my version, is that you can be sure it will still work even if some updates will change the html element structure

src/modules/core/components/Slider/Slider.css
```diff
@@ -17,3 +17,11 @@
 .danger :global(.rc-slider-handle:active) {
   box-shadow: 0 0 5px var(--danger);
 }
+
+/*
+  Removes `left` property from the dot for the correct visual position.
+  I tried many options but that's the only one I found working
+*/
```
rdig, line 24: Appreciate leaving a comment here, because of your use of `!important`

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -50,30 +52,48 @@ const Slider = ({
     }
   }, [sliderValue, value, setSliderValue]);
 
-  const gradientStopPercentage = useMemo(() => {
-    return limit ? Math.round((limit / max) * 100) : 0;
-  }, [limit, max]);
+  const limitValue = useMemo(() => {
+    return limit ? limit * 100 : 0;
```
rdig, line 56: Limit might not always be `100`, I believe this should use the `max` value

src/modules/core/components/Slider/Slider.tsx
```diff
@@ -50,30 +52,48 @@ const Slider = ({
...
+  }, [limit]);
+
+  const gradientPercentage = useMemo(
+    () => (limitValue >= 100 ? 100 : limitValue),
```
rdig, line 60: Same here

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
```diff
@@ -66,8 +68,15 @@ const StakingSlider = ({
   minUserStake,
   canUserStake,
   appearance,
+  userActivatedTokens,
   isObjection,
-}: Props) => {
+}: // userInactivatedTokens,
```
rdig, line 73: Leftover comment

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
```diff
@@ -119,14 +135,18 @@ const StakingSlider = ({
         <Slider
           name="amount"
           value={values.amount}
-          limit={parseFloat(userStakeLimitPercentage.toFixed(2))}
+          limit={parseFloat(userStakeLimitPercentage.toFixed(5))}
```
rdig, line 138: Why do you pass here 5 decimals after the point, when all across the slider we use 2 ?

Just look at the `step` value, that's what dictates the number of decimal points

rdig (approved):
In testing it seems to be working as expected, and I'm also good with most of the code.

I'm approving it, but please make sure to address the things I've pointed out in the comments.

ArmandoGraterol (approved):
Works like a charm!

src/modules/core/components/Slider/Slider.tsx
chinins: Limit is done in %, so it's passed like that: 0.36 and we need to multiply by 100 to get 36%. I don't think it should be max value here as it doesn't depend on it.

src/modules/core/components/Slider/Slider.tsx
chinins: same here

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
chinins: Otherwise it will be too imprecise - the limit value will be less or more than the activated tokens. This is just for the more precise limit value and it actually in my opinion supplies proper functionality.
In this example below (changed toFixed(2)), I have 50 activated tokens, but the limit is calculated as 50.1 and it goes passed the number of my activated tokens on the slider.
Screenshot 2021-05-24 at 11 31 02

There might be other instances where it will be less and then the user won't be able to stake all of his activated tokens.

src/modules/core/components/Slider/Slider.tsx
rdig: Yes, but you don't know that it will be always a percent. _(Hence this is `core` component)_

So if sometime down the line we end up using this in some different scenario, we might use it with different numbers and not a percent based value.

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
rdig: Sure

PR #2512, opened 2021-05-25
Add incompatible extension warning

alicjakujawa:
## Description

This PR adds in incompatible extension warning to extension page

**New stuff** ✨

* Warning core component

**Changes** 🏗

* Refactored ExtensionDetails to show extension warning (currently with temp flag is true)
* Refactored ExtensionDetails to use new Warning component also for modal content warning
* Displayed warning icon next to installed extension label

Screenshot 2021-05-25 at 12 01 54

Screenshot 2021-05-25 at 12 34 58

Resolves DEV-364

### DIFF
--- a/src/modules/core/components/Warning/Warning.css
+++ b/src/modules/core/components/Warning/Warning.css
@@ -1,0 +1,14 @@
+.warning {
+  display: flex;
+  flex-direction: column;
+  padding: 25px 34px;
+  border-radius: var(--radius-large);
+  background-color: var(--pink);
+  font-weight: var(--weight-bold);
+  color: rgba(255, 255, 255);
+}
+
+.warning button {
+  align-self: flex-end;
+  margin-top: 30px;
+}
--- a/src/modules/core/components/Warning/Warning.css
+++ b/src/modules/core/components/Warning/Warning.css
@@ -1,14 +1,18 @@
 .warning {
   display: flex;
   flex-direction: column;
-  padding: 25px 34px;
+  padding: 12px;
   border-radius: var(--radius-large);
   background-color: var(--pink);
   font-weight: var(--weight-bold);
   color: rgba(255, 255, 255);
 }
 
+.content {
+  padding: 13px 22px;
+}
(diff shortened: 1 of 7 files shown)


src/modules/core/components/Warning/Warning.css
```diff
@@ -0,0 +1,14 @@
+.warning {
+  display: flex;
+  flex-direction: column;
+  padding: 25px 34px;
```
ArmandoGraterol, line 4: I think this padding should be `12px` , and then the text inside the container should have its own padding, so something like `padding: 13px 22px 0 22px`. 

This is because I noticed that in the design the button doesn't have the same padding as the text

ArmandoGraterol (approved):
LGTM :ok_hand:  Just noticed a small thing, and well, I'm assuming that flag will be replaced for the real deal at a later date.

src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
```diff
@@ -184,6 +207,9 @@ const ExtensionDetails = ({
     !(extesionCanBeInstalled || extesionCanBeEnabled) &&
     latestNetworkExtensionVersion > extension.currentVersion;
 
+  // TEMP flag
+  const shouldShowWarning = false;
```
chinins, line 211: In your PR description you say that the temp flag should be true. Which version is correct?

chinins (approved):
Do I understand correctly that the check will be wired in in a different branch/PR and for now you are using a temp flag? From the name of the PR I was just expecting to see the check wired in. So want to just confirm that.

Other than that it works & looks good.

src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
rdig: Yes, I will wire this via https://linear.app/colony/issue/DEV-365/wire-up-incompatible-extension-warning

This issue is UI only

PR #2514, opened 2021-05-25
Use Decimal to get token values for action/motion page

ArmandoGraterol:
Added `getFormattedTokenValue` util function to action/motion page that uses Decimal in order to correctly get the decimals of the displayed values

Resolves DEV-333


src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -313,7 +316,10 @@ const ActionsPageEvent = ({
   const colonyNativeToken = tokens.find(
     ({ address }) => address === nativeTokenAddress,
   );
-
+  const decimalStakeAmount = getFormattedTokenValue(
+    values?.stakeAmount || 0,
+    getTokenDecimalsWithFallback(colonyNativeToken?.decimals),
```
alicjakujawa, line 321: i think it would be better to pass just decimals in all of those places and use getTokenDecimalsWithFallback in getFormattedTokenValue util itself

src/utils/tokens.ts
```diff
@@ -46,3 +47,17 @@ export const getTokenDecimalsWithFallback = (
...
   return DEFAULT_TOKEN_DECIMALS;
 };
+
+export const getFormattedTokenValue = (
+  value: BigNumberish,
+  decimals: any,
+): string => {
+  const tokenDecimals = getTokenDecimalsWithFallback(decimals);
+  const safeDecimals = bigNumberify(10)
+    .pow(getTokenDecimalsWithFallback(tokenDecimals))
```
alicjakujawa, line 57: Im lost a bit, why you fire getTokenDecimalsWithFallback twice?

alicjakujawa (changes requested):

src/utils/tokens.ts
ArmandoGraterol: My bad, this slipped my mind and I forgot to remove it

alicjakujawa (approved):

PR #2515, opened 2021-05-26
Remove staking error messages when user is not logged in

chinins:
## Description

This PR adds another check to staking error messages in order to show them only when the user is logged in.

**New stuff** ✨

none

**Changes** 🏗

- added `hasRegisteredProfile` check to staking error messages

**Deletions** ⚰️

none

## TODO

none

Resolves DEV-369
Screenshot 2021-05-26 at 10 58 14


alicjakujawa (changes requested):
When wallet is connected we should see errors, why we cannot stake  (even if we dont have profile)

alicjakujawa (approved):
🥇

ArmandoGraterol (approved):
LGTM :ok_hand:

PR #2516, opened 2021-05-26
Fix edit/create domain motion

alicjakujawa:
## Description

This PR fixes create/edit domain motion. When we didint pass domain purpose, we wasn't able to see Action Page.


chinins (approved):
Good to go!

PR #2518, opened 2021-05-26
Add 1 to required stake

alicjakujawa:
## Description

This PR fixes not being able to fully staked on motion. From time to time, our staking buttons was showing "fully staked" even if motion was still in stake required state. It was because super small amount of stake was missing. 

**Changes** 🏗

* Added 1 to required stake

Resolves DEV-366


rdig (approved):
Good to go

PR #2519, opened 2021-05-26
Add missing motion titles to the action list

ArmandoGraterol:
## Description

- Add `action.title` for the just recently wire up motions

**New stuff** ✨

* Added handling for the wire up motions in the `args` resolver
* Added messages for motions in action list

**Changes** 🏗

* Added `getUpdatedDecodedMotionRoles` for reusability puposes
* Updated useFormatRolesTitles to get a more dynamic title id
*Fixed motion tag being clipped on action list

FireShot Capture 328 - Colony - localhost

Resolves DEV-370

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -973,6 +973,12 @@ export const motionsResolvers = ({
         };
       }
 
+      if (actionValues.name === 'addDomain') {
+        return {
+          metadata: actionValues.args[3],
+        };
+      }
+
       const tokenAddress = colonyClient.tokenClient.address;
       const {
         symbol,
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -920,11 +920,39 @@ export const motionsResolvers = ({
       const actionValues = colonyClient.interface.parseTransaction({
         data: action,
       });
+
+      if (!actionValues) {
+        const oneTxPaymentClient = await colonyManager.getClient(
+          ClientType.OneTxPaymentClient,
+          colonyAddress,
+        );
+
+        const paymentValues = oneTxPaymentClient.interface.parseTransaction({
+          data: action,
+        });
+
+        const tokenClient = await colonyManager.getTokenClient(
+          paymentValues.args[5][0],
+        );
+        const { symbol, decimals } = await tokenClient.getTokenInfo();
+
+        return {
+          amount: paymentValues.args[6][0].toString(),
+          recipient: paymentValues.args[4][0],
+          token: {
+            id: paymentValues.args[5][0],
+            symbol,
+            decimals,
+          },
+        };
+      }
+
       const tokenAddress = colonyClient.tokenClient.address;
       const {
         symbol,
         decimals,
       } = await colonyClient.tokenClient.getTokenInfo();
+
       /*
        * @TODO Return argumnents for the other motions as well, as soon
        * as they get wired into the dapp
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -1011,16 +1011,13 @@ export const motionsResolvers = ({
         };
       }
 
+      // MintTokenMotion - default
       const tokenAddress = colonyClient.tokenClient.address;
       const {
         symbol,
         decimals,
       } = await colonyClient.tokenClient.getTokenInfo();
 
-      /*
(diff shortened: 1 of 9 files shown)


src/i18n/en-actions.ts
```diff
@@ -6,19 +6,27 @@ const actionsMessageDescriptors = {
...
+      ${ColonyMotions.PaymentMotion} {Pay {recipient} {amount} {tokenSymbol}}
       ${ColonyActions.MoveFunds} {Move {amount} {tokenSymbol} from {fromDomain} to {toDomain}}
+      ${ColonyMotions.MoveFundsMotion} {Move {amount} {tokenSymbol} from {fromDomain} to {toDomain}}
       ${ColonyActions.MintTokens} {Mint {amount} {tokenSymbol}}
       ${ColonyMotions.MintTokensMotion} {Mint {amount} {tokenSymbol}}
       ${ColonyActions.CreateDomain} {New team: {fromDomain}}
+      ${ColonyMotions.CreateDomainMotion} {New team: {fromDomain}}
       ${ColonyActions.VersionUpgrade} {Upgrade Colony to Version {newVersion}!}
       ${ColonyActions.ColonyEdit} {Colony details changed}
+      ${ColonyMotions.ColonyEditMotion} {Change colony details}
```
chinins, line 18: It's a nitpick but why the copies of action and motion for edit colony different? Shouldn't they be the same?

src/i18n/en-actions.ts
```diff
@@ -6,19 +6,27 @@ const actionsMessageDescriptors = {
...
       ${ColonyActions.EditDomain} {{fromDomain} team details edited}
+      ${ColonyMotions.EditDomainMotion} {Edit {fromDomain} team details}
```
chinins, line 20: Same for edit domain

chinins (approved):
Looks good!

src/i18n/en-actions.ts
ArmandoGraterol: For edit domain and edit colony it doesn't sit well with me to use the original because they are in the past. 

You create an action and the change is instantly done, the details were _changed_ or the colony was _edited_. It's rightfully so in the past, the change/edit was already made the moment you created the action.

But with the motion versions, you create a motion to _edit_ domain or colony, you haven't edited the domain or colony until and _if_ the motion passes, so for me it doesn't make sense to use "changed" or "edited" like in the action version because the change hasn't been made.

This doesn't happen with the other action/motion titles because the others aren't in the past, only these 2.

Does that make sense to you? Or am I crazy? :D

src/i18n/en-actions.ts
ArmandoGraterol: Explanation above :sunglasses:

src/i18n/en-actions.ts
chinins: I didn't think about it initially, but it makes sense

src/data/resolvers/motions.ts
```diff
@@ -920,15 +921,103 @@ export const motionsResolvers = ({
...
         data: action,
       });
+
+      // PaymentMotion
+      if (!actionValues) {
+        const oneTxPaymentClient = await colonyManager.getClient(
+          ClientType.OneTxPaymentClient,
+          colonyAddress,
+        );
+
```
alicjakujawa, line 931: its quite a big assumption, because we cant be sure that if there is no actionValues above it means that its PaymentMotion. Did you check if there is any other way to find out if this is actually PaymentMotion? what action include?

alicjakujawa (changes requested):

src/data/resolvers/motions.ts
ArmandoGraterol: Hmmm, from what I saw I don't remember seeing something that could indicate if this is a PaymentMotion other than the actionValues coming null/undefined but I'll check again

src/data/resolvers/motions.ts
```diff
@@ -920,15 +921,103 @@ export const motionsResolvers = ({
...
+        const paymentValues = oneTxPaymentClient.interface.parseTransaction({
+          data: action,
+        });
```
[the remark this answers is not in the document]
ArmandoGraterol: How would that look like? Return some default `null`/ 0 values as we do for in `resolvers/colonyActions`?

alicjakujawa (approved):
Good to merge!

PR #2521, opened 2021-05-27
Bug/dev 371 voting extension params

ArmandoGraterol:
Fixed not being able to use decimal percentages, and added extra validation to period and percentage fields

Resolves DEV-371


chinins (commented):
Is it decided somewhere that we can't put any periods for less than an hour?  It seems a bit restrictive but maybe I missed some communication with the product?

ArmandoGraterol:
> Is it decided somewhere that we can't put any periods for less than an hour? It seems a bit restrictive but maybe I missed some communication with the product?

In this PR it's not allowed due to decimals not being a valid value for period params, but now that you mentioned it... Knowing this limitation, I should go ask product if they want to change the period fields from hours to minutes

alicjakujawa:
@ArmandoGraterol did you clarify things with product? is this ready for review?

ArmandoGraterol:
> @ArmandoGraterol did you clarify things with product? is this ready for review?

Not yet, still talking about it in the product channel

ArmandoGraterol:
@alicjakujawa @chinins Ready for re-review! Added changes according to what we concluded in the product channel

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
```diff
@@ -76,11 +76,13 @@ const ExtensionSetup = ({
...
-              formattedPayload[paramName] = payload[paramName] * 3600; // Seconds in 1 hour
+              formattedPayload[paramName] = new Decimal(payload[paramName])
+                .mul(3600) // Seconds in 1 hour
+                .toFixed(0, Decimal.ROUND_HALF_UP);
             } else {
-              formattedPayload[paramName] = bigNumberify(
-                payload[paramName],
-              ).mul(bigNumberify(10).pow(16));
+              formattedPayload[paramName] = new Decimal(payload[paramName])
+                .mul(new Decimal(10).pow(16))
```
chinins, line 84: Why did you use the `Decimal` instead of `bigNumberify` here? I thought that the problem was just for the periods.

chinins (commented):
It's working well, I just have one question (please, have a look at my comment)

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
ArmandoGraterol: We also use `Decimal` here due to one of the reasons we use it above,  because `bigNumberify` doesn't accept decimal values as valid values, so if someone decided to input `50.5%` then that would throw an error

rdig (approved):
Yeah this is all good

PR #2524, opened 2021-05-28
Set domain color to null if undefined

alicjakujawa:
## Description

This pr fixes problem with action page not loading correctly when no color was set up. We need to return null instead of undefined for graphql to be happy


chinins (approved):
Good to go!

PR #2526, opened 2021-05-31
Fix `Passed` tag

chinins:
## Description

This PR fixes the `Passed` tag on the motions page

**Changes** 🏗

- subtract `1` from `requiredStake` in `getMotionState` function

Resolves DEV-374

### DIFF
--- a/src/utils/events/index.ts
+++ b/src/utils/events/index.ts
@@ -523,7 +523,7 @@ export const getMotionState = async (
     motion.skillRep,
     totalStakeFraction,
     18,
-  );
+  ).sub(1);
   switch (motionNetworkState) {
     case NetworkMotionState.Staking:
       return bigNumberify(motion.stakes[1]).gte(bigNumberify(requiredStakes))


src/utils/events/index.ts
```diff
@@ -523,7 +523,7 @@ export const getMotionState = async (
     motion.skillRep,
     totalStakeFraction,
     18,
-  );
+  ).sub(1);
```
alicjakujawa, line 526: hmm can you explain it? 🤔

chinins: In the `getMotionRequiredStake` we add `1` so it screws the comparison of the requiredStake and nayStakes/yayStakes - it's off by `1`. That's why we have motion tag shown as `Failed` - because it never achieves required stake by `1`.
```
export const getMotionRequiredStake = (
  skillRep: BigNumber,
  totalStakeFraction: BigNumber,
  decimals: number,
): BigNumber => {
  const requiredStake = skillRep
    .mul(totalStakeFraction)
    .div(bigNumberify(10).pow(decimals))
    .add(1);

  return requiredStake;
};
```

alicjakujawa: hope that we don't have more places like that.. i felt from the beginning that this adding 1 can bite us

alicjakujawa (approved):

PR #2527, opened 2021-06-01
Maintenance/dev 331 decimal package lighter version

chinins:
## Description

This PR replaces `decimal.js` with lighter version of the package - `decimal.js-light`

**Changes** 🏗

- change package
- remove round() method as it's done automatically when doing toString()

Resolves DEV-331


ArmandoGraterol (changes requested):
I got this error while trying to stake:

FireShot Capture 334 - Colony - localhost

Also, I saw some places where we are still using `decimal.js` and where the light version might be of use (`utils/colonyMotions` and `utils/tokens`)

chinins:
@ArmandoGraterol, good catch about utils files - my code editor search somehow missed it.

Can you please retest again? I can't replicate... I've also made some minor tweaks so maybe that would work...

ArmandoGraterol:
> @ArmandoGraterol, good catch about utils files - my code editor search somehow missed it.
> 
> Can you please retest again? I can't replicate... I've also made some minor tweaks so maybe that would work...

It's strange because for some values it works if I stake the minimum or the maximum but if I want to stake something between those 2 then I can't

FireShot Capture 337 - Colony - localhost

From the error, I see that a decimal is being sent over to the network, which from what I've gathered from another PR I did, doesn't fly.

ArmandoGraterol (approved):
It seems to be working fine now :) No error through the whole process. Good job!

alicjakujawa:
Not sure if we will merge it as we probably get rid of decimal.js at all :D so lets keep on hold and we will see what we will decide

rdig:
@chinins I'm thinking of closing this as it's pretty outdated at this point. 

Want to keep it around in the hopes of salvaging some of your work here ? Or are you ok loosing it ?

chinins:
> @chinins I'm thinking of closing this as it's pretty outdated at this point.
> 
> Want to keep it around in the hopes of salvaging some of your work here ? Or are you ok loosing it ?

Let's close it. I think it's too outdated. If need be it's easier to just redo it again.

PR #2529, opened 2021-06-01
Update copy for motion finalized event

alicjakujawa:
## Description

This PR changes copy for MotionFinalizedEvent. This event doesn't keep information about person which fire this action, so we can't display userName/userAddress in copy

**Changes** 🏗

* Updated copy for MotionFinalizedEvent

Screenshot 2021-06-01 at 17 10 04

Resolves DEV-377


chinins (approved):
Good to go!

PR #2535, opened 2021-06-04
Fix user wallet UI

chinins:
## Description

This PR updates user wallet (yet again)

**Changes** 🏗

- changed font-weight from 700 to 600
- updated alignment

Resolves DEV-389

Screenshot 2021-06-04 at 22 51 03

### DIFF
--- a/src/modules/pages/components/RouteLayouts/UserNavigation.css
+++ b/src/modules/pages/components/RouteLayouts/UserNavigation.css
@@ -96,7 +96,7 @@
 }
 
 .walletAddress {
-  padding: 2px 9px 3px;
+  padding: 3px 9px 2px;
 
   /*
    * @NOTE We need to do border sizing shenanigans due to how the border color
@@ -129,8 +129,9 @@
 }
 
 .buttonsWrapper {
+  display: flex;
   margin-right: 12px;
-  padding: 2px;
+  padding: 1px;
   height: 26px;
   position: relative;
   border-radius: var(--radius-normal);
--- a/src/modules/pages/components/RouteLayouts/UserNavigation.css
+++ b/src/modules/pages/components/RouteLayouts/UserNavigation.css
@@ -116,11 +116,6 @@
   cursor: pointer;
 }
 
-.walletAddress span {
-  font-size: var(--size-smallish);
-  font-weight: var(--weight-bold);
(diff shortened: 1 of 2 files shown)


alicjakujawa (changes requested):
Screenshot 2021-06-07 at 12 40 47
Doesn't seems to be aligned on my machine 🤦‍♀️  Also worth check on @rdig machine

ArmandoGraterol (commented):
It seems like it's kind of aligned in my machine:

FireShot Capture 350 - Colony - localhost

I would add 1 more pixel to the wallet address, or subtract 1 pixel out of the tokens but perhaps I'm being nitpicky

src/modules/pages/components/RouteLayouts/UserNavigation.css
```diff
@@ -96,7 +96,7 @@
 }
 
 .walletAddress {
-  padding: 2px 9px 3px;
+  padding: 3px 9px 2px;
```
rdig, line 99: For me it looks "aligned" if I change the top padding to `0`

```suggestion
  padding: 0 9px 2px;
```

rdig (changes requested):
Sadly, same here :(

Screenshot from 2021-06-14 11-57-37

Another thing, but most likely not related to your issue, there's a leftover artifact from the tokens/address wrapped that should not be visible when the wallet is not connected:

Screenshot from 2021-06-14 11-57-59

Anyway, I think the best way to "solve" these issues is to pair program and fix this "live" so we can avoid all this back and fourth as it's costing us time.

Just ping me at some point _(after you come back from your time off)_ and we can take it from there

rdig:
What's the state of this ?

chinins:
> What's the state of this ?

I think that we were planning to do a session together to find a fix for both of our systems (what looks ok on my machine doesn't look ok on yours) but were distracted by more urgent things... 

There is actually an open issue that touches on  this as well:
https://github.com/JoinColony/colonyDapp/issues/2686

I can close this PR and reopen another one (just to avoid a rebase), and try to fix it again (plus everything from the newer issue) and then we can try having a live session to check/update my fix...

rdig:
Yep. Closing / re-opening sounds like the best approach here.

PR #2538, opened 2021-06-04
Add extra condition for motion failed system message

ArmandoGraterol:
Fixes bug where if the motion failed and nobody voted then the "motion failed" system message doesn't appear.

### To test:

1. Stake both sides of a motion
2. Don't vote and just let the motion fail due to timeout
3. Check if the system message for motion failed appears

Resolves DEV-384


alicjakujawa (approved):

chinins (approved):
Good to go!

PR #2539, opened 2021-06-07
Improve event feed items spacing

ArmandoGraterol:
Updated paddings and margins of FeedItems so that they look more like the design.

FireShot Capture 351 - Colony - localhost

Resolves DEV-385


alicjakujawa (approved):
Good to merge

PR #2540, opened 2021-06-07
Update spacing and alignment on VotingWidget

ArmandoGraterol:
* Increased space between VotingWidget title and voting results
* Added `drop-shadow` to voter avatars
* Updated border of voter avatars

FireShot Capture 352 - Colony - localhost

Resolves DEV-383

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.css
@@ -30,6 +30,10 @@
   border-radius: var(--radius-large);
 }
 
+.voteResultsWrapper > div {
+  margin-top: 20px;
+}
+
 .escalation {
   flex: 1 1 auto;
   padding-bottom: 1px;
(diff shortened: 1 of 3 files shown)


alicjakujawa (approved):
Screenshot 2021-06-08 at 10 13 49
One more thing, space between progress bar and avatars is a bit too big (in compare do designs) after you update it, its good to merge

chinins (approved):
Good to go without changing the distance to the avatar - in my opinion (see me separate comment).

chinins:
> Screenshot 2021-06-08 at 10 13 49
> 
> One more thing, space between progress bar and avatars is a bit too big (in compare do designs) after you update it, its good to merge

@alicjakujawa, I think that in figma the space is smaller because there are several avatars and it will be the case in this instance. When there is only one, it's normal/ok to have it larger - so that the bar doesn't change length depending on how many avatars we have... 

Let me know what you both think.

ArmandoGraterol:
> > Screenshot 2021-06-08 at 10 13 49
> > One more thing, space between progress bar and avatars is a bit too big (in compare do designs) after you update it, its good to merge
> 
> @alicjakujawa, I think that in figma the space is smaller because there are several avatars and it will be the case in this instance. When there is only one, it's normal/ok to have it larger - so that the bar doesn't change length depending on how many avatars we have...
> 
> Let me know what you both think.

@chinins You are correct! I just tested it with 3 avatars, and the space between the avatars and the progress bar is small as in the design. So I think it's good as it is. 

I will not merge this tho as I think there's still discussion on whether the avatars should be aligned to the progress bar instead of the icon, so I'll wait.

PR #2541, opened 2021-06-08
Update voting widget UI

chinins:
## Description

THis PR updates the UI for the voting widget.

**New stuff** ✨

- add thumbs-up/down outlined icons

**Changes** 🏗

- update voting progress bar border-radius and alignment with the question mark tooltip
- change the RadioGroup icon when not selected & the hover state border
- add `borderRadius` prop to the `ProgressBar` & corresponding UI

## TODO

Check potential misalignment on other machines

Resolves DEV-386

Screenshot 2021-06-08 at 12 04 37
Screenshot 2021-06-08 at 12 02 27


ArmandoGraterol (commented):
It seems to be looking good! The widget just needs the disabled icon version you mentioned already through Discord.

As always, the alignment looks different in my machine:

FireShot Capture 353 - Colony - localhost

alicjakujawa (approved):
https://user-images.githubusercontent.com/13069555/121373828-aeefb600-c93f-11eb-9d6c-1c1e7a1cf6a5.mov

Looks good on my machine, the only one thing i would think of is progress bar is moving if counter is taking less space (see video) I think we want this to be fixed. Approving either way

chinins:
@alicjakujawa it's a good point - i'll clarify it with Karol. I thought it should change the position otherwise the spaces will be too large when there is not much time left... But i'll confirm with Karol.

rdig (approved):
I'm good with this.

However I do want to note that on my machine the help icon is still misaligned:

Screenshot from 2021-06-09 20-10-43
Screenshot from 2021-06-09 20-11-38

But at this point I'm begining to suspect this is a issue with the help icon since it is happening in a lot of place, unrelated to one another

chinins:
@alicjakujawa & the rest, I've contacted Karol and he said that the countdown timer should then move to the right. I've update the PR accordingly.

image

PR #2542, opened 2021-06-08
Fix being able to claim when total is 0

ArmandoGraterol:
Self-explanatory title, now shouldn't be able to claim when your payout is 0, as that produces a transaction error.

Resolves DEV-397


chinins (approved):
Looking good!

PR #2543, opened 2021-06-09
Update `TransferFundsDialog` form to avoid validation  error on `to` domain field

chinins:
## Description

This PR fixes the validation error in the `to domain` field.

Please, test both variants: when there is only root domain in the colony & when there are several domains.

**Changes** 🏗

- update `TransferFundsDialog`
- update `TransferFundsDialogForm`

Resolves DEV-394

### DIFF
--- a/src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialog.tsx
+++ b/src/modules/dashboard/components/TransferFundsDialog/TransferFundsDialog.tsx
@@ -1,10 +1,11 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { FormikProps } from 'formik';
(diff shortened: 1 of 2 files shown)


ArmandoGraterol (commented):
When I do it when there's another domain then it works as expected, but if the `Root` domain is the only one created then when I go to the dialog it doesn't throw a validation error until I interact with fields.

Perhaps it's better that the error appears as soon as you enter the dialog and you have only 1 domain instead of it waiting for the user's interaction to show it.

rdig (approved):
This works as expected, with both:
- multiple domains
- only the root domain

However, you did mis-spell the PR's title. It's not `TransformDialog`, it's `TransferFundsDialog` :)

Anyway, nice work! :tada:

PR #2544, opened 2021-06-09
Enable extension without reloading

alicjakujawa:
## Description

This PR updates extension manager to be able to enable voting reputation extension without reloading page (after installing)

**New stuff** ✨

* reinitializeColonyManager util saga

**Changes** 🏗

* refactored setupUserContext to use new saga
* refactored colonyExtensionEnable saga to clear client which we want to enable from extensionClients and set whenever it will be enabled

Resolves DEV-393


chinins (commented):
The installing & enabling without refreshing works fine. But when I then create a motion, I am redirected to a not found page. I then referesh it and the motion details page appears fine. I wonder if it's linked to your PR. It doesn't happen on the `feature/reputation-voting` branch

alicjakujawa:
> The installing & enabling without refreshing works fine. But when I then create a motion, I am redirected to a not found page. I then referesh it and the motion details page appears fine. I wonder if it's linked to your PR. It doesn't happen on the `feature/reputation-voting` branch

Can be! good catch i will investigate it :)

src/modules/dashboard/sagas/extensions.ts
```diff
@@ -126,10 +131,17 @@ function* colonyExtensionEnable({
...
   const initChannelName = `${meta.id}-initialise`;
   const setPermissionChannelName = `${meta.id}-setUserRoles`;
+  const colonyManager = TEMP_getContext(ContextModule.ColonyManager);
 
   if (!extension) {
     throw new Error(`Extension with id ${extensionId} does not exist!`);
   }
+
+  const key = `${colonyAddress}-${extensionId}`;
+  // Remove old exentions client if exist
```
ArmandoGraterol, line 141: Typo on comment `extensions`

ArmandoGraterol (approved):
It seems to be working as advertised :ok_hand:

rdig (approved):
Tested and works very nicely.

A very nice refactor which fixes both the Extensions as well as the clients rehydration.

Way to go :100:

chinins (approved):
Working well now! Great job!

PR #2545, opened 2021-06-09
Fix MotionHasFailedFinalizable system message

ArmandoGraterol:
Fixed conditional for one of the MotionHasFailedFinalizable cases where the message appeared when the motion has passed.

Resolves DEV-399


chinins (approved):
Looking good!

rdig (approved):
Works as expected. I've tested this on QA and works properly.

Nice work.

PR #2546, opened 2021-06-09
Fix CountdownTimer copies

ArmandoGraterol:
- Fix CountdownTimer select for copies not being updated to the latest MotionState changes
- Also an attempt to fix the countdown timer sometimes not appearing on the QA env. From what I managed to debug from QA, it may be due to it not being fast enough sometimes, and that would get the `useEffect` in charge of updating the period stuck when on a staking phase. I've updated the conditional of that `useEffect` to prevent that from happening.

Resolves DEV-400

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
+++ b/src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
@@ -122,14 +122,16 @@ const CountDownTimer = ({
    */
   useEffect(() => {
     if (
-      (prevStateRef.current === null && isStakingPhaseState) ||
-      !isStakingPhaseState
+      data &&
+      state &&
+      ((prevStateRef.current === null && isStakingPhaseState) ||
+        !isStakingPhaseState)
     ) {
       const period = currentStatePeriod() / 1000;
       setTimeLeft(period > 0 ? period + 5 : period);
-      prevStateRef.current = state || null;
+      prevStateRef.current = state;
     }
-  }, [currentStatePeriod, prevStateRef, state, isStakingPhaseState]);
+  }, [data, currentStatePeriod, prevStateRef, state, isStakingPhaseState]);
 
   /*
    * Count it down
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
```diff
@@ -122,14 +122,16 @@ const CountDownTimer = ({
...
+      ((prevStateRef.current === null && isStakingPhaseState) ||
+        !isStakingPhaseState)
     ) {
       const period = currentStatePeriod() / 1000;
       setTimeLeft(period > 0 ? period + 5 : period);
-      prevStateRef.current = state || null;
+      prevStateRef.current = state;
     }
-  }, [currentStatePeriod, prevStateRef, state, isStakingPhaseState]);
+  }, [data, currentStatePeriod, prevStateRef, state, isStakingPhaseState]);
```
alicjakujawa, line 134: I dont think we need to check for data in here, we already have currentStatePeriod dependency which has data as dependency, so whenever data will change, currentStatePeriod change, and this will run as well, also we're handling cases if data is null in currentStatePeriod callback, so its unnecessary additional logic

alicjakujawa (changes requested):

chinins (commented):
I agree with Alicja, it should be updated when state or data is updated. Maybe the problem is somewhere else...
I did cherry-pick your commit for the copies for my countdown timer PR - thanks for that!

ArmandoGraterol:
@alicjakujawa @chinins The problem was that the data wasn't being retrieved _sometimes_ fast enough, and due to that, in the case of the staking phase state (Staked and Staking) running only 1 time, it would retrieve the `-1` and leave it like that (Again, In the case of the staking phases), therefore, the countdown timer wouldn't appear. That's why now I'm checking there's data before getting the periods

chinins (approved):
I guess it makes sense if this resolves the problem...

ArmandoGraterol:
If you guys have any suggestions on how could this be better I'm open to that, maybe you can see a different, better way of solving this :) I almost fried my brain thinking about alternative solutions for this bug

src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
rdig: I think the opposite, that checking the data in here is actually what makes this work, since every time it changes, it will also trigger this

(I suspect, I haven't properly tested it)

rdig (approved):
Seems good to go.

The thing I haven't tested is that issue, where it wouldn't show up sometimes on QA, but I guess we'll see that when we deploy.

PR #2547, opened 2021-06-12
Add countdown timer to motions in actions list

chinins:
## Description

This PR adds countdown timer to a motion in the actions list

**New stuff** ✨

- add tiemoutPeriods to motions subscription, types & resolvers

**Changes** 🏗

- updated transformers accordingly
- update ActionsListItem css accordingly
- update CountdownTimer css accordingly
- cherry-picked Armando's commit for CountdownTimer copies
- updated FormattedAction type accordingly

Resolves DEV-398

Screenshot 2021-06-16 at 14 16 20

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -928,6 +928,44 @@ export const motionsResolvers = ({
         bigNumberify(fundamentalChainId),
       );
     },
+    async timeoutPeriods({
+      fundamentalChainId: motionId,
+      associatedColony: { colonyAddress },
+    }) {
+      try {
+        const extensionClient = await colonyManager.getClient(
+          ClientType.VotingReputationClient,
+          colonyAddress,
+        );
+
+        const blockTime =
+          (await getBlockTime(networkClient.provider, 'latest')) || 0;
+
+        const escalationPeriod = await extensionClient.getEscalationPeriod();
+
+        const { events } = await extensionClient.getMotion(motionId);
+
+        const timeLeftToStake = events[0] * 1000 - blockTime;
+        const timeLeftToSubmit = events[1] * 1000 - blockTime;
+        const timeLeftToReveal = events[2] * 1000 - blockTime;
+        const timeLeftToEscalate =
+          timeLeftToReveal + escalationPeriod.toNumber() * 1000;
+
+        return {
+          __typename: 'MotionTimeoutPeriods',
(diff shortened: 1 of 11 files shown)


src/data/resolvers/motions.ts
```diff
@@ -928,6 +928,44 @@ export const motionsResolvers = ({
         bigNumberify(fundamentalChainId),
       );
     },
+    async timeoutPeriods({
```
rdig, line 931: Why did you add this logic in again ?

We already have it in the `motionTimeoutPeriods` resolver.

If you need to use it in a different resolver as well, just extract it somewhere you can use in both places, please don't duplicate code.

src/modules/dashboard/components/ActionsPage/CountDownTimer/CountDownTimer.tsx
```diff
@@ -4,6 +4,7 @@ import React, {
   useCallback,
   useRef,
   MutableRefObject,
+  useMemo,
 } from 'react';
 import { FormattedMessage, defineMessage } from 'react-intl';
 import { useDispatch } from 'redux-react-hook';
```
rdig, line 10: There's a issue in this component, after the timeout runs out, it tries to trigger a motion values update (by calling `dispatch`)

But the `motionId` is just a `string`, at it's expecting a `bigNumber`

Screenshot from 2021-06-16 16-25-07

Please convert it to a `bigNumber` using `bigNumberify` before passing it to `dispatch`

rdig (changes requested):
So in terms of functionality this works as expected.

Screenshot from 2021-06-16 16-28-31

The ellipsis display works as expected too:

Screenshot from 2021-06-16 16-30-54

However I am not happy with duplicating the resolver code, and I want you to fix this by extracting the logic and generalizing it so that we reduce code repetition.

Besides that, very nice work @chinins :100:

src/data/resolvers/motions.ts
chinins: I did it for testing and I was meaning to extract it but then forgot

chinins: I've done it now and you can rereview.

rdig (commented):
Good to go! Very nice work :100: 

PS: Make sure to rebase your branch on `feature/reputation-voting` since your colonyNetwork version is behind

Screenshot from 2021-06-17 20-15-47

rdig (approved):
Damn it Github! I wanted to approve it!

PR #2548, opened 2021-06-14
Remove Required Stake Off-by-one Compensation Hack

rdig:
## Description

This PR removes all instances where we added `1`, then subtracted `1` to the `requiredStake` value, in order to compensate to a bug in the contracts that was present in `v1` of the Voting Reputation extension.

This is no longer needed since all colonies will be running Voting Reputation V2

My fix was tested against the Xdai QA environment, started from my local machine. 

Please take some time and test it locally, and if the stake value is not calculated correctly, let me know.

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -146,7 +146,6 @@ export const motionsResolvers = ({
           ClientType.VotingReputationClient,
           colonyAddress,
         );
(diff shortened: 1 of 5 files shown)


ArmandoGraterol (approved):
Looks like it's working as expected :ok_hand:

chinins (approved):
Looks good, works well!

alicjakujawa (approved):
Tested for different extension settings and work great 🔥

PR #2550, opened 2021-06-14
Update styles of UserPicker core component

ArmandoGraterol:
Update styles and icons of UserPicker component

**New stuff** ✨

* Added "filled-circle-person" icon
* Added placeholder text for user picker

**Changes** 🏗

* Updated styles of User Picker field and item list

**Deletions** ⚰️

* Removed "Dismiss" header from user picker

Resolves DEV-401


alicjakujawa (approved):
I think all of the problems from issue was addressed. The only 2 things i see is:
* arrow is a bit too high
* color of placeholder is wrong
Approving it, just make sure to address those problems

chinins (approved):
Looking good now! Nice work!

PR #2551, opened 2021-06-15
Voting rewards updates

alicjakujawa:
## Description

This PR updates displaying rewards section for voting state

**Changes** 🏗

* Replaced fixed CLNY token logo with dynamic logo based on native token
* Displaying range widget only if min/max are not equal
* Fixed wrongly displayed rewards values in voting state (min value bigger then max)

Screenshot 2021-06-15 at 10 08 47
Screenshot 2021-06-15 at 10 03 30

Resolves DEV-408

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -431,7 +431,9 @@ export const motionsResolvers = ({
           ClientType.VotingReputationClient,
           colonyAddress,
         );
-        const { domainId } = await votingReputationClient.getMotion(motionId);
+        const { domainId, rootHash } = await votingReputationClient.getMotion(
+          motionId,
+        );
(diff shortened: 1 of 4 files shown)


ArmandoGraterol (approved):
Looks good to me :ok_hand: 

I did saw at some point the `minReward` being higher than the `maxReward` hahaha I'm assuming that's on network.

rdig (approved):
Very nice work @alicjakujawa :tada: 

I've also tested the case where you are the last person to vote, or your vote will take the total vote count to 100%, in which case the reward values are the same, so we just show one of them and not the range. 

Everything seems to be working as expected except that error we keept seeing when the `minReward` value was higher than the `maxReward` value.

However, you can go ahead and merge this, and I will investigate that issue with Alex.

PR #2552, opened 2021-06-16
Finalize Motion Manual Gas Estimation

rdig:
## Description

This problem is very subtle and happens behind the scenes. If you don't provide a big enough gas limit to the _finalize motion_ transaction, while the transaction goes successfully, the contracts, behind the scenes won't actually create your action. So even though you finished your motion successfully, you don't get your action on the other side.

The fix for this is to manually estimate the gas, not for the _finalize motion_ transaction, but for the actual action that the motion will be creating. For this we add a flat rate of `55k` `WEI` to the estimation, to provide a large enough overhead to cover both the _Finalize Motion_ transaction as well as the actual action transaction that will happen behind the scenes.

The problem with testing this is that while you can check that the _finalize motion_ transaction fires, it's not really relevant since this problem only shows up on QA, since the estimation mechanism on the QA / production networks are different from the local ganache one. That's why we found about his so late.

Resolves DEV-406

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
@@ -503,7 +503,6 @@ const DefaultMotion = ({
               colony={colony}
               actionType={actionType}
               motionId={motionId}
-              motionDomain={motionDomain}
               scrollToRef={bottomElementRef}
               motionState={motionState}
             />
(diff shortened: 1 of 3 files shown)


chinins (approved):
The code looks correct and I've tested on development as I could (I understand that fully we can test only in prod).

ArmandoGraterol:
Yo, now I'm getting this error when I try to finalize the motion:

FireShot Capture 355 - Colony - localhost

Is this something expected?

rdig:
@ArmandoGraterol nope it's not expected. Your error suggests it needs more gas, although I need to check with Alex on how much we can increase that value

rdig:
@ArmandoGraterol and however ends up testing this.

I've fixed this _(by increasing the "flat" estimate rate -- this is calculated, not a random value)_, but please test locally the following motions, since they are know to eat more gas:
- payment
- mint tokens
- create team

Let me know if any of them fails, otherwise I will consider this as "working" and will deploy to QA where the "real" test will be

chinins (approved):
I've tested all 3 of the motions and it all worked fine.

ArmandoGraterol (approved):
It seems to be working great now :ok_hand: Now time to test it on QA

PR #2554, opened 2021-06-16
Update NotEnoughReputation component

ArmandoGraterol:
Updated NotEnoughtReputation component copies and styles

**New stuff** ✨

* Added an appearance prop to NotEnoughReputation in order to control when to apply the negative margin top

**Changes** 🏗

* Updated copies of NotEnoughReputation

FireShot Capture 354 - Colony - localhost

Resolves DEV-410

### DIFF
--- a/src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
+++ b/src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
@@ -21,7 +21,7 @@ const MSG = defineMessage({
     id: 'NotEnoughReputation.title',
     defaultMessage: `There is no reputation in this {onRootDomain, select,
       true {colony}
-      false {domain}
+      false {team}
     } yet.`,
   },
   description: {
--- a/src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
+++ b/src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
@@ -1,37 +1,54 @@
+import { ROOT_DOMAIN_ID } from '@colony/colony-js';
 import React from 'react';
 import { defineMessage, FormattedMessage } from 'react-intl';
 
+import { getMainClasses } from '~utils/css';
 import DialogSection from '~core/Dialog/DialogSection';
 
 import styles from './NotEnoughReputation.css';
 
+interface Appearance {
+  marginTop?: 'negative';
+}
+
+interface Props {
+  domainId?: number;
+  appearance?: Appearance;
+}
+
 const MSG = defineMessage({
   title: {
     id: 'NotEnoughReputation.title',
-    defaultMessage:
-      'There is not enough reputation in colony/domain to create this motion.',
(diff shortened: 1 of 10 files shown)


alicjakujawa (changes requested):
Firstly, are we sure that this is correct design?
Screenshot 2021-06-17 at 10 03 26

Secondly, is this component use somewhere where it should have this margin? 
And lastly, shouldn't we have this component in "base" state as component with margin, and adding margin (not negative margin) by appearance object?  Unless this component is displayed everywhere same, then we dont need to appearance object at all

src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
```diff
@@ -1,37 +1,54 @@
...
+}
+
 const MSG = defineMessage({
   title: {
     id: 'NotEnoughReputation.title',
-    defaultMessage:
-      'There is not enough reputation in colony/domain to create this motion.',
+    defaultMessage: `There is no reputation in this {onRootDomain, select,
+      true {colony}
+      false {domain}
```
chinins, line 24: Should be `team` here as we don't use domains in the copies.

chinins (approved):
@alicjakujawa, the designs are what Karol provided at the time (pink title being a different color than the permissions warning). I've just had a call with him on a different matter and he confirmed that we can change the color to the red that is used for warnings.

As for the margin in the `NotEnoughReputation`, the margin comes from the dialog section. Depending on the dialog, we don't need the negative margin, so it is sometimes needed and sometimes not. So it's either we conditionally add/remove margin on the dialog section of the above part and also add/remove the margin-top of that above part as well, or we do a negative margin prop on the component. Another option would be to wrap it in a div & apply the negative margin on that div. Not sure which one is the best. Maybe the prop as it doesn't add another div level.

So I'm pre-approving, just please change the color of the red title & correct `domain` to `team` in the copy.

src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
rdig: Good catch! :1st_place_medal:

ArmandoGraterol:
@alicjakujawa 

1. From what I know, I think so. A screenshot of the design was added on the notion issue and Karol posted a link to the design.
2. If you're talking about the negative margin, yes. The negative margin is used on 4 instances where the component is used.
3. The margin that you're probably thinking about comes from the DialogSections that are above this component because apart of the negative margin I just added, the NotEnoughReputation doesn't any.

Sometimes the DialogSections above the NotEnoughReputation component have a lot of margin/padding on the bottom (It depends on the Dialog), the NotEnoughReputation component is used in 8 dialogs, in which 4 of them a negative margin is needed in order to compensate for the larger margin/padding of the sections.

I didn't want to add another div to the component, so that's why I changed the position of the previous container and added an appearance object for the cases where the negative margin is needed.

Let me know what you think, and if that can be done in a better way

alicjakujawa (approved):

PR #2556, opened 2021-06-16
Fix and update objection annotation

ArmandoGraterol:
- Increase objection annotation limit to 4000
- Fix objection annotation missing the annotation message param on transform func

Resolves DEV-411
Resolves DEV-412


alicjakujawa (changes requested):
There is one problem in here. When i object with annotation message i need to refresh page to see annotation. It cannot be like that.

chinins (commented):
I agree with Alicja. You just need to add `MotionObjectionAnnotation` to `updateMotionValues` so that it gets populated automatically.

alicjakujawa (approved):
Now it works correctly!

PR #2557, opened 2021-06-16
Fix event transaction link

ArmandoGraterol:
Fixed every event transaction link using the transaction hash of the transaction that created the motion instead of their own transaction hash.

Can't _really_ be tested on local as no explorers are used, but I'm 100% sure that this is it. (The correct event transaction hash is reaching `getBlockExplorerLink` but a `#` is returned because it's a local env)

Resolves DEV-413


alicjakujawa (approved):
Yep it should solve the problem, but lets make sure to test it on QA when we will do new deploy

chinins (approved):
It seems that it should work.

PR #2559, opened 2021-06-17
Fix Successful Permissions Management Motions Titles (Permissions Comparisons)

rdig:
## Description

Up until now, the way the _Permission Management_ motions titles would be generated was that we took the "current" user permission and compare them with what the motion was changing through the bitMask _(you can't tell otherwise which one was changed, you can only see the "final" state of the permission after the motion passes)_

In theory this works ok, however, once the motion to change the permissions passes, and you refresh the page, you'll end up comparing the "current" user permissions _(which were already changed by the motion)_ to the ones that the motion is trying to set. _(which now are the same, so we can't actually tell what changed)_

This PR fixes this by get **historic user roles** by limiting the events we look for to determine roles to the block that the motion was created in.

This way we get all the roles _up to_ the block in which the motion was started, and compare them with the roles that the motion is setting currently. These values will always be different so we will always be able to determine what changed.

To make this work, I had to add a new helper to `colonyJS` called `getHistoricColonyRoles`

**Testing**

- **Make sure to run `npm i` to bring in the new version of `colonyJS`**
- Create a  _Permission Management_ motion and pass it  _(the titles both in the motion page and in the actions list, should be the same both before the motion passed and after the motion passed)_

To test the opposite:
- go to a different branch, that doesn't have this fix _(eg: `feature/reputation-voting`)_
- Create a  _Permission Management_ motion and pass it _(after it passes, the title will change to "Generic motion we don't have any information about...")_

Resolves DEV-395

### DIFF
--- a/src/utils/colonyMotions.ts
+++ b/src/utils/colonyMotions.ts
@@ -2,9 +2,10 @@ import { defineMessage } from 'react-intl';
 import { BigNumber, bigNumberify } from 'ethers/utils';
 import { Decimal } from 'decimal.js';
 import { isNil } from 'lodash';
+import { ColonyRoles } from '@colony/colony-js';
 
-import { getUserRolesForDomain } from '../modules/transformers';
-import { Colony, AnyUser } from '~data/index';
+import { getRolesForUserAndDomain } from '../modules/transformers';
+import { AnyUser } from '~data/index';
 import { ActionUserRoles } from '~types/index';
 
 export enum MotionState {
@@ -184,18 +185,18 @@ export interface MotionValue {
 }
 
 export const getUpdatedDecodedMotionRoles = (
-  colony: Colony,
   recipient: AnyUser,
   fromDomain: number,
-  roles: ActionUserRoles[],
+  currentRoles: ColonyRoles = [],
+  setRoles: ActionUserRoles[],
 ) => {
-  const userCurrentRoles = getUserRolesForDomain(
-    colony,
-    recipient.profile.walletAddress,
+  const currentUserRoles = getRolesForUserAndDomain(
+    currentRoles,
+    recipient.id,
     fromDomain,
   );
-  const updatedRoles = roles.filter((role) => {
-    const foundCurrentRole = userCurrentRoles.find(
+  const updatedRoles = setRoles.filter((role) => {
+    const foundCurrentRole = currentUserRoles.find(
       (currentRole) => currentRole === role.id,
     );
     if (!isNil(foundCurrentRole)) {
--- a/package-lock.json
+++ b/package-lock.json
@@ -4348,9 +4348,9 @@
       }
     },
     "@colony/colony-js": {
-      "version": "4.0.0-rc.8",
-      "resolved": "https://registry.npmjs.org/@colony/colony-js/-/colony-js-4.0.0-rc.8.tgz",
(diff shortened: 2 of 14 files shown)


chinins (commented):
It seems to be working only sometimes (as per my screen picture)

In one motion I have added Recovery permission, in the other one I removed it. Before I had the second motion passed, the first one (which has the correct title now) had the generic title.

Screenshot 2021-06-17 at 20 05 48

ArmandoGraterol (changes requested):
Yeah, I think something funky is going on, perhaps it's just the recovery role the one who is causing troubles?

FireShot Capture 356 - Colony - localhost

In the first motion (Going from down to up), I only removed the funding permission, then in the second motion, I added it back and removed the recovery role, and now it's saying the recovery role was assigned on the first and on the second there's no mention that I removed the recovery role.

Also, when I created the motions, for like 1 second the page showed a wrong title:
Screenshot from 2021-06-17 17-16-45

Before I guess loading more and actually showing the correct title:

Screenshot from 2021-06-17 17-16-47

src/utils/colonyMotions.ts
```diff
@@ -184,18 +185,18 @@ export interface MotionValue {
 }
 
 export const getUpdatedDecodedMotionRoles = (
-  colony: Colony,
   recipient: AnyUser,
   fromDomain: number,
-  roles: ActionUserRoles[],
+  currentRoles: ColonyRoles = [],
+  setRoles: ActionUserRoles[],
```
alicjakujawa, line 191: what "setRoles" means? when i see set prefix i usually think of setter methods

alicjakujawa: but its just nitpick :P

alicjakujawa (changes requested):
Okey something seems to be wrong. Flow:
1. I  had administration, architecture and funding permission
2. I create motion to remove administration, architecture and add recovery permission
3. At first title said that im removing administration, architecture  and adding recovery permission (as well details widget showing administration, architecture and funding
4. When motion passed, title says: "Remove the administration, architecture permissions in Root from b" and Details widget shows only administration and architecture

src/utils/colonyMotions.ts
rdig: So we have a prop that's coming from the contracts that is called `setTo`, which can be `true` or `false`

Roles that are inside the `setRoles` array, are roles which have that prop.

That's where the naming comes from :)

rdig:
@chinins @ArmandoGraterol @alicjakujawa Please re-test this when you have the time

A few things to keep in mind:
- after pulling the latest changes on this branch, please run `npm i` since the `colonyJS` package version was updated
- If you create two consecutive motions for the same thing (eg: removing a role), they will both list "remove..." since the actual event never gets emmited in the background -- this is expected and correct functionality _(see screenshot below)_
- even though a motion might have the `passed` tag, it will not fire it underlying action, until you actually finalize it. So make sure to actually send the "finalize" transaction

Screenshot from 2021-06-23 17-25-54

ArmandoGraterol (approved):
Tested it again and now everything seems to be working as expected :ok_hand: 

All the titles seem to remain the same after passing and finalizing them

alicjakujawa (approved):
Works great for me!

chinins (approved):
Working well now!

PR #2560, opened 2021-06-17
Fix wrongly represented big numbers while sending staking value

alicjakujawa:
## Description

This PR fixes wrongly sending big number values while staking. 

**Changes** 🏗

* Overrides Decimal constructor to handle bigger numbers. (now it handles values bigger than the biggest value on chain)

**How to test** 🏗
* You need to large value of active tokens before firing motion
* Run motion and stake fully for any side, you should not see any errors 

Resolves DEV-348


rdig (approved):
Seem to be working. 

Good to go!

PR #2561, opened 2021-06-17
Motion `args` Resolver handle native chain token  payments (ETH/XDAI)

rdig:
## Description

This PR solves a issue that originally presented itself as motions not showing up in the list, but only for _some_ colonies.

What was actually happening behind the scenes was that for payments made with the native chain token (Ie: Xdai or Eth), which is represented with the `0x0000....` token address would fail and break the resolver, hence preventing it from returning the resolver's values.

This was happening because we were trying to instantiate **all** payment token address with a `tokenClient`. However the `tokenClient` only works for colony standard ERC-20 tokens _(meaning ETH/XDAI or non-standard tokens fail initialization)_

We solve this by manually detecting if the payment was made with the native chain token, and if so, use the default values available to us in the `constants` file.

While at it, we also put in a `try/catch` block to test if tokens (which are not native to the chain), are colony standard or not, so that we can attempt to instantiate the `tokenClient` with them.

**Testing**
- Transfer ETH or XDAI to your colony
- Create a payment with ETH or XDAI
- Make sure it has enough stake to go over the 10% threshold
- If the motion shows up in the list (along with your other motions) then everything is working correctly

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -10,6 +10,7 @@ import {
   ROOT_DOMAIN_ID,
 } from '@colony/colony-js';
 import { bigNumberify, LogDescription, hexStripZeros } from 'ethers/utils';
+import { AddressZero } from 'ethers/constants';
 import { Resolvers } from '@apollo/client';
 
 import { Context } from '~context/index';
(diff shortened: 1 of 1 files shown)


alicjakujawa (approved):
🔥

chinins (approved):

PR #2565, opened 2021-06-21
Improve formatting of TokenActivationPopover

ArmandoGraterol:
Improved token values formatting on the TokenActivationPopover.

The issue Jack reported about not being able to stake when he reached the limit it's not a bug inside the staking slider, more like a bug inside the TokenActivationPopover.

While testing the staking slider limit, I've encountered that the slider could have a limit of `3.96` based on the reputation of the user and the popover would show me that I've exactly `3.96` activated so I would believe that I can stake up to the limit but I can't, because that `3.96` is actually `3.959`, so less than `3.96` 

**What I did to reproduce the case above**
1. Create a colony
2. Have a total of 3 members
3. Pay 2 members `4` tokens each
4. Pay 1 member `2` tokens
5. Initialize the motion extension with a `50%` required stake and a `20%` minimum stake
6. Create motion and watch if the limit of the staking slider and the error that it shows corresponds with the amount of token that you have activated.

FireShot Capture 370 - Colony - localhost

Resolves DEV-423

### DIFF
--- a/src/modules/constants.ts
+++ b/src/modules/constants.ts
@@ -141,3 +141,5 @@ export const ALLDOMAINS_DOMAIN_SELECTION = {
 export const TOKEN_LOGOS_REPO_URL = `https://raw.githubusercontent.com/trustwallet/assets/master/blockchains`;
 
 export const NETWORK_RELEASES_URL = `https://github.com/JoinColony/colonyNetwork/releases/tag`;
+
+export const SMALL_TOKEN_AMOUNT_FORMAT = '0.00000...';
--- a/src/modules/core/components/Fields/Input/InputComponent.tsx
+++ b/src/modules/core/components/Fields/Input/InputComponent.tsx
@@ -10,9 +10,9 @@ import { defineMessages } from 'react-intl';
 import Cleave from 'cleave.js/react';
 import { CleaveOptions } from 'cleave.js/options';
 import { ChangeEvent } from 'cleave.js/react/props';
-
 import { isNil } from 'lodash';
 import Decimal from 'decimal.js';
+
 import Button from '~core/Button';
 
 import { getMainClasses } from '~utils/css';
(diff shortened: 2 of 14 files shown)


chinins (approved):
Seems to be working well.

alicjakujawa (changes requested):
Noticed one bug which probably wasn't introduced here, but its important to fix before merge
Screenshot 2021-06-28 at 17 59 18
So i have all tokens active, but wallet dots is saying that i have some of them inactive

Also, wondering if this is correct:
Screenshot 2021-06-28 at 18 07 21
Screenshot 2021-06-28 at 18 07 07

So we're saying to the user "you can stake max 2.96" and then we actually take a bit more :D Maybe its worth to update also formatting for slider max value? Please ask product

alicjakujawa (commented):
Screenshot 2021-06-29 at 09 12 08
Did you address wrongly displayed dots next to wallet balance?

chinins:
It all looks good to me. When I hit withdraw using `Max` button, then the dot changes to green and it's all working properly. I can reproduce what Alicja is seeing if I deactivate manually tokens with lots of decimal numbers & then activate also manually by just copy-pasting the number that I see in the inactive part. Then there is a tiny amount left but we don't see it because we round only to 5 decimals and the dot stays red.

I think that the functionality is all good like this but it is a bit confusing. Maybe we should ask product what they want to show in this situation?

alicjakujawa (changes requested):
Screenshot 2021-06-30 at 15 16 22
Page is crashing when im trying to move slider

alicjakujawa (changes requested):
Screenshot 2021-07-01 at 09 21 28
Correct behaviour for staking for "yey side"
Wrong for staking for "ney"

alicjakujawa (approved):
Now it works 🔥

PR #2566, opened 2021-06-22
Improve tag base styles

ArmandoGraterol:
Removed base margin styles of the Tag component as those margins are right now really only used in 1 instance and for every other instance where we use the tags we don't want them.

FireShot Capture 358 - Colony - localhost

FireShot Capture 359 - Colony - localhost

Resolves DEV-422

### DIFF
--- a/src/modules/core/components/Tag/Tag.css
+++ b/src/modules/core/components/Tag/Tag.css
@@ -7,10 +7,6 @@
   color: var(--colony-white);
 }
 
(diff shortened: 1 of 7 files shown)


alicjakujawa (approved):
Like this refactor! Dump/Visual components shouldn't handle spaces/margins in my opinion. If we want any margin it should be handled by parent.

chinins (approved):
Good refactor! I prefer it much better.

PR #2567, opened 2021-06-22
Update alignment of countdown timer

ArmandoGraterol:
FireShot Capture 360 - Colony - localhost

Resolves DEV-421

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.css
@@ -79,7 +79,6 @@
   composes: details;
   display: flex;
   align-items: baseline;
-  justify-content: flex-end;
 }
 
 .addressInTitle div {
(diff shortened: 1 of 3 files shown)


chinins (changes requested):
If you just remove `flex-end`, then when you have voting progress bar there it will switch to the left like this:

Screenshot 2021-06-22 at 21 19 59


However, I confirmed the specs with Karol and he sad it should move to the right, like below. So I think that would need some conditional styles or something like this to accommodate both options.
image

ArmandoGraterol:
> If you just remove `flex-end`, then when you have voting progress bar there it will switch to the left like this:
> 
> Screenshot 2021-06-22 at 21 19 59
> 
> However, I confirmed the specs with Karol and he sad it should move to the right, like below. So I think that would need some conditional styles or something like this to accommodate both options.
> image

AH indeed, I didn't occurred to me that the progress bar would need to be aligned to the right. Thanks for pointing it out

chinins (approved):
Looks all correct now!

alicjakujawa (approved):

PR #2568, opened 2021-06-22
Update domain description styles on DetailsWidget

ArmandoGraterol:
Update styles of DetailsWidget to allow the domain description to be shown completely

FireShot Capture 361 - Colony - localhost

Resolves DEV-429

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
@@ -189,7 +189,7 @@ const DetailsWidget = ({
           <div className={styles.label}>
             <FormattedMessage {...MSG.domainDescription} />
           </div>
-          <div className={styles.value}>
+          <div className={`${styles.value} ${styles.descriptionValue}`}>
             <div
               className={styles.domainDescription}
               title={values.fromDomain.description || ''}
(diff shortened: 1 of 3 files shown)


src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
```diff
@@ -189,7 +189,7 @@ const DetailsWidget = ({
           <div className={styles.label}>
             <FormattedMessage {...MSG.domainDescription} />
           </div>
-          <div className={styles.value}>
+          <div className={`${styles.value} ${styles.descriptionValue}`}>
```
chinins, line 192: I think it's better to extend the `descriptionValue` class with `value` class and then you can apply only `descriptionValue` class to the div

chinins (changes requested):
Can you please refactor a bit based on my comment? (Or explain why/if you would prefer not to refactor)

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
ArmandoGraterol: I like that! Will do

chinins (approved):
The code looks cleaner now I think :)

alicjakujawa (approved):

PR #2571, opened 2021-06-23
Remove XDAI conversion to USD

chinins:
## Description

This PR removes conversion of XDAI to USD from several dialogs & `UnclaimedTransfers`.

Testing it is difficult as it is the logical check if the network is xdai. You can just change the check to local and if the conversion to ETH (addressZero for the local) is not there it's working.

**Deletions** ⚰️

- remove XDAI to USD conversion in the `EthUds` components

Resolves DEV-428


ArmandoGraterol (approved):
It seems to be working as expected :+1:

rdig (approved):
Good to go

PR #2572, opened 2021-06-23
Fix issue with not reseting timer after fully object

alicjakujawa:
## Description

This PR fixes issue with not updating timer when staked fully on nay side

**Changes** 🏗

* Refactored CountdownTimer component
* Updated ActionsListItem component
* Updated getActionsListData transformer to return requiredStake and totalNayStake

Resolves DEV-427


ArmandoGraterol (approved):
It seems to be working as advertised :ok_hand:

chinins (approved):
Working well. That's great that you did the refactoring as well!

PR #2573, opened 2021-06-23
Fix transfer funds action description copy

ArmandoGraterol:
Every copy that references "domains" will face the wall eventually, they can hide but we will find them all

FireShot Capture 366 - Colony - localhost

Resolves DEV-431


chinins (approved):
I like your PR description :)

alicjakujawa (approved):

PR #2574, opened 2021-06-24
Fix countdown timer loading UI with voting progress bar

chinins:
## Description

This PR fixes the countdown timer loading UI when it's together with the progress bar

Resolves DEV-430

Screenshot 2021-06-24 at 18 35 22
Screenshot 2021-06-24 at 18 32 57


rdig (approved):
Very nice. This makes it look more consistent

PR #2575, opened 2021-06-24
Fix nay side staking calculation

ArmandoGraterol:
The NAY staking amount on the `RaiseObjectionDialog` was being calculated differently from the way we do it in the `StakingWidget`, the result would be that staking the NAY side through the Dialog would have a really small difference (For example, in the widget 100% could be `100 000 006 000` and in the DIalog, it would be `100 000 005 000`) compared to the amount calculated on the widget, which would lead to cases like the 99% when you actually wanted to stake 100%

Resolves DEV-432


chinins (approved):
Seems like a correct fix!

rdig (approved):
Should be good

PR #2576, opened 2021-06-24
Fix ColonyFromName messing up Home, Funding and Members

ArmandoGraterol:
The ColonyFromName query uses a `@export` directive in order to get and use the colony's address to get the `processedColony` data. The problem is that the query's fields are async (and are also executed in no particular order) so it would execute the `processedColony` field before the colony address is retrieved and/or exported. 

In that case, the first result (and the one that crashes the pages) would be `data: { colonyAddress: ..., processedColony: null }`. Technically there's data, but `processedColony` is `null` so what Home, Funding, and Members do is redirect to the 404 Page, even though the colony exists.

To combat this, we could check in the `Loading` state check if we have a colony address but  `processedColony` is `falsy`, from that we could assume that the colony exists but the `processedColony` is not loaded in yet.

Resolves DEV-379

### DIFF
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -375,7 +375,7 @@ export const colonyResolvers = ({
           : null;
       } catch (error) {
         console.error(error);
-        return null;
+        return error;
       }
     },
     async historicColonyRoles(_, { colonyAddress, blockNumber }) {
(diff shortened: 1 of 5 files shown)


chinins (approved):
All good! It's 🔥

alicjakujawa (approved):
Nice work!

PR #2577, opened 2021-06-28
Increase Finalize Motion Saga Gas Estimate

rdig:
## Description

This PR just increases the `finalizeMotion` saga gas estimate, so now the flat increase is enough to cover all payment motion scenarios:
- colony to colony
- colony to user
- colony to user who didn't have the particular token type before
- colony to colony who didn't have the particular token type before

**Testing**:
Just create a payment motion, if you also see the payment action (after the motion has passed and was finalized), then this works

However, locally this has always worked, the real test is on QA, which is why it took me so long to find the "correct" value

Resolves DEV-402
Also resolves the similar issue, but with user transfers https://www.notion.so/colony/QA-008ecbf3acb84c6281ef8b619f5fcb34?p=850931d7db6040b2946d62636112d6e3

### DIFF
--- a/src/modules/dashboard/sagas/motions/finalizeMotion.ts
+++ b/src/modules/dashboard/sagas/motions/finalizeMotion.ts
@@ -51,13 +51,13 @@ function* finalizeMotion({
     });
 
     /*
-     * Increase the estimate by 80k WEI. This is a flat increase for all networks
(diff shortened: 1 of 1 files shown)


alicjakujawa (approved):
Works well 🔥

ArmandoGraterol (approved):
Works as expected :ok_hand:

chinins (approved):
Looks good!

PR #2578, opened 2021-06-28
Fix avatar count on ColonyHome page

chinins:
## Description

This PR fixes math logic in `useAvatarDisplayCounter` to display avatars correctly.

To test you need either to have 16 or more members in your colony or just manually provide an array of 16 or more addresses in the `ColonyMembers` component (i just use the same address over & over again) instead of manually fetching it:
```
const members = {
  colonyMembersWithReputation: ['<user address>' (repeat 16 times)]
}
```

Resolves DEV-440

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyMembers/ColonyMembers.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyMembers/ColonyMembers.tsx
@@ -58,7 +58,11 @@ const ColonyMembers = ({
   const {
     avatarsDisplaySplitRules,
     remainingAvatarsCount,
-  } = useAvatarDisplayCounter(maxAvatars, members?.colonyMembersWithReputation);
+  } = useAvatarDisplayCounter(
+    maxAvatars,
+    members?.colonyMembersWithReputation,
+    false,
+  );
 
   const BASE_MEMBERS_ROUTE = `/colony/${colonyName}/members`;
   const membersPageRoute =
--- a/src/utils/hooks/useAvatarDisplayCounter.ts
+++ b/src/utils/hooks/useAvatarDisplayCounter.ts
@@ -14,7 +14,7 @@ const useAvatarDisplayCounter = (
       return members.length;
     }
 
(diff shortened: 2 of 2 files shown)


alicjakujawa (changes requested):
Please not change useAvatarDisplayCounter logic as this is used not only in ColonyMembers, just set maxAvatar = 14 inside component

ArmandoGraterol (commented):
Tested `useAvatarDisplayCounter in both of the instances where it's used and it seems like on VotingResultsItem is showing 2 avatars instead of 3

chinins:
> Please not change useAvatarDisplayCounter logic as this is used not only in ColonyMembers, just set maxAvatar = 14 inside component

Unfortunately, it's not that simple, the logic will be broken. I've added another solution.

chinins:
> Tested `useAvatarDisplayCounter in both of the instances where it's used and it seems like on VotingResultsItem is showing 2 avatars instead of 3

Good catch. I didn't realise `VotingResultsItem` uses different logic for displaying avatars.

ArmandoGraterol (approved):
Tested it and it seems to be working great now in both places where the hook is used! Good job :ok_hand:

alicjakujawa (changes requested):
It doesn't make sense. maxAvatars is a props and value of this props should be defined outside of this component. Knowledge about how many avatars we want display should be in parent power. Please revert all changes in useAvatarDisplayCounter and pass proper maxAvatar value

chinins:
> It doesn't make sense. maxAvatars is a props and value of this props should be defined outside of this component. Knowledge about how many avatars we want display should be in parent power. Please revert all changes in useAvatarDisplayCounter and pass proper maxAvatar value

As I explained on discord, then the case when there is exactly 15 avatars is wrong:
Screenshot_2021-07-05_at_13 03 25

I still think that my solution is correct as it allows for both options and takes the logic away from the components. Wer can then reuse it if we have other cases like this.

alicjakujawa (approved):

PR #2583, opened 2021-07-01
Update UserPicker size and validation

ArmandoGraterol:
* Added a standard size for UserPicker component
* Updated validation of UserPicker value
* Updated conditional in charge of showing the errors on Input component

Resolves DEV-449


src/modules/core/components/Fields/InputStatus/InputStatus.tsx
```diff
@@ -35,18 +37,20 @@ const InputStatus = ({
   error,
   status,
   statusValues,
+  touched,
 }: Props) => {
   const { formatMessage } = useIntl();
   const errorText = typeof error === 'object' ? formatMessage(error) : error;
   const statusText =
     typeof status === 'object' ? formatMessage(status, statusValues) : status;
+  const showErrorText = !isNil(touched) && touched;
```
chinins, line 46: Why are you using isNil from lodash here?

chinins (commented):
It all seems good and working well. I just have a question about isNil (please, have a look at my comment).

src/modules/core/components/Fields/InputStatus/InputStatus.tsx
ArmandoGraterol: There are some instances of `InputStatus` where `touched` is not needed, therefore I first check that is not `undefined` or `null` to determine if `touched` is worth checking for when hiding the error message.

I actually just noticed that there's something wrong with this conditional, I'll fix it but the `!isNil(touched)` will remain the same so my explanation above still stands

src/modules/core/components/Fields/InputStatus/InputStatus.tsx
```diff
@@ -1,6 +1,7 @@
 import { MessageDescriptor, useIntl } from 'react-intl';
 import React from 'react';
 
+import { isNil } from 'lodash';
```
alicjakujawa, line 4: Could you please at least use import isNil from 'lodash/isNil' import? It will be less costly

alicjakujawa (approved):
One comment, apart of this works nice so pre-approving

src/modules/core/components/Fields/InputStatus/InputStatus.tsx
chinins: I would just do a direct check but that's a valid check too.

chinins (approved):
Good to go!

PR #2584, opened 2021-07-01
Add whitelist policy selectors

ArmandoGraterol:
Add ColonyPolicySelector component

**New stuff** ✨

* Added ColonyPolicySelector component
* Added `greyWithCircle` theme for CustomRadio component
* Added `gap` appearance property with `medium` value to CustomRadioGroup

FireShot Capture 372 - Colony - localhost

Resolves DEV-436


chinins (commented):
Great job! I am mostly ok with it but as we discussed on discord. let's clarify with Alicja about extensionEnbaled check (relevant for my PR as well).

src/modules/core/components/Fields/Radio/CustomRadio.tsx
```diff
@@ -10,7 +11,7 @@ import { UniversalMessageValues } from '~types/index';
 import styles from './CustomRadio.css';
 
 export interface Appearance {
-  theme?: 'primary' | 'danger';
+  theme?: 'primary' | 'danger' | 'traditional';
```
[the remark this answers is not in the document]
ArmandoGraterol: Hahahaha I just named it `traditional` because the theme makes the CustomRadio look like what a "normal" or default radio would look like. 

If you have a better name for this theme then I'll glady change it, I just didn't wanted to waste too much time thinking about it (I was stuck for like 10 minutes thinking about a better name, but nothing came to my mind besides `traditional`).

alicjakujawa (changes requested):
Please see my review in Olga PR - https://github.com/JoinColony/colonyDapp/pull/2590
You both mixed concept. We treat initParams like for all other Extensions, so those changes need to be in ExtensionSetup. ExtensionBody will be content of our extension when it will be enabled (upload addresses, and list of addresses)

src/data/staticData/extensionData.ts
```diff
@@ -477,8 +497,46 @@ const extensions: { [key: string]: ExtensionData } = {
...
+            value: 'KYC_AND_AGREEMENT',
+            label: MSG.whitelistColonyPolicySelectorAgreementAndKYC,
+            name: 'policy',
+            appearance: {
+              theme: 'traditional',
+            },
+            checked: false,
+          },
+        ],
+      },
```
rdig, line 538: So this is more complicated that you might initially think.

For one, the first argument of the whitelist, must always be a `bool`, so the value must always be `true/false`

Second, the logic to determine what type of policy we have, is a bit convoluted from the network side:

```
useApproval = true, no agreement hash -> KYC only
useApproval = false, agreement hash -> Agreement only
useapproval = true, agreement hash -> KYC and agreement
```

So that means that your values, need to look something like this:
```suggestion
      {
        paramName: 'policy',
        validation: yup.string().required(),
        defaultValue: '',
        title: MSG.whitelistColonyPolicySelectorTitle,
        type: ExtensionParamType.ColonyPolicySelector,
        options: [
          {
            value: true,
            label: MSG.whitelistColonyPolicySelectorAgreementOnly,
            name: 'policy',
            appearance: {
              theme: 'traditional',
            },
            checked: false,
          },
          {
            value: false,
            label: MSG.whitelistColonyPolicySelectorKYCOnly,
            name: 'policy',
            appearance: {
              theme: 'traditional',
            },
            checked: false,
          },
          {
            value: true,
            label: MSG.whitelistColonyPolicySelectorAgreementAndKYC,
            name: 'policy',
            appearance: {
              theme: 'traditional',
            },
            checked: false,
          },
        ],
      },
```

src/modules/core/components/Fields/Radio/CustomRadio.tsx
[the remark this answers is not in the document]
rdig: Yeah, this is not a good name :(

Please find something more suitable _(I don't know, maybe `fatCircles`, which while still sounding silly, has the benefit of better describing how the theme looks)_

src/modules/dashboard/components/Extensions/ColonyPolicySelector/ColonyPolicySelector.tsx
```diff
@@ -0,0 +1,34 @@
...
+    <CustomRadioGroup
+      appearance={{ direction: 'horizontal', gap: 'medium' }}
+      options={options}
+      currentlyCheckedValue={selectedPolicy}
+      name="policy"
+    />
+  </div>
+);
+
+export default ColonyPolicySelector;
```
rdig, line 34: Missing `displayName`

src/modules/dashboard/components/Extensions/ColonyPolicySelector/ColonyPolicySelector.tsx
```diff
@@ -0,0 +1,34 @@
...
+  policy: string;
+}
+
+interface Props {
+  selectedPolicy: string;
+  title: MessageDescriptor | string;
+  options: CustomRadioProps[];
+}
+
+const ColonyPolicySelector = ({ selectedPolicy, title, options }: Props) => (
```
rdig, line 19: Please put this (sub)component under the `Whitelist` extension's folder, not the `Extensions` folder

src/modules/dashboard/components/Extensions/WhitelistExtensionBody/WhitelistExtensionBody.tsx
```diff
@@ -1,7 +1,55 @@
...
+          <>
+            {extensionInitParams.map((extensionParam) => (
+              <div key={extensionParam.paramName}>
+                {extensionParam.type ===
+                  ExtensionParamType.ColonyPolicySelector && (
+                  <ColonyPolicySelector
+                    selectedPolicy={values.policy}
+                    title={extensionParam.title}
+                    options={extensionParam.options || []}
+                  />
```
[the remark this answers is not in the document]
rdig: I mean, you shouldn't be able to get to this page when you either don't have a wallet connected, or don't have the correct permissions. 

This should only be visible on the "enable extension" page, the one which shows the initialization params

src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
```diff
@@ -375,7 +375,10 @@ const ExtensionDetails = ({
                   )}
                   {extensionEnabled &&
                     extension.enabledExtensionBody &&
-                    extension.enabledExtensionBody()}
+                    extension.initializationParams &&
+                    extension.enabledExtensionBody(
+                      extension.initializationParams,
+                    )}
```
rdig, line 381: Why are the initialization params passed here to the extension body ?

Params should be visible only when enabling the extension, the body is visible **after** you enable it

rdig: Yeah, this doesn't work. At all!

You need the initialization params available on the extension enable page, otherwise the values won't be available to you to be able to send to the "enableExtension" saga.

rdig (changes requested):
Styles-wise this looks good, however I'm not satisfied with how you implemented this from a logic standpoint as it does nothing from what it's supposed to do.

Screenshot from 2021-07-08 13-49-01

src/data/staticData/extensionData.ts
ArmandoGraterol: In that case, I'll use numbers as values: `0` for KYC only, `1` for agreement only, and `2` for agreement and KYC, because if we use `true/false` then the UI won't be able to tell the difference between selecting agreement only and agreement and KYC.

Then when the saga is connected, we just do something like `!!policy` to transform the numbers into boolean before sending them.

src/modules/dashboard/components/Extensions/ColonyPolicySelector/ColonyPolicySelector.tsx
ArmandoGraterol: A `WhitelistExtension` folder doesn't exist, I'll create it then

rdig: I think Alicja already created it.

Make sure to check after you rebase.

alicjakujawa (approved):
Now its fine :) In future we probably would need to make this selector more reusable, because im sure it will be used often, not only for "policy". But lets worry about it when we will have other need :)

chinins (approved):
It seems that we both misunderstood the flow. Working as expected now :)

PR #2586, opened 2021-07-04
Force OneTxPayment Extension Upgrade to v2

rdig:
## Description

This PR forces a colony that has a `OneTxPayment` extension `v1` installed, to upgrade to version `2` before being allowed to do any more actions on said colony _(colony version upgrades take precendence)_

This is because `v1` of `OneTxPayment` is incompatible with motions and disputes _(ie: `VotingReputation`)_

This is not a problem for newly created colonies, since they will all install `v2` by default, with no option to fall back to `v1`.

**Testing**:

Just replace the version returned by the resolver to return `1`, if the extension is `OneTxPayment` in here: https://github.com/JoinColony/colonyDapp/blob/77ee839925fc68f03b5e8d170d488a9be93ff1fc/src/data/resolvers/extensions.ts#L127
With:
```js
version: extensionId === Extension.OneTxPayment ? 1 : version.toNumber(),
```

**Changes**

- [x] Added `ExtensionUpgrade` component
- [x] Added `oneTxMustBeUpgraded` check
- [x] Refactored: `colonyUpgrade` checks
- [x] Wired above upgrade checks into `ColonyFundingMenu`, `ColonyDomainSelector`, `ColonyMembers`, `ColonyHomeActions`

**Demo**

Screenshot from 2021-07-04 16-19-03

Screenshot from 2021-07-04 16-19-15

DEV-415

### DIFF
--- a/src/modules/dashboard/checks.ts
+++ b/src/modules/dashboard/checks.ts
@@ -6,18 +6,24 @@ import { hasRoot } from '../users/checks';
 /*
  * Colony
  */
-export const canBeUpgraded = (colony?: Colony, networkVersion?: string) =>
+export const colonyCanBeUpgraded = (colony?: Colony, networkVersion?: string) =>
   colony?.version &&
   networkVersion &&
   parseInt(networkVersion, 10) > parseInt(colony.version, 10);
 
-export const mustBeUpgraded = (colony?: Colony, networkVersion?: string) =>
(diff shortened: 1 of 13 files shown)


alicjakujawa (approved):
works fine :)

chinins (approved):
All good! Working as expected.

PR #2592, opened 2021-07-07
Wallet Auto-Login 

### LINKED ISSUE #2591: Wallet auto-login
- If the user connected a wallet before
- auto-login to it without the need for an extra click
- this applies to both metamask and software wallets (dev mode)

rdig:
## Description

This PR refactors both the Metamask wallet and the Software Wallet _(Ganache for local development)_ to auto-login, removing the need for an extra click. This however, only if the user connected a wallet previously.

**Changes**

- [x] export `getAccounts` method from `StepGanache`
- [x] Refactor `useWalletAutoLogin` to also handle the Ganache wallet
- [x] Refactor `UserNavigation` to trigger a wallet autologin

**Demo**

demo-wallet-autologin

Resolves #2591


alicjakujawa (approved):
No problem here 🔥

src/modules/pages/components/RouteLayouts/UserNavigation.css
```diff
@@ -168,3 +168,20 @@
   border-radius: var(--radius-normal);
   background-color: color-mod(var(--temp-grey-blue-7) alpha(10%));
 }
+
+.walletAutoLogin {
+  margin-right: 12px;
+  padding: 3px 10px 0;
+  height: 24px;
```
chinins, line 175: The user reputation on the left to the `connecting wallet` part is 26px - 2 px more than walletAutoLogin (and it does look a bit bigger as per the picture).
Is it intentional? I didn't find the figma specs to check.

Screenshot 2021-07-12 at 11 58 57

chinins (approved):
Working well! It would be great if we could also add this to the Whitelist branch (cherry pick or rebase on master) as it does help when you develop and test things. I can do that after you merge if you want.

src/modules/pages/components/RouteLayouts/UserNavigation.css
rdig: _Ahahahaha specs!_ There's no specs :)

But no, it's not intentional, I just did what looked good to me :P

But I'll happily increase it to `26` so that it's the same height

rdig:
@chinins 

> It would be great if we could also add this to the Whitelist branch (cherry pick or rebase on master) 

After I merge it, you can rebase the Whitelist feature branch on `master`, so that it will bring in this change

PR #2595, opened 2021-07-08
Wire up Enable Whitelist extension

alicjakujawa:
## Description

This PR wire up enabling whitelist extension

**New stuff** ✨

* Added WhitelistClient to ColonyManager
* Added message descriptors for whitelist initialisation
* Added validation for texarea & policy selector combo
* Added PolicyType enum

**Changes** 🏗

* Updated colonyExtensionEnable saga to handle Whitelist extension

Resolves DEV-438

### DIFF
--- a/src/data/staticData/extensionData.ts
+++ b/src/data/staticData/extensionData.ts
@@ -544,7 +544,7 @@ const extensions: { [key: string]: ExtensionData } = {
       },
       {
         paramName: 'agreement',
-        validation: yup.string(),
+        validation: yup.string().min(100),
         defaultValue: '',
         title: MSG.agreementTitle,
         description: MSG.agreementDescription,
--- a/src/data/staticData/extensionData.ts
+++ b/src/data/staticData/extensionData.ts
@@ -29,6 +29,7 @@ export interface ExtensionInitParams {
   validation: object;
   type: ExtensionParamType;
   options?: CustomRadioProps[];
+  disabled?: (props: any) => boolean;
 }
 
 export interface ExtensionData {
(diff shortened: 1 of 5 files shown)


src/data/staticData/extensionData.ts
```diff
@@ -15,6 +15,12 @@ export enum ExtensionParamType {
   ColonyPolicySelector = 'ColonyPolicySelector',
 }
 
+export enum PolicyType {
```
rdig, line 18: Very nice. I like this :+1:

src/modules/dashboard/sagas/extensions.ts
```diff
@@ -165,18 +167,43 @@ function* colonyExtensionEnable({
       throw new Error('Extension not installed');
     }
 
+    /*
+     * Upload whitelist policy to IPFS
+     */
+    let agreementHash = '';
+    if (
+      extensionId === Extension.Whitelist &&
```
rdig, line 175: So I'm not particularly happy with this, but right now it's the best we can do.

However, if we'll have to do this for more extensions, we'll need to think about refactoring this saga.

The way I'm thinking about this to work in the future, is something like each extension could have it's own individual _enable_ saga _(same saga for installation, just the enable saga to be different)_

But we can see about that, if we get to that point.

rdig (approved):
Very nice work @alicjakujawa! :100: :1st_place_medal: 

The only think I you need to pay attention on, is that I think that some rebase didn't work properly for this PR, as I don't think the first 3 commits on this branch were actually added in this branch:

Screenshot from 2021-07-13 14-47-35

Please investigate this before merging.

PR #2597, opened 2021-07-08
Add CSV uploader functionality

ArmandoGraterol:
## Description

- Add CSV uploader and parser

**New stuff** ✨

* Added `papaparse` package for CSV parsing
* Added `textSpace` appearance prop to InputStatus
* Added appearance interface of InputStatus as export
* Added `customErrorMessage` and `inputStatusAppearance` as props for FileUpload error customization.
* Added loading state  to `FileUpload`, `CSVUploader` and item, and widget for when a CSV file is being processed.
* Added `layout` appearance prop to `SpinnerLoader`.
* Added CSVUploader component
* Added CSVUploaderItem sub-component
* Added validation for CSVUploader component on UploadAddressesWidget

**Changes** 🏗

* Changed accepted and rejected files data structures on FileUpload to include errors inside the file object itself

FireShot Capture 380 - Colony - localhost

FireShot Capture 381 - Colony - localhost

FireShot Capture 382 - Colony - localhost

FireShot Capture 383 - Colony - localhost

FireShot Capture 384 - Colony - localhost

Resolves DEV-447


src/modules/dashboard/components/Whitelist/CSVUploader/CSVUploader.tsx
```diff
@@ -0,0 +1,95 @@
...
+        header: true,
+      });
+    } else if (!CSVFile && parsedCSV) {
+      setParsedCSV(null);
+    }
+  }, [CSVFile, parsedCSV]);
+
+  useEffect(() => {
+    if (parsedCSV && value[0] && isNil(value[0].parsedData)) {
+      const headerExpectedName = 'Whitelist Address';
```
rdig, line 50: We don't really expect this exact header name to be in the file, we just expect **a header** to be available on the first line

src/modules/dashboard/components/Whitelist/UploadAddressesWidget/UploadAddressesWidget.tsx
```diff
@@ -35,20 +36,38 @@ const MSG = defineMessages({
...
+});
+
+const validationSchema = yup.object({
+  whitelistCSVUploader: yup.array().of(
+    yup.object().shape({
+      parsedData: yup
+        .array()
+        .of(yup.string())
+        .min(1, () => MSG.badFileError)
+        .max(1000, () => MSG.uploadError),
```
rdig, line 52: How did you arrive at the max value here, since by my calculations it should be somewhere around 4200 _(42 chars in an address x 100 addresses)_

Although the min/max values don't really work, as the min value should be at least the length of an address which is 42

src/modules/dashboard/components/Whitelist/CSVUploader/CSVUploader.tsx
```diff
@@ -0,0 +1,95 @@
...
+      if (
+        parsedCSV.meta.fields?.length === 1 &&
+        parsedCSV.meta.fields[0] === headerExpectedName
+      ) {
+        const validAddresses = parsedCSV.data.flatMap(
+          (CSVRow: Record<string, any>) => {
+            const potentialAddress: string = CSVRow[headerExpectedName];
+            if (potentialAddress) {
+              return isAddress(potentialAddress) ? [potentialAddress] : [];
```
rdig, line 59: We need a dedicated message when one of the addresses in the file is invalid. Something like:
- ```.csv file contains an invalid address`` ... or something along those lines _(maybe ask product)_

I think you can use `setError` available from the `useField` hook to just set that error message manually.

rdig (changes requested):
Very good work here. Way to go @ArmandoGraterol :tada: 

Overall the code is very good, the only thing I rejected this PR for is that I think we need a bit more messages for what exactly is wrong with the file validation. _(see my comments below)_

Also, check this message:
Screenshot from 2021-07-15 14-26-08
As I think it's a bit weird _(maybe double-check with product)_

Other than a few messages to be added / changed, this PR is good to go!

src/modules/core/components/FileUpload/types.ts
```diff
@@ -11,9 +11,10 @@ export interface FileReaderFile {
 
 export interface UploadFile {
   file: File;
-  uploaded?: string;
+  uploaded?: string | boolean;
```
alicjakujawa, line 14: why sometimes its a string and sometimes boolean? cant we make it more generic?

src/modules/dashboard/components/Whitelist/CSVUploader/CSVUploader.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const handleUploadError = async () => {
+    if (isNil(value[0].parsedData)) {
+      await setValue([{ ...value[0], parsedData: [] }]);
+    }
+  };
+
+  useEffect(() => {
+    if (CSVFile && !parsedCSV) {
+      Papa.parse(CSVFile, {
```
alicjakujawa, line 39: Papa 😆 i feel that not only i have problem with naming, but authors of libs have same problem xD

alicjakujawa (approved):
I really like this, great job 🔥  Apart of Raul comments i would consider to move out this component from Whitelist folder, as it can be easily reused in future CSV uploads :P

chinins (commented):
Great work especially for such a difficult  problem to solve. No more comments other than what Raul has already mentioned. I am happy to have final look and approve once you solve those.

src/modules/dashboard/components/Whitelist/CSVUploader/CSVUploader.tsx
ArmandoGraterol: I'm gonna make it dynamic then :ok_hand:

src/modules/dashboard/components/Whitelist/UploadAddressesWidget/UploadAddressesWidget.tsx
ArmandoGraterol: Those limits are for the array of parsed data itself. So if we upload a file then the array with all of the parsed and validated addresses must be within 1 - 1000, otherwise, we throw an error.

The validation of the addresses themself are done in another file

src/modules/core/components/FileUpload/types.ts
ArmandoGraterol: Hmmm I added boolean because in other instances where the `uploaded` is used (AvatarUploadItem and UploadItem) that prop stores a string that represents the uploaded file, for whatever reason that I don't really know because from what I've seen this prop is used to check if the file was already uploaded. (Perhaps @rdig knows more about why this prop is used as a `string`)

Seeing that the prop is used on my case only the check if the .csv file was uploaded I decided it would be more appropriate for it to be a boolean, but I could change it to string and store whatever in it just for consistency sake.

src/modules/dashboard/components/Whitelist/CSVUploader/CSVUploader.tsx
ArmandoGraterol: You should check the `FAQ` of this lib, the author really likes the names and it shows on the puns that he shamelessly does

rdig (approved):
Very nice work once again! :100:

chinins (approved):
Great job! 🔥

PR #2606, opened 2021-07-13
Estimate Gas Cost prevent gasLimit Race Condition

rdig:
## Description

This PR fixes the `estimateGasCost` saga for cases where we both manually estimate a transaction _(ie: `finalizeMotion`)_ as well as rely on the automatic gas price estimator coming from the various gas station oracles (EthGasStation or BlockScout)

The fix here involves checking if the `gasLimit` value is already set, prevent overwriting it.


chinins (approved):
Looking good!

PR #2607, opened 2021-07-13
Add download template

chinins:
## Description

This PR adds download template for Whitelist Addresses.

Raul's guidance was NOT to use the file hosted on the app (he revised the initial solution as it's just easier to create the file on the go).

**New stuff** ✨

- add `DownloadTemplate` component
- add `download` prop in `ExternalLink`

**Changes** 🏗

- add `DownloadTemplate` to `UploadAddressesWidget`
- update css in `UploadAddressesWidget`

Resolves DEV-443

The second picture is how the template should look like (opened with Numbers on my mac)

Screenshot 2021-07-14 at 11 27 53

Screenshot 2021-07-13 at 13 41 26


src/modules/core/components/ExternalLink/ExternalLink.tsx
```diff
@@ -50,6 +52,7 @@ const ExternalLink = ({
       target="_blank"
       rel="noreferrer noopener"
       title={title}
+      {...(download !== undefined ? { download } : {})}
```
ArmandoGraterol, line 55: What's happening here? I don't quite get it

ArmandoGraterol (approved):
LGTM :sunglasses:

src/modules/core/components/ExternalLink/ExternalLink.tsx
rdig: Should't this use `typeof` ?

```suggestion
      {...(typeof download === 'string' ? { download } : {})}
```

rdig (approved):
Very nice work @chinins! :100:

src/modules/core/components/ExternalLink/ExternalLink.tsx
chinins: if the prop is not passed, then it's not taken into account. I prefer to have this extra check so that it doesn't screw other links.

PR #2612, opened 2021-07-20
Move extensions sagas to separate files

chinins:
## Description

This PR refactors moves all of the extensions saga to separate files in `extensions` folder - similar to `actions` & `motions`.

To test you will need to restart the webpack (at least it was the case for me).

**Changes** 🏗

- move sagas
- simplify `updateMotionsValues` imports

Resolves DEV-460


src/modules/dashboard/sagas/extensions/index.ts
```diff
@@ -0,0 +1,25 @@
...
+    ActionTypes.COLONY_EXTENSION_DEPRECATE,
+    colonyExtensionDeprecate,
+  );
+  yield takeEvery(
+    ActionTypes.COLONY_EXTENSION_UNINSTALL,
+    colonyExtensionUninstall,
+  );
+  yield takeEvery(ActionTypes.COLONY_EXTENSION_UPGRADE, colonyExtensionUpgrade);
+  yield takeEvery(ActionTypes.WHITELIST_UPDATE, updateWhitelist);
+}
```
rdig, line 25: Each individual `takeEvery` for each individual action, needs to be added to each own's saga.

This is so that the action _(ex: `COLONY_EXTENSION_UPGRADE`)_ is in the same file as it's used so that it will make debugging easier.

The structure should look something like this:

Saga file (example `colonyExtensionUpgrade`):
```js
function* colonyExtensionUpgrade({ ... }); // notion that there's no export for this function

export default function* colonyExtensionUpgradeSagas() { // this is what it's actually exported
  yield takeEvery(ActionTypes.COLONY_EXTENSION_UPGRADE, colonyExtensionUpgrade);
};
```
Inside the `sagas/extensions/index`:
```js
import { colonyExtensionUpgradeSaga } from './colonyExtensionUpgrade';

export default function* extensionsSagas() {
  yield all([
    call(colonyExtensionUpgradeSaga),
  ]);
}
```

Then inside `sagas/index` add the `extensionsSagas` to the `setupDashboardSagas`:
```js
import extensionsSagas from './extensions';

export default function* setupDashboardSagas() {
  yield all([
    ...
    call(extensionsSagas),
  ]);
}
```

rdig (changes requested):
So this is all good, except for the fact that I want the actions `takeEvery`'s inside each saga's file. _(I've left a comment about this)_

You can follow the example from `motions` about how this is set up properly.

I just rejected it because of that, once you refactor it, this is good to go.

rdig (approved):
Very nice refactor :100: Good to go now!

PR #2613, opened 2021-07-20
Fix Extension Installation on Allowed Networks

rdig:
## Description

This PR fixes a issue introduced in #2605 that prevents installation _(actually all control: install / enable / upgrade / uninstall)_ of Extensions, when the user has the wallet connected to one of the allowed networks: ethereum mainnet or xdai

**Changes**

- [x] Update `colonyNetwork` to the current `dwlss` release
- [x] Fix `ExtensionDetails` to allow extension controls when on an _"Allowed Network"_


chinins (approved):
Seems to be working fine!

ArmandoGraterol (approved):
It's lit :fire:

PR #2615, opened 2021-07-21
Properly estimate "Create Token" and "Deploy Token Authority"  transactions when creating a new Colony

### LINKED ISSUE #2602: Colony Creation Token Transactions are not being gas estimated
It seems that two of transactions that we fire when creating a new colony:
- Create Token
- Set Token Authority
- _(please check if they are the only ones)_

Are not properly gas estimated as they are showing the default `20` gwei each time.

rdig:
## Description

This PR _"fixes"_ an issue we were seeing with two of the colony create transactions not being properly estimated in terms of gas prices:
- Create Token
- Deploy Token Authority

The actual fix actually comes from `colonyJS` since this was caused by our "extra" methods in `colonyJS` not supporting transactions overrides. See the changes here for more context: https://github.com/JoinColony/colonyJS/pull/496/files

_While I was at it, I cleaned up some token-related sagas and actions that were no longer used by the dapp_

**Testing**
- Connect a Metamask wallet
- Create a colony
- Check gas prices for the transactions involved _(since you're on `ganache` they should all be `1`)_

**Changes**

- [x] Updated `colonyJS` to the latest release _(this what actually fixed the issue)_
- [x] Removed **ununsed** `token` saga
- [x] Removed **ununsed** `token` redux actions
- [x] Removed **ununsed** `token` redux action types

Resolves #2602


chinins (approved):
Looks good! 🔥

PR #2616, opened 2021-07-21
Wire up adding whitelist addresses

chinins:
## Description

This PR wires up adding addresses via direct input & a csv file.

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

- move agreement dialog & confirm buttons inside `UploadAddressesWidget`
- updated payload args in `whitelist_update` actions & saga
- update css & html

**Deletions** ⚰️

- remove unused adding addresses actions

Resolves DEV-445


alicjakujawa (changes requested):
Its not yet done. Like in description - this need to handle batch logic. You need to update whitelistUpdate saga to handle batch logic

alicjakujawa (approved):
Works fine 🔥  (im sure product team on qa will find out, that we actually need some state for uploader while transactions are sending) but we will then create separate issue!

src/modules/dashboard/components/Whitelist/Whitelist.tsx
```diff
@@ -41,32 +22,10 @@ const Whitelist = ({ colony: { colonyAddress }, colony }: Props) => {
...
-        agreementHash: agreementHashData?.whitelistAgreementHash,
-      }),
-    [openAgreementDialog, agreementHashData],
-  );
 
   return (
     <div>
-      <UploadAddressesWidget />
+      <UploadAddressesWidget colony={colony} />
       {usersLoading && <MiniSpinnerLoader loadingText={MSG.loadingText} />}
```
rdig, line 29: So my only note here is that I would like to see a loading spinner here, after the last address upload transaction succeeds, so that I know the list needs to be updated.

However, that is actually hard to implement in the current state of the dapp. We'll have to wait and see how long it takes _(in production and on QA)_ from the time you add addresses to the list, to the time it takes for them to show up.

If the gap is not that big, we might not want to bother.

src/modules/dashboard/components/Whitelist/UploadAddressesWidget/UploadAddressesWidget.tsx
```diff
@@ -35,10 +53,6 @@ const MSG = defineMessages({
     id: `dashboard.Whitelist.UploadAddressesWidget.uploadError`,
     defaultMessage: `We do not accept more than 100 addresses at a time, please upload a smaller amount.`,
   },
-  inputError: {
-    id: `dashboard.Whitelist.UploadAddressesWidget.inputError`,
-    defaultMessage: `TODO`,
-  },
```
rdig, line 41: Is this not used anymore?

rdig (approved):
All good from my side as well, with a few notes I left at the bottom.

Very nice work @chinins :100: 

Screenshot from 2021-07-27 17-55-14

ArmandoGraterol (approved):
Works great for me :+1:

src/modules/dashboard/components/Whitelist/UploadAddressesWidget/UploadAddressesWidget.tsx
chinins: no, there is a validation error coming from the form so not used any more.

PR #2445, opened 2021-04-18
Coin Machine Buy Tokens Widget

rdig:
## Description

This PR implements the "Buy Tokens" widget to the Coin Machine page, that allows the user the ability to purchase a certain amount of tokens in the current sale interval.

**Note** that this escalated is scope from when it was initially spec'd out, due to the previous team's failure to implement it's UI originally. See #2206 if you think you can find it.

**Testing**
- Create a colony
- Mint tokens _(1000+)_
- Unlock the token
- Install Coin Machine 
- Enable with _Default Values_
- Send 100+ tokens to the Coin Machine contract _(copy the contract address and create a payment action, with the address as a recipient)_
- Go buy some tokens on the Coin Machine page
- **Note that you can only buy a max of 100 tokens per period, so if you reach that limit, the buy widget will not allow you to buy more, until the next period starts _(and since we don't have UI for that yet, you won't be able to tell)_**

**Changes**
- [x] Cherry picked `QuestionMarkTooltip` from the #2413 branch
- [x] Added global `learnMore` message descriptor _(and replaced it where it was being used)_
- [x] Fixed `Chat` component shadow
- [x] Fixed core `Heading` `medium` appearance size letter spacing _(to be in line with the spec)_
- [x] Added `BuyTokens` component
- [x] Added Coin Machine sagas layout
- [x] Added Coin Machine `buyTokens` saga
- [x] Added Coin Machine redux actions and types
- [x] Added Coin Machine resolvers
- [x] Added Coin Machine GraphQL queries and typedefs

**Demo**
demo-cm-buy-tokens

Resolves DEV-245

PR #2618, opened 2021-07-22
Add SaleStateWidget

alicjakujawa:
## Description

This PR adds in SaleStateWidget

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves DEV-454


ArmandoGraterol (changes requested):
Solid work :ok_hand: I think I just noticed some small things:

Firstly, I noticed is that I think the button displayed at the bottom of the widget (in all of the states) should have more space between itself and the text above.

 Design:
FireShot Capture 400 - CoinMachine – Figma - www figma com

Implementation:
FireShot Capture 401 - Colony - localhost

Secondly, the buttons have less width, and also have different widths between themselves, when in the specs all of them have a standard width of `160px`.

Lastly, and least important (in my opinion),  is that the correct _final_ format for the timer? In the design, the timer is shown in a format like `00:00:00`. It's just a question, as I'm happy with this format (because it's consistent with the format we use for the timer of the motions), but it's also different from the design.

alicjakujawa:
> Solid work 👌 I think I just noticed some small things:
> 
> Firstly, I think the button displayed at the bottom of the widget (in all of the states) should have more space between itself and the text above.
> 
> Design:
> FireShot Capture 400 - CoinMachine – Figma - www figma com
> 
> Implementation:
> FireShot Capture 401 - Colony - localhost
> 
> Secondly, the buttons have less width, and also have different widths between themselves, when in the specs all of them have a standard width of `160px`.
> 
> Lastly, and least important (in my opinion), is that the correct _final_ format for the timer? In the design, the timer is shown in a format like `00:00:00`. It's just a question, as I'm happy with this format (because it's consistent with the format we use for the timer of the motions), but it's also different from the design.

@ArmandoGraterol yes, timer is correct, you can see my talk with Karol on product-private channel.

chinins (approved):
Other than Armando's comments, everything seems to be good! Great job!

ArmandoGraterol (approved):
Looks really good to me :ok_hand: Nice work!

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
```diff
@@ -100,17 +103,32 @@ const CoinMachine = ({
...
-        </div>
+        {(saleStarted && (
+          <div className={styles.saleStarted}>
+            <SaleStateWidget
+              state={SaleState.Success}
+              price="1234234340000000"
+              amount="123423434000000000000"
+              timeLeftToNextSale={864000}
+              nativeToken={{ symbol: 'CLNY', decimals: 18, name: 'CLNY' }}
+              transactionToken={{ symbol: 'ETH', decimals: 18, name: 'ETH' }}
```
rdig, line 114: Just a heads up that these tokens might not actually be the colony's native token or ETH/XDAI. They can be _any_ token the colony has.

So I guess, better names for these props could be `soldToken` and `receivedToken`

_(this is just a nit-pick I noticed while rebasing on your changes, it's not something major, so you can change it when you wire in the logic)_

rdig: Never mind, I fixed it in my #2445 

Screenshot from 2021-07-27 19-30-33

PR #2628, opened 2021-07-29
Fix motion feed comment box margin

chinins:
## Description

This PR fixes the margin between the motion feed & the comment box

**Changes** 🏗

- added margin to the comment box in the motion feed

For the change to work I had to restart the webpack - just in case you can't see it straight away.

Resolves issue 2627

Screenshot 2021-07-29 at 15 08 07


rdig (approved):
Thank you for digging into this to find the missing code :100:

PR #2634, opened 2021-07-30
Update fetch policy of useWhitelistedUsers query

### LINKED ISSUE #2630: Whitelist extension - if uninstalled / reinstalled, old data persists (temporarily)
## Steps to reproduce

1. Install and enable whitelist extension
2. Add some whitelisted addresses
3. Deprecate and uninstall extension
4. Install and enable extension again 

## Expected behaviour

No whitelisted addresses are shown

## Actual behaviour

The addresses that were whitelisted on the previous instance of the contract are shown once more on this new instance. Upon a refresh, they are gone.

ArmandoGraterol:
Updated fetch policy to `network-only` to avoid the list not updating when you uninstall the extension and install it again without reloading the page. 

Hypothetically, it also wouldn't update in cases where another user apart from yourself adds an address to the list and you don't reload the page when re-entering the extension page to see the added users.

Resolves #2630

### DIFF
--- a/src/modules/dashboard/components/Whitelist/Whitelist.tsx
+++ b/src/modules/dashboard/components/Whitelist/Whitelist.tsx
@@ -21,6 +21,7 @@ interface Props {
 const Whitelist = ({ colony: { colonyAddress }, colony }: Props) => {
   const { data: usersData, loading: usersLoading } = useWhitelistedUsersQuery({
     variables: { colonyAddress },
+    fetchPolicy: 'network-only',
   });
 
   return (


alicjakujawa (commented):
Shouldn't cache be updated when we refetch query after each whitelist update? I would avoid network-only policy for bigger list of objects, unless there is no other option

ArmandoGraterol:
> Shouldn't cache be updated when we refetch query after each whitelist update? I would avoid network-only policy for bigger list of objects, unless there is no other option

@alicjakujawa  From what I'm seeing, the query is actually executed again after the initial execution only if you reload the page or if you manually tell the query to refetch itself like we do every time a user is added to the list. The only thing I can think of right now that could be an alternative to this would be to refetch the whitelist data when we uninstall the extension, what do you think about that?

alicjakujawa:
> > Shouldn't cache be updated when we refetch query after each whitelist update? I would avoid network-only policy for bigger list of objects, unless there is no other option
> 
> @alicjakujawa From what I'm seeing, the query is actually executed again after the initial execution only if you reload the page or if you manually tell the query to refetch itself like we do every time a user is added to the list. The only thing I can think of right now that could be an alternative to this would be to refetch the whitelist data when we uninstall the extension, what do you think about that?

sure lets try like this!

chinins:
> > Shouldn't cache be updated when we refetch query after each whitelist update? I would avoid network-only policy for bigger list of objects, unless there is no other option
> 
> @alicjakujawa From what I'm seeing, the query is actually executed again after the initial execution only if you reload the page or if you manually tell the query to refetch itself like we do every time a user is added to the list. The only thing I can think of right now that could be an alternative to this would be to refetch the whitelist data when we uninstall the extension, what do you think about that?

Yeah, I think it might be a better approach.

alicjakujawa:
How is going with that?

PR #2635, opened 2021-08-02
Auto filled whitelist address in coin machine init params

### LINKED ISSUE #2631: Automatically add installed Whitelist extension address to Coin Machine initialization params
Currently, when initializing the Coin Machine extension, the _"whitelist address"_ param is set to the `null` address. This was because we didn't have the Whitelist extension when first started out building Coin Machine.

Screenshot from 2021-07-30 13-04-05

Now, that we have, we want to detect if Whitelist is installed _(and enabled, but not deprecated)_, and if it is, get it's contract address and add it to the init params of Coin Machine.

To do this, we need to do a bit of code maintenance. We need to build a `getClientAddress` local resolver, to which you pass the in the `clientType` _(it's the enum client type from `colonyJS`)_ and the current colony's `address`, and based on those values it returns that specific clients's address, if it exists _(you'll need a try/catch block)_

After this is done, you're going to wire that resolves inside `ExtensionSetup`, then pass that  value to the `initialValues` memo hook. 

We already have something like this setup for the native token address _(if the extension is Coin Machine)_ as well as a `@TODO` comment for exactly the thing this issues is handling:

https://github.com/JoinColony/colonyDapp/blob/feature/coin-machine-redux/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx#L111-L130

alicjakujawa:
## Description

This PR updates init params of coin machine extension. It fills whitelistAddress for coin machine extension IF whitelist extension was installed&enabled, if not it set whitelistAddress to AddressZero

**Changes** 🏗

* Updated useEnabledExtension hook
* Updated ExtensionSetup component

Resolves #2631

### DIFF
--- a/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
+++ b/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
@@ -133,7 +133,13 @@ const ExtensionSetup = ({
       };
     }
     return defaultValues;
-  }, [extensionId, initializationParams, nativeTokenAddress]);
+  }, [
+    extensionId,
+    initializationParams,
+    nativeTokenAddress,
+    isWhitelistExtensionEnabled,
+    whitelistAddress,
+  ]);
 
   if (
     installedExtension.details.deprecated ||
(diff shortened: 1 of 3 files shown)


chinins (approved):
Looks good!

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
```diff
@@ -108,6 +112,13 @@ const ExtensionSetup = ({
...
   );
 
+  const {
+    isWhitelistExtensionEnabled,
+    whitelistAddress,
+  } = useEnabledExtensions({
+    colonyAddress,
+  });
+
   const initialValues = useMemo(() => {
```
ArmandoGraterol, line 122: There are a few missing dependencies in the dependency array according to the linter

ArmandoGraterol (approved):
Works as expected! Just left a small comment but I'm approving anyways

PR #2638, opened 2021-08-03
User is whitelisted logic

alicjakujawa:
## Description

This PR adds in logic for checking if user is whitelisted, or signing agreement is required

**New stuff** ✨

* WhitelistPolicy resolver
* WhitelistPolicy queries

**Changes** 🏗

* Updated BuyTokens component to wire in logic which determine if user was whitelisted
* Updated updateWhitelist saga to refetch whitelistPolicy query after each change

**What to test** ⚰️
* When whitelist extension is NOT enabled, we shouldn't see "get whitelisted" button or "your account is whitelisted" text
* When whitelist extension is ENABLED - whitelisted on UI means, that user needs to fulfil policy requirement (KYC only, Agreement only, or BOTH, depend on policy) so just test all combination and check if correct state is displaying (Get whitelisted, Buy, Your account is whitelisted)

Resolves DEV-462

### DIFF
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -323,6 +323,7 @@ export type Query = {
   votingState: VotingState;
   whitelistAgreement: Scalars['String'];
   whitelistAgreementHash?: Maybe<Scalars['String']>;
+  whitelistPolicy?: Maybe<WhitelistPolicy>;
   whitelistedUsers: Array<User>;
 };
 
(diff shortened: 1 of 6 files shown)


chinins (approved):
Looks good! Nice work!

ArmandoGraterol (approved):
It seems to be working as expected! Solid work :+1:

PR #2640, opened 2021-08-04
Wire up time remaining widget

ArmandoGraterol:
Added logic to time remaining widget

FireShot Capture 407 - Colony - localhost

FireShot Capture 408 - Colony - localhost

FireShot Capture 409 - Colony - localhost

Resolves DEV-465

### DIFF
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -77,7 +77,7 @@ export const coinMachineResolvers = ({
         return null;
       }
     },
-    async coinMachinePeriodTimeRemaining(_, { colonyAddress }) {
+    async coinMachineSalePeriod(_, { colonyAddress }) {
       try {
         const { networkClient } = colonyManager;
         const coinMachineClient = await colonyManager.getClient(
@@ -86,10 +86,14 @@ export const coinMachineResolvers = ({
         );
         const periodLength = await coinMachineClient.getPeriodLength();
         const blockTime = await getBlockTime(networkClient.provider, 'latest');
-        const timeRemaining = bigNumberify(blockTime).mod(
-          periodLength.mul(1000),
+        const periodLengthInMs = periodLength.mul(1000);
+        const timeRemaining = periodLengthInMs.sub(
+          bigNumberify(blockTime).mod(periodLengthInMs),
         );
-        return timeRemaining.toString();
+        return {
+          periodLength: periodLength.toString(),
+          timeRemaining: timeRemaining.toString(),
+        };
       } catch (error) {
         console.error(error);
         return null;
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -1,5 +1,6 @@
-import { ClientType } from '@colony/colony-js';
+import { ClientType, getBlockTime } from '@colony/colony-js';
 import { Resolvers } from '@apollo/client';
+import { bigNumberify } from 'ethers/utils';
 
 import { Context } from '~context/index';
 import { createAddress } from '~utils/web3';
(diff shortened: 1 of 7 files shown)


src/modules/dashboard/components/CoinMachine/RemainingDisplayWidget/RemainingDisplayWidget.tsx
```diff
@@ -101,7 +108,10 @@ const RemainingDisplayWidget = ({
 
     return <FormattedMessage {...widgetText.placeholder} />;
   }, [displayType, splitTime, value, widgetText]);
-
+  const showValueWarning =
+    displayType === DataDisplayType.Time && appearance.theme !== 'danger'
+      ? (splitTime?.hours || 0) < 1
+      : false;
```
alicjakujawa, line 114:
```suggestion
      displayType === DataDisplayType.Time && appearance.theme !== 'danger' && (splitTime?.hours || 0) < 1
```

alicjakujawa (changes requested):
Few things here:
1. Changing to danger needs to be depended on periodLength variable (asked on product, about exact rules, so just take a look there, when they response)
2. We need to again refetch queries after extension installation, as currently if i uninstall and install coin machine with different period length, if i will go to buy tokens page without reload i will see old time. I think that the best option will be to do something similar to updateMotionValues method, as we have more and more cases when we need to refetch query after reinstall
3. Last thing: Currently if times go to 0, and you wont reload page you will see "N/A". Instead we should see new period started to count down

src/data/resolvers/coinMachine.ts
```diff
@@ -76,5 +77,23 @@ export const coinMachineResolvers = ({
...
+    async coinMachinePeriodTimeRemaining(_, { colonyAddress }) {
+      try {
+        const { networkClient } = colonyManager;
+        const coinMachineClient = await colonyManager.getClient(
+          ClientType.CoinMachineClient,
+          colonyAddress,
+        );
+        const periodLength = await coinMachineClient.getPeriodLength();
+        const blockTime = await getBlockTime(networkClient.provider, 'latest');
+        const timeRemaining = bigNumberify(blockTime).mod(
```
chinins, line 89: I think that the logic is not complete here. How can we now how much time left if we don't take into account the time of the last period change? 
It seems to me that the formula need to be something like:

`periodLength - (latest block time - latest period change block)` 

Or am I missing something?

chinins (commented):
I agree with Alicja's comments. Plus, I have a question (please, see my code comment).

src/data/resolvers/coinMachine.ts
ArmandoGraterol: You were right in that something was missing here, the formula would actually be `period length - (latest block timestamp % period length)`. Just asked Alex for clarification on this.

As for how do we know when did the last period change occur, that is what the `latest block timestamp % period length` does, if the result of that is `0`, then that means that the period just began. 

That's at least what I understood from what Alex explained to me.

ArmandoGraterol:
> 2. We need to again refetch queries after extension installation, as currently if i uninstall and install coin machine with different period length, if i will go to buy tokens page without reload i will see old time. I think that the best option will be to do something similar to updateMotionValues method, as we have more and more cases when we need to refetch query after reinstall

So, what I'm gonna do is update the fetch policy to network-only, as that fixes this problem and also fixes the problem of the timer  possibly resetting itself if the user leaves the page and comes back later. Leaving it in `network-only` shouldn't be a problem as it's not a heavy query. Should kill 2 birds with 1 stone

src/modules/dashboard/components/CoinMachine/RemainingDisplayWidget/RemainingDisplayWidget.tsx
```diff
@@ -101,7 +111,11 @@ const RemainingDisplayWidget = ({
 
     return <FormattedMessage {...widgetText.placeholder} />;
   }, [displayType, splitTime, value, widgetText]);
-
+  const showValueWarning =
+    appearance.theme !== 'danger' &&
+    typeof value === 'number' &&
+    periodLength &&
+    ((value / 1000) * 100) / periodLength <= 10;
```
alicjakujawa, line 118: it should be last 10% not last 10 seconds from what i read on product-private

alicjakujawa (approved):
Nice work, please add periodLength to dependencies array as eslint is talking :P

src/data/resolvers/coinMachine.ts
chinins: So it still seems to me that something is off in your current code. Your current formula is this:
`const timeRemaining = periodLengthInMs.sub(bigNumberify(blockTime).mod(periodLengthInMs),);`

I don't understand how that formula is calculating time remaining. It doesn't take into account when last period change happened (and in my kitchen logic we need that timestamp).

I think that my testing confirms it. When I installed coin machine and go to the Buy tokens page, I saw about 30 min time remaining although I suppose it should be 60 min left (I enabled with default values and the period length is 60 min).

I don't want to be difficult but I don't want to approve before I understand this. Maybe we can have a call together with Alex to go through that as it's an important point.

chinins (commented):
Everything looks good except for the timeRemaining formula (as per my separate comment).

chinins (approved):
Thank you for organising the call with Alex to explain me the periods time keeping - very useful!

All good now! Nice work

PR #2644, opened 2021-08-05
Add Sale started/Sale paused logic to coin machine

ArmandoGraterol:
Added logic to detect if the sale started or was paused due to a lack of tokens to sell in Coin Machine

- [x] TODO: Update this PR once #2642 is merged, as that PR has reusable code for this one

* Coin machine with no token balance
FireShot Capture 417 - Colony - localhost

Resolves DEV-469

### DIFF
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -198,8 +198,6 @@ export const coinMachineResolvers = ({
         // eslint-disable-next-line max-len
         const targetPerPeriodTokens = await coinMachineClient.getTargetPerPeriod();
 
-        const tokenPeriodBalance = await coinMachineClient.getTokenBalance();
-
         return {
           maxPerPeriodTokens: maxPerPeriodTokens.toString(),
           activeSoldTokens:
@@ -207,8 +205,21 @@ export const coinMachineResolvers = ({
               ? activeSoldTokens.toString()
               : '',
           targetPerPeriodTokens: targetPerPeriodTokens.toString(),
-          tokenPeriodBalance: tokenPeriodBalance.toString(),
         };
+      } catch (error) {
+        console.error(error);
+        return null;
+      }
+    },
+    async coinMachineTokenBalance(_, { colonyAddress }) {
+      try {
+        const coinMachineClient = await colonyManager.getClient(
+          ClientType.CoinMachineClient,
+          colonyAddress,
+        );
+        const tokenBalance = await coinMachineClient.getTokenBalance();
+
+        return tokenBalance.toString();
       } catch (error) {
         console.error(error);
         return null;
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -198,6 +198,8 @@ export const coinMachineResolvers = ({
         // eslint-disable-next-line max-len
         const targetPerPeriodTokens = await coinMachineClient.getTargetPerPeriod();
 
+        const tokenPeriodBalance = await coinMachineClient.getTokenBalance();
+
         return {
           maxPerPeriodTokens: maxPerPeriodTokens.toString(),
           activeSoldTokens:
@@ -205,6 +207,7 @@ export const coinMachineResolvers = ({
               ? activeSoldTokens.toString()
               : '',
           targetPerPeriodTokens: targetPerPeriodTokens.toString(),
+          tokenPeriodBalance: tokenPeriodBalance.toString(),
         };
       } catch (error) {
         console.error(error);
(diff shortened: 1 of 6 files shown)


alicjakujawa (changes requested):
Screenshot 2021-08-23 at 12 12 26
i think its not enough. We show this state, when we sold token for current period.

src/data/resolvers/coinMachine.ts
```diff
@@ -198,13 +198,16 @@ export const coinMachineResolvers = ({
         // eslint-disable-next-line max-len
         const targetPerPeriodTokens = await coinMachineClient.getTargetPerPeriod();
 
+        const tokenPeriodBalance = await coinMachineClient.getTokenBalance();
```
chinins, line 201: Can you please confirm what this method returns - `getTokenBalance`: is it the balance of tokens available for sale for the period? How can we then use it to decide if the sale has started? What if it's zero and we are just sold out for the period?

The logic (it's more for the `CoinMachine` component later as well) doesn't seem to be correct. I also remember Raul saying it was a bit more tricky than just that....

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
```diff
@@ -90,9 +87,12 @@ const CoinMachine = ({
     variables: { colonyAddress },
     fetchPolicy: 'network-only',
   });
+  const saleStarted = !bigNumberify(
```
chinins, line 90: Just a nitpick but I prefer `hasSaleStarted` style (kinda more grammatically correct & sound more complete), but as you prefer since it's your code.

src/modules/dashboard/components/CoinMachine/RemainingDisplayWidget/RemainingDisplayWidget.tsx
```diff
@@ -217,9 +219,13 @@ const RemainingDisplayWidget = ({
           [styles.valueWarning]: isValueWarning || showValueWarning,
         })}
       >
-        {displayedValue}
+        {saleStarted ? (
```
chinins, line 222: we don't need this check here as this logic is already covered in `displayedValue` - line 143.

src/modules/dashboard/components/CoinMachine/RemainingDisplayWidget/RemainingDisplayWidget.tsx
```diff
@@ -217,9 +219,13 @@ const RemainingDisplayWidget = ({
...
+          displayedValue
+        ) : (
+          <FormattedMessage {...widgetText.placeholder} />
+        )}
       </p>
-      {periodTokens && widgetText.footerText && (
+      {saleStarted && periodTokens && widgetText.footerText && (
```
chinins, line 228: we don't need the `saleStarted` check here either as it's already in `periodTokens` in CoinMachine. If the sale hasn't started it will return undefined and hence `periodTokens` will be converted to `false`.
This way (together with my previous comment) we can remove `saleStarted` prop from the whole component.

chinins (changes requested):
There is a couple of small things in my comments to the files but the main one of course is the hasSaleStarted logic which seems to be off. Please, either clarify the logic or update it.

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
ArmandoGraterol: Sure, sounds good

src/modules/dashboard/components/CoinMachine/RemainingDisplayWidget/RemainingDisplayWidget.tsx
ArmandoGraterol: The logic there doesn't cover the case of the remaining time widget, I'll update it tho as it should

src/data/resolvers/coinMachine.ts
ArmandoGraterol: `getTokenBalance` gets you the balance of the coin machine, so if the coin machine is empty then a sale hasn't been "started". So, yeah I should actually change the name and perhaps move it to a separate resolver as this is not the balance of a period but of the coin machine in general

ArmandoGraterol:
> Screenshot 2021-08-23 at 12 12 26
> 
> i think its not enough. We show this state, when we sold token for current period.

That state is to be expected if the coin machine still has tokens on the balance, if the token balance is 0 then I think we can assume that the sale is paused / ended

chinins (approved):
If the `isSale` is indeed just to check the token balance, then it all looks good.

rdig (approved):
Looks good. Nice work!

PR #2645, opened 2021-08-06
Add whitelist users update to refreshExtension helper

### LINKED ISSUE #2630: Whitelist extension - if uninstalled / reinstalled, old data persists (temporarily)

alicjakujawa:
## Description

This PR fixes problem with old whitelist users are showing after extension reinstall

Resolves #2630


ArmandoGraterol (approved):
Damn, I actually had almost the same code but I didn't think of adding the `userLoading` there to avoid the old users from appearing while it was updating. 

So simple and I missed that! Good one :ok_hand:

chinins (approved):
Nice solution!

PR #2649, opened 2021-08-11
Wire up pevious sales table

ArmandoGraterol:
Wire up previous sales table with real data and logic

The data comes from the subgraph (https://github.com/JoinColony/subgraph/pull/28) due to the dapp version of this logic being too heavy-duty in all meanings of the word.

The dates displayed on the table can be strange but that would be nothing new considering how times advance on the dev env using the reputation miner.

FireShot Capture 418 - Colony - localhost
FireShot Capture 419 - Colony - localhost

## TODO

- [x] Reuse code available on #2642 to display the price status and token remaining formatting

Resolves DEV-467


alicjakujawa (commented):
Screenshot 2021-08-23 at 12 29 52
How you was actually able to test it? i bought 1 token, price should go down, but like you see on screen price SUPER GROW and i wasnt able to do another sale.. it think something is super wrong with buy token. Also i think we miss refetching table here, because i had to refresh page to see change

src/data/resolvers/coinMachine.ts
```diff
@@ -211,5 +212,136 @@ export const coinMachineResolvers = ({
...
+    async coinMachinePreviousSales(_, { colonyAddress }) {
+      try {
+        const { networkClient } = colonyManager;
+        const coinMachineClient = await colonyManager.getClient(
+          ClientType.CoinMachineClient,
+          colonyAddress,
+        );
+
+        const periodLength = await coinMachineClient.getPeriodLength();
+        const windowSize = await coinMachineClient.getPeriodLength();
```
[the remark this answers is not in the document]
ArmandoGraterol: Good catch! Didn't notice this, prob an error from my rebases

ArmandoGraterol:
@alicjakujawa I tested it by starting up coin machine, filling it with tokens, buying some of them, and then making a payment action to myself (or to anyone else really) with the reputation miner _on_ to advance time. A lot of periods are going to pass but you should be able to see the one where you bought the tokens and see the difference in comparison to the rest where you didn't.

Also, I agree with you, I think there's something fishy with how the BuyTokens widget is working/displaying values

src/data/resolvers/coinMachine.ts
```diff
@@ -211,5 +212,136 @@ export const coinMachineResolvers = ({
         return null;
       }
     },
+    async coinMachinePreviousSales(_, { colonyAddress }) {
```
rdig, line 215: So while I'm generally ok with having a local resolver for stuff like this, this a **big one**.

There's too much going on, too many calls to the network being made, as well as too many events being fetched and parse.

Keeping in mind that sales are basically unlimited this has the potential of breaking the app _(by failing to load, due to the vast number crunching this resolver has to)_

So, as I said initially when we started Coin Machine, this needs to be put on the subgraph, and just fetch the sales data from there _(previous sales, current sale can still come from events)_

Talk to @area about ways to achieve this, otherwise this won't work in production.

rdig (changes requested):
I firmly believe that this won't work until we put the previous sales data on the subgraph, due to the large amount of data we need to process and the potential of any sale to be infinite.

If you think that's not true, and you believe I missed something, then please argue you point, otherwise see the comments I left about this.

chinins:
@alicjakujawa, @ArmandoGraterol, might the problem with the BuyTokens be linked to the Coin machine init params. The target per period is a weird value (much much smaller than max per period) and if it's taken into account on how to calculate the price, this might have an impact... Just a thought.... I usually manually change this value when I enable the extension.

I also agree with Raul that the resolver might cause the problem in the future...

src/utils/colonyCoinMachine.tsx
```diff
@@ -0,0 +1,56 @@
+import { FormattedMessage, defineMessages } from 'react-intl';
```
rdig, line 1: No, this is not ok. Util files should be `.ts` not `.tsx`.

Only pure components should be a `.tsx` file, otherwise they are not utils.

I saw you wanted to return a message descriptor, but you can also do so programatically, without involving a React component. You can import `formatMessage` from `react-intl` and use that to parse your values.

Please refactor this.

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
```diff
@@ -82,6 +83,12 @@ const CoinMachine = ({
   const { transactionHash } = useParams<{
     transactionHash: string;
   }>();
+  const {
+    data: salePeriodsData,
+    loading: salePeriodsLoading,
+  } = useSubgraphCoinMachinePeriodsQuery({
+    variables: { colonyAddress: colonyAddress.toLowerCase() },
+  });
```
rdig, line 91: I couldn't get this to display anything. I make two periods go by:
Screenshot from 2021-08-30 14-31-08

Please notice the buy price, that's because I've bought all tokens in the previous periods.q

However, looking at query directly, it does seem that it's returning values:

Screenshot from 2021-08-30 14-33-31

Please investigate what's going on

rdig (changes requested):
This needs work, since it's not displaying any values for sale periods.

Please be more thorough in your PRs as, while it's all there, and the code seems fine, it doesn't work properly. This to me looks like you didn't test your PR properly before submitting.

Please pay more attention next time.

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
rdig: I've re-tested this and now seems to be working, no idea what the issue was the first time around.

Screenshot from 2021-08-30 17-44-56

You can consider this fixed

src/utils/colonyCoinMachine.tsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  if (bigNumberify(tokensBought).gt(targetPeriodTokens)) {
+    return TokenPriceStatuses.PRICE_UP;
+  }
+
+  return undefined;
+};
+
+export const getFormattedTokensRemaining = (
```
chinins, line 42: I agree w/Raul that it should be either refactored or you can also create a little component & put it in a separate file and reuse it like this...

chinins (commented):
I think that it is all looking good and working well. I had a small suggestion about Raul's comment about the utils file. I am happy to take a quick look after you refactor it in either way and approve after that. I want to have a look because it might impact a rebase for one of my PRs.

chinins (approved):
Looks much better with that small reusable component!

rdig (approved):
Looks good now. Very nice work. :100: 

Screenshot from 2021-08-31 21-22-15

PR #2650, opened 2021-08-17
Allow claim action for every logged in person

### LINKED ISSUE #2647: Example: Unable to log in with MetaMask wallet
## Steps to reproduce

1. Send tokens to a colony that you don't have permissions in

## Expected behaviour

* Should be able to claim tokens for the colony

## Actual behaviour

* Can't claim tokens
image

alicjakujawa:
## Description

This PR updates logic for claim button. 

* Every logged in person can claim tokens

Resolves #2647

### DIFF
--- a/src/modules/dashboard/components/UnclaimedTransfers/UnclaimedTransfersItem.tsx
+++ b/src/modules/dashboard/components/UnclaimedTransfers/UnclaimedTransfersItem.tsx
@@ -16,15 +16,11 @@ import {
   ColonyTransaction,
   useUsernameQuery,
   useLoggedInUser,
-  Colony,
 } from '~data/index';
 import { ActionTypes } from '~redux/index';
 import { mergePayload } from '~utils/actions';
-import { useTransformer } from '~utils/hooks';
 import { getTokenDecimalsWithFallback } from '~utils/tokens';
 import { tokenIsETH } from '../../../core/checks';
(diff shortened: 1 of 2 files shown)


ArmandoGraterol (approved):
It's working as expected! Good work

src/modules/dashboard/components/UnclaimedTransfers/UnclaimedTransfersItem.tsx
```diff
@@ -136,7 +126,7 @@ const UnclaimedTransfersItem = ({
           error={ActionTypes.COLONY_CLAIM_TOKEN_ERROR}
           success={ActionTypes.COLONY_CLAIM_TOKEN_SUCCESS}
           transform={transform}
-          disabled={!userHasPermission || !isNetworkAllowed}
+          disabled={!isNetworkAllowed || ethereal}
```
rdig, line 129: So this needs to check if the user has a registered profile as well, not just that it's not an ethereal wallet.

Something like this:
```js

  const hasRegisteredProfile = !!username && !ethereal;
```

This is not necessarily something in the contracts, but it something we don't allow in the dapp, if you don't have your profile registered first.

rdig (approved):
Other than change I pointed out, this is good to go

chinins (approved):
Good job!

PR #2651, opened 2021-08-17
Allow upgrade only if new version is supported (by colonyJS)

### LINKED ISSUE #2588: New Network version advertised even if not supported
## Steps to reproduce

Make a new version of the colony contracts available that is not supported in colonyJS. 

## Expected behaviour

If a new network version is deployed that is not supported, the app should remain silent on that front, and not advertise or allow a colony to be upgraded.

## Actual behaviour

The upgrade banner will be shown in the app when it should not be. If the colony is upgraded (not 100% sure that's actually possible in that state, we stumbled in to this with a new Dapp on staging and the old one on production), it will fail to load. The banner also fails to link to the release notes correctly (which is expected, as it doesn't know the release name).

rdig:
## Description

This PR refactors the Dapp to only allow upgrades _(allow in the modal, or show the upgrade banner)_ if the version is actually supported by colonyJS _(ie: it has a client for it already)_

Otherwise, after the upgrade, the colony breaks and cannot be accessed trough the dapp. _(until the dapp is re-deployed with a `colonyJS` version that supports said version)_

This is needed so that the network can freely deploy versions in the background without affecting the dapp.

**Testing**

Just change the `networkVersion` _(or `newVersion` for `NetworkContractUpgrade`)_ to something larger than the current version _(`'8'`, `'9'`, `'10'` -- note it must be a string)_, as you are testing the logic inside the component, not the way we actually fetch the version value.

If you don't see nothing _(as the version is not yet supported by `colonyJS`)_, it means all is good.

**Changes** 

- [x] `ColonyUpgrade` component check if version is supported before showing the upgrade banner
- [x] `NetworkContractUpgrade` component check if version is supported before allowing the upgrade
- [x] `NetworkContractUpgrade` component add error message if version is available, but not supported

Resolves #2588

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx
@@ -1,5 +1,6 @@
 import React, { useCallback } from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
+import { ColonyVersion } from '@colony/colony-js';
 
 import { useDialog } from '~core/Dialog';
 import NetworkContractUpgradeDialog from '~dashboard/NetworkContractUpgradeDialog';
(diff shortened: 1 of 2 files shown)


alicjakujawa (approved):
Works good 👌

chinins (approved):
Good to go!

PR #2652, opened 2021-08-18
Add proper coin machine init params

chinins:
## Description

This PR adds updated coin machine init parameters and also fixes a couple of bugs in the BuyTokens widget.

I am not 100% sure what is the best way to deal with links & 2 types of init parameters plus the ordering. So I am open to suggestions if you have a simpler/better way to do that.

Figma specs copies are not fully updated. There is a notion doc for that (both links below).

My UI doesn't have everything from Figma specs. Karol introduced new looks (some fonts, colors etc) and we agreed that I will keep the same stuff as in all other extensions for now. There is an issue on Linear to update everything based on these new specs.

Link to the figma specs:
https://www.figma.com/file/fw2pDIUlkhCkJE79ZyTbjp/CoinMachine?node-id=2769%3A22579

Notion doc:
https://www.notion.so/colony/Coin-Machine-Params-4e29e47d84d64b1b92673a0cedd68f7c

**New stuff** ✨

- add `TokenSelector` to `ExtensionParamType`
- add several extra fields to `ExtensionData` & `ExtensionInitParams`
- add `TokenSelector` to `ExtensionSetup`
 
**Changes** 🏗

- change `userLimit` from tokens to fraction
- move input complementary labels to `extensionData`
- make the use of links in extension description consistent & replicable across extensions
- update `colonyExtensionEnable` saga
- update the way `maxPurchase` is calculated in `BuyTokens` (before it showed the max balance denominated in purchase token and not in sellable token)
- replace `formatEther` with `getFormattedTokenValue` as the purchase token might not be etherium

Resolves Dev-470
Resolves Dev-476


src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
```diff
@@ -360,19 +358,15 @@ const ExtensionDetails = ({
                           text={chunks}
                         />
                       ),
+                      link0: extension.descriptionLinks?.[0],
```
rdig, line 361: This is a bit of a nit-pick but I really hate the name of the link variable :)

src/data/staticData/extensionData.tsx
```diff
@@ -339,81 +387,126 @@ const extensions: { [key: string]: ExtensionData } = {
...
+    ],
     currentVersion: 1,
     createdAt: 1603915271852,
     neededColonyPermissions: [ColonyRole.Root],
-    initializationParams: [
+    tokenContractAddress: (
+      <FormattedMessage
+        {...MSG.coinMachineTokenContractAddress}
+        values={{
+          link: <ExternalLink href={BLOCKSCOUT_LINK} text="Blockscout" />,
```
rdig, line 412: This is not ok.

- If we're on ethereum mainnet this should point to etherescan
- if we're on xdai should point to blockscout
- if on localhost should point to #

Also, it should point directly to the token contract, not just to the general home page. _(that's the point, so the user can actually verify the token contract)_

We actually have a utility that helps in this case: `getBlockExplorerLink`  so you don't have to deal with all the complexity.

Just call it like:
```js
getBlockExplorerLink({
  linkType: 'token',
  addressOrHash: tokenAddress,
});
```

And it will generate the link for you

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
```diff
@@ -178,6 +237,107 @@ const ExtensionSetup = ({
...
+                    <FormattedMessage {...fieldName} />
+                  </p>
+                )}
+                <TokenSymbolSelector
+                  tokens={tokens}
+                  name={paramName}
+                  appearance={{ alignOptions: 'right', theme: 'grey' }}
+                  elementOnly
+                  label={paramName}
+                />
```
rdig, line 326: There's a bug in the `TokenSymbolSelector` component. Either only here, or is a global problem.

When changing the token, the token's avatar doesn't change.

Here's a demo of the bug:
bug-token-selector-same-token-avatar

rdig (changes requested):
This is mostly good, however a did find some minor stuff that needs to be addresses. Please see the comments below.

Other than that, very nice work!

src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
chinins: I can change if you have a preference for another name? How about `externalLink`? I am happy with any other if you have any suggestions. It needs to be generic as there are several in the array...

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
rdig: I'll check again, maybe it was something on my side

src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
rdig: I know, that's why it was a nit-pick. There's not really a good way to do naming here.

Leave it like it is, but I still don't like it :)

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
rdig: Re-tested and it's the same :(

bug-token-selector-same-token-avatar-2
bug-token-selector-same-token-avatar-3
bug-token-selector-same-token-avatar-4

Tell you what. I'll approve this PR, since this bug is not really related to this issue, but after you merge this, open a issue to track the Token Selector bug and fix it separately

rdig (approved):
I'm pre-approving this, but before you merge:
- Make sure to refactor the block explorer link to use: `getBlockExplorerLink`
- Open a issue to track the the `Token Selector` bug

ArmandoGraterol (approved):
Looks good to me :ok_hand: Nice work!

src/data/staticData/extensionData.tsx
```diff
@@ -405,11 +406,24 @@ const extensions: { [key: string]: ExtensionData } = {
...
+              href={getBlockExplorerLink({
+                linkType: 'token',
+                addressOrHash: tokenAddress,
+              })}
+              text={
+                DEFAULT_NETWORK === Network.Xdai ||
+                DEFAULT_NETWORK === Network.XdaiFork
+                  ? 'Blockscout'
+                  : 'Etherscan'
+              }
```
rdig, line 424: Or localhost if `DEFAULT_NETWORK === Network.Local`

But that's just mostly for us...

rdig (approved):
Very nice work. Good to go!

PR #2665, opened 2021-08-29
Comments Subscriptions

rdig:
## Description

This PR integrates the newly added server GraphQL subscriptions for the comments so that actions, motions, as well as the colony home list will all update instantly once a new comment has been posted

Note that this makes use of the GrapQL Subscriptions PR from [JoinColony/colonyServer#139](https://github.com/JoinColony/colonyServer/pull/139)

**Installing**
- Run `npm run provision` since the server submodule was updated

**Testing**
Just comment on an action or motion, and, in a different browser window, see it update in real time

**Changes** 
- [x] Update the `colonyServer` submodule
- [x] Refactored the apollo client route spliting so that it now takes into account our server's WebSocket path
- [x] Refactored the graphql code structure so that we can have multiple files _(copied what I've implemented in #2407)_
- [x] Added `Comments` and `CommentCount` subscriptions
- [x] Refactored `ColonyActions`, `ActionPageFeed` and `ColonyEvents` to use the new comments subscriptions

**Demo**
demo-subscription-comments

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -18,6 +18,9 @@ NODE_ENV=development
 # HTTP endpoint for graphql server
 SERVER_ENDPOINT=http://127.0.0.1:3000
 
+# WebSocket endpoint for our own graphql server
+SERVER_WS_ENDPOINT=ws://127.0.0.1:3000
+
 # HTTP endpoint for "The Graphs" graphql server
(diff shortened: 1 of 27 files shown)


ArmandoGraterol (approved):
Very nice work! Tested it on local and QA env and it's working as expected

chinins (approved):
Looking very good!

PR #2679, opened 2021-09-01
Add Tokens section to MembersInfoPopover

### LINKED ISSUE #2657: Add tokens section to MemberInfo
### Description

Implement a section on MemberInfo that shows how many tokens does the user has. (This includes both the styles and the logic to get the number of tokens that the user has)

**Questions for Product:**

1. How does this section should look like if the user doesn't have any tokens?
    R: If the user doesn't have any tokens then we should hide the section

FireShot Capture 425 - Colony Simplified – Figma - www figma com

Figma file: https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=422%3A10616

ArmandoGraterol:
FireShot Capture 430 - Colony - localhost

Resolves #2657


alicjakujawa (approved):

chinins (approved):
All good!

src/modules/core/components/InfoPopover/MemberInfoPopover/UserTokens.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  return (
+    <div className={styles.tokensContainer}>
+      <Heading
+        appearance={{
+          size: 'normal',
+          margin: 'none',
+          theme: 'grey',
+          weight: 'bold',
+        }}
```
rdig, line 40: I mentioned this in the other PR as well:
- color is not properly applied _(theme `grey` still renders as `--dark`)_
- please check if the font size corresponds to the Figma specs

rdig (approved):
Same mentions as in the other PR that I've reviewed, please check the heading size and color.

Other than that, everything's good, so I'm pre-approving.

PR #2680, opened 2021-09-02
Update extension setup styles based on new specs

chinins:
## Description

This PR updates styles for the enable extensions page. Please, check all of the extensions when you test as they all should be the same for consistency.

Here is the link to Figma. Please, note that there might be slight differences vs. Figma specs, e.g.: the font size in the whitelist address field is smaller than the specs because otherwise it won't fit in (aligned with Karol).

https://www.figma.com/file/fw2pDIUlkhCkJE79ZyTbjp/CoinMachine?node-id=2769%3A22579

**Changes** 🏗

- updated html & css of the ExtensionSetup

Resolves Dev-474


alicjakujawa (changes requested):
Few small UI things here i noticed:
Screenshot 2021-09-02 at 14 00 51
Screenshot 2021-09-02 at 14 00 46
1. Explorer link has different color than on designs
2. We need a bit more space between explorer and text
3. Line height need to be set to 18px
4. Breadcrumbs on designs aren't bold (Setup should be same weight like others)

Screenshot 2021-09-02 at 14 03 47
Space above and below text "Select the token you wish to sell." should be equal and should have different color

ArmandoGraterol (commented):
Reviewed, but I think I just noticed the same things Alicja is mentioning already, so I'll wait until changes are made

alicjakujawa (approved):

src/modules/dashboard/components/Extensions/ExtensionDetails.css
```diff
@@ -4,6 +4,10 @@
   grid-template-columns: minmax(460px, 2fr) minmax(340px, 1fr);
 }
 
+.main a {
```
ArmandoGraterol, line 7: I think this selector is making the explorer links look like this:

FireShot Capture 432 - Colony - localhost

ArmandoGraterol (approved):
Nice work! I just noticed some small thing so I'm gonna approve anyway :)

src/modules/dashboard/components/Extensions/ExtensionDetails.css
chinins: good catch - fixed now.

PR #2691, opened 2021-09-06
Add smite action type and saga

### LINKED ISSUE #2664: Add actions, action types & action saga for Smite action
Implement logic for Smite action.

Permission for action - `Arbitraiton`

ArmandoGraterol:
In order to test without the smite UI you could use any UAC dialog, like the PermissionManagement one, and paste this on the transform func:

```
      mapPayload(
        ({ user, domainId, annotationMessage }) => ({
          domainId,
          userAddress: user.profile.walletAddress,
          annotationMessage,
          amount: bigNumberify(NEGATIVE_NUMBER_HERE),
        }),
      ),
      mergePayload({
        colonyAddress,
        colonyName,
      }),
```
And then change the action type to the smite one: `COLONY_ACTION_SMITE`

Also, the negative number must be equivalent *at least* to 0.01% or *more* of the `total colony rep amount` in order to see an actual difference in the UI reputation percentage. It's not that the reputation doesn't change if you use a lower value, but you will not be able to tell the difference just looking at the UI
 
Resolves #2664


src/modules/dashboard/sagas/actions/smite.ts
```diff
@@ -0,0 +1,193 @@
...
+      {
+        query: UserReputationDocument,
+        variables: {
+          colonyAddress,
+          address: AddressZero,
+          domainId,
+        },
+        fetchPolicy: 'network-only',
+      },
+    );
```
rdig, line 173: Why are we calling this refresh query with `AddressZero` ?

rdig (approved):
Seems good. Nice work. :100:

chinins (approved):
Looks good! Nice work!

@rdig, I think that the addressZero is used to get update on the total reputation in the colony that is used to then calculate the % of reputation for the user.

src/modules/dashboard/sagas/actions/smite.ts
ArmandoGraterol: Like Olga already mentioned, the UserReputation is called with `AddressZero` when we want to fetch the colony's total reputation, which should change after the penalization.

PR #2692, opened 2021-09-06
Add Smite action to ActionsPage

### LINKED ISSUE #2670: Smite Action Page Wiring
Add types, resolver, proper values, icons etc

ArmandoGraterol:
Wired up smite action with the ActionsPage

FireShot Capture 449 - Colony - localhost

Resolves #2670

### DIFF
--- a/src/i18n/en-events.ts
+++ b/src/i18n/en-events.ts
@@ -18,7 +18,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.MotionFinalized} {{motionTag} was finalized. Stakes may be claimed.}
       ${ColonyAndExtensionsEvents.ObjectionRaised} {{staker} raised an {objectionTag}}
       ${ColonyAndExtensionsEvents.MotionRewardClaimed} {{staker} claimed their stake.}
-      ${ColonyAndExtensionsEvents.ArbitraryReputationUpdate} {{initiator} smites {recipient} with {reputationPenalty} reputation penalty}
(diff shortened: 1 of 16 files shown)


rdig (approved):
Very nice work here! :100: 

However, there is one thing I want to point out, which either you or @chinins should bring up with Product.

When we deduct reputation from the Smite Modal, we are using a percentage based value:
Screenshot from 2021-09-30 20-23-52

But when we are showing it on a page, we are using points (pts), which, while calculated correctly in this case, are different from percentage:
Screenshot from 2021-09-30 20-24-00

I consider this to be an inconsistency, which we need to address.

rdig:
O the other hand, never mind, I see #2756 already addresses that it changes the modal to using points as well.

In that case, good to go! :100:

chinins (approved):
Great job! Good to go.

src/i18n/en-events.ts
```diff
@@ -18,6 +18,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.MotionFinalized} {{motionTag} was finalized. Stakes may be claimed.}
       ${ColonyAndExtensionsEvents.ObjectionRaised} {{staker} raised an {objectionTag}}
       ${ColonyAndExtensionsEvents.MotionRewardClaimed} {{staker} claimed their stake.}
+      ${ColonyAndExtensionsEvents.ArbitraryReputationUpdate} {{initiator} smote {recipient} with a {reputationPenalty} reputation penalty}
```
chinins, line 21: Small typo here - should be `smite` I think, not `smote`

PR #2696, opened 2021-09-09
Fix User Profile page not loading

### LINKED ISSUE #2632: User profile not loading
The current user profile page `/user/username` is in an infinite loading state:

Screenshot from 2021-07-30 12-51-00

However, the settings page, `/edit-user`, which uses basically the same logic in the background, loads fine.

This needs to be investigated as to what exactly is causing the infinite loader to show _(most likely a resolver not returning proper data)_, then fixed.

This fix should go directly into the `master` branch.

alicjakujawa:
## Description
This PR fixes not loading user profile page

**Changes** 🏗

* Refactored UserProfile component to use data from resolver instead of calling saga

Resolves #2632


src/modules/users/components/UserProfile/UserProfile.tsx
```diff
@@ -20,21 +19,23 @@ const UserProfile = ({
     params: { username },
   },
 }: Props) => {
-  const { userAddress, error: userAddressError } = useUserAddressFetcher(
```
rdig, line 23: We're also using the `useUserAddressFetcher` custom hook in the core `UserMention` component.

It would be good if you refactored that one in the same style as you did here with `useUserAddressQuery`, and after that you can remove `useUserAddressFetcher` from the code entirely since it won't be used anymore.

rdig (approved):
This is very nice work.

I do however have one more request, to refactor `core/UserMention` and after remove `useUserAddressFetcher` entirely

chinins (commented):
Other than extra request from Raul it's all good. I am happy to have a look when you do extra part.

rdig (approved):
Very nice refactor and cleanup. Top marks! :1st_place_medal:

PR #2698, opened 2021-09-09
Update not enough reputation copy

### LINKED ISSUE #2672: Insufficient reputation warning message copy is incomplete
## Steps to reproduce

1. Attempt to send an expenditure without reputation via a motion

## Expected behaviour

* Copy: "There is no reputation in this team yet.
Please create the motion in a parent of this domain where there is reputation. Alternatively, if you have the necessary permissions to take this action, you may toggle the "Force" switch in the top right corner of this modal to bypass the governance process."
* It should be noted that the "There is reputation in this team yet." should be red with a line break following it (as it is now). 

## Actual behaviour

Screenshot_2021-08-30_at_17 46 26

alicjakujawa:
## Description

This PR updates not enough reputation copy

Screenshot 2021-09-09 at 15 59 08

Resolves #2672

### DIFF
--- a/src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
+++ b/src/modules/dashboard/components/NotEnoughReputation/NotEnoughReputation.tsx
@@ -26,7 +26,7 @@ const MSG = defineMessage({
   },
   description: {
     id: 'NotEnoughReputation.description',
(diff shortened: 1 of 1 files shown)


rdig (approved):
Good to go! :tada:

chinins (approved):
All good!

Issue #2697, opened 2021-09-09
Add logic for action/motion & amount calculation in the Smite dialog


chinins:
This is to add logic for if we do action or motion.

Also to calculate the absolute number of the smiting based on the % input.

PR #2685, opened 2021-09-02
Rework reputation section of MembersInfoPopover

ArmandoGraterol:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #2656

PR #2708, opened 2021-09-12
Add smite action/motion & transform logic

chinins:
## Description

This PR adds transform function & logic to calculate the amount of the reputation to pass to the action. It also adds the motion/action logic, toggle & motion domain id selection & logic.

I had to remove the `prefix: '%',` from the amount Input as it was converting the number to a string & would fail validation. Added the `%` manually

Resolves issue #2697

### DIFF
--- a/src/modules/core/components/Fields/Input/InputComponent.tsx
+++ b/src/modules/core/components/Fields/Input/InputComponent.tsx
@@ -37,7 +37,7 @@ type CleaveHTMLInputElement = HTMLInputElement & { rawValue: string };
 
 interface MaxButtonParams {
   setFieldValue: (field, value) => void;
-  maxAmount: string | number;
+  maxAmount: string;
   fieldName: string;
 }
 
--- a/src/modules/core/components/Fields/Input/InputComponent.tsx
+++ b/src/modules/core/components/Fields/Input/InputComponent.tsx
@@ -37,7 +37,7 @@ type CleaveHTMLInputElement = HTMLInputElement & { rawValue: string };
 
 interface MaxButtonParams {
   setFieldValue: (field, value) => void;
-  maxAmount: string;
+  maxAmount: string | number;
   fieldName: string;
 }
 
--- a/src/modules/core/components/MemberReputation/MemberReputation.tsx
+++ b/src/modules/core/components/MemberReputation/MemberReputation.tsx
@@ -41,7 +41,7 @@ enum ZeroValue {
 
 type PercentageReputationType = ZeroValue | number | null;
 
-const calculatePercentageReputation = (
+export const calculatePercentageReputation = (
   decimalPlaces: number,
   userReputation?: Reputation,
   totalReputation?: Reputation,
@@ -80,6 +80,7 @@ const MemberReputation = ({
 }: Props) => {
   const { data: userReputationData } = useUserReputationQuery({
     variables: { address: walletAddress, colonyAddress, domainId, rootHash },
+    fetchPolicy: 'network-only',
   });
 
   const { data: totalReputation } = useUserReputationQuery({
@@ -87,8 +88,8 @@ const MemberReputation = ({
       address: AddressZero,
       colonyAddress,
       domainId,
-      rootHash,
     },
+    fetchPolicy: 'network-only',
   });
 
   const userPercentageReputation = calculatePercentageReputation(
--- a/src/modules/core/components/MemberReputation/MemberReputation.tsx
+++ b/src/modules/core/components/MemberReputation/MemberReputation.tsx
@@ -1,14 +1,17 @@
 import React from 'react';
 import { defineMessages } from 'react-intl';
-
 import { ROOT_DOMAIN_ID } from '@colony/colony-js';
 import { AddressZero } from 'ethers/constants';
-import { bigNumberify } from 'ethers/utils';
 
 import { useUserReputationQuery } from '~data/index';
 import { Address } from '~types/index';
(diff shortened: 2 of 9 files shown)


src/modules/core/components/Fields/Input/InputComponent.tsx
```diff
@@ -37,7 +37,7 @@ type CleaveHTMLInputElement = HTMLInputElement & { rawValue: string };
 
 interface MaxButtonParams {
   setFieldValue: (field, value) => void;
-  maxAmount: string;
+  maxAmount: string | number;
```
rdig, line 40: This is really confusing. Either one or the other.

As far as I understand you made this change because the type wouldn't work in your dialog form. But since this is a core component, the actual display component should adapt the types to it, not the other way around.

So please refactor this to have a more strict type.

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
```diff
@@ -58,30 +82,59 @@ const SmiteDialog = ({
...
+        const totalReputation = bigNumberify(totalReputationData || '0');
+        const repuationChangeAmount = totalReputation.mul(amount).div(100);
+
+        return {
+          colonyAddress,
+          colonyName,
+          domainId,
+          userAddress: user.profile.walletAddress,
+          annotationMessage: annotation,
+          amount: `-${repuationChangeAmount.toString()}`,
```
rdig, line 105: I would have written it as:
```suggestion
          amount: String(-1 * repuationChangeAmount),
```
just to be more readable

src/modules/core/components/MemberReputation/MemberReputation.tsx
```diff
@@ -41,7 +41,7 @@ enum ZeroValue {
 
 type PercentageReputationType = ZeroValue | number | null;
 
-const calculatePercentageReputation = (
+export const calculatePercentageReputation = (
   decimalPlaces: number,
```
rdig, line 45: I think the `decimalPlaces` prop should have a default value of `2` since that is about what we use everywhere in the dapp.

And just pass a different value if the display we require is non-standard.

Also, I think this could benefit from being extracted into it's own `util`

src/modules/dashboard/components/SmiteDialog/SmiteDialogForm.tsx
```diff
@@ -158,23 +297,54 @@ const SmiteDialogForm = ({
...
               numeral: true,
-              prefix: '%',
               // @ts-ignore
               tailPrefix: true,
             }}
             elementOnly
             maxButtonParams={{
               fieldName: 'amount',
-              maxAmount: '10',
+              maxAmount: userPercentageReputation || 0,
```
rdig, line 306: Can't this be:
```suggestion
              maxAmount: String(userPercentageReputation || 0),
```
to satisfy the `string` type requirement in `MaxButtonParams` ?

rdig (changes requested):
Mostly good, however I found places where you can improve this.

src/modules/core/components/MemberReputation/MemberReputation.tsx
ArmandoGraterol: On the user profile popover, I extracted this to utils: https://github.com/JoinColony/colonyDapp/pull/2685/commits/6fec45db81419366adc33d0be50371dbac26f925

Perhaps Olga can cherry-pick that commit or do something similar?

chinins: Yeah, I noticed that Armando did it in his PR and I thought that would be fixed during the rebase. But i've cherry-picked the commit, adapted to my code & also made the decimal places optional with a default value.

ArmandoGraterol (changes requested):
Solid work! I noticed that when no user is selected the max is shown as `null%`, perhaps we should hide that or display a 0 instead.

FireShot Capture 444 - Colony - localhost

Also, I can go past the max percentage that is displayed there. For example, it said that the max was 89% but I still can input and submit 100% instead

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
```diff
@@ -58,30 +82,59 @@ const SmiteDialog = ({
...
 
   const { data: colonyMembers } = useMembersSubscription({
     variables: { colonyAddress },
   });
 
+  const transform = useCallback(
+    pipe(
+      mapPayload(({ amount, domainId, annotation, user, motionDomainId }) => {
+        const totalReputation = bigNumberify(totalReputationData || '0');
+        const repuationChangeAmount = totalReputation.mul(amount).div(100);
```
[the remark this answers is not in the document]
chinins: Good point about decimals. I have a little get around though without Decimal it as the BigNumber is more straight forward to use.

chinins:
> Solid work! I noticed that when no user is selected the max is shown as `null%`, perhaps we should hide that or display a 0 instead.
> 
> FireShot Capture 444 - Colony - localhost
> 
> Also, I can go past the max percentage that is displayed there. For example, it said that the max was 89% but I still can input and submit 100% instead

Good catch! I tried it before but then max would appear and disappear constantly as you change the selection. I have implemented displaying `0` instead.

ArmandoGraterol (approved):
Looks good to me! Nice work

rdig (approved):
I like it. Very nice work! :1st_place_medal: 

Sorry this took so long to review :facepalm:

PR #2709, opened 2021-09-13
Filter out duplicated forced events for motions

### LINKED ISSUE #2703: Remove Corresponding "Forced" Action when Motions and Disputes is Enabled
Currently, after creating a payment motions, and the motion succeeds, and new, "forced" action will show up the Actions List _(this is technically correct, since that's the actual payment transaction fired by the OneTx extension contract)_

Screenshot from 2021-09-09 21-54-55

This needs to be hidden in the list when a corresponding motion was created prior _(basically filter out forced actions created by the OneTx contract)_

**To do:**
- Check all motion types if they create an additional "forced" action
- Filter out all corresponding "forced" actions for a motion 

Note that "real" forced actions, still need to show up in the list

See this archived thread for the origins of this change: https://discord.com/channels/562263648173555742/611581703026245689/885594249607929896

alicjakujawa:
## Description

This PR updates action list on home page

**Changes** 🏗

* Filtered out forced events emitted by contract when motion extension is enabled

Resolves #2703

### DIFF
--- a/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
+++ b/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
@@ -9,6 +9,7 @@ import ActionsList, {
 import { Select, Form } from '~core/Fields';
 import LoadMoreButton from '~core/LoadMoreButton';
 import { SpinnerLoader } from '~core/Preloaders';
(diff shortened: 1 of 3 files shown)


ArmandoGraterol (approved):
LGTM :+1:

rdig (approved):
Looks good locally, and hopefully will perform the same on QA :)

PR #2716, opened 2021-09-13
Update motion page

### LINKED ISSUE #2677: Add "Motion created in" field in metadata list
Reputation_message_disambiguation

1. In this situation the motion had accidentally been created in Root but there is no way to tell from the screen that it wasn't created in Info & Globalization. We should have a "Motion created in" field in that metadata list.

2. The error message here contains too little information, which made it unclear to the user why it hadn't worked. The user thought they did have reputation in the relevant team but became confused because a) the motion was created in Root so the stakes were much higher than expected, and b) they didn't understand that they didn't have enough reputation. I suggest that a richer error message could fix this. e.g.:

`The minimum stake requirement in this team is 33.03908 FOX. You only have 0.49 FOX activated, so cannot provide the minimum stake. Please activate more tokens.`

`The minimum reputation requirement to stake a motion in this team is 33.03908 points. You only have 0.49 points and therefore do not meet the minimum Reputation requirement to be permitted to stake.`

Requested via Jack

ArmandoGraterol:
Add "motion created in" field to DetailsWidget and reworked logic and messages of some of the error messages of the `StakingSlider`.

FireShot Capture 442 - Colony - localhost
FireShot Capture 441 - Colony - localhost

Resolves #2677

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
+++ b/src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
@@ -1,12 +1,13 @@
 import React, { useState, useMemo } from 'react';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { Decimal } from 'decimal.js';
+import { bigNumberify } from 'ethers/utils';
 
 import Heading from '~core/Heading';
 import Slider, { Appearance } from '~core/Slider';
 import StakingValidationError from '~dashboard/ActionsPage/StakingValidationError';
 
-import { Colony } from '~data/index';
+import { Colony, useLoggedInUser } from '~data/index';
 import { getFormattedTokenValue } from '~utils/tokens';
 
 import styles from './StakingWidget.css';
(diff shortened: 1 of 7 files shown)


src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
```diff
@@ -138,10 +166,29 @@ const StakingSlider = ({
           max={100}
           disabled={!canUserStake}
           appearance={appearance}
-          exceedLimit={exceedLimit}
+          handleLimitExceeded={onLimitExceeded}
         />
       </div>
-      {showError && <StakingValidationError stakeType={errorStakeType} />}
+      {errorStakeType && (
```
rdig, line 172: After I fully stake a motion, I'm seeing this error: 
Screenshot from 2021-09-15 14-52-06

This is not necessary something that you added, but since you're working on this, might as well take a look.

Screenshot from 2021-09-15 14-51-41

rdig (approved):
The motions details seem to work as expected, but I did find an error that shows up when it shouldn't. 

Please take a look at it.

Other than that it's all good, so I'm pre-approving, but please fix that error before merging.

Screenshot from 2021-09-15 14-50-46
Screenshot from 2021-09-15 14-51-41

chinins (approved):
Looks good

PR #2722, opened 2021-09-14
Add copy button to coin machine contract address

### LINKED ISSUE #2720: In the extension manager add a "copy" icon next to the coinmachine extension address as that's where funds need to be sent to start the sale
### Story

As a user setting up a token sale, I want to easily & reliably copy the address where the funds need to be sent to start the sale, so that I know what to do & successfully do it.

Screenshot on 2021-09-14 at 15-37-47

A little copy icon next to the contract address.

- [ ] There's a copy icon next to the contract address
- [ ] When clicking the copy icon the address is copied into your clipboard

ArmandoGraterol:
FireShot Capture 443 - Colony - localhost

Resolves #2720

### DIFF
--- a/src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
+++ b/src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
@@ -265,11 +265,9 @@ const ExtensionDetails = ({
             >
               <MaskedAddress address={installedExtension.address} />
             </ExternalLink>
-            {installedExtension.extensionId === Extension.CoinMachine && (
-              <span className={styles.copyButton}>
-                <ClipboardCopy value={installedExtension.address} />
-              </span>
-            )}
+            <span className={styles.copyButton}>
+              <ClipboardCopy value={installedExtension.address} />
+            </span>
           </div>
         ),
       },
--- a/src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
+++ b/src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
@@ -43,6 +43,7 @@ import {
   COLONY_EXTENSION_SETUP_ROUTE,
 } from '~routes/index';
 import { DEFAULT_NETWORK_INFO, ALLOWED_NETWORKS } from '~constants';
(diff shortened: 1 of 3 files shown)


alicjakujawa (changes requested):
Screenshot 2021-09-15 at 09 07 14

it seems to be not aligned on my machine :<
Also, i dont think we should introduce it ONLY for coin machine extension, but for all extensions, please recheck with product

src/modules/dashboard/components/Extensions/ExtensionDetails.tsx
```diff
@@ -258,11 +259,18 @@ const ExtensionDetails = ({
...
+            <ExternalLink
+              href={`${contractAddressLink}/${installedExtension.address}`}
+            >
+              <MaskedAddress address={installedExtension.address} />
+            </ExternalLink>
+            {installedExtension.extensionId === Extension.CoinMachine && (
+              <span className={styles.copyButton}>
+                <ClipboardCopy value={installedExtension.address} />
+              </span>
+            )}
```
rdig, line 272: This needs to be enabled for all extensions not just Coin Machine

rdig (changes requested):
I'm good with this, apart from the fact that this needs to be enable for all extensions

rdig (changes requested):
I'm good with your changes but now I'm reconsidering if this is the correct route to take.

We now have two ways of copying the address in the same side:
- hover + click for the user address
- just click for the extension contract address

demo-different-types-of-address-copy

We should standardize on just using one of them. Either the hover + click, or just showing the "Copy" and letting the user click it to copy the address into the clipboard.

Show the gif to product and tell them we need to standardize here, and that they need to pick on of them.

After that, refactor this so that it reflects those changes.

Sorry about not picking up on this earlier.

alicjakujawa (approved):

chinins (commented):
The code & functionality are all good but I do agree with Raul about having it the same way. Should we also have it similar to other places in the app?

rdig (approved):
Good to go! :100:

chinins (approved):
All good!

PR #2731, opened 2021-09-16
Update BuyTokens validation

### LINKED ISSUE #2724: Should be able to purchase fractional tokens
## Steps to reproduce

Try to buy a fractional number of tokens

Screenshot 2021-09-15 at 10 22 04

## Expected behaviour

I should be able to buy a fractional amount of tokens.

## Actual behaviour

I am unable to buy a fractional amount of tokens. 

## Suggestions (optional)

Note that the decimals value for any tokens in question should also be considered here. If a token is indivisible (i.e. 0 decimals), then I should not be able to buy fractional numbers of tokens.

ArmandoGraterol:
Basically:

- Replaced parseInt for parseFloat in order to accurately get the max value.
- Updated validationSchema to pass the balance as it's returned from `getFormattedTokenValue` instead of passing it first through parseFloat as otherwise values like `0.00000...` (for low values), will not be displayed on the error message correctly.
- Took into account that the amount must be an integer if the decimals of the sold token is `0` 

Resolves #2724

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
+++ b/src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
@@ -5,6 +5,7 @@ import * as yup from 'yup';
 import { FormikProps } from 'formik';
 import { AddressZero } from 'ethers/constants';
 import { bigNumberify } from 'ethers/utils';
+import isNil from 'lodash/isNil';
 
 import Heading from '~core/Heading';
 import QuestionMarkTooltip from '~core/QuestionMarkTooltip';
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
```diff
@@ -286,6 +300,10 @@ const BuyTokens = ({
     );
   }
 
+  const sellableTokenDecimals = !isNil(sellableToken)
+    ? sellableToken.decimals
+    : 18;
```
rdig, line 305: We have a `DEFAULT_TOKEN_DECIMALS` value in `~constants` for this

src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
```diff
@@ -5,6 +5,7 @@ import * as yup from 'yup';
 import { FormikProps } from 'formik';
 import { AddressZero } from 'ethers/constants';
 import { bigNumberify } from 'ethers/utils';
+import isNil from 'lodash/isNil';
```
rdig, line 8: This might be a nitpick, but I really hate importing crap from `lodash`

There must be a easier way just to check if a value is `null` or `undefined` without importing an extra library for that.

Why not cast it to a boolean ?

```js
 const sellableTokenDecimals = !!sellableToken ? sellableToken.decimals : 18;
```

rdig (approved):
Seems to work good, however I don't like some of the code decisions you've made.

Screenshot from 2021-09-17 12-20-32

I'm pre-approving, but please look at the comments I've left and refactor them appropriately

chinins (approved):
Other than  Raul's comments, looking good.

src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
ArmandoGraterol: :thinking: That's true, didn't occur to me that I could just try casting it to boolean

Issue #2713, opened 2021-09-13
Switch the chat entry functions (CMD+Return = new line | Return = submit)


thycommodore:
Note: This is a UX bug, not a technical bug.

## Steps to reproduce

1. View a coin machine sale
2. Send a chat message

## Expected behaviour

* Hitting Return after typing feels like it will submit
* Hitting CMD+Return feels like it will do a new line

## Actual behaviour

* The opposite

## Suggestions (optional)

* Flip the functionalities so that:

- [ ] Hitting Return after typing submits your message
- [ ] Hitting CMD+Return after typing creates a new line

rdig:
The keyboard shortcuts were @jakzilla 's picks, so you guys have. To decide amongst yourselves if you want this changed.

Also note that this will change will be dapp-wide and will affect both actions and motions

rdig:
Converted into a feature request which we need to decide how we act upon.

rdig:
Decided to change the inputs, so make sure that all places that use comments have these keyboard shortcuts changed, not just Coin Machine.

Also, make sure they are also reflected in the tip that shows up after entering 3 chars in the comment box.

PR #2741, opened 2021-09-21
Add autofocus to comment box input

### LINKED ISSUE #2719: Autofocus back on chat after submitting (currently have to reselect the box)
### Story

As a user chatting in a coin machine sale, I want to be able to continue to type in the chat box after submitting so that I can stay engaged with the chat session.

The workflow to send 2 chat messages works like this:

1. Click the input field
2. Type your message
3. Submit it
4. Click the input field
5. Type your message
6. Submit it

It feels like the common UX pattern is:

1. Click the input field
1. Type your message
1. Submit it
1. Type your message
1. Submit it

On step (3) the active focus remains on the text input field. 

- [ ] After submitting the chat entry, typing without clicking anything results in characters appearing in the input field

ArmandoGraterol:
The TextareaAutosize is only used on 2 components, the comment box or chat, and on the storage inputs of the recovery mode actions page. So to kill 2 birds with 1 stone while testing I would just enter recovery mode, all the instances where the component is used are there (Apart from CoinMachine's chat of course).

- Added autofocus to comment box input after you submit a message
- Changed how to comment box input ref is used on the `InputStorageWidget` to accommodate the changes made to `TextareaAutosize`
- Changed the ref returned by `TextareaAutosize` from the ref of the react component to specifically the ref to the HTML element
- Updated `react-textarea-autosize` lib in order to solve some weird `ref` issues I encountered while working on this. Also to get that sweet TS support

Resolves #2719

### DIFF
--- a/package.json
+++ b/package.json
@@ -188,7 +188,7 @@
     "react-redux-promise-listener": "^1.0.0",
     "react-router-dom": "^5.1.2",
     "react-tabs": "^3.1.0",
-    "react-textarea-autosize": "^7.1.2",
+    "react-textarea-autosize": "^8.3.3",
     "recompose": "^0.30.0",
     "redux": "^4.0.1",
     "redux-immutable": "^4.0.0",
--- a/src/modules/dashboard/components/ActionsPage/InputStorageWidget/InputStorageWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/InputStorageWidget/InputStorageWidget.tsx
@@ -83,97 +83,78 @@ const InputStorageWidget = ({
   }, [storageSlot, colonyAddress, fetchStorageSlotValue]);
 
   /*
-   * We need to handle these manually using a ref, since the TextAreaAutoresize
-   * component doesn't give us access to the native React ones :(
+   * Prevent entering a new line
+   *
+   * Based on specs we need to **fake** a multi-line input, but one
+   * you can't actually create a new line manully, just if the text
+   * overflows...
+   *
+   * While we're at it we also disable the space bar, since no value
+   * that can be entered requires a space (it's just a litte nice-to-have)
    */
-  const handleStorageSlotLocationBlur = useCallback(
-    (
-      autoResizeTextarea: HTMLElement | null,
-      currentValue,
-      fieldError,
-      updateValues,
-    ) => {
-      if (autoResizeTextarea) {
-        /*
-         * Add 0x prefix on blur and set the storage slot in state
-         */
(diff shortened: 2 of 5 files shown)


rdig (changes requested):
So the autofocus part of the comment box works _(I haven't tested it on the Coin Machine chat though)_

But your changes broke the `onBlur` handler in the `InputStorageWidget`, basically breaking all it's functionality.

The onblur, would fetch the input storage slot current value, as well as append the hex prefix if the value didn't have it.  Here is a  demo of that:

demo-input-storage-widget

But your current changes made the input storage widget `onBlur` not trigger anymore. Here's how it behaves:

bug-input-storage-widget-onblur

Please go back and fix this, because as it is right now, it breaks the Recovery Mode functionality almost in it's entirety.

src/modules/dashboard/components/ActionsPage/InputStorageWidget/InputStorageWidget.tsx
```diff
@@ -82,88 +82,79 @@ const InputStorageWidget = ({
...
+   * you can't actually create a new line manully, just if the text
+   * overflows...
+   *
+   * While we're at it we also disable the space bar, since no value
+   * that can be entered requires a space (it's just a litte nice-to-have)
+   */
+  const handleOnKeyDown = useCallback((event) => {
+    if (event.key === ENTER || event.code === SPACE) {
+      event.preventDefault();
+    }
```
rdig, line 98: Note that this change will likely cause a merge conflict for @chinins who started working on #2713

package.json
```diff
@@ -188,7 +188,7 @@
     "react-redux-promise-listener": "^1.0.0",
     "react-router-dom": "^5.1.2",
     "react-tabs": "^3.1.0",
-    "react-textarea-autosize": "^7.1.2",
+    "react-textarea-autosize": "^8.3.3",
```
rdig, line 191: I haven't checked the new version properly, but it seems to be working without issues.

Let's hope this upgrade doesn't break something we're not aware of :)

rdig (approved):
Works fine now. Very nice work @ArmandoGraterol :100:

src/modules/dashboard/components/ActionsPage/InputStorageWidget/InputStorageWidget.tsx
chinins: Well, I guess I'll be rebasing the coin-machine-redux on master anyway :)

chinins (approved):
Very nice work!

PR #2742, opened 2021-09-22
Fix ipfs domain error showing "Ooops, action not found"

### LINKED ISSUE #2739: Changing team name (via force) causing  `Oops, Action Not Found`
## Steps to reproduce

1. Take an existing team & change the name
1. Force the action

A user reported this via Canny, I replicated on my test Colony https://xdai.colony.io/colony/test-kh

## Expected behaviour

* No error page

## Actual behaviour

* User is taken to a "Oops, action not found"

Example transaction hash: 0x0cc12e1162ee921dbfb9634d5ba6ab7b8e2612ca2ceb964dbda6120abed14cef

ArmandoGraterol:
Managed to replicate this once on QA.

It seems like the "Ooops, action not found" page would show if the IPFS failed to retrieve the domain data due to a time-out. With these changes, we can at least show the action/motion page even when the IPFS fails, so the user is no longer left confused about whether the transaction was created or not.

The changes should affect specifically the edit domain action. create domain and edit domain motion.

Resolves #2739

### DIFF
--- a/src/utils/events/index.ts
+++ b/src/utils/events/index.ts
@@ -205,6 +205,51 @@ const getPaymentActionValues = async (
   return paymentActionValues;
 };
 
+const getDomainValuesFromIPFS = async (ipfsHash: string) => {
+  let ipfsMetadata: any = null;
+  let domainName = null;
+  let domainColor = null;
+  let domainPurpose = null;
+
+  try {
+    ipfsMetadata = await ipfs.getString(ipfsHash);
+  } catch (error) {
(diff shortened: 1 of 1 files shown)


rdig (approved):
Couldn't test this properly as I can't really replicate this reliably either locally or on QA. 

However, the changes seems valid from a logic standpoint, so I'm just going to trust that they work in production as well.

Good to go!

chinins (approved):
Looks good!

PR #2747, opened 2021-09-27
Remove whitelist address field when whitelist extension is not installed

### LINKED ISSUE #2743: Estimation error when enabling CoinMachine extension (network does not support ENS) without whitelist
## Steps to reproduce

1. Create new colony
2. Only install Coin Machine
3. Enable Coin Machine

## Expected behaviour

* Coin Machine can be installed without whitelist extension enabled

## Actual behaviour

screencapture-qaxdai-colony-io-colony-colonyurl-extensions-CoinMachine-setup-2021-09-22-17_11_01 (1)

chinins:
## Description

This PR removes the whitelists address field from Coin Machine extension setup if the whitelist extension is not enabled. This is agreed with Jack and Chad in the product-private channel. In this case the address zero is submitted as whitelist address so that there is no error during transaction.

**Changes** 🏗

- update `ExtensionSetup` component.

resolves #2743

### DIFF
--- a/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
+++ b/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
@@ -201,7 +201,7 @@ const ExtensionSetup = ({
     whitelistAddress,
   ]);
 
-  const showWhitelistAddressField = useCallback(
+  const showInputField = useCallback(
     (paramName) => {
       return (
         paramName !== 'whitelistAddress' ||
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
```diff
@@ -201,6 +201,17 @@ const ExtensionSetup = ({
     whitelistAddress,
   ]);
 
+  const showInputField = useCallback(
```
rdig, line 204: It works, but this is kinda super specific for Whitelist.

I which there was a better way to extension specific cases like this somewhere else, but I can't really think of an alternative :(

rdig (approved):
Seems to be working fine

src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
chinins: I know, I don't like it too much myself but I thought about other options and couldn't come up with a better solution...

ArmandoGraterol (approved):
Looks good to me :ok_hand:

PR #2758, opened 2021-09-29
Remove ETH mainnet as a allowed network

### LINKED ISSUE #2740: Wrong chain still works on MetaMask
## Steps to reproduce

1. Open a Colony
2. Switch MetaMask to ETH Mainnet
3. Create an Expenditure

## Expected behaviour

* Blocked or heavily warned that you're trying to do an action on ETH

## Actual behaviour

* UX acts as if it's expected behavior

ArmandoGraterol:
FireShot Capture 454 - Colony - localhost

Resolves #2740


rdig (changes requested):
This is not the correct approach.

We need the codebase to support both XDAI and ETH, and that's how we built it so far. Removing ETH now, will break that, so when we'll need to re-deploy to ETH mainnet we'll have to refactor everything. Not fun.

The correct approach here is to detect where the current instance of the dapp is deployed by querying the `NETWORK` environment variable, and based on that, filter the `ALLOWED_NETWORKS` object to only include that network

src/modules/dashboard/components/ColonyHome/WrongNetworkDialog/WrongNetworkDialog.tsx
```diff
@@ -13,12 +14,15 @@ const MSG = defineMessages({
   },
   description: {
     id: 'dashboard.ColonyHome.WrongNetworkDialog.description',
-    defaultMessage: 'Please connect to the appriopriate Ethereum network.',
+    defaultMessage: 'Please connect to xDai chain. <a>Learn more.</a>',
```
rdig, line 17: Again, this is not correct. What if we deploy this to mainnet. Then the user has the wrong network selected in the wallet. Then they receive e message saying: "Please connect to the xDai chain", while they are on ethereum mainnet.

See the problem ?

So this should be dynamic, just as everything. And this should use the same logic of detecting the network as `checkIfNetworkIsAllowed` _(but not the same util method)_

In this case, just take the current network where the dapp is deployed (from `process.env.NETWORK`, map it using the `NETWORK_DATA` constant to get it's "official" name _(the `name` prop)_ and then use that as a value in this message.

```suggestion
    defaultMessage: 'Please connect to {networkName}. <a>Learn more.</a>',
```

src/modules/constants.ts
```diff
@@ -116,7 +116,7 @@ export const TOKEN_DATA = {
   [Network.Mainnet]: ETHER_TOKEN,
 };
 
-export const ALLOWED_NETWORKS = {
+export const SUPPORTED_NETWORKS = {
```
rdig, line 119: I don't really know if this variable name change was necessary, but I'm good with it either way.

src/modules/dashboard/components/ColonyFundingMenu/ColonyFundingMenu.tsx
```diff
@@ -53,7 +53,7 @@ const ColonyFundingMenu = ({
   selectedDomainId,
 }: Props) => {
   const { walletAddress, networkId, ethereal, username } = useLoggedInUser();
-  const isNetworkAllowed = !!ALLOWED_NETWORKS[networkId || 1];
+  const isNetworkAllowed = checkIfNetworkIsAllowed(networkId);
```
rdig, line 56: Very nice refactoring all these files to use your `checkIfNetworkIsAllowed` util

rdig (changes requested):
Almost there, but you've done some good work nonetheless

src/modules/constants.ts
ArmandoGraterol: I changed the name because in my mind, while yes,  we support all of those networks, they _might_ not be  `allowed` at the time, depending on what chain is the dapp deployed. Nothing serious, I just think semantically speaking, it makes more sense

PR #2693, opened 2021-09-07
Add `Smite` feature

chinins:
## Description

This PR is for `Smite` feature implementation.

Corresponds to `Smite` project & all cascading issues.

PR #2770, opened 2021-10-04
Add award reputation dialog

### LINKED ISSUE #2699: Add award reputation part to the smite dialog & make UI conditional on which type of dialog is opened
The UI is pretty much the same so I guess it should be a prop that indicates which type of dialog is shown (for copies etc) & hence an argument in the action on which motion/action to trigger.

ArmandoGraterol:
Added award reputation dialog UI and form validation, which most of it is shared with the SmiteDialog 

**New stuff** ✨

* Added AwardDialog to UAC

**Changes** 🏗

* Created AwardAndSmiteDialogForm components for things that are reusable between the SmiteDialog and AwardDialog
* Updated SmiteDialog with components that are shared with AwardDialog

FireShot Capture 463 - Colony - localhost

## TODO

- [x] Rebase this to award reputation's feature branch when it exists

Resolves #2699


src/modules/dashboard/components/AwardDialog/AwardDialog.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+import { DEFAULT_TOKEN_DECIMALS } from '~constants';
+import { ActionTypes } from '~redux/index';
+import { pipe, withMeta, mapPayload } from '~utils/actions';
+
+import DialogForm, {
+  AwardAndSmiteDialogProps,
+  AwardAndSmiteDialogFormValues,
+  AwardAndSmiteFormValidationSchema,
+} from '../AwardAndSmiteDialogForm';
```
rdig, line 19: All of `Smite`, `Award` and `AwardAndSmite` componets should live in the same folder.

Something like `src/modules/dashboard/components/AwardAndSmiteDialogs` or something...

Please move them to a single location

rdig (approved):
This is all good. I do want the dialogs moved to a central folder, but other than that, this is good to go! :heavy_check_mark: 

However, there's another discussion to have here, as I saw this is rebased on the `Smite` feature.

We're planning on putting Smite up on QA, then merging it into master. If we wait for Award to finish, we'll delay this.

So how do we want to proceed?
- wait until award is done, then put on QA, then merge all ?
- separate this into it's own separate feature branch ?

I'd vote for the latter, since I want Smite merged faster, but this is @chinins 's feature, so she has the ultimate say on how we proceed.

src/modules/dashboard/components/AwardAndSmiteDialogForm/AwardAndSmiteDialogForm.tsx
```diff
@@ -283,12 +257,12 @@ const SmiteDialogForm = ({
         <div className={styles.singleUserContainer}>
           <SingleUserPicker
```
rdig, line 258: As I commented in Olga's #2693 this also needs the first user in the colony, pre-selected in the user picker _(same way we do in the Permissions modal)_

Currently it is not:
Screenshot from 2021-11-02 16-44-56

chinins (commented):
In my opinion, there is too much of repetitive code in Award & Smite dialog components. I think it should be one component/file & there you distinguish between the 2.

I am happy to reconsider if there are any impediments that I didn't think of or any other things to consider...

Regarding the rebase, I am happy for the Smite to be shipped first, then Award reputation to be rebased on the master & put on QA separately.

For the user picker, I have created a new issue int he Smite feature:
https://github.com/JoinColony/colonyDapp/issues/2829

src/modules/dashboard/components/AwardAndSmiteDialogForm/AwardAndSmiteDialogForm.tsx
ArmandoGraterol: Once #2830 is merged I'll make sure the changes are compatible with the award dialog

ArmandoGraterol:
> In my opinion, there is too much of repetitive code in Award & Smite dialog components. I think it should be one component/file & there you distinguish between the 2.
> 
> I am happy to reconsider if there are any impediments that I didn't think of or any other things to consider...

@chinins  Normally, you would think that you can easily pass down a _dynamic_ boolean prop that tells the dialog what action to use/display, but I didn't find a way of telling `useNaiveBranchingDialogWizard` that. That's why I didn't combine `AwardDialog` and `SmiteDialog` as I did with the form component. 

I tested a few ways of combining both dialogs which I thought could work but as of now, I haven't found a working solution other than making 2 separate dialogs and minimizing as much as I can the repetition (Once this PR is merged with #2827 we might have more opportunities to remove code repetition)

chinins:
> > In my opinion, there is too much of repetitive code in Award & Smite dialog components. I think it should be one component/file & there you distinguish between the 2.
> > I am happy to reconsider if there are any impediments that I didn't think of or any other things to consider...
> 
> @chinins Normally, you would think that you can easily pass down a _dynamic_ boolean prop that tells the dialog what action to use/display, but I didn't find a way of telling `useNaiveBranchingDialogWizard` that. That's why I didn't combine `AwardDialog` and `SmiteDialog` as I did with the form component.
> 
> I tested a few ways of combining both dialogs which I thought could work but as of now, I haven't found a working solution other than making 2 separate dialogs and minimizing as much as I can the repetition (Once this PR is merged with #2827 we might have more opportunities to remove code repetition)

You can have 2 separate dialog shallow wrappers for each one of Smite & Award reputation where the only thing you do is injecting this boolean prop and potentially the separate translation strings object (might be good to separate that). And then you have a one dialog component with all of the same logic in one place that you call from both of these wrappers. Let me know if you need any more explanation on how to reduce the code repetition.

chinins (changes requested):
As per what I've mentioned in my comment.

ArmandoGraterol:
> You can have 2 separate dialog shallow wrappers for each one of Smite & Award reputation where the only thing you do is injecting this boolean prop and potentially the separate translation strings object (might be good to separate that). And then you have a one dialog component with all of the same logic in one place that you call from both of these wrappers. Let me know if you need any more explanation on how to reduce the code repetition.

That's a good idea, didn't think about it that way! Definitely worth a try

chinins (approved):
I think it's much better like this :) Nice work

src/modules/dashboard/components/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -70,16 +63,18 @@ const SmiteDialog = ({
 
   const getFormAction = useCallback(
     (actionType: 'SUBMIT' | 'ERROR' | 'SUCCESS') => {
+      // @TODO: To be updated. This won't be needed once #2827's changes are integrated.
+      const actionName = isSmitingReputation ? 'SMITE' : 'AWARD';
```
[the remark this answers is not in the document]
ArmandoGraterol: If I merge #2827 first, which is what I've planned, then yeah I'll come back and update this

PR #2773, opened 2021-10-05
Fix staking slider no more rep error

### LINKED ISSUE #2768: Motion tells me I don't have enough reputation to stake more when I do.
1. This message shows before I've tried to stake anything at all, so even if it were true it would be the wrong message.
2. I have enough reputation to stake it completely, and I am able to stake it completely if I attempt to.

Screenshot 2021-10-04 at 18 17 01

https://xdai.colony.io/colony/swashbuckle/tx/0x03dc6fdb56f76245afd2e1398f6ae501dedeed99ccd72b152da8efad63033fc0

ArmandoGraterol:
- Fixed token activation popover not validating correctly some token amounts due to Javascript being weird with type coercion
- Fixed staking slider error showing up in cases where the user had more active tokens than reputation points, but had enough reputation to stake the entire remaining stake, and therefore the error making no sense.
- FIxed staking slider sometimes having a wrong limit due to the precision of the numbers used to calculate it

Resolves #2768

### DIFF
--- a/src/modules/core/components/Slider/Slider.tsx
+++ b/src/modules/core/components/Slider/Slider.tsx
@@ -1,6 +1,7 @@
 import React, { useState, useMemo, useCallback, useEffect } from 'react';
 import ReactSlider from 'rc-slider';
 import { useField } from 'formik';
+import Decimal from 'decimal.js';
 
 import 'rc-slider/assets/index.css';
 
@@ -15,7 +16,7 @@ interface Props {
   value: number;
   max?: number;
   min?: number;
-  limit?: number;
+  limit?: Decimal;
   appearance?: Appearance;
   onChange?: (val: any) => void;
   name: string;
@@ -53,18 +54,18 @@ const Slider = ({
   }, [sliderValue, value, setSliderValue]);
 
   const limitValue = useMemo(() => {
-    return limit ? limit * 100 : 0;
+    return limit ? limit.times(100) : new Decimal(0);
   }, [limit]);
 
   const gradientPercentage = useMemo(
-    () => (limitValue >= 100 ? 100 : limitValue),
+    () => (limitValue.gte(100) ? new Decimal(100) : limitValue),
     [limitValue],
   );
 
   const onSliderChange = useCallback(
     (val): void => {
       if (
-        (limit !== undefined && sliderValue < limitValue) ||
+        (limit !== undefined && limitValue.gt(sliderValue)) ||
         val < sliderValue ||
         !limitValue
       ) {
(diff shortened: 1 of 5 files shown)


chinins (dismissed):
Looks good!

ArmandoGraterol:
@rdig After rebasing, I couldn't replicate the first case and the second case seems to be correct so please test this again and notify if you see something weird again.

rdig (changes requested):
Sadly this is still the same:

Screenshot from 2021-11-08 15-19-36

Error shows up, however the stake button was not disabled and I could click it an generate the transactions _(which went through successfully)_

So I guess we'll just have to replicate the logic the button is using for disabled/enabled states to show/hide the specific errors.

ArmandoGraterol:
> Error shows up, however the stake button was not disabled and I could click it an generate the transactions _(which went through successfully)_
> 
> So I guess we'll just have to replicate the logic the button is using for disabled/enabled states to show/hide the specific errors.

@rdig Wait, isn't this "error" (More of a warning than an error) correct? It's saying `You do not have enough active tokens to stake more`, which is true, you only have `0.045` active tokens and therefore you can only stake up to that amount, but if you activate more tokens, you could stake more. Basically, the error is more of an informative message or a warning

At least that's what I understand about how this error works

chinins:
> > Error shows up, however the stake button was not disabled and I could click it an generate the transactions _(which went through successfully)_
> > So I guess we'll just have to replicate the logic the button is using for disabled/enabled states to show/hide the specific errors.
> 
> @rdig Wait, isn't this "error" (More of a warning than an error) correct? It's saying `You do not have enough active tokens to stake more`, which is true, you only have `0.045` active tokens and therefore you can only stake up to that amount, but if you activate more tokens, you could stake more. Basically, the error is more of an informative message or a warning
> 
> At least that's what I understand about how this error works

@rdig, I agree with Armando. This was my understanding as well and it looks like expected behaviour. You have 0.045 active tokens and this is how much you could stake - seems correct to me

rdig:
I agree that it's more like a "info" message than anything else, however it is still styled as an error, and to the layman, they can't be differentiated.

So I'd suggest talking to product about styling it differently, and maybe even a different copy message. Something like: "this is the max you can stake right now", or something along those lines, because as it currently stands it's confusing as hell.

chinins:
> I agree that it's more like a "info" message than anything else, however it is still styled as an error, and to the layman, they can't be differentiated.
> 
> So I'd suggest talking to product about styling it differently, and maybe even a different copy message. Something like: "this is the max you can stake right now", or something along those lines, because as it currently stands it's confusing as hell.

We can revisit it with product but that was initially what was given by them...

rdig (approved):
All good now! :+1: 

I still think we should have different styling for "info"-type messages, but if product is fine with them, let's not spend more time here.

chinins (approved):
All good, nice work!

PR #2788, opened 2021-10-14
Expand coin machine extension enabling

### LINKED ISSUE #2761: Coin Machine needs to be able to sell locked tokens.
Currently, Coin Machine can only sell unlocked tokens. This will be a problem for token sales, including ours, where it's important that tokens are not tradable while the token sale is in progress.

When Coin Machine is deployed/enabled, we also need to deploy a new authority for the token giving the coinmachine permissions to transfer the locked token.

ArmandoGraterol:
Added setup of token authority on coin machine enabling in order to control whether the token can be sold locked or not

Resolves #2761

### DIFF
--- a/src/modules/dashboard/sagas/extensions/colonyExtensionEnable.ts
+++ b/src/modules/dashboard/sagas/extensions/colonyExtensionEnable.ts
@@ -97,6 +97,7 @@ function* colonyExtensionEnable({
       });
 
     if (!initialized && extension.initializationParams) {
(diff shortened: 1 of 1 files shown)


ArmandoGraterol:
> It's all working well and the code looks good. I was thinking about this PR this weekend and maybe I am missing something but do we always have to do this arbitrary transaction? Is it worth checking if the token is locked or not before?

Asked Alex, it doesn't need to be so I'll add an extra condition for this later

PR #2793, opened 2021-10-15
Improve token remaining widget data refresh

### LINKED ISSUE #2782: Refactor Tokens Remaining Widget Data Fetching
Based on my work on #2765 I realized we can improve this in two ways:
- fetch events from the subrgaph rather than directly from the network _(this cuts down on node requsts)_
- refresh data automatically after a period ends
- refresh data automatically when other users buy tokens

### Fetching events from the subgraph

Rather then using `client.filter.EventName` we can just construct a custom subgraph query which can aggregate all the events we need _(and if not add them, so that the subgraph handles them)_.

For Coin Machine I've already done most of the work here, since we only need the `TokensBought` and `Transfer` events, so please feel free to either copy-paste and modify my queries/resolvers or just strait up use them as they are.

The events query example:
https://github.com/JoinColony/colonyDapp/blob/2bb47f7bbca70c48fe56b8173f78af6bb379a5a3/src/data/graphql/queries/coinMachine.graphql#L63-L111

And here is how I am using it inside the resolver itself:
https://github.com/JoinColony/colonyDapp/blob/2bb47f7bbca70c48fe56b8173f78af6bb379a5a3/src/data/resolvers/coinMachine.ts#L279-L290
https://github.com/JoinColony/colonyDapp/blob/2bb47f7bbca70c48fe56b8173f78af6bb379a5a3/src/data/resolvers/coinMachine.ts#L338-L348

### Refresh data automatically after a period ends

This is done inside the component, with an interval set to the period's length _(plus an initial one that is set to the time remaining in the current period, if it's already started)_

This is needed because if a user stays on the page when a period rolls over, they will still operate with the old period's data, so that the number of available tokens will wrong, same as with the price, and most certainly the validation will fail.

I've done all of the work here, so this is just a copy-paste job and adapting it to your specific component:
https://github.com/JoinColony/colonyDapp/blob/4e6868d3ae91388b96ed18996c0c9d725d7ae0e0/src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.tsx#L91-L99
https://github.com/JoinColony/colonyDapp/blob/4e6868d3ae91388b96ed18996c0c9d725d7ae0e0/src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.tsx#L147-L174

# Refresh data automatically when other users buy tokens

This is the most tricky part here, as you'll need to create a subgraph `subscription` for the `TokensBought` event for the current Coin Machine contract.

Once new items are received through the subscription websocket, you'll know that another user bought tokens, so you'll need to trigger a data refresh to reflect the new values.

This is needed since other people might buy tokens and we need to reflect that to the user _(not allowing them to buy more than available, otherwise the transaction will fail)_

The `subscription` is written exactly the same way as the query I've showed you above, just that it has the keyword `subscription` in front. Note that you'll have to create new types for the subscription, as they are not share with queries.

If you run into any issues with all of this let me know, and I'll do my best to help.

ArmandoGraterol:
## Description

- Added token remaining widget data refreshing when other users buy tokens and when the period ends

**New stuff** ✨

* Added `SubgraphTokenBoughtEvents` subgraph subscription
* Added useEffect on `CoinMachine` to refetch period tokens data when the period ends
* Added useEffect on `CoinMachine` to refetch period tokens data when a new `tokenBoughts` event is emitted

Resolves #2782

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
+++ b/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
@@ -1,9 +1,10 @@
-import React, { useMemo } from 'react';
+import React, { useEffect, useMemo, useState } from 'react';
 import { Redirect, useParams } from 'react-router-dom';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { Extension } from '@colony/colony-js';
 import { bigNumberify } from 'ethers/utils';
 
+import { AddressZero } from 'ethers/constants';
 import ExternalLink from '~core/ExternalLink';
 import { SpinnerLoader } from '~core/Preloaders';
 import BreadCrumb, { Crumb } from '~core/BreadCrumb';
@@ -16,6 +17,7 @@ import {
   useCoinMachineSalePeriodQuery,
   useCoinMachineTokenBalanceQuery,
   useSubgraphCoinMachinePeriodsQuery,
+  useSubgraphTokenBoughtEventsSubscription,
 } from '~data/index';
 
 import Chat from './Chat';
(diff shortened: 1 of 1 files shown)


rdig:
I am planning to merge colonyDapp#2782 when I get around to fixing it :(

But I guess that will be much later than this one, and It's me who's going to have to deal with the rebase :)

Go you can go ahead an merge

Issue #2668, opened 2021-08-30
Fetch Motion Data from The Graph


rdig:
_placeholder_

PR #2810, opened 2021-10-20
Fetch events using subgraph in motions resolvers

### LINKED ISSUE #2797: Fetch events using the Subgraph in the `motions` resolvers
We need to refactor most _(if not all)_ of the places inside the Dapp _(and it's external libraries, like `colonyJS`)_ and change the way we fetch colony related events.

Currently we are fetching events directly from the network, using the client's own filters, but this is proving to be problematic as we already hit the limit of historic blocks that the RPC endpoints are wiling to return:

Screenshot from 2021-10-14 09-30-37

_(this is on `xdai`, however on `mainnet` this problem should be much worse)_

To solve this we need to refactor places where me make those calls to instead fetch events from the [Subgraph](https://github.com/joinColony/subgraph). This should be relatively strait forward as it's already set up to do so, plus we have all the helpers needed inside the dapp to process the specific events so that they will be a drop-in replacement from the network events.

For this, you'll have to:
- [ ] Add the events to the subgraph, or ensure they there
- [ ] Add copy and types for events inside the dapp
- [ ] Refactor the place we're fetching network events to call the apollo query _(most likely in the resolvers)_
- [ ] Make sure the new events don't put a too much sync strain on the Subgraph

## Branching 
**Please note to branch your dapp and subgraph from the following branches (not `master`). This is very important, don't skip this step:**
- Dapp should be branched from `refactor/user-roles-from-subgraph-events` --  #2794
- Subgraph should be branch form `feature/handle-user-roles-related-events` -- [JoinColony/subgraph#34](https://github.com/JoinColony/subgraph/pull/34)

## Add events to the subgraph

We already track most events in the subgraph so this step might be already done for you. However do have a look and ensure you are tracking the correct event _(and if needed, it's older version)_.

_See this example on how older versions of events are handled: [9a30c07](https://github.com/JoinColony/subgraph/pull/34/commits/9a30c072c1f22f2897fdcdbe564838455979115c)_

If however you need to add new events make sure you pass the correct `address` argument to the `handleEvent` handler as that might impede your ability to filter them out properly.

If you have to handle a network level event make sure you first check with Alex or me, on the proper address to pass down to the handler.

## Add copy and types for the new event

If you ended up adding a new event on the subgraph, it will automatically picked up by our dapp in the general `events` query.

For this to be properly reflected, make sure you add the new event's type inside: `src/types/ColonyActions.ts`, in the `ColonyAndExtensionsEvents` enum

And the copy for it goes into `src/i18n/en-events.ts`

## Refactor the Dapp to fetch events from the subgraph

In this step you'll change places were the client `filters` are used with direct subgraph queries. Here's and example on how to do that: https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/resolvers/colony.ts#L458-L470

_NOTE: please check and decide if it's worth to add the `fetch-policy` to your query, as it incurs a large penalty. Only add it if **you specifically have a use for it**_

For `SubgraphEvent`s you will already have the types declared so you'll just have to write the query that fetches then and re-generate the hooks. **Make sure there isn't a similar query to the one you're writing to avoid code repetition. If there is, try and see if we you can reuse it**

If you have to write a new query make sure to place it inside `data/graphql/queries/<query-category>.graphql`

Events from the subgraph will come unformatted and unsorted, so you'll have to use the following two helpers:
- `parseSubgraphEvent` for parsing, it will parse and format your events so that they will be exactly as the one coming from the network _(with minor differences of course, but they won't affect you here)_
- `sortSubgraphEventByIndex` for sorting, if your care about that, in places where you would need to replay the events historically 

Here's an usage example for both:  https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/resolvers/colony.ts#L478-L485

Note, that for sorting to work you'll have to fetch both the event's `id` and the block's `number` inside the query, as the unique event index is constructed from those two values. Here's how the query should look for that to work: https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/graphql/queries/roles.graphql#L2-L21

## Make sure the Subgraph syncs in a timely fashion

Currently a new subgraph deployment on QA takes about ~41 minutes to fully synchronize. Every new event being handled puts strain on this and adds to the sync time.

We don't want this to get out of hand _(it already takes about 1 hour to do a production deployment, so this limits our ability to proactive)_, so **it's up to you to raise the issue, if you notice the subgraph syncs slower after your changes**.

To test this, we'll do a QA build and see how long it takes for it to be fully sync'd, **after** you've ensured that it works as it should locally. 

## Specific notes for this issue:

https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/resolvers/motions.ts#L41-L941

chinins:
## Description

This PR changes calls to get events directly from the Network with subgraph queries in motions resolvers.

There was quite a lot of type errors which I tired to resolve nicely. Please, let me know if you think there is a better way.

If it is reviewed on Friday I think I can still make updates if necessary and merge on Friday as well.

**New stuff** ✨

- added lots of motions subgraph queries

**Changes** 🏗

- updated motions resolvers
- updated types for events

## TODO

- change  `SubgraphMotionVoteSubmittedEventsQuery` to fetch user specific events to avoid filtering in the resolver

Resolves #2797
Contributes to #2668

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -30276,12 +30276,14 @@
     },
     "node_modules/gc-stats/node_modules/abbrev": {
       "version": "1.1.1",
+      "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
       "inBundle": true,
       "license": "ISC",
       "optional": true
     },
     "node_modules/gc-stats/node_modules/ansi-regex": {
       "version": "2.1.1",
+      "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
       "inBundle": true,
       "license": "MIT",
       "optional": true,
@@ -30291,12 +30293,14 @@
     },
     "node_modules/gc-stats/node_modules/aproba": {
       "version": "1.2.0",
+      "integrity": "sha512-Y9J6ZjXtoYh8RnXVCMOU/ttDmk1aBjunq9vO0ta5x85WDQiQfUF9sIPBITdbiiIVcBo03Hi3jMxigBtsddlXRw==",
       "inBundle": true,
       "license": "ISC",
       "optional": true
     },
     "node_modules/gc-stats/node_modules/are-we-there-yet": {
       "version": "1.1.5",
+      "integrity": "sha512-5hYdAkZlcG8tOLujVDTgCT+uPX0VnpAH28gWsLfzpXYm7wP6mp5Q/gYyR7YQ0cKVJcXJnl3j2kpBan13PtQf6w==",
       "inBundle": true,
       "license": "ISC",
       "optional": true,
@@ -30307,12 +30311,14 @@
     },
     "node_modules/gc-stats/node_modules/balanced-match": {
       "version": "1.0.0",
+      "integrity": "sha1-ibTRmasr7kneFk6gK4nORi1xt2c=",
       "inBundle": true,
       "license": "MIT",
       "optional": true
     },
     "node_modules/gc-stats/node_modules/brace-expansion": {
       "version": "1.1.11",
+      "integrity": "sha512-iCuPHDFgrHX7H2vEI/5xpz07zSHB00TpugqhmYtVmMO6518mCuRMoOYFldEBl0g187ufozdaHgWKcYFb61qGiA==",
       "inBundle": true,
       "license": "MIT",
       "optional": true,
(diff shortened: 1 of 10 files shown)


rdig (approved):
Yep, working as expected now. Good to go! :checkered_flag: 

Screenshot from 2021-11-02 15-50-08

ArmandoGraterol (approved):
It's working as expected to me! Great work

PR #2811, opened 2021-10-20
Update recovery resolvers with subgraph queries

### LINKED ISSUE #2798: Fetch events using the Subgraph in the `recovery` resolver
We need to refactor most _(if not all)_ of the places inside the Dapp _(and it's external libraries, like `colonyJS`)_ and change the way we fetch colony related events.

Currently we are fetching events directly from the network, using the client's own filters, but this is proving to be problematic as we already hit the limit of historic blocks that the RPC endpoints are wiling to return:

Screenshot from 2021-10-14 09-30-37

_(this is on `xdai`, however on `mainnet` this problem should be much worse)_

To solve this we need to refactor places where me make those calls to instead fetch events from the [Subgraph](https://github.com/joinColony/subgraph). This should be relatively strait forward as it's already set up to do so, plus we have all the helpers needed inside the dapp to process the specific events so that they will be a drop-in replacement from the network events.

For this, you'll have to:
- [ ] Add the events to the subgraph, or ensure they there
- [ ] Add copy and types for events inside the dapp
- [ ] Refactor the place we're fetching network events to call the apollo query _(most likely in the resolvers)_
- [ ] Make sure the new events don't put a too much sync strain on the Subgraph

## Branching 
**Please note to branch your dapp and subgraph from the following branches (not `master`). This is very important, don't skip this step:**
- Dapp should be branched from `refactor/user-roles-from-subgraph-events` --  #2794
- Subgraph should be branch form `feature/handle-user-roles-related-events` -- [JoinColony/subgraph#34](https://github.com/JoinColony/subgraph/pull/34)

## Add events to the subgraph

We already track most events in the subgraph so this step might be already done for you. However do have a look and ensure you are tracking the correct event _(and if needed, it's older version)_.

_See this example on how older versions of events are handled: [9a30c07](https://github.com/JoinColony/subgraph/pull/34/commits/9a30c072c1f22f2897fdcdbe564838455979115c)_

If however you need to add new events make sure you pass the correct `address` argument to the `handleEvent` handler as that might impede your ability to filter them out properly.

If you have to handle a network level event make sure you first check with Alex or me, on the proper address to pass down to the handler.

## Add copy and types for the new event

If you ended up adding a new event on the subgraph, it will automatically picked up by our dapp in the general `events` query.

For this to be properly reflected, make sure you add the new event's type inside: `src/types/ColonyActions.ts`, in the `ColonyAndExtensionsEvents` enum

And the copy for it goes into `src/i18n/en-events.ts`

## Refactor the Dapp to fetch events from the subgraph

In this step you'll change places were the client `filters` are used with direct subgraph queries. Here's and example on how to do that: https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/resolvers/colony.ts#L458-L470

_NOTE: please check and decide if it's worth to add the `fetch-policy` to your query, as it incurs a large penalty. Only add it if **you specifically have a use for it**_

For `SubgraphEvent`s you will already have the types declared so you'll just have to write the query that fetches then and re-generate the hooks. **Make sure there isn't a similar query to the one you're writing to avoid code repetition. If there is, try and see if we you can reuse it**

If you have to write a new query make sure to place it inside `data/graphql/queries/<query-category>.graphql`

Events from the subgraph will come unformatted and unsorted, so you'll have to use the following two helpers:
- `parseSubgraphEvent` for parsing, it will parse and format your events so that they will be exactly as the one coming from the network _(with minor differences of course, but they won't affect you here)_
- `sortSubgraphEventByIndex` for sorting, if your care about that, in places where you would need to replay the events historically 

Here's an usage example for both:  https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/resolvers/colony.ts#L478-L485

Note, that for sorting to work you'll have to fetch both the event's `id` and the block's `number` inside the query, as the unique event index is constructed from those two values. Here's how the query should look for that to work: https://github.com/JoinColony/colonyDapp/blob/dbd95d023aea22b59ed415a1f510df568aa6341f/src/data/graphql/queries/roles.graphql#L2-L21

## Make sure the Subgraph syncs in a timely fashion

Currently a new subgraph deployment on QA takes about ~41 minutes to fully synchronize. Every new event being handled puts strain on this and adds to the sync time.

We don't want this to get out of hand _(it already takes about 1 hour to do a production deployment, so this limits our ability to proactive)_, so **it's up to you to raise the issue, if you notice the subgraph syncs slower after your changes**.

To test this, we'll do a QA build and see how long it takes for it to be fully sync'd, **after** you've ensured that it works as it should locally.

ArmandoGraterol:
**New stuff** ✨

* Added `SubgraphRecoveryModeEventsQuery` subgraph query

**Changes** 🏗

* Updated `getSessionRecoveryEvents`
* Updated `recoveryAllEnteredEvents` resolver
* Updated `recoveryRolesUsers` resolver
* Updated `legacyNumberOfRecoveryRoles` resolver

**Yapa**
* Fixed a bug where entering recovery mode would throw Apollo related errors if the list of users with recovery roles contained a user without a profile due to `getMinimalUser` not returning the users with `__typename`(s) 
* Fixed some recovery mode events on the action feed using the wrong user as the initiator

Resolves #2798

### DIFF
--- a/src/i18n/en-events.ts
+++ b/src/i18n/en-events.ts
@@ -10,9 +10,9 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.DomainAdded} {{initiator} added Team: {fromDomain}}
       ${ColonyAndExtensionsEvents.ColonyUpgraded} {This colony has upgraded to {newVersion}}
       ${ColonyAndExtensionsEvents.RecoveryModeEntered} {{initiator} activated Recovery Mode}
-      ${ColonyAndExtensionsEvents.RecoveryStorageSlotSet} {{initiator} submitted new value for slot {storageSlot}}
-      ${ColonyAndExtensionsEvents.RecoveryModeExitApproved} {{initiator} approved exiting}
-      ${ColonyAndExtensionsEvents.RecoveryModeExited} {{initiator} exited Recovery Mode}
+      ${ColonyAndExtensionsEvents.RecoveryStorageSlotSet} {{user} submitted new value for slot {storageSlot}}
+      ${ColonyAndExtensionsEvents.RecoveryModeExitApproved} {{user} approved exiting}
+      ${ColonyAndExtensionsEvents.RecoveryModeExited} {{user} exited Recovery Mode}
       ${ColonyAndExtensionsEvents.MotionCreated} {{initiator} created a {motionTag}}
       ${ColonyAndExtensionsEvents.MotionStaked} {{staker} backed the {backedSideTag} by staking {amountTag}}
       ${ColonyAndExtensionsEvents.MotionFinalized} {{motionTag} was finalized. Stakes may be claimed.}
--- a/src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
+++ b/src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
@@ -363,6 +363,11 @@ const ActionsPageEvent = ({
                     <FriendlyName user={initiator} autoShrinkAddress />
                   </span>
                 ),
+                user: (
+                  <span className={styles.userDecoration}>
+                    <FriendlyName user={initiator} autoShrinkAddress />
+                  </span>
+                ),
                 storageSlot: values?.slot,
                 amountTag: (
                   <div className={styles.amountTag}>


alicjakujawa (dismissed):

ArmandoGraterol:
@rdig I updated the PR's description with an explanation of the bug and the fix, it's a really _really_ weird and annoying bug, don't know if there's an alternative or something that I'm missing, but alas it should be fixed.

rdig (changes requested):
Sadly it's even more broken now :

Entering recovery mode still fails _(first time when you give permissions, I can't test the second time around, since it's broken and doesn't let me exit recovery mode)_

Screenshot from 2021-11-02 15-15-31

And after a page refresh, when entering the recovery mode action page, gives out the same error, and prevents you from exiting recovery mode (does not show correct number of users with permissions, and does not show the "exit recovery" button once approvals are set)

Screenshot from 2021-11-02 15-16-24

Oh! I think I know what it is. I gave "Recovery" permissions to a user that doesn't have a profile created:
(Screenshot from 2021-11-02 15-19-53

So this might be the problem this time around. However this is a edge case we need to fix, as this will most likely happen in production. _(I know colonies that do this)_

rdig:
Yep. I confirm this works as expected when all users have a profile created.

Screenshot from 2021-11-02 15-23-53

So I guess this is just a case of handling users who don't have a profile yet

rdig (changes requested):
This is all good, however I think I found bug that was here before and we never noticed. So this is not on you, it's basically on me for not finding it sooner.

- I have two users with recovery permissions (`a` and `b`)
- they both create a recovery action (update slot value)
- both actions show up as user `a`

Here's a demo:
bug-same-user-for-all-recovery-actions

I think this is a case of the message descriptors values not being passed around properly, because other than that, everything seems to be tracked correctly.

ArmandoGraterol:
@rdig Should be fixed now! It was indeed a case of the wrong value being used on the message descriptors

rdig:
@ArmandoGraterol let me know if you want me to merge this for you.

chinins (approved):
Working as expected - well done! it was not an easy PR.

PR #2827, opened 2021-11-01
Add award reputation action/motion saga

### LINKED ISSUE #2700: Add actions, types & extend motion & action sagas for reward reputation feature

ArmandoGraterol:
Added award reputation action and motion saga and action type by reusing the smite saga to include both smite and award action.

### To test:
1. Go to `SmiteDialog`'s transform
2. Remove the `.mul(-1)` which is not needed when awarding rep
3. Change `isSmitingReputation` to `false` instead of true
4. Now you can make an award action by using the smite dialog, using it the same way you would use it to smite a user.
5. Bon appetite!

### TODO

- [ ] Connect action/motion type to award dialog (#2770)
- [ ] Rebase this to award rep's own feature branch

Resolves #2700


src/modules/dashboard/sagas/actions/manageReputation.ts
```diff
@@ -71,16 +68,18 @@ function* smiteAction({
         },
       });
 
-    yield createGroupTransaction(emitDomainReputationPenalty, {
+    yield createGroupTransaction(manageReputation, {
       context: ClientType.ColonyClient,
-      methodName: 'emitDomainReputationPenaltyWithProofs',
+      methodName: isSmitingReputation
+        ? 'emitDomainReputationPenaltyWithProofs'
+        : 'emitDomainReputationReward',
```
rdig, line 75: Why doesn't reward require proofs when sending. I think it should.

Please check with Alex and then let me know, as if it does, I need to build it into `colonyJS`

src/modules/dashboard/sagas/motions/manageReputationMotion.ts
```diff
@@ -33,23 +33,28 @@ function* smiteMotion({
...
+}: Action<ActionTypes.COLONY_MOTION_MANAGE_REPUTATION>) {
   let txChannel;
   try {
     /*
      * Validate the required values
      */
     if (!userAddress) {
-      throw new Error('A user address is required to smite this user');
+      throw new Error(
+        'A user address is required to manage the reputation this user',
```
chinins, line 48: Needs to be: `to manage reputation of the user`

src/modules/dashboard/sagas/motions/manageReputationMotion.ts
```diff
@@ -33,23 +33,28 @@ function* smiteMotion({
...
+      );
     }
     if (!domainId) {
-      throw new Error('Domain id not set for Smite transaction');
+      throw new Error(
+        'Domain id not set for manage the reputation transaction',
```
chinins, line 53: I am nitpicking but  I would say `Domain id not set for the manage reputation transaction`. Otherwise it sounds a bit off.

src/modules/dashboard/sagas/motions/manageReputationMotion.ts
```diff
@@ -33,23 +33,28 @@ function* smiteMotion({
...
+      );
     }
     if (!amount) {
-      throw new Error('Amount not set for Smite transaction');
+      throw new Error('Amount not set for manage the reputation transaction');
```
chinins, line 57: same here

chinins (approved):
Just a couple of wording nitpicks. Other than that & Raul's comment, all good!

src/modules/dashboard/sagas/actions/manageReputation.ts
ArmandoGraterol: In the words of our local wizard, Alex:

> Reward requires the root permission, which you do not need to provide a proof when you use
Penalty requires the arbitration permission, which does require a proof to leverage

With that being said, I thought rewarding required the arbitration permission, so I need to do some changes in my award dialog PR to reflect that it's the root permission the one that is required

rdig: Great to hear then. Less work for me :P

PR #2830, opened 2021-11-03
Add default user to user picker in `Smite` dialog

### LINKED ISSUE #2829: Add first colony user as a default in user picker
The Smite dialog should have the first colony user already selected in the picker (as we do in the Permissions dialog).

chinins:
## Description

This PR adds a default user to the user picker field in `SmiteDialog`. If there is just one user in the dropdown, this user will be selected. If it's more than 1 user, then we pick the first user that's not a logged in user (assuming that the logged in user doesn't want to smite himself)

**Changes** 🏗

- update `SmiteDialog` & `SmiteDialogForm` components

resolves #2829 

smite-dialog

### DIFF
--- a/src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
+++ b/src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useMemo, useState } from 'react';
+import React, { useCallback, useState } from 'react';
 import { FormikProps } from 'formik';
 import * as yup from 'yup';
 import { ROOT_DOMAIN_ID } from '@colony/colony-js';
@@ -12,9 +12,10 @@ import { ActionForm } from '~core/Fields';
 import { DEFAULT_TOKEN_DECIMALS } from '~constants';
 import { Address } from '~types/index';
 import { ActionTypes } from '~redux/index';
-import { useMembersSubscription, useLoggedInUser } from '~data/index';
+import { useMembersSubscription } from '~data/index';
 import { pipe, withMeta, mapPayload } from '~utils/actions';
 import { WizardDialogType } from '~utils/hooks';
+import { useSelectedUser } from '~utils/hooks/useSelectedUser';
 
 import DialogForm from './SmiteDialogForm';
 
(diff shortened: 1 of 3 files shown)


src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
```diff
@@ -126,6 +126,22 @@ const SmiteDialog = ({
     [totalReputationData],
   );
 
+  const { walletAddress: loggedInUserWalletAddress } = useLoggedInUser();
+
+  const selectedUser = useMemo(() => {
```
ArmandoGraterol, line 131: I think this function could be converted to a util function/hook, as it may be needed in other dialogs. You could do that in this PR, otherwise, you could merge it, and then I convert it in #2770

ArmandoGraterol (approved):
Working as expected! Good job

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
rdig: I agree with @ArmandoGraterol, as this should be a hook, since he's going to use it as well for the "Awards" modal.

The more hooks the better :)

But it's up to you guys where / who decides to implement it.

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
```diff
@@ -126,6 +126,22 @@ const SmiteDialog = ({
...
+      return undefined;
+    }
+
+    if (colonyMembers.subscribedUsers.length === 1) {
+      return colonyMembers.subscribedUsers[0];
+    }
+
+    return colonyMembers.subscribedUsers[0].id === loggedInUserWalletAddress
+      ? colonyMembers.subscribedUsers[1]
+      : colonyMembers.subscribedUsers[0];
```
rdig, line 142: So this is just for clarity / better readability, not to do with your code logic (which is fine).

But I would have written this as:
```js

const [firstSubscriber, secondSubscriber] = colonyMembers?.subscribedUsers;

if (!secondSubscriber) {
  return firstSubscriber;
}

return firstSubscriber.profile.walletAddress === loggedInUserWalletAddress 
  ? secondSubscriber
  : firstSubscriber;
```

rdig (approved):
Good to go from me as well! :+1:

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
chinins: done

rdig: Actually, the first line should be:

```js
const [firstSubscriber, secondSubscriber] = colonyMembers?.subscribedUsers || [];
```

To have a fallback in case `colonyMembers?.subscribedUsers` returns `undefined`

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
chinins: done

src/modules/dashboard/components/SmiteDialog/SmiteDialog.tsx
chinins: good catch! done

PR #2833, opened 2021-11-04
Fixes funding link alignment

### LINKED ISSUE #2814: Fix 'Manage Funds' link alignment
The arrow symbol to the left of the the _Manage Funds_ link on the Colony Home is misaligned for some _(all?)_ machines.

This is a cross OS, cross Machine or cross Resolution issue, so please ensure this will render properly no matter on which machine we try to render.

Note that we only support Chorme officially _(and Chrome based browsers unofficially)_ so you don't really have to bother with how it renders on anything else other than Chrome.

The component you need to be targeting is `ColonyTotalFunds`

Here's some examples of the problem:

I'd say this one is "ok-ish" in terms of alignment _(Linux, 1x Pixel Density, 100% Zoom)_
Screenshot from 2021-10-24 22-58-35

But when you zoom in, it starts to break down _(Linux, 1x Pixel Density, 250% Zoom)_
Screenshot from 2021-10-24 23-10-40

However, this one is clearly misaligned _(Linux, 2x Pixel Density, 100% Zoom)_
Screenshot from 2021-10-24 22-58-24

agustif:
## Description
- [x] Fix the alignment of the funding link

**New stuff** ✨
Added a gap: 15px to separate each button on itself Colony total balance / Manage funds

**Changes** 🏗
Moved from margin to flex based align-items: center; for alignment of items.
**Removed** 
Margin for the rightarrowicon

## TODO

- [ x]  Fixes the alignment of the funding link

## Screenshots **Updated**
image
image
image

Resolves #2814

### DIFF
--- a/src/data/resolvers/events.ts
+++ b/src/data/resolvers/events.ts
@@ -1,4 +1,4 @@
-import { ClientType, ColonyClientV6, getLogs } from '@colony/colony-js';
+import { ClientType, ColonyClientV6 } from '@colony/colony-js';
 import { AddressZero } from 'ethers/constants';
 import { Resolvers } from '@apollo/client';
 
@@ -58,12 +58,10 @@ export const eventsResolvers = ({
         )) as ColonyClientV6;
 
         const colonyInRecoveryMode = await colonyClient.isInRecoveryMode();
+        const enterRecoveryEvents =
+          recoveryStartedEvents?.data?.recoveryAllEnteredEvents || [];
 
-        const enterRecoveryEvents = await getLogs(
-          colonyClient,
-          colonyClient.filters.RecoveryModeEntered(null),
-        );
-        const [mostRecentRecoveryStarted] = enterRecoveryEvents.reverse();
+        const [mostRecentRecoveryStarted] = [...enterRecoveryEvents].reverse();
 
         const recoveryApprovalsForCurrentSession = await apolloClient.query<
           RecoveryRolesAndApprovalsForSessionQuery,
@@ -91,6 +89,7 @@ export const eventsResolvers = ({
         const {
           recoveryRolesAndApprovalsForSession: approvalsForSession,
         } = recoveryApprovalsForCurrentSession.data;
+
         const currentUserNeedsToApprove = approvalsForSession.find(
           ({ id: userWalletAddress, approvedRecoveryExit }) =>
             !approvedRecoveryExit &&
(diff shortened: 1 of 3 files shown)


alicjakujawa (commented):
https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=825%3A10764

Im not sure, was it a real bug? I see in designs that arrow should be more on top of text

chinins (changes requested):
It's looking fine on my machine with 100% or higher zoom. But it's not looking ok when I zoom out (as per the screens shot). On your screen shots it doesn't look ok on #2 & #3 (the arrow is too high in both). Please, fix this as well.
Screenshot 2021-11-05 at 16 17 27

chinins:
> https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=825%3A10764
> 
> Im not sure, was it a real bug? I see in designs that arrow should be more on top of text

It's good point to clarify with Karol. But what Alicja shared looks a bit strange to me - the text is not aligned either, so I'd just confirm with Karol.
Screenshot 2021-11-05 at 16 20 09

agustif:
> > https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=825%3A10764
> > Im not sure, was it a real bug? I see in designs that arrow should be more on top of text
> 
> It's good point to clarify with Karol. But what Alicja shared looks a bit strange to me - the text is not aligned either, so I'd just confirm with Karol. Screenshot 2021-11-05 at 16 20 09

Yeah I agree, I will look into using something else instead of margins, like flex or vertical-align or whatever to try to not have to put actual pixels so it works well in *all cases*

agustif:
I added some changes and updated the screenshots, could you re-review? Thanks

New approach should work better on any os/browser/resolutions/dpi whatever I think, no longer using margin moved to flex.

agustif:
> > https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=825%3A10764
> > Im not sure, was it a real bug? I see in designs that arrow should be more on top of text
> 
> It's good point to clarify with Karol. But what Alicja shared looks a bit strange to me - the text is not aligned either, so I'd just confirm with Karol. Screenshot 2021-11-05 at 16 20 09

Can you guys re-review, I took a new approach which should look good on most setups I hope!

agustif:
Moved https://github.com/JoinColony/colonyDapp/pull/2848

PR #2834, opened 2021-11-04
Fix ETH to USD conversion placement

### LINKED ISSUE #2813: Fix ETH to USD conversion placement
Current conversion rate estimates are shown to the right of the Token Selection dropdown, but the placement is off, as well as it being a bad place to show that.

Note that this only happens when deployed to a network which is using ETH as it's native token _(since we only provide conversion rates for that)_, but the principle still stands regarding this being a problem.

This needs to be refactored, to displayed in the same way / placement as we do for _"Available funds"_, below the dropdown, and aligned to the right. It needs to also have the same font styles and colors, but for the exact specifications look at the code and find them out from there.

Also note, that the new conversion rate alignment should not interfere with any input validation errors that might show up.

There's two modals that we display this, so both need to be fixed:
- Create Payment _(`CreatePaymentDialog`)_
- Transfer Funds _(`TransferFundsDialog`)_

Examples:

Screenshot from 2021-10-24 22-31-52
Screenshot from 2021-10-24 22-32-10
Screenshot from 2021-10-24 22-32-19
Screenshot from 2021-10-24 22-33-17

agustif:
## Description
Fixes the position of the converted amount in usd to be at the bottom of the selector input.
And makes it use the same styling than available funds further up it.
**New stuff** ✨
Added a new container tokenAmountContainer css class.
Which we wrap-in with a div our current EthUsd and Select Input form in each dialog in both form tsx files

**Changes** 🏗
* EthUsd > Appearance > size? making size optional. so it respects the parents if not given in order to make it be 11px and not 12px.
* for both CreatePaymentDialogForm.css and TransferFundsDialogForm.css, add the font size and weight to tokenAmountUsd class.

**Deletions** ⚰️
* Size small Appearance props on EthUsd component in the dialog forms.

## Screenshots
image
image

## TODO

- [x] Fix ETH to USD conversion placement in create payment dialog
- [x] Fix ETH to USD conversion placement in transfer funds dialog
- [x] It needs to also have the same font styles and colors than available funds too

Resolves #2813

### DIFF
--- a/src/modules/core/components/EthUsd/EthUsd.tsx
+++ b/src/modules/core/components/EthUsd/EthUsd.tsx
@@ -17,7 +17,7 @@ const MSG = defineMessages({
 
 interface Appearance {
   theme: 'primary' | 'grey' | 'dark';
-  size: 'medium' | 'large' | 'small';
+  size?: 'medium' | 'large' | 'small';
 }
 
 interface Props extends NumeralProps {
--- a/src/modules/core/components/FriendlyName/FriendlyName.css
+++ b/src/modules/core/components/FriendlyName/FriendlyName.css
@@ -1,6 +1,5 @@
 .main {
   composes: inlineEllipsis from '~styles/text.css';
-  display: inline-block;
+  display: contents;
   max-width: 150px;
-  vertical-align: text-bottom;
 }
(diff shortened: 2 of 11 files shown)


agustif:
moved to https://github.com/JoinColony/colonyDapp/pull/2844

PR #2844, opened 2021-11-09
Fix ETH to USD conversion

### LINKED ISSUE #2813: Fix ETH to USD conversion placement

agustif:
## Description
Fixes the position of the converted amount in usd to be at the bottom of the selector input.
And makes it use the same styling than available funds further up it.
**New stuff** ✨
Added a new container tokenAmountContainer css class.
Which we wrap-in with a div our current EthUsd and Select Input form in each dialog in both form tsx files

**Changes** 🏗
* EthUsd > Appearance > size? making size optional. so it respects the parents if not given in order to make it be 11px and not 12px.
* for both CreatePaymentDialogForm.css and TransferFundsDialogForm.css, add the font size and weight to tokenAmountUsd class.

**Deletions** ⚰️
* Size small Appearance props on EthUsd component in the dialog forms.

## Screenshots
image
image

## TODO

- [x] Fix ETH to USD conversion placement in create payment dialog
- [x] Fix ETH to USD conversion placement in transfer funds dialog
- [x] It needs to also have the same font styles and colors than available funds too

Resolves #2813

### DIFF
--- a/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
+++ b/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
@@ -22,7 +22,6 @@
   font-size: 11px !important;
   font-weight: 700;
   text-align: right;
-  justify-self: right;
 }
 
 .domainPotBalance {
--- a/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
+++ b/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
@@ -9,7 +9,19 @@
   margin-left: 20px;
 }
 
+.tokenAmountContainer {
+  display: flex;
+  flex-direction: column;
+  align-items: flex-start;
+  margin-top: 20px;
+  gap: 20px;
+}
+
 .tokenAmountUsd {
+  width: 100%;
+  font-size: 11px !important;
+  font-weight: 700;
+  text-align: right;
   justify-self: right;
 }
 
--- a/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
+++ b/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
@@ -19,7 +19,7 @@
 
 .tokenAmountUsd {
   width: 100%;
-  font-size: 11px !important;
+  font-size: 11px;
   font-weight: 700;
   text-align: right;
 }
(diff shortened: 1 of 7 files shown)


chinins:
Please, rebase on latest master (and let me know once done)

src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
```diff
@@ -9,8 +9,19 @@
...
+  flex-direction: column;
+  align-items: flex-start;
+  margin-top: 20px;
+  gap: 20px;
+}
+
 .tokenAmountUsd {
-  justify-self: right;
+  width: 100%;
+  font-size: 11px !important;
```
chinins, line 22: I don't like the use of important. Have you tried other ways of raising specificity like: 
`.tokenAmountUSD > span`

chinins (approved):
It's all good - very nice work! Just please remove `important!` (as per my earlier comment).

I don't think you need it at all (this is without important tag)
Screenshot 2021-11-10 at 13 44 35

src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.css
agustif: I've removed the !important from both tokenAmountUsd, since I made size optional and we're no longer passing small size appearance as props to EthUsd component, the size is correct at 11px without need of further specyfiying.

Does that work?

chinins: yes, very good

PR #2857, opened 2021-11-09
Wire up award action and motion page

### LINKED ISSUE #2701: Award reputation action page wiring

ArmandoGraterol:
Integrated award action/motion by adding the `EmitDomainReputationReward` related action types to the mappings and reusing the existing smite action/motion page.

FireShot Capture 468 - Colony - localhost

FireShot Capture 469 - Colony - localhost

Resolves #2701


src/i18n/en-events.ts
```diff
@@ -73,7 +79,7 @@ const eventsMessageDescriptors = {
       ${ColonyAndExtensionsEvents.MotionEscalated} {{agent} escalated motion {motionId} from {domain} to {newDomain}}
       ${ColonyAndExtensionsEvents.MotionRewardClaimed} {{agent} claimed their stake in motion {motionId}}
       ${ColonyAndExtensionsEvents.MotionEventSet} {Motion {motionId} fast-forwarded to the next lifecycle}
-      ${ColonyAndExtensionsEvents.ArbitraryReputationUpdate} {{agent} smote {recipient} with a {reputationPenalty} reputation penalty}
+      ${ColonyAndExtensionsEvents.ArbitraryReputationUpdate} {{agent} smote {recipient} with a {reputationChange} reputation penalty}
```
chinins, line 82: I think it should be here smote or awarded as well depending on the case. Below is actually an award action that should up as smite:
Screenshot 2021-11-10 at 09 42 50

Also it seems to me that the events lit copy is missing.

Screenshot 2021-11-10 at 09 44 55

chinins (changes requested):
It's mostly all good. You just need to fix the actions list & events list copies.

src/i18n/en-events.ts
ArmandoGraterol: About the first thing you mention, I think the responsibility of changing that might fall on to #2702 as that is the issue that should wire up the award action/motion to the action/event list, right?

chinins: fair enough :)

chinins (approved):
All good! Sorry, it took me a while - there are always os many coin machine PRs to review

PR #2865, opened 2021-11-11
UI - add hover state with actions to a comment

### LINKED ISSUE #2850: UI - add hover state with actions to a comment
This should be added to all comments (actions/motion pages & coin machine chats). 

Figma spec:

https://www.figma.com/file/fw2pDIUlkhCkJE79ZyTbjp/CoinMachine?node-id=3003%3A24280

The story behind:

https://github.com/JoinColony/colonyDapp/issues/2711

arrenv:
## Description

Adding hover state and actions to each comment.
For the purpose of moderation of actions/motions & coin machine chats.

NOTES:
1. The action kebab icon will appear small, however, this is due to the incorrect icon within the Dapp.
I have corrected this locally and will raise another issue and PR for this.
2. There is an aside element blocking part of the hover, this should be addressed by issue #2869

**New stuff** ✨

* Added hover state to comments.
* Added kebab menu for actions
* Added permission based actions menu

## TODO

- [x] Add hover state to comments
- [x] Add actions kebab menu
- [x] Should not appear if user is not logged in
- [x] Add popover menu
- [x] Popover menu actions should be dependent on permissions

Resolves #2850


chinins (commented):
Very nice and complete work!

As discussed on discord, please confirm with Karol a couple of functionality points:

- how should `CommentActionsPopover` be closed? When there is no hover state (currently) or on click on the kebab menu or outside the popover?
- can user who created a comment delete it themselves (current) or only admins can do that?

I've also noticed that `CommentActionsPopover` is 6-7 px smaller (height) than the specs. Please, increase the height as per the spec.

arrenv:
> I've also noticed that `CommentActionsPopover` is 6-7 px smaller (height) than the specs. Please, increase the height as per the spec.

I will clarify this one with Karol as well, as the height is coming from the DropdownMenuItem component. Which is being used elsewhere.

arrenv:
> * how should `CommentActionsPopover` be closed? When there is no hover state (currently) or on click on the kebab menu or outside the popover?

Confirming that this should be toggle based and not hover based as it is currently. I will update.

> * can user who created a comment delete it themselves (current) or only admins can do that?

This is okay to have according to Karol, but needs to be confirmed if it is okay with Raul.

> I've also noticed that `CommentActionsPopover` is 6-7 px smaller (height) than the specs. Please, increase the height as per the spec.

Karol confirmed that this is okay to have as is, using the existing DropdownMenu component paddings.

PR #2870, opened 2021-11-11
Fix aside element covering up other elements

### LINKED ISSUE #2869: Investigate and potentially remove <aside> element from extensions & buy tokens pages
It seems that this element is messing up with tooltips in the extension details & with hover state on messages in buy tokens chat.

Screenshot 2021-11-11 at 15 16 20

ArmandoGraterol:
Removed empty aside element which appeared when the page displayed by `ColonyHomeLayout` didn't have a sidebar like the action list or event list did. With this change, the extension (extension list, details, setup, and enabling) and coin machine page should be affected.

From what I've tested, the removal of the `aside` doesn't negatively impact the UI of the pages that were using it and gets rid of the bug, but that could just be from my side of things.

We will find out if it's working as expected once you guys test it on your env :)

Resolves #2869

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyHomeLayout.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyHomeLayout.tsx
@@ -93,7 +93,7 @@ const ColonyHomeLayout = ({
           )}
           {children}
         </div>
-        {showSidebar ? (
+        {showSidebar && (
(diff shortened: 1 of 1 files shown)


chinins (approved):
It works well for me! Good fix!

alicjakujawa (approved):

PR #2874, opened 2021-11-11
Add banned tag UI

### LINKED ISSUE #2853: UI - add `Banned` tag
First you need to add a new them to the Tag component & then add the `Banned` version that includes the emoji.

Figma spec:

https://www.figma.com/file/fw2pDIUlkhCkJE79ZyTbjp/CoinMachine?node-id=3195%3A36646

The story behind:

#2711

agustif:
## Description
Create a new BannedTag component that leverages the new bannedTheme in Tag and uses the new emoji-goblin svg.
It also handles special letter-spacing and different paddings/margins

In order to review this, please add this code wherever you want on the UI for the Tag to show up.
It defaults to text = `Banned`
```tsx
 <BannedTag />
```
**New stuff** ✨
* added new css variable themeBanned that:
* uses pink color for text, and mod pink, 20% transparent (The 20% opacity on the background, makes this tag blend with the bg color under it, so it will look as needed on white components, but not otherwise?

**Screenshot** 📸 
image

# TODOs
 
- [x] Create new BannedTag component in core
- [x] adds a new theme 'banned' to the Tag component.
- [x] adds new emoji as svg, for the emoji-goblin
- [x] has correct margins / paddings  and font styles

Resolves #2853


src/modules/core/components/Tag/Tag.css
```diff
@@ -46,6 +46,12 @@
   composes: baseStyles;
 }
 
+.themeBanned {
+  composes: baseStyles;
+  background-color: var(--banned);
+  color: var(--danger);
```
chinins, line 52: It should be --pink (by the look from the specs)

src/styles/variables.css
```diff
@@ -8,6 +8,7 @@
   --dark-blue: rgb(69, 85, 230);
   --sky-blue: rgb(90, 145, 255);
   --danger: rgb(248, 43, 101); /* user mentions color */
+  --banned: rgb(248, 43, 101, 0.2); /* banned tag bg color */
```
chinins, line 11: you don't beed an extra css variable just for that, please remove.

You should use `color-mod` directly in the component css file, eg:
`fill: color-mod(var(--dark) alpha(70%));`

And it's --pink with 0.2, not danger

src/modules/core/components/Tag/Tag.css
```diff
@@ -46,6 +46,12 @@
   composes: baseStyles;
 }
 
+.themeBanned {
+  composes: baseStyles;
+  background-color: var(--banned);
```
chinins, line 51: as mentioned in my comment for the variables file, it should use color-mod and --var-pink directly here

chinins (changes requested):
Unfortunately, it's not done properly. Please, see my comments & correct.

In addition if you create a banned tag as you suggest (with `👺 Banned`) then the margins and padding are not the same as specs. Please create a small component that is `BannedTag` that addresses this inside of it
spces:
Screenshot 2021-11-12 at 09 54 23

your implementation:
Screenshot 2021-11-12 at 09 54 12

src/styles/variables.css
agustif: Yup you're right olga thank you!

src/modules/core/components/Tag/Tag.css
agustif: Fixed on next push!

src/modules/core/components/Tag/Tag.css
agustif: Done!

agustif:
Archiving in lieu of https://github.com/JoinColony/colonyDapp/pull/2882

PR #2885, opened 2021-11-15
Added Userflow to index.html

thycommodore:
I'm not 100% if this should go here, but my goal is to add this snippet just before the </body> tag.

## Description

- [] Add Userflow to index.html

**New stuff** ✨

* Integrated Userflow which gives us a no-code tool to improve onboarding conversion & reduce churn.

### DIFF
--- a/src/templates/index.html
+++ b/src/templates/index.html
@@ -8,5 +8,11 @@
 </head>
 <body>
     <div id="root"></div>
+    <script>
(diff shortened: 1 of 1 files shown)


arrenv:
@thycommodore It looks to be in the right spot. Although, there may be preferences as to the installation method, for example using the npm package (their recommended) - https://userflow.com/docs/dev/userflow-js-installation

Or, using Google Tag Manager, so tracking scripts can be added and removed without a re-build and re-deployment.

rdig:
@arrenv @thycommodore there's more to this. This came out of the blue. Was this discussed somewhere and we missed that ?

I am personally against just plopping in external services in without a good reason.

thycommodore:
@rdig this did come out of the blue. Jack wanted to make sure we have a best shot at high conversion of onboarding for our public beta users & a reduction of churn for those colonies. He found Userflow & Stonly, asked me to dive deep into them. I think Userflow is a pretty powerful tool that offers a no-code (aside from adding this snippet) solution to give Product & Marketing the ability to rapidly add onboarding flows or highlight key new features. 

I've built out about 6 flows & looking to add another 4~6 more. I've been testing via Userflow's preview feature & I think the functionality is pretty non-disruptive & helpful. I suspect that we can scale our onboarding velocity (ie not doing manual onboardings) and decrease churn with adding this.

The cost from an engineering perspective (from my naive perspective) is that it's minimal to add the snippet to the site.

The cost from another external service provider perspective is limited due to this sitting as a visual overlay ontop of our existing html/css. There's no server uptime to worry about, no material integration to tackle. The downsides seemed driven by a principle (as you stated) & perhaps the financial cost. 

If my assumptions are true, is your main objection (without contemplating the reasons I've outlined) a principled one? If so, does the goal pass your bar for a good reason?

arrenv:
Sure thing, I will look after this one. I will be able to get to it in a couple of days once I arrive.

thycommodore:
@arrenv just saw that you're off this week, I think we're targeting for this to be live w/ the public beta release on Monday. Is there any possibility for this to be done by then or should we move it to a different engineer? Alternatively, I can sync w/ Jack on if this is a requirement for the public beta opening.

arrenv:
> @arrenv just saw that you're off this week, I think we're targeting for this to be live w/ the public beta release on Monday. Is there any possibility for this to be done by then or should we move it to a different engineer? Alternatively, I can sync w/ Jack on if this is a requirement for the public beta opening.

@thycommodore, I missed your message while I have been travelling, I have it on my list to do first thing in the morning. However, I believe the build is already done for the go live.

rdig:
Don't worry, if it's not done by when we deploy tomorrow, we can redeploy Tuesday to include this

PR #2891, opened 2021-11-16
Replace kebab menu icon in dapp

### LINKED ISSUE #2875: Replace kebab menu icon in Dapp
## Steps to reproduce

* The Colony kebab menu icon and comments kebab action icon need to be updated.

## Expected behaviour

* The following image shows the correct icon in each case (hover & active state in left image).
(It only appears in two instances)

kebab-menu-new

## Actual behaviour

* The following image shows what is currently in place. Notice the size and the usage of square dots instead of circles.

kebab-menu

arrenv:
## Description

Updates Kebab menu icon to match design.
Also adds active state to colony menu icon.

NOTE: The kebab icon does look slightly skewed down on the spec, however, the centered version of this commit has been confirmed okay by Karol.

**Changes** 🏗

* Icon SVG image changed.
* Added logic to handle active state.

menu_icon-dev

Resolves #2875

### DIFF
--- a/src/img/icons/three-dots-row.svg
+++ b/src/img/icons/three-dots-row.svg
@@ -1,6 +1,5 @@
 <svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
 	 viewBox="0 0 30 30" style="enable-background:new 0 0 30 30;" xml:space="preserve">
-<circle cx="15" cy="3" r="3"/>
-<circle cx="15" cy="15" r="3"/>
-<circle cx="15" cy="27" r="3"/>
+<path d="M15,0c1.7,0,3,1.3,3,3s-1.3,3-3,3s-3-1.3-3-3S13.3,0,15,0z M15,12c1.7,0,3,1.3,3,3s-1.3,3-3,3s-3-1.3-3-3S13.3,12,15,12z
+	 M15,24c1.7,0,3,1.3,3,3s-1.3,3-3,3s-3-1.3-3-3S13.3,24,15,24z"/>
 </svg>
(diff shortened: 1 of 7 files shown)


src/img/icons/three-dots-row.svg
```diff
@@ -1,11 +1,6 @@
...
-	C8.903,6.997,9,7.095,9,7.215V8.78z"/>
-<path d="M9,12.785c0,0.12-0.097,0.217-0.217,0.217H7.217C7.097,13.003,7,12.906,7,12.785V11.22c0-0.12,0.097-0.217,0.217-0.217
-	h1.565C8.903,11.003,9,11.1,9,11.22V12.785z"/>
-</svg>
+<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 30 30" style="enable-background:new 0 0 30 30;" xml:space="preserve">
+<circle cx="15" cy="3" r="3"/>
+<circle cx="15" cy="15" r="3"/>
+<circle cx="15" cy="27" r="3"/>
+</svg>
```
rdig, line 6: No way. Please convert these to `path`s before merging. Not all browsers know how to render native elements, so the safest route is to convert

arrenv: Thanks @rdig I had not seen this issue before. After looking into it briefly, it is a worth avoiding. Thank you pointing it out. I will update.

arrenv: Updated to use paths now. It reduced file size too.

ArmandoGraterol (approved):
Nice! It's looking much better than what we had before.

My only nitpick would be that it could look a _bit_ better and closer to design if you added a `margin-left: 10px` to the icon

chinins (approved):
Nice work!

arrenv:
> Nice! It's looking much better than what we had before.
> 
> My only nitpick would be that it could look a _bit_ better and closer to design if you added a `margin-left: 10px` to the icon

Thank you, I did play around with that a little. To get it right, I will take another look.

PR #2895, opened 2021-11-16
Update maximum per period field description

### LINKED ISSUE #2883: Copy error in Coin Machine enable params
Copy is the same as `Target per Period`. 

Screenshot 2021-11-15 at 19 46 38

If I did not provide this copy previously, try this:

`The maximum number of tokens that can be sold per period. If this limit is reached, the batch will appear sold out and prospective purchasers will be advised to wait for the next period.`

ArmandoGraterol:
FireShot Capture 487 - Colony - localhost

Resolves #2883


chinins (approved):
Good to go!

PR #2898, opened 2021-11-16
Add total sold vs available widget

### LINKED ISSUE #2868: Add "Total sold / available" section to Coin Machine.
The lack of this feature is a considerable irritation during testing which I worry is going to cause complaints.

We've tried to make the change minimally disruptive to the rest of the UI. I hope it's okay!

Please note:

1. Number font size in both sold vs available boxes should have a max of 18pts and scale down to fit on a single line when the numbers get big.
2. Numbers should be to 5 sig figs. i.e. have no decimal places if the number is 5 figures or larger.
3. New titles on each card in the new three-card layout.
4.  Changed location of the "Price next batch" notification which is shown when the target is reached.

Running Out@2x

chinins:
## Description

This PR adds total sold vs. available widget and adds some updates to tokens display in the remaining widget.

**New stuff** ✨

- add new widget & corresponding queries
- add react-textfit package
- add updated numbers formatting

## TODO

- the numbers in sold vs. available widgets should go into 2 lines when they hit 10px - this is now done! :)

Screenshot 2021-11-23 at 14 12 18

Resolves #2868

### DIFF
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -262,6 +262,49 @@ export const coinMachineResolvers = ({
         return null;
       }
     },
+    async coinMachineTotalTokens(_, { colonyAddress }) {
+      try {
+        const coinMachineClient = await colonyManager.getClient(
+          ClientType.CoinMachineClient,
+          colonyAddress,
+        );
+        const tokenClient = (await colonyManager.getClient(
+          ClientType.TokenClient,
+          colonyAddress,
+        )) as TokenClient;
+
+        const currentTokenBalance = await coinMachineClient.getTokenBalance();
+
+        /*
+         * We use the `FromChain` suffix to make these events more easily
+         * recognizable when reading the code
+         */
+        const transferLogsFromChain = await getLogs(
+          tokenClient,
+          tokenClient.filters.Transfer(null, coinMachineClient.address, null),
+        );
+
+        const transferEventsFromChain = await Promise.all(
+          transferLogsFromChain.map(async (log) =>
+            tokenClient.interface.parseLog(log),
+          ),
+        );
+        const totalAvailableTokens = transferEventsFromChain.reduce(
+          (acc, event) => event.values[2].add(acc),
+          '0',
+        );
+
+        return {
+          totalAvailableTokens: totalAvailableTokens.toString(),
+          totalSoldTokens: totalAvailableTokens
+            .sub(currentTokenBalance)
+            .toString(),
+        };
+      } catch (error) {
+        console.error(error);
+        return null;
+      }
+    },
     /*
      * WARNING!
      *
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -275,6 +275,23 @@ export const coinMachineResolvers = ({
 
         const currentTokenBalance = await coinMachineClient.getTokenBalance();
 
+        const subgraphData = await apolloClient.query<
+          SubgraphCoinMachinePeriodsQuery,
+          SubgraphCoinMachinePeriodsQueryVariables
+        >({
+          query: SubgraphCoinMachinePeriodsDocument,
+          variables: {
+            colonyAddress: colonyAddress.toLowerCase(),
+            extensionAddress: coinMachineClient.address.toLowerCase(),
+            limit: 100,
+          },
+          fetchPolicy: 'network-only',
+        });
+
+        const [extensionInitialised] = (
+          subgraphData?.data?.extensionInitialisedEvents || []
+        ).map(parseSubgraphEvent);
+
         /*
          * We use the `FromChain` suffix to make these events more easily
          * recognizable when reading the code
@@ -282,6 +299,9 @@ export const coinMachineResolvers = ({
         const transferLogsFromChain = await getLogs(
           tokenClient,
           tokenClient.filters.Transfer(null, coinMachineClient.address, null),
+          {
+            fromBlock: extensionInitialised?.blockNumber,
+          },
         );
 
         const transferEventsFromChain = await Promise.all(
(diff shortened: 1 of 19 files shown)


ArmandoGraterol (commented):
Nice! I like how it looks and the new info is quite useful when testing the coin machine. Surely it's also gonna be useful for normal users.

When I enter the coin machine I'm seeing some warnings on the console:
Screenshot from 2021-11-23 20-06-53

Can you check them out?

arrenv:
Looks like the perfectFit property was dropped - See reference: https://github.com/malte-wessel/react-textfit/blob/fc7aa18e61a2397fcef838dc8b16c8dc027dce26/CHANGELOG.md#v100

arrenv (commented):
Looks to be working well.

tokens-available

tokens-available-big-size

chinins:
@ArmandoGraterol, it's a leftover prop I forgot to remove - thank you for spotting it! It should be all good now - without any warning.

arrenv (approved):
Working well, without the error now for me!

ArmandoGraterol (approved):
Yep, from what I've tested it's working as expected and without warnings! Good job

chinins:
@ArmandoGraterol , @arrenv , can you please review my latest commit and test again? 

I have added the dynamic switching to 2 lines when the font-size reaches 10px.

arrenv (approved):
Nice one, the split line looks to be working well.

number-newline-1

number-newline-3

number-newline-4

number-newline-2

ArmandoGraterol (approved):
The multiline seems to be working as expected!

FireShot Capture 499 - Colony - localhost

src/data/resolvers/coinMachine.ts
```diff
@@ -262,6 +262,49 @@ export const coinMachineResolvers = ({
...
+        const currentTokenBalance = await coinMachineClient.getTokenBalance();
+
+        /*
+         * We use the `FromChain` suffix to make these events more easily
+         * recognizable when reading the code
+         */
+        const transferLogsFromChain = await getLogs(
+          tokenClient,
+          tokenClient.filters.Transfer(null, coinMachineClient.address, null),
+        );
```
rdig, line 285: This is very expensive, but I guess it's what we have to do in order to fetch the token transfers into the coin machine.

However, you can filter it even more, by only fetching blocks from the time the current instance of the coin machine was installed.

See this for an example:
https://github.com/JoinColony/colonyDapp/blob/9ba1bbf9606d3afd7ee897ce2ae5fdee1eb75f71/src/data/resolvers/coinMachine.ts#L329-L356

src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
```diff
@@ -201,7 +201,7 @@ const BuyTokens = ({
   const getFormattedCost = useCallback(
     (amount) => {
       const decimalCost = new Decimal(amount)
-        .times(salePriceData.coinMachineCurrentPeriodPrice)
+        .times(salePriceData?.coinMachineCurrentPeriodPrice || '0')
```
rdig, line 204: Shouldn't this actually be _OR `1`_, because if you multiply by `0`, you'll always get `0`, but if you multiply by `1` you'll just get the amount.

Depends on what you want the fallback to be.

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
```diff
@@ -311,6 +346,7 @@ const CoinMachine = ({
             }}
             sellableToken={sellableToken}
             purchaseToken={purchaseToken}
+            periodTokens={periodTokens}
```
rdig, line 349: I really that you put the period tokens back in here, as I worked really hard to refactor this so that it expressly doesn't use this specific value as it's not quite optimized.

However, I do understand why it's here, because you had to move the "Price next batch" display into the table.

I really don't like this, but this is a side-effect of just refactoring design willy nilly, you end up with content you need to move around, and you just slap it somewhere....

_(sorry for the rant :) )_

rdig (changes requested):
Congrats on putting this PR together. Very nice work! :100: 

There still is a issue in here. _(but which was inherited from the previous iteration)_

If the coin machine has less tokens that the batch available, it will still show the batch as having available the "max" allowed, even though the coin machine doesn't have them anymore.

I'll show you some pictures since it's easier to explain.

Screenshot from 2021-11-26 01-18-49

In here, there are only 10 tokens available in the coin machine _(which the "Total sold v. available correctly shows -- 70/80)_, the buy widget correctly only allows you to buy 10 _(notice the error)_, but the "Batch sold v. available" widget still shows 0 / 40 which is incorrect in this case as there are only 10 tokens left. It should show 0 / 10.

Notice, that if I pass the period, the previous batches table shows the period, correctly, as having just 0 / 10

Screenshot from 2021-11-26 01-19-03

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
chinins: I am not sure I understand why it's so bad as I just moved the `Price next batch` logic from one component to another and added one prop to `TokenSalesTable`...
I am happy to discuss this & refactor once I understand the ideal state you would like to see.

src/data/resolvers/coinMachine.ts
chinins: Fair enough - done now.

src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
chinins: I think the default value should be `0`. If it's `0` it will give an indication that something is off more that if it defaults to `1`.... But I am happy to hear a counter-argument

chinins (commented):
@rdig updated the PR. Not sure about your comment regarding the `periodTokens` prop passing, other than that it's ready for re-review.

rdig (approved):
Working as expected now. Very nice turnaround. :1st_place_medal: 

Screenshot from 2021-11-26 22-11-56

On Monday I'll put it up on QA

PR #2917, opened 2021-11-18
Improve Coin Machine period timer

### LINKED ISSUE #2912: Improve Coin Machine sale countdown timer
Basically try to re-fresh the base time periodically (and show a spinner while doing so)

rdig:
Basically this PR adds a 1 minute poll interval to the salePeriod query _(the one that fetches the period remaining time, and the period length)_, so that it will keep (mostly) in sync with the values on the chain. This will help us achieve a (somewhat) accurate timer. _(it cannot be trully 100% accurate since times in the client, and on the chain deviate)_

While I was at it, I also did some improvements to the query and to the `RemainingDisplayWidget` _(to actually display when it's syncing)_

Resolves #2912

### DIFF
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -1292,8 +1292,8 @@ export type SaleTokens = {
 };
 
 export type CurrentSalePeriod = {
-  periodLength: Scalars['String'];
-  timeRemaining: Scalars['String'];
+  periodLength: Scalars['Int'];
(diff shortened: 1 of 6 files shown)


chinins (approved):
Looks good! Nice solution!

ArmandoGraterol (approved):
LGTM :ok_hand: It looks like a nice solution to all of this timer accuracy ordeal

PR #2922, opened 2021-11-18
Fix long dates occupying 2 lines on Sale End column

### LINKED ISSUE #2913: Display longer dates in Previous Sale Period Table in the same line
The dates in the "Sale End" column, need to be displayed in the same line, even if both the day and month contain a 2 character value

Screenshot from 2021-11-18 15-52-04

ArmandoGraterol:
FireShot Capture 495 - Colony - localhost

Resolves #2913

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.css
+++ b/src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.css
@@ -34,11 +34,11 @@
 }
 
 .tableHeaderCell:first-child {
-  width: 16%;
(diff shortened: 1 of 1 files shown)


rdig (approved):
Maybe we should look into making the table columns actually responsive, rather than just percentage values, but it seems to works with large values as well, so I say it's good enough for now:

Screenshot from 2021-11-18 23-17-18

Large values will however break the "Tokens Remaining" widget, but that's not the scope of your issue, since most likely it will be fixed by @chinins when she implements #2898 

Screenshot from 2021-11-18 23-17-23

PR #2924, opened 2021-11-18
Add polling logic to CoinMachine

### LINKED ISSUE #2916: Change CoinMachine queries to use `pollInterval`
The current `refetch` logic we have implemented is good from an optimization perspective, but it won't work so nicely when people will be rushing to buy tokens. This is because the refresh timeouts are too long _(basically the remaining period length, which is a lot)_

Since we don't have access to `subscriptions` here, we're doing the next best thing, and using a `pollInterval` value of `1000` to refresh a couple of queries:
- `CurrentPeriodTokens`
- `CoinMachineCurrentPeriodPrice`
- `CoinMachineCurrentPeriodMaxUserPurchase`

To do this, remove all the `refetch`/`startPolling` logic, and just pass in the `pollInterval` to those queries.

Note however to leave in the `stopPolling` callback, and run in on component unmount via `useEffect`

ArmandoGraterol:
Replaced `refetch` logic in `CoinMachine` with `polling` logic.

Affected queries: `CurrentPeriodMaxUserPurchase`, `CurrentPeriodTokens` and `CurrentPeriodMaxUserPurchase`

Resolves #2916

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
+++ b/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
@@ -1,9 +1,10 @@
 import React, { useEffect, useMemo } from 'react';
 import { Redirect, useParams } from 'react-router-dom';
 import { defineMessages, FormattedMessage } from 'react-intl';
-import { Extension } from '@colony/colony-js';
+import { Extension, Network } from '@colony/colony-js';
 import { bigNumberify } from 'ethers/utils';
 
+import { DEFAULT_NETWORK } from '~constants';
 import ExternalLink from '~core/ExternalLink';
 import { SpinnerLoader } from '~core/Preloaders';
 import BreadCrumb, { Crumb } from '~core/BreadCrumb';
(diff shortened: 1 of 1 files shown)


alicjakujawa (approved):

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
```diff
@@ -115,13 +100,12 @@ const CoinMachine = ({
...
   const {
     data: periodTokensData,
-    refetch: refetchCurrentPeriodTokensData,
-    startPolling: startPollingCurrentPeriodTokensData,
     stopPolling: stopPollingCurrentPeriodTokensData,
     loading: periodTokensLoading,
   } = useCurrentPeriodTokensQuery({
     variables: { colonyAddress },
     fetchPolicy: 'network-only',
+    pollInterval: 1000,
```
rdig, line 108: One second is waaaay too agressive.

It should be based on blocktime, so about 5 seconds for xdai, and about 20-ish for mainnet

Note: this applies to all three

rdig (approved):
Good to go, but make sure to update the polling interval values

PR #2925, opened 2021-11-19
Refactor coinMachineBoughtTokens to use chain events

### LINKED ISSUE #2914: Refactor `coinMachineTransactionAmount` and `coinMachineBoughtTokens` resolvers to use chain events
This is needed to determine the sale status after you've bought tokens.

Currently it uses subgraph events and apparently they are not reliable, so we need to use chain events _(in theory this should be not that bad, if we fetch just the events we need)_

This needs to be super reliable as it's dealing with people buying tokens, and you don't fuck with that.

alicjakujawa:
## Description

This PR refactor coinMachineBoughtTokens resolver to use chain events instead of subgraph

**Changes** 🏗

* Refactored coinMachineBoughtTokens

Resolves #2914

### DIFF
--- a/src/data/resolvers/coinMachine.ts
+++ b/src/data/resolvers/coinMachine.ts
@@ -102,6 +102,8 @@ export const coinMachineResolvers = ({
     async coinMachineBoughtTokens(_, { colonyAddress, transactionHash }) {
       const { provider } = colonyManager.networkClient;
       try {
+        const { blockHash } = await provider.getTransaction(transactionHash);
+
         const coinMachineClient = await colonyManager.getClient(
           ClientType.CoinMachineClient,
           colonyAddress,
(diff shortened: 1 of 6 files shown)


src/data/resolvers/coinMachine.ts
```diff
@@ -99,41 +99,47 @@ export const coinMachineResolvers = ({
...
-            /*
-             * The limit doesn't matter in this case as we're not using that data,
-             * we're just interested in the TokensBought events
-             */
-            limit: 1,
-          },
-          fetchPolicy: 'network-only',
-        });
+        const boughtTokensFilter = coinMachineClient.filters.TokensBought(
+          null,
```
rdig, line 111: The `buyer` argument is sadly not `indexed`, otherwise we could have filtered by that too _(which would have been amazing)_

rdig (approved):
Seems to be working good now. Very nice work!

:100:

ArmandoGraterol (approved):
LGTM :ok_hand:

PR #2929, opened 2021-11-20
Improve token displays of SaleStateWidget and BuyTokens

### LINKED ISSUE #2910: Increase the cut-off to 5 significant figures in the `for` section of `Success` of buying tokens
Currently if the amount is small it will show only zeros, so we need to increase it to 5 significant figures.
Screenshot_2021-11-17_at_19 20 29

ArmandoGraterol:
- Improved formatting of tokens in the SaleStateWidget and BuyTokens `Cost` field.
- Also, while working with the above changes, I noticed that the "max buy amount" validation could be improved. As it was before, the validation could've been problematic when trying to buy a really small amount (0.00000...) or sums with really detailed decimals (12.18429...). Please make sure to test it.
- Improved `getFormattedTokenValue` code, should be a bit easier to read now.

Resolves #2910


rdig (approved):
I like this. Very high quality work! :100:

PR #2946, opened 2021-11-23
Fix countdown timer on motion finished

### LINKED ISSUE #2944: Fix 'Timeout' display in `CountDownTimer`
Currently, when the timer runs out, the `CountDownTimer` widget will display the copy "Timeout"

That copy is only intended for error states, not normal functionality.

So in this case, where the timer finishshes (eg: counts down to zero) it should not show anything. (no copy at all)

ArmandoGraterol:
Fixed the countdown timer appearing as `Timeout` when the motion failed/passed/failed but not finalizable

Resolves #2944


arrenv:
Hi @ArmandoGraterol, I have only explored motions a little bit. What is the best way to test this?

chinins (approved):
Very good - working as expected.

arrenv (approved):
Reviewed with @chinins, working as required.

PR #2948, opened 2021-11-23
Time remaining fix

### LINKED ISSUE #2930: 'Time remaining' box is resetting to incorrect duration
## Steps to reproduce

1. Enable coin machine
2. Set duration to .025
3. Buy tokens
4. Hit buy again
5. Observe

## Expected behaviour

* Time remaining is accurate

## Actual behaviour

* Example: A `WALLET_CREATE_ERROR` action is dispatched and the user cannot log in

Colony: https://qaxdai.colony.io/colony/krausehouse112021d/buy-tokens/ 
Video: https://share.descript.com/view/aVgNNposgBe

alicjakujawa:
## Description
This PR fixes wrongly displayed time remaning. When timeLeft was 0 we was setting timeLeft to periodLength (instead of periodLength/1000) 

**Changes** 🏗

* Updated useSplitTime hook
* Updated TokenSalesTable (was throwing error because not needed refetch)

Resolves #2930


chinins:
It's working correctly (at least what I tested) but I don't understand why we don't need refetch anymore. Could you, please, explain?

src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.tsx
```diff
@@ -169,7 +168,6 @@ const TokenSalesTable = ({
   useEffect(() => {
     if (periodRemainingMS > 1000 && periodRemainingMS < periodLengthMS) {
       setTimeout(() => {
-        refetchSalePeriodsData(salePeriodQueryVariables);
         startPollingSalePeriodsData(periodLengthMS);
       }, periodRemainingMS);
     } else {
```
ArmandoGraterol, line 173: If we don't need the refetch anymore, I think this code can be further simplified like this:

```
  useEffect(() => {
    startPollingSalePeriodsData(periodLengthMS);
    return () => stopPollingSalePeriodsData();
  }, [
    periodLengthMS,
    startPollingSalePeriodsData,
    stopPollingSalePeriodsData,
  ]);
```

As far as I understand this code, the `else` would still be doing the same thing as the first conditional

ArmandoGraterol (changes requested):
Working as expected for me! It's fetching the correct time :ok_hand: There's just something that I want to address before giving it the approve

src/utils/hooks/useSplitTime.ts
```diff
@@ -19,7 +19,7 @@ const useSplitTime = (
       }
 
       if (timeLeft === 0 && periodLength) {
-        setTimeLeft(periodLength);
+        setTimeLeft(periodLength / 1000);
```
rdig, line 22: Why do we divide by 1000 here, the period length is already in ms ?

src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.tsx
```diff
@@ -169,7 +168,6 @@ const TokenSalesTable = ({
   useEffect(() => {
     if (periodRemainingMS > 1000 && periodRemainingMS < periodLengthMS) {
       setTimeout(() => {
-        refetchSalePeriodsData(salePeriodQueryVariables);
```
rdig, line 172: But this is the tokens sales table, we need to refetch the periods data each time a period passes. 

What's reason to remove this ?

src/utils/hooks/useSplitTime.ts
alicjakujawa: is it? So something else need to be wrong with conversion because with just periodLength it set time wrongly 🤔

src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.tsx
alicjakujawa: If we poll data every "period length" then why we also want to refetch? 🤔

rdig: Let's say the period is 1 hour.

You load the component when the period has only 30 minutes until it ends.

Your poll interval will run every hour, but it will not run in 30 minutes, when the period ends, and you want to refresh the data.

That's why we also use `refetch`, but be able to manually get new data, in-between periods, if we loaded the component in the time between periods

src/utils/hooks/useSplitTime.ts
rdig: Now I'm confused as well :)

Please check to be sure, I might be wrong :)

src/utils/hooks/useSplitTime.ts
```diff
@@ -7,7 +7,7 @@ const useSplitTime = (
   activeTimer: boolean,
   periodLength?: number,
 ) => {
-  const [timeLeft, setTimeLeft] = useState<number>(initialTime / 1000);
```
ArmandoGraterol, line 10: With this change, the timer in `SaleStateWidget` is showing the wrong time:
FireShot Capture 500 - Colony - localhost

Correct time:

FireShot Capture 501 - Colony - localhost

So I think we may be receiving different time units in another place, weird.

ArmandoGraterol (changes requested):
The remaining time widget is working as expected as far as I can tell, but now the `SalesStateWidget` isn't (It uses `useSplitTime`)

ArmandoGraterol (approved):
It seems like both timers are displaying the correct time after refetching! Nice job

src/modules/dashboard/components/CoinMachine/TokenSalesTable/TokenSalesTable.tsx
```diff
@@ -183,8 +183,8 @@ const TokenSalesTable = ({
...
         startPollingSalePeriodsData(periodLengthMS);
+        refetchSalePeriodsData(salePeriodQueryVariables);
```
rdig, line 187: Don't we also need `salePeriodQueryVariables` as dependencies to the `useEffect` ?

rdig (approved):
Other than the missing dependency, this is good to go

PR #2953, opened 2021-11-23
Font weight - colony wallet address

### LINKED ISSUE #2892: Font weight - colony wallet address
## Steps to reproduce

1. Log in with a MetaMask wallet and go to home page

## Expected behaviour

Make sure to use the right font weight for colony wallet address and the correct distance between the wallet and the name of the colony

https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=563%3A1013

## Actual behaviour

colony-menu

arrenv:
## Description

Improves the alignment of the Colony name and address in the colony menu component.
Also fixes the colony menu hover state.

NOTE: There will still be slight visual differences in the fonts, due to different systems rendering the fonts differently.

fixed-alignment-colony-name

colony-menu-hover-state

**Changes** 🏗

* Positioning of the colony menu.
* Colony menu address hover state color.

Resolves #2892


ArmandoGraterol (approved):
LGTM :ok_hand:

chinins (approved):
I think that colony name should have font-weight of 600 (that's what I see in specs) and currently it's 400. Can you please fix this and then it's good to go.
This is the specs:
Screenshot 2021-11-24 at 09 18 51

arrenv:
> I think that colony name should have font-weight of 600 (that's what I see in specs) and currently it's 400. Can you please fix this and then it's good to go.

We currently don't have a 600 font size in the app, only 400 and 700. Introducing the 600 would add about 18kb to 40kb to the app initial load. The following screenshot shows the comparison on my screen, I will check with Karol.

menu-alignment-compare

rdig:
You are correct. That's why the dapp is not design perfect.

We won't be adding the 600 weight font. Just choose the one that looks closest to the designs

arrenv:
This was discussed with product and dev and it was decided that no new custom font weight would be introduced.

The bold font weight has been used based on various chats, to match the design and also better balance the title with the bold address below it.

menu-alignment-final

PR #2957, opened 2021-11-24
Fixes coin machine footer padding issue

### LINKED ISSUE #2902: Adjust padding - Coin Machine
## Steps to reproduce

1. Log in with a MetaMask wallet
2. Open coin machine
3. Scroll to the bottom

## Expected behaviour

There should be a proportional padding in-between the grid and the bottom area. 
Suggest to keep it as 20px. 
Sizing reference:
https://www.figma.com/file/fw2pDIUlkhCkJE79ZyTbjp/CoinMachine?node-id=3102%3A29565

image

Note - please check Suggestions section below for details

## Actual behaviour

Notice, there is no bottom padding at all. Should be at least 20px 

qacm

## Suggestions

Make sure to check distances between the chat and middle grids and sizes of the following 
UI elements: chat width, middle boxes height, the previous sales width.

agustif:
## Description

- [x] Fixes padding in the bottom of the coin machine, So 'We love feedback' doesn't get on top of it.

**New stuff** ✨

* Adds bottom padding value to mainContent class in ColonyHomeLayout.css

Screenshot 2021-11-24 at 01 50 23

Resolves #2902

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/CoinMachine.css
+++ b/src/modules/dashboard/components/CoinMachine/CoinMachine.css
@@ -1,5 +1,14 @@
 .main {
   display: block;
+  margin: 0 0 20px 20px;
+  height: 100%;
+  width: calc(100% + 360px);
+}
+
+@media (max-height: 1024px) {
+  .main {
+    margin-bottom: 20%;
+  }
 }
 
 .loadingSpinner {
--- a/src/modules/dashboard/components/CoinMachine/CoinMachine.css
+++ b/src/modules/dashboard/components/CoinMachine/CoinMachine.css
@@ -5,7 +5,7 @@
   width: calc(100% + 360px);
 }
 
-@media (max-height: 1024px) {
+@media (max-height: 960px) {
   .main {
     margin-bottom: 20%;
   }


agustif:
Moved to not touching colonyLayout with the padding hack, instead removed height from grid class in CoinMachine css, and moved margin-top to margin adding also a margin-bottom of 20px to it

Screenshots at different zoom levels:
Screenshot 2021-11-24 at 15 44 09
Screenshot 2021-11-24 at 15 43 40

Screenshot 2021-11-24 at 15 44 21

chinins (changes requested):
There is problem with the implementation. Please, see my gif.

The coin machine should be the height of the screen minus the margins and content around. It should not overflow for more than the screen height. I think you'll have to adapt Raul's calculation of the height, not just remove it.

I think it would also be good to have margin-right of 20-30px for the whole coin-machine since we are on the subject of margins here.
coin-machine-scroll

agustif:
> There is problem with the implementation. Please, see my gif.
> 
> The coin machine should be the height of the screen minus the margins and content around. It should not overflow for more than the screen height. I think you'll have to adapt Raul's calculation of the height, not just remove it.
> 
> I think it would also be good to have margin-right of 20-30px for the whole coin-machine since we are on the subject of margins here. 

Moved implementation to coin machine css, not touching height now, only adding proper marigins.

Added margin right of 30px

Can you re-review @chinins thank you!

chinins (changes requested):
There is no margin bottom on the coin machine which is one of the issues to solve here.
Screenshot 2021-11-26 at 17 19 21

agustif:
image
image

@chinins thanks! Fixed now, 
Rebasing!

src/modules/dashboard/components/CoinMachine/CoinMachine.css
```diff
@@ -1,5 +1,6 @@
 .main {
   display: block;
+  margin-bottom: 100px;
```
chinins, line 3: Why 100px? I think it's very wrong...

src/modules/dashboard/components/CoinMachine/CoinMachine.css
```diff
@@ -19,7 +20,7 @@
 
 .grid {
   display: grid;
-  margin-top: 40px;
+  margin: 40px 0 20px;
```
chinins, line 23: I think this 20px margin is not applied, I think it's useless here.

src/modules/dashboard/components/CoinMachine/CoinMachine.css
```diff
@@ -60,12 +61,14 @@
 
 .sales {
   composes: container;
-  min-height: 100%;
+  margin-bottom: 50px;
```
chinins, line 64: Why 50px? Specs shared by Karol says 30px. And why do we apply on one of the elements only and not on the larger container?

src/modules/dashboard/components/CoinMachine/CoinMachine.css
```diff
@@ -60,12 +61,14 @@
...
+  height: calc(100% - 50px);
   grid-column: buy / comments;
   grid-row: sales / last-line;
 }
 
 .comments {
+  margin: 0 30px 50px 0;
```
chinins, line 71: Same here - why do we apply on one of the elements only and not on the larger container?

chinins (changes requested):
Both the margin (of 50 px instead of 30px) and the implementation are incorrect...

rdig:
Please fix the issues pointed out by @chinins after which I'll review as well :+1:

agustif:
> Please fix the issues pointed out by @chinins after which I'll review as well 👍

I moved implementation to main class css now.

I had to make it's height and width bigger due to it getting smaller than its coin machine grid contents. Let me know if this is not appropiate on how I should best handle this (the weird margin-top in main before was due to the same issue)

I also added a little margin on the left, so in smaller resoutions doesnt clash with the side menu Buy Tokens or Extensions.

src/modules/dashboard/components/CoinMachine/CoinMachine.css
agustif: Removed it

src/modules/dashboard/components/CoinMachine/CoinMachine.css
agustif: Removed it, applying styles only now to main container

src/modules/dashboard/components/CoinMachine/CoinMachine.css
agustif: Done!

src/modules/dashboard/components/CoinMachine/CoinMachine.css
agustif: I've moved to using width and height with calc, otherwise main class is smaller than it's child grid, and margin's don't get applied on lower resolutions?

Hope it's good now, let me know. Thank you

chinins (approved):
It seems like a strange way to achieve the goal but it's working properly. I would like for @rdig  to have a look as well and check on a larger screen that it stretches all the way (it should but it still needs to be tested).

rdig (changes requested):
This is still not good. I shouldn't have a scrollbar to the side on any height. (but I do currently, notice the screenshots)

Both the "Previous Batches" and "Chat" should grow and shrink with the viewport, but not show  a srollbar 

Screenshot from 2021-12-01 20-03-16
Screenshot from 2021-12-01 20-03-21

rdig:
Also, why does this PR have 7 commits!!!! for such a small change ?

agustif:
> Also, why does this PR have 7 commits!!!! for such a small change ?

I tried two different approaches and didn't clean up the history, will do and squash 'em into one commit after fixing new feedback!

agustif:
Fixed scroll issue.
By adding a media query and only adding a marging of 20% (could also be a height: 120% but that didn't seem better) so in case of smaller than 1024px height viewports the padding still shows.

Also contained into 1 commit

Can you guys please re-review thx

rdig (approved):
I'm good now with all the changes, however I would argue that the 20% margin when the media query triggers is bit to agressive.

See the screenshots I posted below, as it adds a scroll bar a bit too early (look at when the window is half the size, that's when you see it)

On the other hand, you've spent too much time on this already, so this will do for now.

Screenshot from 2021-12-02 10-11-05
Screenshot from 2021-12-02 10-11-44
Screenshot from 2021-12-02 10-11-49

chinins (approved):
Agree with Raul that 20% padding is not ideal but good enough for now.

rdig (approved):
Works better after the media query change

PR #2958, opened 2021-11-24
Hide `userLimitFraction` during coin machine installation when there is no whitelist

### LINKED ISSUE #2934: Maximum per person purchase limit is not enforced.
I enabled Coin Machine with a 10% of the total supply. However, I was able to buy 100%.

When I bought them I was given the message that I tried to purchase 100 but someone beat me to it.

The params I installed:
Screenshot 2021-11-20 at 16 22 07

I then funded Coin Machine with 100MEEP.

My purchase of 100 MEEP.
Screenshot 2021-11-20 at 16 24 25

chinins:
## Description

This PR hides per person limit field when there is no whitelist. This limit is not enforced when there is no whitelist (you can see the explanation in the thread of the corresponding issue) so Jack said to hide it (like we do with the whitelist address field).

To test - try enabling coin machine with whitelist & without.

resolves #2934

### DIFF
--- a/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
+++ b/src/modules/dashboard/components/Extensions/ExtensionSetup.tsx
@@ -301,7 +301,8 @@ const ExtensionSetup = ({
   const showInputField = useCallback(
     (paramName) => {
       return (
-        paramName !== 'whitelistAddress' ||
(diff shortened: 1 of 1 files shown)


ArmandoGraterol (approved):
LGTM :100: Working as expected

PR #2961, opened 2021-11-25
Updated button size, copy and removed microcopy

### LINKED ISSUE #2932: Unban button copy is not centred in the button.
Hardly top priority, but catching it anyway.

Change copy to `Remove ban` too.

Also remove the microcopy as it doesn't make any sense.

Screenshot 2021-11-20 at 16 00 16

arrenv:
## Description

- Fixes center alignment of 'Unban the user' button.
- Updates the button copy to 'Remove ban'
- Removes the irrelevant microcopy note text under the referenced user.

unban-user

Resolves #2932

### DIFF
--- a/src/modules/core/components/Comment/Dialogs/BanComment/BanComment.tsx
+++ b/src/modules/core/components/Comment/Dialogs/BanComment/BanComment.tsx
@@ -55,7 +55,7 @@ const MSG = defineMessages({
   confirmButtonText: {
     id: 'core.Comment.BanComment.confirmButtonText',
     defaultMessage: `{unban, select,
-      true {Unban the user}
+      true {Remove ban}
       other {Ban the troll}
     }`,
   },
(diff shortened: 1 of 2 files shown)


src/modules/core/components/Comment/Dialogs/BanComment/BanComment.tsx
```diff
@@ -131,9 +131,11 @@ const BanComment = ({
                 />
               </div>
             </div>
-            <p className={styles.note}>
-              <FormattedMessage {...MSG.note} values={{ unban }} />
-            </p>
+            {!unban && (
+              <p className={styles.note}>
+                <FormattedMessage {...MSG.note} values={{ unban }} />
```
ArmandoGraterol, line 136: Now that we don't have a note for the `unban` version of this dialog, the values prop shouldn't be needed and therefore the select and `unban` copy inside `MSG.note` is also not needed

ArmandoGraterol (approved):
LGTM :100: Just consider removing the now unused unban note copy as per my comment below.

chinins (approved):
Good to go, nice fix!

src/modules/core/components/Comment/Dialogs/BanComment/BanComment.tsx
arrenv: Good pickup, I have cleaned it up now.

PR #2964, opened 2021-11-25
Update font weight of BuyTokens info tooltip

### LINKED ISSUE #2950: Tooltip hover text has different stylings in Coin Machine
## Steps to reproduce

1. Login, Enable, Create, and got to your CoinMachine page.

image
image

## Expected behaviour

Texts to have same styling

## Actual behaviour

*  Each text has a different size/weight?

## Suggestions 

* All popups shuold probably have same styling? Not sure if this is done on purpose because one of the texts is larger than the other.

ArmandoGraterol:
BuyTokens's info tooltip styling should now be consistent with the rest of the tooltips in the coin machine

Resolves #2950


chinins (approved):
Good to go!

alicjakujawa (approved):

PR #2969, opened 2021-11-29
"I agree" button disabled, changing zoom makes button enabled

### LINKED ISSUE #2965: "I agree" button disabled, changing zoom makes button enabled
I can't replicate due to QA not having Whitelist extension enabled. However, there seems to be a pattern to resolve this (and it's a frequent enough issue) that I'm hoping a dev can replicate locally. 

This is a fairly common issue folks are running into. It's not particularly urgent at this moment, but as we get closer to the opening of Coin Machine — it will become much more urgent + important. 

## Steps to reproduce

1. For a Coin Machine, with Whitelist enabled (KYC + Agreement option)
2. Add the user to the whitelist
3. Attempt to agree to the terms of the Coin Machine

## Expected behaviour

* The "I agree" button is enabled to click

## Actual behaviour

* The "I agree" button is disabled
* If the user zooms in/out this seems to get the button to flip to enabled

image

image

arrenv:
## Description

Fixes the agreement 'Got it' / 'I agree' button not working in some cases on scroll.

NOTE: The only way I was able to replicate the issue was to zoom in within the browser window.

### Before fix:

Colony-agreement-scroll

### After fix:

Colony-agreement-scroll-fix

Resolves #2965

### DIFF
--- a/src/modules/dashboard/components/Whitelist/AgreementDialog/AgreementDialog.tsx
+++ b/src/modules/dashboard/components/Whitelist/AgreementDialog/AgreementDialog.tsx
@@ -68,7 +68,7 @@ const AgreementDialog = ({
 
   const handleScroll = (e) => {
     const bottom =
-      e.target.scrollHeight - e.target.scrollTop <= e.target.clientHeight;
+      e.target.scrollHeight - 1 - e.target.scrollTop <= e.target.clientHeight;
 
     if (bottom) {
       setHasBeenScrolled(true);


alicjakujawa (approved):

ArmandoGraterol (approved):
Couldn't replicate this with my screen, but the button is also working as expected with your fix applied, so I guess it's good to go

rdig (approved):
Yup all good, tested In all scenarios I could think of:

Screenshot from 2021-11-29 19-31-58
Screenshot from 2021-11-29 19-32-02
Screenshot from 2021-11-29 19-32-06
Screenshot from 2021-11-29 19-32-11

Very nice work and turnaround! :100:

PR #2972, opened 2021-11-29
Improved Subgraph Events Sorting

rdig:
This PR introduces events sorting on the subgraph side, by adding a `timestamp` prop on all event entities on that are coming from the subgraph, so that we can pass query selection fields like `orderBy` and `orderDirection`

Besides the above, I've also refactored the `ColonyEvents` component _(it actually started this)_ to properly have pagination _(before this, we would just fetch the first 100 events, then split them up locally)_

_Note that we cannot do the same for the `ColonyActions` component due to the way we are combining 4 data streams into one, and not always symmetrical (this will become a problem once a colony surpasses `1000` combined actions)_

Since sorting is now done directly, prior to the data reaching our data, I also removed the `sortSubgraphEventByIndex` util, as it was no longer required

**Testing note**

Since this is using an update version of the `subgraph`, make sure to run `npm run provision` prior to testing this


arrenv (commented):
This is a nice improvement to the SubGraph data and handling sorting.

I did notice that the 'Load more' button still appears on the last item when there are no more results.

event-sorting

arrenv (commented):
Sorting looks to be working much better.

(I'm assuming the 5th Dec date is unrelated and a result of the dev environment?)

### Previous

events-sorting-before

### New

events-sorting-after

rdig:
> (I'm assuming the 5th Dec date is unrelated and a result of the dev environment?)

Two things about dates in the local dev environment to keep in mind:
- account for them
- but don't trust them

They can and are manipulated either by us directly or by various services we have running in the background

chinins (approved):
Very good!

ArmandoGraterol (approved):
It's working as expected from what I could see! Well done

PR #2981, opened 2021-11-30
Members List Poor Man's Pagination

rdig:
This PR adds a _"pagination"_ implementation to the `Members` list for the colony. This is **not a proper** implementation of pagination as the metadata server where we fetch this data from does not have this capability. As I've mentioned in my [comment](https://github.com/JoinColony/colonyDapp/blob/34371d0229b41ea312522fc0ae4a59d7117b69f6/src/modules/dashboard/components/Members/Members.tsx#L256-L266), the only good way to implement this, would be to add this logic to the server, then just fetch the chunks of users data we need.

But due to time constraints we have to do the next best thing. Just split the users list into chunks, and then render just render specific chunks. While the data fetching cost still remains high, it will alleviate some memory problems that React will incur trying to render a huge list.

While I was at it, I also fixed a problem when the user wasn't being redirected properly back to the colony, when first registering their account.

### 1. Members list 

Just replace the `skelethonUsers` array with one that duplicates your existent data a stupid amount of times_(ignore React duplicate keys error)_

```js
 const skelethonUsers = useMemo(() => {
    let displayMembers =
      allMembers?.subscribedUsers.map(
        ({ profile: { walletAddress } }) => walletAddress,
      ) || [];
    if (currentDomainId !== COLONY_TOTAL_BALANCE_DOMAIN_ID) {
      displayMembers = membersWithReputation?.colonyMembersWithReputation || [];
    }

    return [
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
      ...displayMembers,
    ].map((walletAddress) => ({
      id: walletAddress,
      profile: { walletAddress },
    }));
  }, [allMembers, membersWithReputation, currentDomainId]);
```

### 2. Redirect after account creation

- Go to an existing colony with a fresh address _(one that doesn't have a profile already registered)_
- Click the top-right avatar dropdown "Get Started" menu entry
- Create your profile
- At the end, after the transaction is confirmed, you should be redirected back to the original colony you started from

### DIFF
--- a/src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
+++ b/src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
@@ -31,9 +31,8 @@ const StepConfirmTransaction = () => {
   const { location } = useHistory<{ colonyURL?: string }>();
 
   if (username) {
-    const { colonyURL } = location.state;
-    if (colonyURL) {
-      return <Redirect to={colonyURL} />;
+    if (location?.state?.colonyURL) {
+      return <Redirect to={location.state.colonyURL} />;
     }
 
     return <Redirect to={LANDING_PAGE_ROUTE} />;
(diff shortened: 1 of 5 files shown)


chinins (approved):
Working as expected.

arrenv (approved):
Working as expected.

members-listing

PR #2982, opened 2021-11-30
Add unwrap and claim tokens routes

### LINKED ISSUE #2976: Create Vesting Unwrap and Claim Tokens Routes
Add two new routes that will display the _Unwrap Token_ and _Claim Token_ components 

Route names:
- `/colony/:colonyName/unwrap-tokens`
- `/colony/:colonyName/claim-tokens`

Requirements:
- user must be logged in and have a profile registered in order to access them
- will only be accessible to the metacolony for now _(this actually might not be doable, and we'll have to redirect from the component itself -- please investigate)_
- It needs to have the layout that has the "back to colony" controls on it _(I think it's `NavBar`)_

ArmandoGraterol:
Added `/:colonyName/unwrap-tokens` and `/:colonyName/claim-tokens` routes to the metacolony. 

**What to test:**
1. You shouldn't be able to go to those routes if the colony from which we are trying to access it is not the meta colony. If you try it, you should be redirected to a 404 page.
2. You shouldn't be able to access the pages if you don't have a wallet connected and username assigned.

In order to test the page with a normal non-meta colony you can just change line `208` inside `WalletProtectedRoute` to `props.match.params.colonyName === METACOLONY_ENS`

Resolves #2976

### DIFF
--- a/src/modules/dashboard/components/ClaimTokensPage/ClaimTokensPage.tsx
+++ b/src/modules/dashboard/components/ClaimTokensPage/ClaimTokensPage.tsx
@@ -1,0 +1,12 @@
+import React from 'react';
+
+const displayName = 'dashboard.ClaimTokensPage';
+
+const ClaimTokensPage = () => {
+  // @TODO: Add actual content for the claim tokens page
+  return <span>Claim tokens now!! Limited time offer</span>;
+};
+
+ClaimTokensPage.displayName = displayName;
+
+export default ClaimTokensPage;
--- a/src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
+++ b/src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
@@ -2,7 +2,6 @@ import React from 'react';
 import { defineMessages } from 'react-intl';
 import { Redirect, useHistory } from 'react-router-dom';
 
-import { LANDING_PAGE_ROUTE } from '~routes/index';
 import { useSelector } from '~utils/hooks';
 import { useLoggedInUser } from '~data/index';
 
@@ -28,14 +27,26 @@ const displayName = 'dashboard.CreateUserWizard.StepConfirmTransaction';
 const StepConfirmTransaction = () => {
   const { username } = useLoggedInUser();
   const transactionGroups = useSelector(groupedTransactions);
-  const { location } = useHistory<{ colonyURL?: string }>();
+  const { location } = useHistory<{
+    colonyURL?: string;
+    redirectTo?: string;
+  }>();
 
   if (username) {
     if (location?.state?.colonyURL) {
       return <Redirect to={location.state.colonyURL} />;
     }
 
-    return <Redirect to={LANDING_PAGE_ROUTE} />;
+    return (
+      <Redirect
+        to={{
+          pathname: location.state.redirectTo,
+          state: {
+            ...location.state,
+          },
+        }}
+      />
+    );
   }
 
   const colonyTransaction = findTransactionGroupByKey(
--- a/src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
+++ b/src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
@@ -12,6 +12,8 @@ import {
 } from '~users/GasStation/transactionGroup';
 import Heading from '~core/Heading';
 import GasStationContent from '~users/GasStation/GasStationContent';
+import { LANDING_PAGE_ROUTE } from '~routes/routeConstants';
+
 import styles from './StepConfirmTransaction.css';
 
 const MSG = defineMessages({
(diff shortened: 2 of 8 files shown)


src/routes/Routes.tsx
```diff
@@ -145,6 +154,28 @@ const Routes = () => {
...
         />
+        <WalletRequiredRoute
+          isConnected={isConnected}
+          didClaimProfile={didClaimProfile}
+          path={UNWRAP_TOKEN_ROUTE}
+          component={UnwrapTokensPage}
+          layout={NavBar}
+          routeProps={{
+            backText: MSG.backToMetacolony,
+            backRoute: `/colony/${METACOLONY_ENS}`,
```
rdig, line 165: These should still be colony generic, and not specific to the metacolony, as we might want to open up this to all colonies at some point.

Plus, you already test this inside the route and redirect if it's a different colony, so the "safety" catch is there

src/routes/Routes.tsx
```diff
@@ -145,6 +154,28 @@ const Routes = () => {
             hasBackLink: false,
           }}
         />
+        <WalletRequiredRoute
```
rdig, line 157: I think we'll need to change this to a "AlwaysAccesible" route_(yes, I know, I told you to require a wallet in the specs)_, but now I'm thinking that the users should be able to browse it without having connected (and controls disabled in the component of course)

This is because these links will be shared publicly, meaning users will go directly to them, and users who didn't log into colony previously will be redirected to 404, which will not go over well

src/routes/WalletRequiredRoute.tsx
```diff
@@ -50,13 +53,22 @@ const WalletRequiredRoute = ({
...
+            redirectTo?: string;
+            colonyURL?: string;
+            isMetaColonyExclusive?: boolean;
+          }
         >,
       ) => {
+        const isMetaColonyExclusive =
+          props.match.path === CLAIM_TOKEN_ROUTE ||
+          props.match.path === UNWRAP_TOKEN_ROUTE;
+
```
rdig, line 71: I like this, but as I mentioned above, I think we need to switch this to use the other type of route, meaning this won't be needed anymore

src/modules/dashboard/components/ClaimTokensPage/ClaimTokensPage.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+
+const displayName = 'dashboard.ClaimTokensPage';
+
+const ClaimTokensPage = () => {
+  // @TODO: Add actual content for the claim tokens page
+  return <span>Claim tokens now!! Limited time offer</span>;
```
rdig, line 7: :)

src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
```diff
@@ -28,15 +27,27 @@ const displayName = 'dashboard.CreateUserWizard.StepConfirmTransaction';
...
     const { colonyURL } = location.state;
     if (colonyURL) {
       return <Redirect to={colonyURL} />;
     }
 
-    return <Redirect to={LANDING_PAGE_ROUTE} />;
+    return (
+      <Redirect
+        to={{
+          pathname: location.state.redirectTo,
```
rdig, line 44: What if there's no `redirectTo` in the location state, where will this redirect to ?

rdig (changes requested):
I'm good with the changes here, however thinking more about it, we need to change it, otherwise we'll run into unpleasant scenarios for users trying to claim (basically being redirected to 404, which will scare them)

src/routes/Routes.tsx
ArmandoGraterol: That scenario makes sense! I agree that it's better to leave it as public as possible

src/routes/WalletRequiredRoute.tsx
ArmandoGraterol: Does this mean that we want to make these routes _not_ metacolony exclusive so that users are not redirected to a 404 page if they try to access the route from another colony?

src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
ArmandoGraterol: When `redirectTo` is `undefined`, 2 things can happen:

1. You get redirected to `/connect` if you don't have a wallet connected.
2. If you are connected and have a username, you get redirected to the landing page but I think that's a side effect of the code we have in `WalletRequiredRoute`

So, I'm gonna add the landing page route as default, just to be sure

src/routes/WalletRequiredRoute.tsx
rdig: Yes

src/modules/dashboard/components/CreateUserWizard/StepConfirmTransaction.tsx
rdig: Great

chinins (approved):
Very good! I've been working off of it in my PR as well.

PR #2984, opened 2021-11-30
Automatically add Xdai Chain when using Metamask

rdig:
This PR adds the ability for the dapp to automatically add Xdai Chain to Metamask _(if the user doesn't already have it)_, then switch to it.

This will only trigger if:
- Using metamask
- If the user doesn't have the network added _(if adding)_
- If not xdai or xdai fork networks _(if switching)_

If the user already has the network added to metamask _(as determined by the `chainId` value)_, metamask will only prompt to switch the network _(if not already on said network)_. If all of the previous are true _(both the network exists and the user is on the current network)_ then nothing will happen.

The bulk of the work here was actually added to `@purser/metamask`, which implements the underlying functionality, and is directly interacting with Metamask, so if curious on what changed, have a look at this diff: https://github.com/JoinColony/purser/compare/71029c6934597146a8eb7d3cf93e58f3e53f992a...3cab89377cfd93b50af482ff5e03ee8ce58d8983

Note, since this branch uses an update version of `@purser/metamask` please make sure to run `npm i` before running it, otherwise none of this functionality will work.

**Testing**

- Make sure Metamask is connected to your local network
- Delete your current Xdai network from metamask _(Metamask Extension -> Settings -> Networks -> Xdai -> Delete)_
- Change this logic check to always return `true`: https://github.com/JoinColony/colonyDapp/blob/56595eee23cbf20935330e73e43c5a7361fdcc92/src/modules/users/sagas/wallet.ts#L57-L60
Something like: `if (true) { ... }`
- Refresh the dapp and connect your Metamask wallet
- _(you should now be prompted to both add the new network, then switch to it)_

**Screenshots**

Screenshot from 2021-12-01 00-53-46

Screenshot from 2021-12-01 00-54-03

Screenshot from 2021-12-01 00-54-45

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,7 +17,7 @@
         "@formatjs/intl-unified-numberformat": "^3.3.5",
         "@popperjs/core": "^2.3.3",
         "@purser/core": "^3.0.0",
-        "@purser/metamask": "^3.1.2",
+        "@purser/metamask": "^3.2.0",
         "@purser/signer-ethers": "^1.0.3",
         "@purser/software": "^2.0.0",
         "@synaps-io/verify.js": "^3.0.5",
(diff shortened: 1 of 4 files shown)


arrenv (approved):
Nice addition! It is working well for me!

xdai-network-switcher

NOTE: for other testers if you get a types error you may need to restart TS server.

chinins (approved):
Nice - working as exoected!

PR #2985, opened 2021-12-01
Fix mint tokens and unlock tokens dialog actions not showing

### LINKED ISSUE #2941: Users w/o permissions can't see Mint Tokens / Unlock Token entries
This is an oversight of #2822 which prevent colonies which don't have a native token from displaying these two entries.

However, this also hides them for users without the required permissions.

This is is not correct, as for colonies which have native tokens, and for users without permissions, they should still show, just be non-accessible _(show the yellow triangle warning icon)_

This is because, when enabling Motions, users without minting permission, could still create a motion to mint said token, and for that the menu entry is required.

ArmandoGraterol:
- Added a `canColonyMintNativeToken` field to the `ProcessedColony` fragment to determine if the colony (independent of the user's permissions) can mint/unlock their native/default token.
- Updated previous `canMintNativeToken` field to `canUserMintNativeToken`, as this field specifically checks if the user can mint tokens, not the colony.
- Updated `TokenMintForm` and `ManageFundsDialog` and `ManageFundsDialog` validation based on changes mentioned above.
- **NEW:** Added a new redux action called `USER_CONNECTED` to let `ColonyHomeActions` know whether the wallet is finished loading in.

**What to test:**
- Colony with a native token as the default:
  - As a root user, you should be able to mint and unlock the token using the UAC or manage funds page. If the motion and disputes extension is installed, you should be able to create a motion to mint or force it.
  - As a non-root user, you should _not_ be able to mint and unlock the token, at all. If the motion and disputes extension is installed, you should be able to create a motion to mint tokens.
- Colony with a non-native token as the default:
  - As ANY user, you should NOT see the `Mint tokens` and `Unlock tokens` entries inside the UAC and the option to mint 
    tokens in the manage funds page should be disabled.
- **NEW:** When the dapp hasn't fully loaded the connected wallet or the extensions, you should be shown a loading state where the `New action` button is supposed to be in order to prevent you from clicking it. This is done to avoid the following scenarios:
  - If the wallet is not fully loaded, then the UAC won't be able to tell if the user can mint, unlock tokens, and potentially won't be able to do other actions that require permissions.
  - If the extensions are not fully loaded, then the UAC won't know, for example, if the motion & disputes extension is installed and therefore show the actions as if the extension wasn't installed. 

Resolves #2941


rdig (changes requested):
Code wise this is very nice! Kudos! :tada: 

However, I think there's still some caching issue going around related to the Mint Tokens check:
- first time you load a colony it doesn't allow you to, telling you, you're missing permissions _(this is made even worse by the fact that I had motions installed, so it should have allowed me to create a motion at least)_
- after you go to a different colony then come back, the cache is refreshes and I can mint.

Check out the demo:
bug-first-time-cannot-mint

ArmandoGraterol:
> Code wise this is very nice! Kudos! tada
> 
> However, I think there's still some caching issue going around related to the Mint Tokens check:
> 
> * first time you load a colony it doesn't allow you to, telling you, you're missing permissions _(this is made even worse by the fact that I had motions installed, so it should have allowed me to create a motion at least)_
> * after you go to a different colony then come back, the cache is refreshes and I can mint.
> 
> Check out the demo: bug-first-time-cannot-mint

@rdig From what I'm seeing in your demo and from what I've tested, I think that what's happening is that you are able to click New Action before your wallet and extensions finished loading in. I've managed to replicate this by hard reloading and instantly clicking `New action`, if I wait for the wallet and extension to load, it doesn't happen.

Any thoughts?

rdig:
Hmmm... good point.

I'll test it again

rdig:
You are correct in your diagnosis that this is because the wallet is not yet loaded.

However this is very confusing UX for users and we'll need to find another solution to alleviate this. We'll also, most likely, need the input of _Product_ here, because it's going to be hard to achieve and we'll most likely have to do some compromises.

The "cleanest" solution here would be to show a "loading" state until the wallet has loaded and the resolvers have refreshed (inside the index modal), but that's almost impossible due to how the data is wired, so we'll need to come up with something else.

Explain this problem to product, maybe they have a different view on things

rdig:
What the state with this one? Has _product_ responded on how they want this handled ?

ArmandoGraterol:
> What the state with this one? Has _product_ responded on how they want this handled ?

Yes, their suggested solution is to add a loading state to the index UAC modal. I will further talk about implementing this with you on Discord.

src/data/cacheUpdates.ts
```diff
@@ -292,7 +292,7 @@ const cacheUpdates = {
             cache.modify({
               id: cache.identify(data.processedColony),
               fields: {
-                canMintNativeToken: () => canMintNativeToken,
+                canUserMintNativeToken: () => canMintNativeToken,
```
chinins, line 295: Shouldn't we add `canColonyMintTokens` here as well?

src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
```diff
@@ -177,7 +174,7 @@ const ManageFundsDialog = ({
       description: MSG.unlockTokensDescription,
       icon: 'emoji-padlock',
       onClick: () => callStep(nextStepUnlockToken),
-      permissionRequired: !canUnlockToken,
+      permissionRequired: !colony.canUnlockNativeToken,
```
chinins, line 177: Just to confirm, we remove the option of raising the motion to unlock token?

src/data/resolvers/colony.ts
```diff
@@ -404,20 +406,46 @@ export const colonyResolvers = ({
...
       return canMintNativeToken;
     },
+    async canUserMintNativeToken({ colonyAddress }) {
+      const colonyClient = await colonyManager.getClient(
+        ClientType.ColonyClient,
+        colonyAddress,
+      );
+      // fetch whether the user is allowed to mint tokens via the colony
+      try {
+        await colonyClient.estimate.mintTokens(bigNumberify(1));
```
chinins, line 443: As talked on discord, how come the `canUserMintNativeToken` doesn't include the user address? Could you please clarify this - for reference.

chinins (commented):
All working properly - vey nice work. Could you just please confirm a couple of points I mentioned in my commentS?

src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
ArmandoGraterol: As far as I know, we haven't implemented the possibility of making a motion to unlock the token. There's even an issue in maintenance to update the modal to support that

src/data/resolvers/colony.ts
ArmandoGraterol: Talked with Raul about this. `.mintTokens` doesn't include the user address because the contracts already extract the address from the wallet that is sending the transaction to check if they can mint or not. 

It's a safe way of checking that the user has permission to mint, otherwise, if we added an address param to the method call any user could just plug in an address with permission that is not necessarily their own and mint tokens like nobody business.

src/data/cacheUpdates.ts
ArmandoGraterol: From what I understood when I talked with Raul about the usefulness of adding `canColonyMintTokens` to this file... It seems like this may not be necessary as this file is the old way in which we determine how we updated the cache after we received the results of a query, but we haven't been using/updating it all this time.

It could be argued that it could alleviate the amount of data that the dapp needs to query or receive, but it could also be detrimental for the performance because it's extra stuff that the client needs to do every time, for example, a user reloads the dapp and it's data that we want to stay as fresh as possible anyways.

Due to that, I think I won't add `canColonyMintTokens` yet. Also because it's not as straight-forward to get as `canUserMintNativeToken`

chinins (approved):
Thank you very much for clarifying the points! It's really good to learn this. Very nice work!

chinins (changes requested):
As discussed in discord, it's missing the similar approach for unlockToken. We should have `canUnlockNativeToken` and `canUserUnlockNativeToken`.

arrenv (commented):
It seems to be much more reliable now.

I think the addition of the pre-loader on the action button is a good idea. It could maybe show the button, with a loader on it, so that the page has less jump in. However, that depends on what you were spec'd.

ArmandoGraterol:
> As discussed in discord, it's missing the similar approach for unlockToken. We should have `canUnlockNativeToken` and `canUserUnlockNativeToken`.

@chinins Done!

chinins (approved):
Very good! Working as expected and it's an important fix.

@arrenv, we need some of these changes in the Unlock token feature branch. It would be good to rebase the feature branch on master once Armando merges this PR. I will then need to update some logic fr the unlock token motion. I raised an issue for that: #3118

arrenv (approved):
@ArmandoGraterol, very nice work!! Works great!

PR #2992, opened 2021-12-01
Fix subscription data being wrongly updated

ArmandoGraterol:
Addresses an issue where the subscription data would be updated when it's not supposed to. Now the affected subscription should only be updated when _new, relevant_ data comes through.

**Subscriptions affected:**
- `transactionMessages`: Used in the chats of the coin machine and action pages, this was fixed before so the changes here are more of an improvement.
- `transactionMessagesCount`: Used to display the comment counter that appears in the action list items. Before, if you had 2 tabs with 2 different colonies open, and one of the subscriptions of any of the colonies fired, it would make the counter of the other colony disappear.
- `subscribedUsers`: Used to get a member list of a colony.  

`colonyServer` side of this PR: https://github.com/JoinColony/colonyServer/pull/147


rdig (changes requested):
I like the changes. I especially like the use of `withFilter` on the server side, that one is actually clever.

However, I was able to replicate the members list problem still on local _(comments seem to work fine, both on colonies and on coin machine)_

Here's a demo replicating this:

bug-members-list-updates-on-all-colonies

rdig (approved):
Turns out my original review test was flawed somehow, which is weird since I tested this two times before submitting that.

Anyway, I've tested yet again, a third time, since all the code changes made sense, and it shouldn't have been broken still.

And yes, upon the latest test, it all seems to be working as expected.

Very nice work! :100: 

Here's the working test:

members-list-no-longer-updating-wrongly

chinins (approved):
Very nice! Working as expected.

PR #2994, opened 2021-12-02
Fix edit domain dialog 

### LINKED ISSUE #2988: Broken `Edit domain` dialog
Has horizontal scroll sidebar. Requires domain description although shouldn't.

Check both motions & actions.

ArmandoGraterol:
Fixed `domainPurpose` validation on the edit domain.

The dialog was overflowing because the error message (that was hidden) below the domain purpose input was really long, due to `domainPurpose` being initially `null/undefined`.

Resolves #2988


alicjakujawa (approved):

arrenv (approved):
I was not able to replicate the original issue, however, this makes sense and does not break anything on my end.

PR #2997, opened 2021-12-02
Fixes permissions bug on unlock modal when token already unlocked

### LINKED ISSUE #2989: After unlocking your token you can no longer access the modal
After you unlock your token, you can no longer access that modal, since it tells you (wrongly) that you don't have the correct permissions. (you do, I just suspect we do a unlock token check which is not correct)

Screenshot from 2021-12-01 20-09-58

There's two ways of fixing this:
- allow you to access the modal, but tell you that the token is already unlocked and there's nothing for you to do (with the controls disabled of course)
- hide the menu entry if the token is unlocked

Please check with Product on which way they'd like us to fix it.

agustif:
## Description

- [x] allow you to access the modal, but tell you that the token is already unlocked and there's nothing for you to do (with the controls disabled of course)

**New stuff** ✨

* Added new unlockedTitle, and unlockedDescription, which are conditionally showed instead of the regular title & description when the NativeToken is Not Locked anymore.

**Changes** 🏗

* Where title and description are called, we check for colony.nativeTokenIsLocked, and if true we show the regular one, if false the unlocked one variants.

**Deletions** ⚰️

* Removed condition check colony.isNativeTokenLocked from canUnlockNativeToken (So permission error doesn't show when it's already Unlocked)

## TODO

- [x] Add new unlocked title and description messages
- [x] Remove permissions check if token unlocked
- [x] Show different versions of text if token alrady unlocked

Resolves #2989

### DIFF
--- a/src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
+++ b/src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
@@ -117,7 +117,7 @@ const ManageFundsDialog = ({
     ? colony.canMintNativeToken
     : colony.canMintNativeToken && hasRoot(allUserRoles);
   const canUnlockToken = isVotingExtensionEnabled
-    ? colony.isNativeTokenLocked && colony.canUnlockNativeToken
+    ? colony.canUnlockNativeToken
     : colony.canUnlockNativeToken && hasRoot(allUserRoles);
 
   const canManageTokens = hasRegisteredProfile && hasRoot(allUserRoles);
--- a/src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
+++ b/src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
@@ -118,9 +118,7 @@ const ManageFundsDialog = ({
     : colony.canMintNativeToken && hasRoot(allUserRoles);
   const canUnlockToken = isVotingExtensionEnabled
     ? colony.isNativeTokenLocked && colony.canUnlockNativeToken
-    : colony.isNativeTokenLocked &&
-      colony.canUnlockNativeToken &&
-      hasRoot(allUserRoles);
+    : colony.canUnlockNativeToken && hasRoot(allUserRoles);
 
   const canManageTokens = hasRegisteredProfile && hasRoot(allUserRoles);
 
--- a/src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenForm.tsx
+++ b/src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenForm.tsx
@@ -40,8 +40,8 @@ const MSG = defineMessages({
       to take this action.`,
   },
   unlockedTitle: {
-    id: 'dashboard.UnlockTokenDialog.UnlockTokenForm.dialogunlockedTitle',
-    defaultMessage: 'Token  Unlocked',
+    id: 'dashboard.UnlockTokenDialog.UnlockTokenForm.unlockedTitle',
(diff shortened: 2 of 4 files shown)


src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenForm.tsx
```diff
@@ -39,6 +39,14 @@ const MSG = defineMessages({
     defaultMessage: `You do not have the {roleRequired} permission required
       to take this action.`,
   },
+  unlockedTitle: {
+    id: 'dashboard.UnlockTokenDialog.UnlockTokenForm.dialogunlockedTitle',
```
ArmandoGraterol, line 43: Typo here, should be `.unlockedTitle`

src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
```diff
@@ -118,9 +118,7 @@ const ManageFundsDialog = ({
     : colony.canMintNativeToken && hasRoot(allUserRoles);
   const canUnlockToken = isVotingExtensionEnabled
     ? colony.isNativeTokenLocked && colony.canUnlockNativeToken
```
ArmandoGraterol, line 120: You also need to change this condition, as of now I can't access the dialog even if I have root permission when the voting reputation extension is enabled:
FireShot Capture 504 - Colony - localhost

ArmandoGraterol (changes requested):
The non-reputation version of the modal seems to be working as expected, however, the voting rep version is still not accesible.

src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenForm.tsx
agustif: Good one, thanks for spotting, fixed on new commit

src/modules/dashboard/components/ManageFundsDialog/ManageFundsDialog.tsx
agustif: I've fixed it by removing `colony.isNativeTokenLocked &&` in line 120.

Tested by enabling motions and working as expected, thanks for catching this as I didn't think of it

ArmandoGraterol (approved):
Yep, now I can access it with or without the voting reputation extension :+1:

rdig (changes requested):
All good, except the warning should not show up at all, when the token is unlocked already.

src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenForm.tsx
```diff
@@ -81,7 +88,11 @@ const UnlockTokenForm = ({
...
       )}
       <DialogSection appearance={{ theme: 'sidePadding' }}>
-        <FormattedMessage {...MSG.description} />
+        {colony.isNativeTokenLocked ? (
+          <FormattedMessage {...MSG.description} />
+        ) : (
+          <FormattedMessage {...MSG.unlockedDescription} />
+        )}
       </DialogSection>
       <DialogSection appearance={{ theme: 'sidePadding' }}>
```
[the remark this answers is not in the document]
agustif: Yeah makes sense was silly to leave it, will hide the notice when unlocked. Thanks Raul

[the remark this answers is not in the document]
agustif: Fixed now on last commit, I added an unlocked style to add margin-bottom 20px, since hiding the notice made the description have no margin-bottom.

image

rdig (approved):
Good to go!

PR #2998, opened 2021-12-02
Add `Unwrap Tokens` and `Claim Tokens` menu entries

### LINKED ISSUE #2977: Create Unwrap Tokens and Claim Tokens Colony Menu Entries
Create two menu entries in the left colony home sidebar for the _Unwrap Tokens_ and _Claim Tokens_ components.

These will render both the aforementioned component, as well as be accessible through their routes  #2976

Names:
- `Uwrap Tokens`
- `Claim Tokens`

Requirements:
- Only visible to the metacolony _(we might enable for the general public at some point)_
- Manual disable flag _(basically a `disabled` prop)_
- If they are disabled or not the metacolony, redirect to the colony home _(note that they can also be accessed through a route)_
- They should placed after the "Buy Tokens" menu entry

Screenshot from 2021-11-29 19-15-37

ArmandoGraterol:
FireShot Capture 508 - Colony - localhost

Resolves #2977


chinins (approved):
Looks good!

PR #2999, opened 2021-12-02
Show member popover when user avatar selected

### LINKED ISSUE #2847: When clicking on a user from within action details view, show the user profile popover
### Story

Example: As a *user creating a colony*, I want to *see a visual indicator of activity/loading*, so that *I know I should wait for it to finish*

As a user interacting with a motion, I want to see the new profile popover when clicking on a user throughout the app, so that I understand the details of the user. 

image

[Video example](https://share.descript.com/view/2OyMAZJzlOm)

- [ ] When looking at a Motion's details, clicking on a user's avatar should show the profile popover

danbr:
## Description

This PR updates Actions, Events, Extension, & Motion pages to show user information popover when clicking on UserAvatar.

Screenshot 2021-12-02 at 13 21 32

## Bugs
Some issues that were identified and corrected along the way:

- Some MemberInfoPopover inherited from their parent and text became right aligned.
- Differences in Width (CSS) between InfoPopover & MemberInfoPopover 
- InfoPopover “Copy” link when clicked, not showing “Copied” fully, due to Avatar size being affected by parent.

Resolves #2847


src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -219,9 +221,21 @@ const ActionsListItem = ({
...
+           * Clicking on UserAvatar would redirect to Actions page and stop
+           * interaction with popover.
+           * stopPropagation prevents event being inherited by child
+           */
+          onClick={stopPropagation}
+          onKeyPress={stopPropagation}
+          role="button"
+          tabIndex={0}
+          className={styles.avatar}
+        >
```
rdig, line 235: I'm good with this change, however there's an `handleOnClick` prop that's coming in, which assumes it can run the click action on the whole item. 

Stopping propagation on the avatar, might impede that to some extent.

But since there's not "proper" way to check the execution requirements of the `onClick` function that's coming in _(well, at least not without going a very deep dark rabbit hole)_, it will have to do for now

src/modules/core/components/DetailsWidgetUser/DetailsWidgetUser.tsx
```diff
@@ -23,11 +24,16 @@ const DetailsWidgetUser = ({ walletAddress }: Props) => {
...
+        colony={colony}
         size="s"
         notSet={false}
         user={userProfile}
         address={walletAddress || ''}
         showInfo
+        popperProps={{
+          showArrow: false,
+          placement: 'bottom',
+        }}
```
rdig, line 36: I think this looks better with a `left` placement and a bit of offset

```suggestion
        popperProps={{
          showArrow: false,
          placement: 'left',
          modifiers: [
            {
              name: 'offset',
              options: {
                offset: [0, 10],
              },
            },
          ],
        }}
```

Screenshot from 2021-12-03 11-28-52

src/modules/core/components/InfoPopover/MemberInfoPopover/MemberInfoPopover.css
```diff
@@ -2,7 +2,10 @@
 
 .main {
   padding: 20px 25px 25px 25px;
-  width: 470px;
+  width: 480px;
+  letter-spacing: normal;
+  text-align: left;
+  cursor: auto;
```
rdig, line 8: I understand the width update (since that was missed when this was implemented)

But why the reset to default for `letter-spacing`, `text-align` and `cursor` ?

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -194,6 +194,8 @@ const ActionsListItem = ({
     motionState === MotionState.Passed ||
     motionState === MotionState.Failed ||
     motionState === MotionState.FailedNoFinalizable;
+
+  const stopPropagation = (event) => event.stopPropagation();
```
rdig, line 198: Why did you make this an external, named, method rather then making it an anonymous defined at render time ?

_(It's good what you did, but I still want to know what was your rationale for it)_

rdig (changes requested):
Very nice work! :100: 

Congratulations on your first PR!

This is all good, I just rejected it because I want you to add the "fancy' popover in one more place: Annotations on the Actions/Motions pages

See this screenshot:

Screenshot from 2021-12-03 11-25-56

The user widget on the right has the correct member popover _(the one that you just added in this PR)_, while the one that appears over the avatar in Annotations, is still the "basic" version.

Please add it in there as well.

_PS: you can drop the "Feature" from your PR's title, we only use "tags" like that for commit messages, PR titles should be more "human friendly" :)_

src/modules/core/components/InfoPopover/MemberInfoPopover/MemberInfoPopover.css
rdig: PS: these are not ordered properly according to our style lint rules, that's why your build is failing.

If stylelint is not set up in your editor, you can run it manually via `npm run stylelint`

src/modules/core/components/ActionsList/ActionsListItem.tsx
danbr: The onClick was being passed to MemberInfoPopover, so it became clickable (same as the parent listItem), taking the user to the DefaultAction page. 
Preventing the propagation allows MemberInfoPopover to behave as expected.

src/modules/core/components/InfoPopover/MemberInfoPopover/MemberInfoPopover.css
danbr: There were several occasions where the popover inherited properties of the parent, screwing up the layout, or behavior. The extra props were added to standardize the memberInfoPopover.

Certainly, I need to spend some time improving my editor config. I ran npm linting & type checking manually, so surprised it failed the build.
I want everything automated - manual checking is too easy to forget!!

src/modules/core/components/ActionsList/ActionsListItem.tsx
danbr: 2 reasons..... easy!!
- Improved performance: Anonymous functions are not persistent, so TypeScript has to create them again and again on every re-render. Verses creating the named function only once. 
- I checked the codebase for similar situations, to see how you deal with such situations..... and followed your example. ;)

rdig (approved):
All good now, however there's a "rogue" merge commit in your commit history for this branch that shouldn't be there _(most likely a rebase gone wrong)_

Screenshot from 2021-12-06 16-50-14


Please clean up your commit history, after which you can merge

I'm pre-approving in the mean time.

_(PS: let me know if you have trouble merging due to permissions issues)_

ArmandoGraterol (approved):
Looking good! Nice first PR

Issue #2974, opened 2021-11-29
Create Vesting Unwrap Component


rdig:
This component will be used by people who bought a wrapped token _(wrapped clny basically)_ during the token sale, to help them unwrap it.

This component will be connected to a route _(#2976)_, and only available if the current colony is the metacolony and will be accessible through the Colony Home left sidebar _(#2977)_

This and the _Claim Tokens_ component will be part of the `Vesting` "feature" _(which is only available for the metacolony currently -- we might release it for others in the future though)_, so  please make sure to place this component inside `~dashboard/Vesting/...`

It needs:
- only accessible if the current colony is the metacolony
- loading state, while we fetch (or refresh) the user's respective token balance
- the wrapped token to be configurable _(basically take the token -- `AnyToken` type -- as a prop)_
- the unwrapped token  will come from a query inside the component (will be added by Raul separately)
- It will send a transaction, so make sure to wrap the button in an `ActionForm` _(don't worry about wiring, this issue is just setting this up)_
- Make sure to ask _Product_ for the correct copy, both for the description, and for the tooltips

_PS: We use the same list styles for the Actions/Motions pages details sidebar, and in some Motions widgets, so you can easily copy the styles from there._

**Screenshot**

![Unwrap](https://user-images.githubusercontent.com/4774274/144281388-7abdd76a-80cd-4f64-b434-7fd91122a81a.png)

**Figma File**
https://www.figma.com/file/lT70SdJVGABrA3DBEhof4Z/Vesting?node-id=0%3A1

rdig:
Closed by #3009

PR #3009, opened 2021-12-06
Add unwrap & claim tokens pages

### LINKED ISSUE #2975: Create Vesting Claim Tokens Component
This component will be used by people who have tokens allocated to them _(basically the team and investors)_ and will periodically claim them, as they unlock over a period of time.

This component will be connected to a route _(#2976)_, and only available if the current colony is the metacolony and will be accessible through the Colony Home left sidebar _(#2977)_

This and the _Unwrap Tokens_ component will be part of the `Vesting` "feature" _(which is only available for the metacolony currently -- we might release it for others in the future though)_, so  please make sure to place this component inside `~dashboard/Vesting/...`

It needs:
- only accessible if the current colony is the metacolony
- loading state, while we fetch (or refresh) the user's respective token balances and amounts
- the vested token will come from a query inside the component (will be added by Raul separately)
- It will send a transaction, so make sure to wrap the button in an `ActionForm` _(don't worry about wiring, this issue is just setting this up)_
- Make sure to ask _Product_ for the correct copy, both for the description, and for the tooltips - no more description or tooltips (deleted by Jack)

_PS: We use the same list styles for the Actions/Motions pages details sidebar, and in some Motions widgets, so you can easily copy the styles from there._

**Screenshot**

Vesting

**Figma File**
https://www.figma.com/file/lT70SdJVGABrA3DBEhof4Z/Vesting?node-id=0%3A1

chinins:
## Description

This PR adds Claim Tokens and Unwrap tokens pages.  Please, note that you see the tooltips in the Unwrap tokens page but Jack said to remove them.

Please, also note that this PR is vs. a `routes` branch but as soon as Armando merges it into vesting branch I will change the base comparison branch.

TO TEST, go to the routes:
`http://localhost:9090/colony/colonyName/unwrap-tokens`
`http://localhost:9090/colony/colonyName/claim-tokens`

**New stuff** ✨

 - add `VestingPageLayout`
 - add `ClaimTokensPage` & `UnwrapTokensPage`

## TODO

To wire the pages (dines separately by Raul)

resolves #2975 #2974 

Screenshot 2021-12-07 at 18 50 51

Screenshot 2021-12-07 at 18 51 07


arrenv (approved):
Looks good to me! Only some slight padding difference with the spec, but likely not important.

src/modules/dashboard/components/Vesting/UnwrapTokensPage.tsx
```diff
@@ -0,0 +1,78 @@
...
+      }
+      tableValues={[
+        {
+          label: (
+            <FormattedMessage
+              {...MSG.wrappedBalance}
+              values={{ tokenSymbol: token?.symbol }}
+            />
+          ),
+          value: token.wrappedBalance,
```
arrenv, line 57: Should there be some checks and fallbacks, in case no token data is retrieved from query?

arrenv: Or, will that be handled by the wiring?

rdig: Handled by wiring, but even then I don't think I'll put in a fallback for this.

If the token doesn't load, there's bigger problems as the colony, prior to this hasn't loaded so the dapp crashes :)

src/modules/dashboard/components/Vesting/VestingPageLayout.tsx
```diff
@@ -0,0 +1,74 @@
...
+      {title}
+      <div className={styles.table}>
+        {tableValues.map(({ label, value }) => (
+          <div className={styles.item}>
+            <div className={styles.label}>{label}</div>
+            <div className={styles.value}>
+              <Numeral value={getFormattedTokenValue(value, tokenDecimals)} />
+            </div>
+          </div>
+        ))}
```
rdig, line 59: Items in this list need a unique `key`

Screenshot from 2021-12-08 16-46-56

src/modules/dashboard/components/Vesting/VestingPageLayout.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+const VestingPageLayout = ({
+  title,
+  tableValues,
+  buttonText,
+  tokenDecimals,
+  buttonTextValues,
+  isLoading,
+}: Props) => {
```
rdig, line 40: How will this component handle the different sagas that we need to send ? (one for unwrapping and one for claiming)

My point is that the Button should not be contained in here, but should be added individually to each component, as to be wrapped in a different `ActionForm`

src/modules/dashboard/components/Vesting/Vesting.css
```diff
@@ -0,0 +1,47 @@
+.main {
```
rdig, line 1: You've implemented according to the spec, but man the design looks horrible on large screens :)

Screenshot from 2021-12-08 16-46-41

_(there's nothing to do here, this is just me ranting)_

rdig (changes requested):
Almost there, however the pages need to handle the "send" button separately, as well as the vesting layout needs unique keys for it's mapped children

chinins:
> Looks good to me!
> 
> Only some slight padding difference with the spec, but likely not important. For reference:
> 
> unwrap-tokens

I applied the same padding as we have in Actions/Motions details widget - Karol said to keep styles as consistent as possible.

src/modules/dashboard/components/Vesting/VestingPageLayout.tsx
chinins: I've added the ActionForm to both pages and I think it doesn't require to have Button component twice. Please, check my implementation and let me know if you have something different in mind.
For now I've addded just Generic actions taking into account that you will change them once you wire this stuff up. But let me know if you want me to add already actual actions.

src/modules/dashboard/components/Vesting/VestingPageLayout.tsx
```diff
@@ -0,0 +1,80 @@
...
+        loadingText={MSG.loading}
+        appearance={{ theme: 'primary', size: 'massive' }}
+      />
+    </div>
+  ) : (
+    <div className={styles.main}>
+      {title}
+      <div className={styles.table}>
+        {tableValues.map(({ label, value }) => (
+          <div className={styles.item} key={value}>
```
rdig, line 57: This can't work properly as a `key` since those values will at some point be equal _(ie: two or more are `0`)_

The approach to take here, is to set up an autoID generator, like we have in places where we don't actually have access to good ids.

See this example:
https://github.com/JoinColony/colonyDapp/blob/f8080fb95a6b3f5c556355870f192edbabf9b41f/src/modules/core/components/IndexModal/IndexModal.tsx#L36-L42

PR #3016, opened 2021-12-07
Modifies SaleFailed state to be more versatile

### LINKED ISSUE #3003: Buying last available tokens at the end of the batch presents incorrect experience
## Steps to reproduce

1. Enable Coin Machine (set time period to like `0.2`)
2. Send in a small amount of tokens (like 100)
3. Buy the max tokens available near the very end of the batch (like 10 secs remaining)

image

## Expected behaviour

* Successful purchase
* OR if unsuccessful then unsuccessful UX (ie not CLNY as the default token listed, note the purchase above was in MEEP)
* No red box or text "Next sale starts in" if the sale is over

## Actual behaviour

* In this case, it was a successful sale (see below). But the UI says it was unsuccessful and says CLNY.

image

arrenv:
## Description

Modifies the SaleFailed state to be more versatile and avoids a blank button when there is no time left on the timer.

salefailed-state

salefailed-state-timer

**Testing** 

- You need to have coinMachine enabled and have it funded.
- You need to initiate a token buy.
- You may like to make the following temporary  modification to more reliably get the SaleFailed state. `setState(SaleState.SaleFailed); return;`

sample-code-2

- You may like to make the following temporary modification to more reliably get nil time value. `timeLeftToNextSale = 0;`

sample-code-1

Resolves #3003

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/SaleStateWidget/SaleStateWidget.tsx
+++ b/src/modules/dashboard/components/CoinMachine/SaleStateWidget/SaleStateWidget.tsx
@@ -66,7 +66,7 @@ const MSG = defineMessages({
   },
   saleFailedTitle: {
     id: 'dashboard.CoinMachine.SaleStateWidget.saleFailedTitle',
-    defaultMessage: `You didn't get any CLNY this time`,
+    defaultMessage: `You didn't get any {token} this time`,
   },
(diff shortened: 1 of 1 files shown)


rdig (approved):
Very nice work! :1st_place_medal:

PR #3019, opened 2021-12-08
Fix prices for Coinmachine periods with no sales

### LINKED ISSUE #2938: Price does not progress until the first purchase has been made.
Started a sale with a high starting price expecting to see the price begin to decrease after the first period had elapsed. Price did not progress as expected, and prices did not show up in the previous batches box until I made a purchase.

Screenshot 2021-11-21 at 22 45 49

Screenshot 2021-11-21 at 22 47 35

After this purchase, the numbers in the price field of previous batches are incorrect.

Screenshot 2021-11-21 at 22 53 18

### LINKED ISSUE #2995: 'Per user purchase limit' is being enforce on the batch instead of total sale
## Steps to reproduce

2. Setup whitelist extension w/ agreement option enabled
1. Setup a fresh Coin Machine
2. Select a `Per user purchase limit` of 10%
3. Set `Target per period` to 50
4. Set `Maximum Per Period` to 100
5. Set `Period Length` to 0.1
6. Send funds to Coin Machine to start sale
7. Attempt to purchase `Max` amount
8. Attempt to purchase `Max` amount in next batch

## Expected behaviour

* The user should be capped against the total token sale quantity

## Actual behaviour

* The user is capped against the batch's quantity 

This is a critical problem because the batches are smaller than the total tokens sold, which means we're restricting them on a % basis on a fraction of our intended cap.

### LINKED ISSUE #2996: Overlapping banner on motions
## Steps to reproduce

1. Create a motion
2. Ensure page length is long enough to scroll

## Expected behaviour

* Banner should be pinned to top of the content area and not overlap scroll bar.

## Actual behaviour

* Banner is overlapping the content and the scrollbar

motion-banner-overlaps

## Suggestions

* Banner should not be fixed,
* Should be fixed to the top of the screen on scroll, or
* Have the main top header stay fixed at the top of the screen along with the banner

### LINKED ISSUE #3006: Previous Coin Machine sale data populates in "Previous batches" box in a redeployment of a Coin Machine sale
UPDATE: This might be working as expected. I'm realizing the replication may have not uninstalled, which means this is operating on the same contract — which would mean this is working as expected. Working to replicate...

## Steps to reproduce

1. Setup a Coin Machine w/ `period length = 0.2`(note your settings)
2. Send 100 tokens to start Coin Machine
2. 1st batch, buy 1 token
3. 2nd batch, buy 1 token
4. 3rd batch, buy 98 tokens
5. Uninstall Coin Machine
6. Reinstall Coin Machine w/ same settings
7. Send 100 tokens to start Coin Machine
8. View "Previous batches" table
9. Wait until the time period passes at least twice

## Expected behaviour

* "Previous batches" table remains blank

## Actual behaviour

* "Previous batches" starts out blank, but upon the reset it seems to pull the previous deployment of Coin Machine's prior data. 

## Priority (optional)

* I'm not sure how important this is due to the low probability of this case being played out. But if the data persists over longer periods of time it could be potentially a more important challenge. Additionally, I'm making the assumption this is a visual bug. I need to test if the old data is impacting the pricing — in that case it would be an important issue.

area:
screenshot-2021-12-08_12-37-21

This PR updates the logic for calculating historical prices for coinmachine periods where no tokens have been sold. I've just lightly touched the maths rather than anything more radical, and as a result there are two wrinkles here that I have not solved, but for the majority of historical periods the prices are now correct. 

1. When calculating the price of periods prior to the first sale, I have assumed that the price is evolving and decays. That is not currently the case, but I believe is the intended behaviour. The 'correct' way to solve this is to look at the `PriceEvolutionSet` event on the contracts, which has a boolean argument to track whether the price is evolving or not.
2. If no tokens have been sold yet, the historical prices will show as 0.

We can either fix these here (I suspect the second point, at least, needs to be fixed, with [Raul's proposed solution](https://github.com/JoinColony/colonyDapp/issues/2938#issuecomment-975307170) the way to go), or elsewhere.

Resolves #3006
Resolves #2995 
Resolves #2938
Resolves #2996

### DIFF
--- a/jest.conf.json
+++ b/jest.conf.json
@@ -14,6 +14,7 @@
     "rootDir": "src",
     "moduleNameMapper": {
         "^\\~constants$": "<rootDir>/modules/constants.ts",
+        "^\\~externalUrls$": "<rootDir>/modules/externalUrls.ts",
         "\\.(jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$": "<rootDir>/__mocks__/fileMock.ts",
         "\\.(css|less)$": "<rootDir>/__mocks__/styleMock.ts",
         "^\\~context(.*)$": "<rootDir>/context$1",
--- a/package-lock.json
+++ b/package-lock.json
@@ -67,6 +67,7 @@
         "react-tabs": "^3.1.0",
         "react-textarea-autosize": "^8.3.3",
         "react-textfit": "^1.1.1",
+        "react-twitter-widgets": "^1.10.0",
         "recompose": "^0.30.0",
         "redux": "^4.0.1",
         "redux-immutable": "^4.0.0",
@@ -42482,6 +42483,11 @@
         "node": ">=4.0.0"
       }
     },
+    "node_modules/loadjs": {
+      "version": "4.2.0",
+      "resolved": "https://registry.npmjs.org/loadjs/-/loadjs-4.2.0.tgz",
+      "integrity": "sha512-AgQGZisAlTPbTEzrHPb6q+NYBMD+DP9uvGSIjSUM5uG+0jG15cb8axWpxuOIqrmQjn6scaaH8JwloiP27b2KXA=="
+    },
     "node_modules/localforage": {
       "version": "1.7.3",
       "resolved": "https://registry.npmjs.org/localforage/-/localforage-1.7.3.tgz",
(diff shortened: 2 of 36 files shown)


rdig:
> When calculating the price of periods prior to the first sale, I have assumed that the price is evolving and decays. That is not currently the case, but I believe is the intended behaviour. The 'correct' way to solve this is to look at the PriceEvolutionSet event on the contracts, which has a boolean argument to track whether the price is evolving or not.

I think we'll have to find a fix for this as well, but I really have no good idea on how

rdig (approved):
Good to go!

Thank you @area for fixing my shitty math!

ArmandoGraterol (approved):
LGTM :ok_hand:

chinins (approved):
All good!

PR #3021, opened 2021-12-08
Fixes wrong datetimes in motions

### LINKED ISSUE #3017: Wrong Dates and Times on Motions/Disputes
## Steps to reproduce

Create a motion. Stake motion. Object to motion. Stake objection. Vote. Reveal. Finalise. Claim. Claim.

## Expected behaviour

While it might be somewhat unorthodox to both support and oppose your own motion, it does happen. Things should proceed normally.

## Actual behaviour

dates

Source:
https://qaxdai.colony.io/colony/topologic/tx/0xc0c10dfd4d9ca56464c465b49852bf37fd167ac1441b1ecaed638b054fb53ec8

The only hunch I have is that everything went too fast. Since I was doing all the staking/objecting/voting/revealing myself maybe too little time had passed and somewhere in the code it assumes implicitly that this whole process takes some minimum amount of time? Maybe that's why the displayed times are off by 1 epoch?

agustif:
## Description
1. Enable motions
2. Create a motion
3. Date times would show wrongly before, should be OK now (Still not real, but I think 10 days in the future is OK for dev env)

- [x] Fix wrong datetimes in motions
**New stuff** ✨

* Handles miliseconds from MotionCreated by dropping all characters after 10 in the createdAt timestamp string.
* Doesn't cast dates before passing to TimeRelative since TimeRelative already handles this.

**Changes** 🏗

* Changed some props number / Date etc..

## TODO

- [ ] Find out why MotionCreated get's a createdAt in miliseconds (more than 10 characters) epoch time.

## Screenshots

Screenshot 2021-12-08 at 23 42 10
Screenshot 2021-12-08 at 23 30 36

Resolves #3017

### DIFF
--- a/src/modules/core/components/Comment/Comment.tsx
+++ b/src/modules/core/components/Comment/Comment.tsx
@@ -40,7 +40,7 @@ export interface Props {
   colony: Colony;
   user?: AnyUser | null;
   annotation?: boolean;
-  createdAt?: Date | number;
+  createdAt?: number;
   showControls?: boolean;
 }
 
--- a/src/modules/core/components/TimeRelative/TimeRelative.tsx
+++ b/src/modules/core/components/TimeRelative/TimeRelative.tsx
@@ -5,7 +5,7 @@ interface Props extends HTMLAttributes<HTMLSpanElement> {
   /** Time in seconds to update */
   updateInterval?: number;
   /** Datetime to calculate relative/difference from. */
-  value: Date;
+  value: Date | number;
 }
 
 const nearestIntervalOf = (value: number, increment: number) =>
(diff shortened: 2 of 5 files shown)


chinins (changes requested):
I think it's not the right way to fix it. You need to go to the root where created at is done/parsed and fix it there. This place is in the resolvers (specifically colonyAction resolver & motions resolvers - in `src/data/resolvers` folder). As I can see createdAt from subgraph is coming in seconds & from the contracts it seems that it's coming in milliseconds (but please check for yourself). So you need to normalise it.

Let me know if you have any questions.

Also I saw that the Comment box has disappeared at some point when staking/voting etc and appeared only after I refreshed the page. Please, make sure it's always there.

agustif:
Already fixed by https://github.com/JoinColony/colonyDapp/commit/5bcddd2c80671419dd29ea3748b8a549b127c518

PR #3023, opened 2021-12-09
Fix interpretation of the decimal points in amount fields

### LINKED ISSUE #3011: Coin Machine: When a user enters a decimal point rather than a number, the screen goes blank.

ArmandoGraterol:
The problem was that `Decimal` and `parseFloat`/`parseInteger` don't interpret '.' as '0.0', so I had to use an alternative like `toFinite` to get a safe value.

- If it's `parseFloat`/`parseInteger`, it would interpret it as `NaN`, resulting in the validation of the form throwing a really long validation error message that could break the interface, and it's not user friendly (`TransferFundsDialog`, `TokenMintDialog`, `CreatePaymentDialog`)
- If it's `Decimal` the one that receives the `.`, it would just crash the page and if it somehow doesn't, it still would show the long weird error (`BuyTokens`) 

Please check all of the places where we use an amount field, all of them should be able to perceive `.` as a `0` without crashing or throwing a user-unfriendly error.

Resolves #3011

### DIFF
--- a/src/data/staticData/extensionData.tsx
+++ b/src/data/staticData/extensionData.tsx
@@ -450,6 +450,7 @@ const extensions: { [key: string]: ExtensionData } = {
         validation: yup
           .number()
           .transform((value) => toFinite(value))
+          .positive()
           .required(),
         title: MSG.coinMachinePeriodLengthTitle,
         description: MSG.coinMachinePeriodLengthDescription,
@@ -463,7 +464,7 @@ const extensions: { [key: string]: ExtensionData } = {
         validation: yup
           .number()
           .transform((value) => toFinite(value))
-          .integer()
+          .positive()
           .required(),
         title: MSG.coinMachineWindowSizeTitle,
         description: MSG.coinMachineWindowSizeDescription,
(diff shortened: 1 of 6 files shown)


arrenv (changes requested):
The fix works well for me, there is just one other location I found that still has the error.

decimal-error-2

All the inputs on Motions and CoinMachine enabling, could probably benefit as well, to make the errors a little friendlier.

decimal-unfriendly-errors

alicjakujawa (commented):
https://github.com/JoinColony/colonyDapp/issues/3011

I was testing this "bug" on master and it wasn't happening for me 🤔

ArmandoGraterol:
@arrenv Thank you for pointing that out! I knew I was forgetting to check some places

@alicjakujawa Weird, don't know what to tell you other than to check that you are writing just a `.` with no numbers alongside it just like in Arren's screenshots or Jack's gif

arrenv:
@ArmandoGraterol Is it just on my end or are there no longer user errors showing up for the extension setup fields, only a console error and a disabled button.

dot-in-extension-installing

ArmandoGraterol:
> @ArmandoGraterol Is it just on my end or are there no longer user errors showing up for the extension setup fields, only a console error and a disabled button.
> 
> dot-in-extension-installing

Yep, the form was missing some validation from before this PR, so I added it and now it should validate as expected. Thank you Arren!

chinins (approved):
Very good! Didn't even realise there was this bug. I tried it on master and it crashed straight away but working as expected now.

arrenv (approved):
Working great, everything I could find is working as expected.

PR #3025, opened 2021-12-10
Fix transfer funds motion

### LINKED ISSUE #2789: Transfer funds motion shows estimation error when transferring funds NOT from root domain
Steps to reproduce: try a motion to transfer funds from non-root domain to root domain.

Screenshot 2021-10-14 at 16 29 24

alicjakujawa:
## Description

This PR fixes transfer funds motion. There was change on network side about which we can read more here:
https://github.com/JoinColony/colonyNetwork/blob/develop/contracts/extensions/VotingReputation.sol#L238

**Changes** 🏗

* Passing root domain id as as motion domain id (currently its the only possibility as MotionDomainSelect is disabled

**How to test** ⚰️
* Create a motion to transfer funds from child domain to root

Resolves #2789

### DIFF
--- a/src/modules/dashboard/sagas/motions/moveFundsMotion.ts
+++ b/src/modules/dashboard/sagas/motions/moveFundsMotion.ts
@@ -145,7 +145,7 @@ function* moveFundsMotion({
       methodName: 'createMotion',
       identifier: colonyAddress,
       params: [
-        fromDomainId,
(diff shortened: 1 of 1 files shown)


chinins (approved):
Short and sweet

ArmandoGraterol (approved):
Nice one! It's working as expected

PR #3043, opened 2021-12-14
Adds loading text to synaps KYC dialog

### LINKED ISSUE #2903: Coin Machine - Microcopy for Synaps Pre-Loading
### Story

As a *user who wants to get whitelisted*, I want to *understand why KYC module is loading*, so that *I can check pending transaction in my wallet*

Please include microcopy below the loading circle as on attached URL. 
Currently, someone might not get a clue that has to open metamask, especially while scrolling page down. 

Make sure to align preloader to the middle (taco). 

[Figma link](https://www.figma.com/file/fw2pDIUlkhCkJE79ZyTbjp/CoinMachine?node-id=3516%3A35286)

Check copy with Jack: "Sign transaction with your wallet" (suggested one)

image## Implementation

- [ ] The activity indicator for the `TransactionStatus` component is shown during any and all async waiting

agustif:
## Description
For testing purposes you might add the spinner loader somewhere else on the dapp with the appropriate text value. 
Btw I had to pass defaultMessage explicitly. not sure if good but is how it worked. otherwise it showed the label instead of the text.

 
- [x] Add loading text when Synaps KYC loading asks for wallet access / confirmation

**How to test:** 
In order to test this, since it's a loading state, I just placed the SpinnerLoader component in the landing page. So you can see how it looks.
`src/modules/pages/components/LandingPage/LandingPage.tsx`
```tsx
<SpinnerLoader
  loadingText="Sign transaction with your wallet"
  appearance={{ size: 'large' }}
/>
```
**New stuff** ✨

* Adds new message for loadingText in KYCDialog
* Adds new .loadingText css class with a padding-top 30px property. I double checked and it seems this is the first instance of using loadingText on SpinnerLoader, so this should not break any existing implementation and I guessued we would want such a padding in all cases of showing the loadingText props. Let me know if you don't think that's a good approach!

```diff
+  loadingText: {
+    id: 'dashboard.CoinMachine.SynapsKYCDialog.loadingText',
+    defaultMesage: 'Sign transaction with your wallet',
+  },
```
**Changes** 🏗

* Adds loadingText prop to SpinnerLoader in KYCDialog component
```diff
-          {isLoading && <SpinnerLoader appearance={{ size: 'large' }} />}
+          {isLoading && (
+            <SpinnerLoader
+              loadingText={MSG.loadingText.defaultMesage}
+              appearance={{ size: 'large' }}
+            />
+          )}
```
Resolves #2903

### DIFF
--- a/src/modules/core/components/Preloaders/SpinnerLoader.css
+++ b/src/modules/core/components/Preloaders/SpinnerLoader.css
@@ -87,3 +87,7 @@
 .layoutHorizontal .loader {
   margin: 0 10px 0 0;
 }
+
+.loadingText {
+  padding-top: 30px;
+}
--- a/src/modules/core/components/Preloaders/SpinnerLoader.css.d.ts
+++ b/src/modules/core/components/Preloaders/SpinnerLoader.css.d.ts
@@ -10,3 +10,4 @@ export const loadingTextContainer: string;
 export const themeGrey: string;
 export const themePrimary: string;
 export const layoutHorizontal: string;
+export const loadingText: string;
--- a/src/modules/core/components/Preloaders/SpinnerLoader.tsx
+++ b/src/modules/core/components/Preloaders/SpinnerLoader.tsx
@@ -34,7 +34,7 @@ const SpinnerLoader = ({
       <div className={styles.loader} />
       {loadingText && (
         <div className={styles.loadingTextContainer}>
-          <div>
+          <div className={styles.loadingText}>
             {typeof loadingText === 'string'
               ? loadingText
               : formatMessage(loadingText, textValues)}
--- a/src/modules/dashboard/components/CoinMachine/SynapsKYCDialog/SynapsKYCDialog.tsx
+++ b/src/modules/dashboard/components/CoinMachine/SynapsKYCDialog/SynapsKYCDialog.tsx
@@ -31,6 +31,10 @@ const MSG = defineMessages({
     id: 'dashboard.CoinMachine.SynapsKYCDialog.verifiedIconTitle',
     defaultMessage: 'You are verified! Congrats',
   },
+  loadingText: {
+    id: 'dashboard.CoinMachine.SynapsKYCDialog.loadingText',
+    defaultMesage: 'Sign transaction with your wallet',
+  },
 });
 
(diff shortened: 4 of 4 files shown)


chinins (changes requested):
You need more margin below the spinner (doesn't have to be precise but just close to what Karol had in his print screen)

What you have:
Screenshot 2021-12-15 at 15 47 01

What Karol is asking for:
Screenshot 2021-12-15 at 15 47 36

Moreover, could you please rebase your branch and also add how to test (and for future PRs with testing that is not so straight forward). I can find my way to it but it feels a bit of a waste of time for every person to do that if we don't remember the exact flow of the dialogs.

rdig:
What's the state of this, and why, does adding some small copy take so long ?

agustif:
> What's the state of this, and why, does adding some small copy take so long ?

Will be pushing fix for olga's requested changes shortly, and it should be good to go.

Sorry for the delay!

agustif:
Added a new loadingText css class with a padding-top of 30px; 

image

Rebasing now

agustif:
I've added a submodule, redoing history to squash all changes in one commit and undo it.

agustif:
good to re-review now!

chinins (changes requested):
This is wrong. `loadingText` prop is used in lots of places - that's why it exists. Doing a simple search in a code editor should show you that:
Screenshot 2021-12-23 at 16 30 20

You broke the UI of loading spinners with loading text, these are just 2 examples (and you can see them straight away when you reload the dapp):
loading-actions

colony-loading

The margin should be applied in the KYC dialog component and NOT on the core component.
Also, you should not use the `defaultMessage`. If you have to this means something is wrong.

agustif:
> This is wrong. `loadingText` prop is used in lots of places - that's why it exists. Doing a simple search in a code editor should show you that:
> 

> 
> 
> You broke the UI of loading spinners with loading text, these are just 2 examples (and you can see them straight away when you reload the dapp):
> 

> The margin should be applied in the KYC dialog component and NOT on the core component.
> 
> Also, you should not use the `defaultMessage`. If you have to this means something is wrong.
> 
> 

Thank you olga, will redo and fix

danbr:
Solved in a different PR.

PR #3050, opened 2021-12-15
Fix sold out conditionals of coin machine

### LINKED ISSUE #3024: Incorrect CoinMachine batch Sold out message when reloading page
## Steps to reproduce

1. Enable CoinMachine and fund it.
2. Purchase a portion of the available batch
3. Reload the page.

## Expected behaviour

* Refreshing the page should not cause the batch to be sold out when it is not.

## Actual behaviour

* Batch is incorrectly marked as sold out after reloading the page.
* Even after refreshing again, or letting the next batch start it is still marked as sold out.
* The only way to fix it is to re-fund the CoinMachine.

soldout-after-refresh-cant-fix-until-funding

I can easily replicate this in my local environment (I have not tested in QA, currently under maintenance), however, let me know if localised to me or a symptom of the local dev environment.

ArmandoGraterol:
Updated the logic that determines the `maxPeriodTokens` amount when the coin machine has fewer tokens than what's the max in a period so that it doesn't show that the period sold out when it hasn't. Also improves consistency with the rest of the batches, to be more specific, how the last batches are portrayed.

FireShot Capture 513 - Colony - localhost

Also:
- Removed unused message descriptor from RemainingTokens
- Added refresh query for coin machine's token balance data

Resolves #3024

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
+++ b/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
@@ -167,17 +167,11 @@ const CoinMachine = ({
     if (!saleTokensData || !periodTokensData || !hasSaleStarted) {
       return undefined;
     }
-    const maxPerPeriodTokens = bigNumberify(maxPerPeriod);
-    const leftAvailableTokens = bigNumberify(
-      coinMachineTokenBalanceData?.coinMachineTokenBalance || '0',
-    );
 
     return {
       decimals: saleTokensData.coinMachineSaleTokens.sellableToken.decimals,
       soldPeriodTokens: bigNumberify(activeSold),
-      maxPeriodTokens: maxPerPeriodTokens.gt(leftAvailableTokens)
-        ? leftAvailableTokens
-        : bigNumberify(maxPerPeriod),
+      maxPeriodTokens: bigNumberify(maxPerPeriod),
       targetPeriodTokens: bigNumberify(targetPerPeriod),
     };
   }, [
@@ -187,7 +181,6 @@ const CoinMachine = ({
     activeSold,
     maxPerPeriod,
     targetPerPeriod,
-    coinMachineTokenBalanceData,
   ]);
 
   const totalTokens = useMemo(() => {
--- a/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
+++ b/src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
@@ -167,11 +167,18 @@ const CoinMachine = ({
     if (!saleTokensData || !periodTokensData || !hasSaleStarted) {
       return undefined;
     }
+    const maxPerPeriodTokens = bigNumberify(maxPerPeriod);
+    const leftAvailableTokens = bigNumberify(
+      coinMachineTokenBalanceData?.coinMachineTokenBalance || '0',
+    );
+    const isPartialMaxPeriodTokens = maxPerPeriodTokens.gt(leftAvailableTokens);
 
     return {
       decimals: saleTokensData.coinMachineSaleTokens.sellableToken.decimals,
       soldPeriodTokens: bigNumberify(activeSold),
-      maxPeriodTokens: bigNumberify(maxPerPeriod),
+      maxPeriodTokens: isPartialMaxPeriodTokens
+        ? leftAvailableTokens.add(bigNumberify(activeSold))
+        : bigNumberify(maxPerPeriod),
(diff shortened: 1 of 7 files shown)


src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
```diff
@@ -170,17 +170,11 @@ const CoinMachine = ({
...
-    const leftAvailableTokens = bigNumberify(
-      coinMachineTokenBalanceData?.coinMachineTokenBalance || '0',
-    );
 
     return {
       decimals: saleTokensData.coinMachineSaleTokens.sellableToken.decimals,
       soldPeriodTokens: bigNumberify(activeSold),
-      maxPeriodTokens: maxPerPeriodTokens.gt(leftAvailableTokens)
-        ? leftAvailableTokens
-        : bigNumberify(maxPerPeriod),
```
chinins, line 183: If you remove this comparison, then if in coin machine there are less tokens than maxPerPeriod, we will still show maxPerPeriod while we should be showing what is actually left in coin machine. Does it make sense? People can't buy more than what we have in coin machine even if maxPerPeriod shows more.

chinins (changes requested):
You removed the logic to check how much tokens is left in the coin machine which should still be there (please, see my comment)

src/modules/dashboard/components/CoinMachine/CoinMachine.tsx
ArmandoGraterol: Ooof my bad, I didn't think about that case. 

I will have to tweak this logic and/or tweak the sold-out logic because this is making the coin machine assume that the tokens are sold out if the `maxPeriodTokens` is less or equal to the sold amount of tokens (Right now we are **always** assuming that `maxPeriodTokens` is the _full_ amount, without considering the possibility that it could also be what's left.) 

Tricky.

arrenv (commented):
I am getting some very strange things happening with the batches at the moment. I am going to completely restart my environment and test a bit more tomorrow.

ArmandoGraterol:
> I am getting some very strange things happening with the batches at the moment. I am going to completely restart my environment and test a bit more tomorrow.

@arrenv If I might ask, strange how?

arrenv:
> > I am getting some very strange things happening with the batches at the moment. I am going to completely restart my environment and test a bit more tomorrow.
> 
> @arrenv If I might ask, strange how? Perhaps I did something wrong :P

Sure, I will make a gif tomorrow if it is still there. The refreshing was working well. Although, with the batches I had the sold tokens per batch persist in the next batch.

ArmandoGraterol:
> Sure, I will make a gif tomorrow if it is still there. The refreshing was working well. Although, with the batches I had the sold tokens per batch persist in the next batch.

Uh yeah, that's really strange. My changes shouldn't affect that but weirder things have happened before

chinins:
There is a bug! And I know where it's coming from.

My maxPerPeriod is 4000, and it's not enough in the coin machine but it doesn't need to be because we already sold a lot in the period. So it correctly compares it to the amount left but then because it's sold quite a lot in the period it adds that amount and it becomes larger than the max per period. So the check should be if what's left in the coin machine minus what's sold in the period is greater than the reaming of the coin machine.
I hope I'm clear. In the case below it should be `3000/4000`.
Screenshot 2021-12-16 at 16 22 36

Additionally I did notice the sales persistence but for me I think it was caused my time not forwarding. So it seems that when I made a payment to coin machine the time didn't fast forward similar to when we do a payment to a user.

ArmandoGraterol:
@chinins Nice catch! I just have 1 question. The conditional you are suggesting is (left tokens - sold tokens) > left tokens? Would it not be something more like (max period tokens - sold tokens) > left tokens?

arrenv:
@ArmandoGraterol the new environment fixed the sold tokens persisting, but I am seeing the addition the same as @chinins is seeing.

For me it seems to be only for a brief second, until the query catches up.

adding-to-batch-limit

chinins:
> @chinins Nice catch! I just have 1 question. The conditional you are suggesting is (left tokens - sold tokens) > left tokens? Would it not be something more like (max period tokens - sold tokens) > left tokens?

Yes, that's what I meant :)

arrenv (approved):
Nice one @ArmandoGraterol , this turned out to be a little more work. Working as expected from all of my tests.

chinins (approved):
Now it's all good! Great idea on how to have it updated on time :)

PR #3053, opened 2021-12-16
Fix duplicate events on events page

### LINKED ISSUE #3045: Duplicate events appearing on Events page.
## Steps to reproduce

1. Go to the events page of the live MetaColony page and leave it loaded for long enough (a couple of minutes should be fine).
2. You should see events being duplicated.

## Expected behaviour

* Only one of each event should be visible on the event page.

## Actual behaviour

* Event items are being duplicated several times.
* Eventually the page crashes from too many duplicated events.

duplicate-events-appearing

duplicate-events-appearing

## Suggestions

* It may be a good idea to have an option to load more events for the user manually.
For example, sitting above the events list, if there are new events, could be a call to action with the text:
'There are 11 new events... Load events'.

example-load-more-option

agustif:
## Description

- [x] Make sure duplicate events are not possible on events list component in events page.
- Sorry team, this one is a hard to reproduce bug from metacolony, if anyone knows how to test it direclty with metacolony prod data that'd be nice..
Otherwise, we can introduce artificially duplicate streamedEvents at the new Set:
```ColonyEvents.tsx
  const {
    data,
    loading: subgraphEventsLoading,
    error,
  } = useSubgraphEventsSubscription({
    variables: {
      skip: dataPage === 1 ? 0 : ITEMS_PER_PAGE * dataPage,
      first: ITEMS_PER_PAGE,
      colonyAddress: colonyAddress.toLowerCase(),
      sortDirection: 'desc',
    },
    onSubscriptionData: ({ subscriptionData: { data: newSubscriptionData } }) =>
      setStreamedEvents([
        ...new Set([
          ...streamedEvents,
          ...streamedEvents,
          ...(newSubscriptionData?.events || []),
        ]),
      ]),
  });
```

and see if Set catches and removes 'em.

**Changes** 🏗

* Wrap current array for events into set, so it filters them by ID.

**Minor cosmetic changes** ⚰️

* setstreamedEvents uppercased to -> setStreamedEvents

## TODO

Resolves #3045

### DIFF
--- a/src/modules/dashboard/components/ColonyEvents/ColonyEvents.tsx
+++ b/src/modules/dashboard/components/ColonyEvents/ColonyEvents.tsx
@@ -44,7 +44,7 @@ const ColonyEvents = ({
 
   const [eventsSort, setEventsSort] = useState<string>(SortOptions.NEWEST);
   const [dataPage, setDataPage] = useState<number>(1);
-  const [streamedEvents, setstreamedEvents] = useState<
+  const [streamedEvents, setStreamedEvents] = useState<
     SubgraphEventsSubscription['events']
   >([]);
 
@@ -65,9 +65,8 @@ const ColonyEvents = ({
       sortDirection: 'desc',
     },
     onSubscriptionData: ({ subscriptionData: { data: newSubscriptionData } }) =>
-      setstreamedEvents([
-        ...streamedEvents,
-        ...(newSubscriptionData?.events || []),
+      setStreamedEvents([
+        ...new Set([...streamedEvents, ...(newSubscriptionData?.events || [])]),
       ]),
   });
 


ArmandoGraterol (approved):
Well, with the example you provided it seems to be working, and in theory, it should work with prod's data, so I guess it's good to go!

agustif:
> Well, with the example you provided it seems to be working, and in theory, it should work with prod's data, so I guess it's good to go!

Thanks for the input @ArmandoGraterol . Do you think I should look in adding this in Actions Page too? Doesn't think the a bug has showed there yet, but might be good for -guard railing- in the future? Will look into it

ArmandoGraterol:
> > Well, with the example you provided it seems to be working, and in theory, it should work with prod's data, so I guess it's good to go!
> 
> Thanks for the input @ArmandoGraterol . Do you think I should look in adding this in Actions Page too? Doesn't think the a bug has showed there yet, but might be good for -guard railing- in the future? Will look into it

I haven't seen that happen in the action list, so I don't think it's necessary right now but if it doesn't hurt performance then you can try adding it there too

chinins (approved):
Seems to be working. I agree with Armando that we shouldn't change anything in the ActionsList because the logic there is a bit different.

PR #3054, opened 2021-12-16
Improve UI of Claiming Tokens for a Colony

### LINKED ISSUE #3044: Improve UI of Claiming Tokens for a Colony
### Story

When a user visits the Events page they my be presented with a button to claim funds. In its current state, this could easily be misunderstood that they are claiming the funds for themselves.

It would likely be better to be a bit clearer here, clarifying that the funds are claimable for the colony.

This is evident in the current MetaColony token sale with so many different users claiming funds for the colony.

### Description

Update this section and/or claim button, with advice from Product, as to how to make it clearer that the funds are claimable for the colony and not for the individual.

### Suggestions

* The button text could simply be updated with the text 'Claim for <COLONY_NAME>', consideration would need to be made as to variable length of a colony's name though.

claim-for-colony

* Another option is that the 'Incoming funds' could be changed to 'Incoming funds for <COLONY_NAME>' with the button text changed to 'Claim for colony', which matches the copy in the event items.

arrenv:
## Description

This PR is intended to address any misunderstandings there may be with the 'Claim' button on the 'Events' page of a colony, as described in #3044.

better-claim-funds-ui

The design and changes have been raised in Product, with approval from Karol.

**Changes** 🏗

* New title for the section.
* New button text for claiming.

**Testing** 

To replicate, you will need to send funds to a colony.

- One way to test is to run a CoinMachine token sale and make some buy orders.
- Go to the events page to view the 'Claim' button appear.

Resolves #3044

### DIFF
--- a/src/modules/dashboard/components/UnclaimedTransfers/UnclaimedTransfers.tsx
+++ b/src/modules/dashboard/components/UnclaimedTransfers/UnclaimedTransfers.tsx
@@ -16,7 +16,7 @@ interface Props {
 const MSG = defineMessages({
   title: {
     id: 'dashboard.UnclaimedTransfers.title',
-    defaultMessage: 'Incoming funds',
+    defaultMessage: 'Incoming funds for {colony}',
   },
   loadingData: {
     id: 'dashboard.UnclaimedTransfers.title',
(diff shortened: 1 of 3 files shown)


alicjakujawa (approved):

chinins (approved):
Looking good!

PR #3055, opened 2021-12-16
Fix extensions page overflowing on lower resolutions

### LINKED ISSUE #3035: Extensions container breaks out of content container
## Steps to reproduce

1. Navigate to an extensions page.
2. Notice horizontal scroll and highlight the extension container

## Expected behaviour

* Extension pages should fit within its container.
* It should not cause a horizontal scroll on only extension pages.

## Actual behaviour

* Extension pages overflows out of its containers.

extensions-sidebar-overflowing

ArmandoGraterol:
Removed unused sidebar from the extensions page to stop the page from overflowing on lower resolutions. 

The page is still overflowing on even lower resolutions but changing that would require a rework of the grid that we use on a lot of pages, a rework which will probably come when we start improving the responsiveness of the dapp.

Resolves #3035

### DIFF
--- a/src/modules/dashboard/components/Extensions/Extensions.css
+++ b/src/modules/dashboard/components/Extensions/Extensions.css
@@ -23,14 +23,6 @@
   display: block;
 }
 
-/*
- * @NOTE We don't realy use the sidebar here, it's just easier this way
- * to properly align the grid with the one on the details page
- */
-.sidebar {
-  display: block;
-}
-
 .availableExtensionsWrapper {
   margin-top: 40px;
 }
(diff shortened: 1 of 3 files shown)


alicjakujawa (approved):
Yeah we are far away from being responsive 😆 but lets say its a fix xD

arrenv (approved):
Looks good to me!

I checked other extension pages and it did not seem to have any impact removing the sidebar.

### Before

remove-extensions-sidebar-before

### After

remove-extensions-sidebar-after

chinins (approved):
Good to go!

PR #3059, opened 2021-12-18
Use public submodule URLs

eugenijusr:
## Description

- [x] Point all submodules to public repo URLs.

**Changes** 🏗

* Change submodule URLs to be a public ones so that non-team members would be able to checkout and build the project. Currently this trigger errors that some remote repositories cannot be cloned.

Goes together with https://github.com/JoinColony/colonyServer/pull/148.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,9 +7,9 @@
 	url = https://github.com/JoinColony/colonyServer.git
 [submodule "src/lib/subgraph"]
 	path = src/lib/subgraph
-	url = git@github.com:JoinColony/subgraph.git
+	url = https://github.com/JoinColony/subgraph.git
 	ignore = dirty
 [submodule "src/lib/graph-node"]
 	path = src/lib/graph-node
-	url = git@github.com:graphprotocol/graph-node.git
+	url = https://github.com/graphprotocol/graph-node.git
 	ignore = dirty


rdig:
This has nothing to do with "team memebers", it's just a matter of security. All the repos used as sub-modules inside the dapp repo are public.

However, they require cloning using a ssh key. Once you add a key to your account, it will work: https://docs.github.com/en/authentication/connecting-to-github-with-ssh/generating-a-new-ssh-key-and-adding-it-to-the-ssh-agent

Closing.

eugenijusr:
@rdig Thanks for the quick feedback.

That's interesting. Can you elaborate on this a bit more? I want to understand why using ssh URL is more secure than https one? Also I noticed some submodules are using http and some ssh. Is there are particular reason for that?

The reason why I suggested this change is I'm using GitHub Codespaces to quickly spin up a dev environment (like many spontaneous contributors would do). The way a Codespace comes preconfigured is it won't fetch ssh URLs and will fail with repo could not be cloned error for submodules like these by default. Took me an hour to figure it out.

rdig:
We use 2FA for authentication on all services (dev or otherwise) we interact to, meaning, if we switch to https authentication, we'll have to all generate new access tokens and enter them manually every time we push something to the repo. That would cause a major overhead for the dev team.

This is beause, in August of this year, github removed password based authernticaton: https://github.blog/2020-12-15-token-authentication-requirements-for-git-operations/

Using ssh allows us to more easily interact with our repos.

eugenijusr:
I see, thanks. You're absolutely right - it does work without a hitch on my local machine where I have the ssh-agent setup. This is only a GitHub Codespaces issue I guess then.

```
% git submodule update --init --recursive
Submodule 'src/lib/colonyNetwork' (https://github.com/JoinColony/colonyNetwork.git) registered for path 'src/lib/colonyNetwork'
Submodule 'src/lib/colonyServer' (https://github.com/JoinColony/colonyServer.git) registered for path 'src/lib/colonyServer'
Submodule 'src/lib/graph-node' (git@github.com:graphprotocol/graph-node.git) registered for path 'src/lib/graph-node'
Submodule 'src/lib/subgraph' (git@github.com:JoinColony/subgraph.git) registered for path 'src/lib/subgraph'
```


Shouldn't the other two submodule URLs for `colonyNetwork` and `colonyServer` be changed to ssh then?

PR #3063, opened 2021-12-20
Adds lock icon to locked native token on funds page

### LINKED ISSUE #3034: Add a UI element if a token is locked
In the "funds" view, adding some UI element (such as a lock icon) on tokens which are locked would be helpful.

arrenv:
## Description

Adds an icon to represent that the native colony token is locked and a tooltip to explain it.

token-locked-icon

lock-icon-1

**Changes**

* Fixes Lock SVG icon to be more compatible, align with dapp icon standards, and allow for color changing.

**New stuff** ✨

* Adds an icon to represent that the native colony token is locked.
* Adds a tooltip to explain the locked icon.
* To account for the multiple locations a new component IconTooltip has been created. Which, can now also be used for any icon that needs a tooltip in future.
* The tooltip text has been added to the translation file for a singular reference.

## TODO

- [x] Determine if the locked icon should appear anywhere else.
- [x] If required elsewhere, create a dedicated component for it.

Resolves #3034


src/modules/core/components/IconTooltip/IconTooltip.tsx
```diff
@@ -0,0 +1,85 @@
...
+          offset: [122, 15],
+        },
+      },
+    ],
+  },
+  tooltipClassName = styles.tooltipWrapper,
+  className,
+  showArrow,
+}: Props) => (
+  <>
```
ArmandoGraterol, line 50: I think this empty bracket might not be needed

src/i18n/en.json
```diff
@@ -237,5 +237,7 @@
     "permission.3": "Architecture",
     "permission.5": "Funding",
     "permission.0": "Recovery",
-    "permission.1": "Root"
+    "permission.1": "Root",
+
+    "dashboard.lockedTokenTooltip": "This token is locked. Colony native tokens are locked and non-transferrable by default to avoid unwanted project token transfer outside of the colony."
```
ArmandoGraterol, line 242: I think this might not need to be `dashboard.` as this may be used in other components other than the ones in dashboard

src/modules/core/components/IconTooltip/IconTooltip.tsx
```diff
@@ -0,0 +1,85 @@
...
+            <div className={tooltipClassName}>
+              <FormattedMessage {...tooltipText} values={tooltipTextValues} />
+            </div>
+          )
+        }
+        trigger="hover"
+        showArrow={showArrow}
+        popperProps={tooltipPopperProps}
+      >
+        <div>
```
ArmandoGraterol, line 70: Is this div around the icon needed? I feel like it might not

ArmandoGraterol (changes requested):
Perhaps the lock could be 1px or 2px more to the bottom, especially in the funds and colony total balance display, in the `Available funds` it looks fine:
FireShot Capture 519 - Colony - localhost
FireShot Capture 520 - Colony - localhost
FireShot Capture 521 - Colony - localhost

However, with large amounts the alignment gets funky and it even breaks:
FireShot Capture 523 - Colony - localhost

This is happening below a 1400px screen size.

src/modules/core/components/IconTooltip/IconTooltip.tsx
arrenv: This is currently needed to account for the innerRef of the Popover component.

src/i18n/en.json
arrenv: I will replace with 'tooltip.lockedToken' that may make more sense for future use cases.

src/modules/core/components/IconTooltip/IconTooltip.tsx
arrenv: Yes, this can be removed now. Thank you.

src/modules/core/components/IconTooltip/IconTooltip.tsx
arrenv: I have used an empty bracket instead though.

arrenv: Actually, this does still need a tag, and not a direct component, seemingly due to the Popover component usage of innerRef.

arrenv:
@ArmandoGraterol I have made some tweaks, which should help force some line heights and make alignments more consistent.

As well as the hover consistency of the Popover.

Let me know if you notice any issues, or if we need to look at some more robust alignment solutions.

chinins (approved):
Great job! I think it looks good and it also helps the UX. As I mentioned it's a great idea to add the IconTooltip component.
If it looks right on Armando's and Raul's machines, I think we should refactor the `QuestionmarkTooltip` component to try & use the `IconTooltip` and have it consistent across the app. This should be of course a separate issue. @agustif maybe you should try this approach in your PR.
Screenshot 2021-12-22 at 17 01 42


Screenshot 2021-12-22 at 17 01 23

ArmandoGraterol (approved):
LGTM :+1:

PR #3066, opened 2021-12-20
Fixes annotation comment hover state and spacing

### LINKED ISSUE #3062: Annotation background becomes transparent on hover
See the demo:
bug-annotation-hover

The correct version is for the background to remain the same color on hover, both for the normal annotation (blue), as well as for the staking annotation (still blue), or the objection annotation (red)

Note that his was introduced when we added hover controls for comment banning, so make sure you don't break those while fixing this

arrenv:
## Description

This PR fixes the hover state of the annotation comment, as well as the spacing below annotation comment for staking or objection.

**New stuff** ✨

* Add a new prop to comments to allow hover state to be disabled.
* Adds the same container around main annotation, around staking and objection annotation.

fixes-annotation-comment-hover

### Additional

For reference, this screenshot shows the spacing issue also fixed in this update.

fix-gap-between-comments

Resolves #3062

### DIFF
--- a/src/modules/core/components/Comment/Comment.css
+++ b/src/modules/core/components/Comment/Comment.css
@@ -7,7 +7,7 @@
   border-radius: var(--radius-large);
 }
 
-.main:hover, .main.activeActions {
+.main:not(.stateDisableHover):hover, .main.stateActiveActions:not(.stateDisableHover) {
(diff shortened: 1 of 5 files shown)


chinins (approved):
Looks good!

ArmandoGraterol (approved):
LGTM :100:

PR #3067, opened 2021-12-20
Add token info popover to Funding and Token Activation

### LINKED ISSUE #3042: Easily add the token of a colony to a users wallet
### Story

A user joins a colony and wants to add a colony's tokens to their wallet. It would be great to have a button that easily adds the token to their wallet.

### Description

This functionality should already be within the Dapp as mentioned by @rdig, it should only require a button that hooks into this functionality. This will likely require a design that incorporates this functionality.

EDIT: Please see @rdig's comment below for context on current implementation.

Some location suggestions:

add-token-to-wallet

add-token-to-wallet-2

### Design 

Needs to be done.

ArmandoGraterol:
Cherry-picked commit added by Raul to extract the "Add token to metamask" button of the token info popover.

According to Product, the current implementation is good enough as it is, so this PR is mostly to add the token info popover to the token action popover and the colony funding page token card as those seen like places where the user would expect to find this popover.

Resolves #3042

### DIFF
--- a/src/modules/core/components/InfoPopover/InfoPopover.css
+++ b/src/modules/core/components/InfoPopover/InfoPopover.css
@@ -106,3 +106,7 @@
   width: 100%;
   overflow: hidden;
 }
+
+.addToWallet {
+  float: right;
+}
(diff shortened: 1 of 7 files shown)


arrenv (approved):
Great addition! Nice one, looks good to me.

add-to-metamask-1

add-to-metamask-4

add-to-metamask-2

add-to-metamask-3

chinins (approved):
Looks very good!

PR #3068, opened 2021-12-20
Add extension re-enabling functionality

### LINKED ISSUE #3061: Extensions Support Re-enabling
Currently extensions only support being deprecated, then un-installed.

When deprecated, it's technically possible (from the contracts), to re-enable said extension, which we currently in the dapp don't support.

This is quite a simple change:
- adapt the UI to allow for re-enabling
- the enable call is the same as the deprecate call, just needs to be passed `false` when calling it

chinins:
## Description

This PR adds extension re-enabling feature. It was already available on the contracts side.

The copy for the confirm dialog and the UI elements have been confirmed with Karol.

Please, note that when transaction is executed there will still be a copy:

`Deprecate Colony Extension`

This is not really possible to change due to the current implementation. But Raul is working on something that will change this implementation, and then we will be able to update a copy specifically to re-enabling extension. But for now it will stay like this.

**Changes** 🏗

- update deprecateExtension saga, action & `ExtensionDetails` component

Resolves #3061

### DIFF
--- a/src/modules/dashboard/components/Extensions/ExtensionDetails.css
+++ b/src/modules/dashboard/components/Extensions/ExtensionDetails.css
@@ -76,7 +76,9 @@
   margin-right: 0px;
 }
 
-.buttonUninstall {
+.actionButtons {
+  display: flex;
+  justify-content: flex-end;
   margin-top: 10px;
   text-align: right;
 }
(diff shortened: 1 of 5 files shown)


ArmandoGraterol (approved):
Nice! Working as expected, great job

arrenv (approved):
Great addition. Working as expected, considering your notes.

Just a comment, not required for this PR, it would be helpful to know what the parameters are that are being re-enabled. What do you think?

chinins:
> Great addition. Working as expected, considering your notes.
> 
> Just a comment, not required for this PR, it would be helpful to know what the parameters are that are being re-enabled. What do you think?

It's a good point. I was actually thinking that this should be visible even when you just check out the extension info/details. This should be a separate issue and I was waiting for the coin machine storm to calm down to raise it with the product.

Issue #2846, opened 2021-11-09
Motion events are inappropriate for what happened


thycommodore:
## Steps to reproduce

Haven't been able to replicate yet. This seems to happening to Jack (user with many transactions) in a colony with many transactions (Roadrunner).

## Expected behaviour

* Only transactions that occurred for that action should appear. 

## Actual behaviour

![image](https://user-images.githubusercontent.com/89165465/140958361-df258a04-f492-4f45-9dc9-935f8545f426.png)

* This motion was not objected to, and no vote took place, yet the event feed says otherwise.
* https://qaxdai.colony.io/colony/roadrunner/tx/0x0abb82c4f206fec195bc929a970e5eb53f60ea6fb04e61699e216f8202888f5a

rdig:
These are not events, these are system messages. Totally different things.

That being said, your bug report still seems to be valid

PR #3071, opened 2021-12-21
Fix motions using old motion events

### LINKED ISSUE #2845: Smite action "attached" to a bunch of other transactions
## Steps to reproduce

1. Go to Roadrunner https://qaxdai.colony.io/colony/roadrunner
2. Smite a user with reputation

## Expected behaviour

* User smites user, only the transactions relevant are referenced in the detail view

## Actual behaviour

image

* This smite action was taken recently, but is referencing >100 day old transactions.

Ex 1: https://qaxdai.colony.io/colony/roadrunner/tx/0x0abb82c4f206fec195bc929a970e5eb53f60ea6fb04e61699e216f8202888f5a

Ex 2: https://qaxdai.colony.io/colony/roadrunner/tx/0x6f5962024a00faaa7393911bed68d3d5fb47c4b5877c7903261f8267bd8f7e64

## Additional Notes

* This seems to be related Motions (I tried replicating on a colony w/o Motions & Disptues).

ArmandoGraterol:
Fixed the motions using events from previous instances of the voting reputation extension. The queries would use old events if the current motion id corresponds to any old motion id, so now we need to filter by the current extension's address.

In order to test, you will need to install the voting reputation extension, do some motions, and then uninstall and install the extension and do more motions and see if there are any events that don't correspond with the motions.

To see the bug in action, aside from following the instructions I wrote above in `master`, you can check the Roadrunner colony on QA: https://qaxdai.colony.io/colony/roadrunner

Resolves #2845 and #2846

### DIFF
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -2206,6 +2206,7 @@ export type SubgraphMotionEventsQuery = { motionEvents: Array<(
 export type SubgraphMotionSystemEventsQueryVariables = Exact<{
   colonyAddress: Scalars['String'];
   motionId: Scalars['String'];
+  extensionAddress: Scalars['String'];
   sortDirection?: Maybe<Scalars['String']>;
 }>;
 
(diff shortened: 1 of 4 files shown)


arrenv (approved):
Seems to be only showing the correct events now, having tested a few cycles of the steps as described with and without your updates.

Good stuff!

### Before

Should show only the annotation and the creation event.

before-motion-events


### After

Correctly shows only the annotation and the creation event.

after-motion-events

chinins (approved):
Working well! Nice job!

PR #3072, opened 2021-12-22
Refresh transactions after sign out

### LINKED ISSUE #3038: GasStation transactions persist after signing out and into another wallet
## Steps to reproduce

1. Sign into a wallet and perform actions.
2. Sign out of wallet and sign into another wallet.
3. Check the GasStation transaction history.

NOTE: Likely not a common use case, so, probably not a high priority.

## Expected behaviour

* GasStation transactions should only show wallets transactions.

## Actual behaviour

* GasStation transactions shows all recent transactions performed in browser.

history-showing-for-different-accounts

ArmandoGraterol:
Added reducer to refresh wallet's transactions after signing out.

To test it you simply need to complete some transactions, like creating a colony or an action, log out and then log in with the same or a different wallet to check if the previous transactions appear in the gas station.

Resolves #3038

### DIFF
--- a/src/modules/core/reducers/transactions.ts
+++ b/src/modules/core/reducers/transactions.ts
@@ -216,6 +216,9 @@ const coreTransactionsReducer: ReducerType<CoreTransactionsRecord> = (
       }
       return state.deleteIn([CORE_TRANSACTIONS_LIST, id]);
     }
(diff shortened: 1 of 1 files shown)


arrenv (approved):
Good fix. Works well!

chinins (approved):
Looks good!

PR #3073, opened 2021-12-23
Fixes motions not appearing in the correct domain in actions list

arrenv:
## Description

Fixes motions correctly appearing in the actions list and actions team filter. See this issue for details #3056.

**New stuff** ✨

* Adds assignment of ethDomainId to motion action item.
* Adds expectation check to actions motion transformer.

### Testing

* Ensure the 'Motions' extension is enabled.
* Create a child team.
* You will need to make sure there is funds and reputation in the team.
* Perform a motion action in the child team and ensure it finishes. You can speed up the time with the following script, this passes 3 days of time. Just run it in your terminal.
`curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[260000],"id": 1}' localhost:8545 && curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[]}' localhost:8545`
* Ensure the motion action appears in the list. It may take a second to load after the other actions load.
* Check the domain label of the action.
* Filter the actions by the team, and ensure the motion appears in the list.

### Before

actions-wrong-team-2

actions-wrong-team

### After  ✨

actions-right-team

actions-right-team-2

Resolves #3056

PR #3077, opened 2021-12-27
Fix logic for motionDomainId in Payment & Permissions dialogs

### LINKED ISSUE #3074: Manage permissions & create payment dialogs send incorrect motionDomainId
If you change from domain the motionDomainId changes in the UI but doesn't change in the formValues.

chinins:
## Description

This PR fixes the motion domain ID that is sent by the dialog to the saga.

To test it you need to create a payment & permissions motion in different domains (at least Root & another domain) by changing  From domain  (and not directly motion domain ID). This should change the motion domain id. You then need to log the values sent to the saga (on master there is another bug that doesn't display correct domain ids in the motions, this is being fixed by a separate PR #3073 

@rdig, I would like a review from you at some point as it touches the code that you were implementing (if I recall correctly). I would like to make sure that I am not misunderstanding the logic and the implementation.

Resolves #3074

### DIFF
--- a/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.tsx
+++ b/src/modules/dashboard/components/CreatePaymentDialog/CreatePaymentDialogForm.tsx
@@ -284,10 +284,7 @@ const CreatePaymentDialogForm = ({
       } else {
         setCurrentFromDomain(ROOT_DOMAIN_ID);
       }
-      if (
-        selectedMotionDomainId !== ROOT_DOMAIN_ID &&
-        selectedMotionDomainId !== fromDomainId
-      ) {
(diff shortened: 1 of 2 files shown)


arrenv (approved):
Fix is working well for me, very nice!

### Before

wrong-team-for-motion-before


### After

wrong-team-for-motion-after

ArmandoGraterol (approved):
Working as expected now! Good job

rdig (approved):
This make sense, but I haven't had time to review all cases.

Initial tests shows everything is working fine, but give it a final test before merging make sure all cases (when creating a motion in a sub-domain) work as expected

chinins:
@rdig, thank you! I've tested again and it's all working fine.

PR #3079, opened 2021-12-29
Fix apollo error in user resolver

### LINKED ISSUE #3078: Apollo error when trying to create a payment
This is coming from a query in createPayment action. The action actually goes through and it appears in the list if I close the dialog.
Screenshot 2021-12-27 at 15 50 48

chinins:
## Description

This PR adds a try catch block around initiation of voting extension client in user resolver. If we don't do it and the voting extension is not installed then we will get an apollo error when creating a payment. This is because we use a query in the create payment saga that uses this resolver.

Resolves #3078

### DIFF
--- a/src/data/resolvers/user.ts
+++ b/src/data/resolvers/user.ts
@@ -140,10 +140,16 @@ const getUserLock = async (
     ClientType.TokenLockingClient,
     colonyAddress,
   );
-  const votingReputationClient = await colonyManager.getClient(
-    ClientType.VotingReputationClient,
-    colonyAddress,
-  );
+  let votingReputationClient;
+
+  try {
+    votingReputationClient = await colonyManager.getClient(
+      ClientType.VotingReputationClient,
+      colonyAddress,
+    );
+  } catch (error) {
+    console.error(error);
+  }
 
   const userLock = await tokenLockingClient.getUserLock(
     tokenAddress,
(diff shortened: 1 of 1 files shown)


arrenv:
@chinins This seems to trigger the error always now. As in, just reloading the page causes the error.

error-with-no-action

arrenv (changes requested):
As of previous comment.

src/data/resolvers/user.ts
```diff
@@ -140,10 +140,16 @@ const getUserLock = async (
...
-  );
+  let votingReputationClient;
+
+  try {
+    votingReputationClient = await colonyManager.getClient(
+      ClientType.VotingReputationClient,
+      colonyAddress,
+    );
+  } catch (error) {
+    console.error(error);
```
arrenv, line 151: Removing the console.log should do the trick to remove the error. Otherwise, this PR fixes the main error.

arrenv (changes requested):

chinins:
> @chinins This seems to trigger the error always now. As in, just reloading the page causes the error.
> 
> error-with-no-action

Yeah, I guess it's not worth even logging it. I've changed it to `silent` error now.

arrenv (approved):
Nice, works well for me, no more errors.

ArmandoGraterol (approved):
Good job :100:

PR #3080, opened 2021-12-29
Fix annotations transaction failure

### LINKED ISSUE #2654: Transaction failure when trying to annotate the Mint Tokens action
## Steps to reproduce

1. Create a token called $UNIT
2. Mint tokens

## Expected behaviour

* If we can't handle this type of symbol, we should throw an error when creating your token symbol.

## Actual behaviour

* User is approved when creating the token symbol.

## Suggestions (optional)

* Another way to solve this quickly would be to add some copy about our expectations, but given the user is effectively stuck once they do this — we should probably prevent it from happening.

chinins:
## Description

This PR fixes the error that we have when ipfs fails to upload an annotation (both for motion & action). The action and motion both go through but because there is an error in annotation (you can't upload null, but need a string), there is no redirect.

The fix I implemented is to try again and if it failed again then just assign an empty string.

I had a problem with types when I pushed my commits from the `versionUpgrade` action so I just copy pasted the code instead of calling a helper. Hopefully will find out why it happens & fix it.

resolves #2654

### DIFF
--- a/src/modules/dashboard/sagas/actions/versionUpgrade.ts
+++ b/src/modules/dashboard/sagas/actions/versionUpgrade.ts
@@ -94,6 +94,7 @@ function* createVersionUpgradeAction({
 
     yield takeFrom(upgrade.channel, ActionTypes.TRANSACTION_SUCCEEDED);
 
+    /* need to check for annotaiton message here again because there is a TS error when pushing */
     if (annotationMessage && supportAnnotation) {
       yield put(transactionPending(annotateUpgrade.id));
 
--- a/src/modules/dashboard/sagas/actions/versionUpgrade.ts
+++ b/src/modules/dashboard/sagas/actions/versionUpgrade.ts
@@ -11,6 +11,7 @@ import {
 import { Action, ActionTypes, AllActions } from '~redux/index';
 import { putError, takeFrom, routeRedirect } from '~utils/saga/effects';
 
+import { uploadIfpsAnnotation } from '../utils';
 import {
   createTransaction,
   createTransactionChannels,
@@ -21,7 +22,6 @@ import {
   transactionPending,
   transactionAddParams,
 } from '../../../core/actionCreators';
-import { ipfsUpload } from '../../../core/sagas/ipfs';
 
 function* createVersionUpgradeAction({
   payload: { colonyAddress, colonyName, version, annotationMessage },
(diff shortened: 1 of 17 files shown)


src/modules/dashboard/sagas/utils/uploadIfpsAnnotation.ts
```diff
@@ -0,0 +1,28 @@
+import { call } from 'redux-saga/effects';
+import { ipfsUpload } from '../../../core/sagas/ipfs';
+
+export function* uploadIfpsAnnotation(annotationMessage: string) {
+  let ipfsHash: string | null = null;
```
arrenv, line 5: Nice work, bringing it all in here, much cleaner.

Is it worth wrapping in a try/catch to add further redundancy here?

arrenv (approved):
This one is a difficult to replicate, unless you have a super fast machine it seems.

However, these updates seem to work well and makes sense. I have added some notes, one which fixes the typing issue you were having in `versionUpgrade`.

src/modules/dashboard/sagas/utils/uploadIfpsAnnotation.ts
chinins: Not really, ipfs just fails without an error so it won't do anything.

src/modules/dashboard/sagas/actions/versionUpgrade.ts
```diff
@@ -96,14 +97,28 @@ function* createVersionUpgradeAction({
     if (supportAnnotation) {
```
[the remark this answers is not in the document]
chinins: Yeah, I was thinking that though a bit redundant it might fix it and it did. Updated now, plus added the comment to track it.

alicjakujawa (commented):
I feel that we could just put call to ipfsUpload in try/catch block and assign ipfsHash to empty string in catch block, not sure if all of this here is needed

chinins:
> I feel that we could just put call to ipfsUpload in try/catch block and assign ipfsHash to empty string in catch block, not sure if all of this here is needed

`ipfsUpload` already uses try-catch block inside and we have it in the saga itself. This is to guard when it fails silently. It's hard to replicate but when it does we have an error in a different transaction because we can't use null as an argument, only a string.

ArmandoGraterol (approved):
I haven't been able to replicate the issue, however, the annotations work as expected and the changes made make sense to me so it's good to go!

PR #3082, opened 2021-12-30
Add Unlock Token Action Page

### LINKED ISSUE #2817: Add Unlock Token Action Page
### * SEE BRIEF NEXT COMMENT DOWN

Outline:
- add action types
- add event to static map
- fetch correct data inside `colonyAction` resolver, from `getActionValues`
- pass down correct values to the message descriptor in the `DefaultAction` component
- add event to `ActionsPageFeed`

arrenv:
## Description

This adds the Token Unlock Action page for the TokenUnlocked event.

All copy used has been confirmed by product. It uses the existing action page component.

**New stuff** ✨

* New descriptors added to describe the action and event
* Action Type label added as well as the action type emoji icon
* Management of action values, despite there being none for the TokenUnlocked event
* The contract event does not record any information about the initiator, so, the annotation is queried using the linked transaction hash

unlock-token-action-page-built

## Testing

NOTE: The SubGraph updates are now included in this PR. So, you will likely need to run provisions again and restart your dev environment to include them and test this PR.

Previous Instructions
* Given that this requires the TokenUnlocked event to be indexed by the SubGraph and for that we need [subgraph#36](https://github.com/JoinColony/subgraph/pull/36) to be merged first.
* So, the only way to test this beforehand is to make the changes from that PR to your SubGraph submodule in your local dev environment.
* Once you have that update you will be able to just create a standard action, which should redirect to this new page.

## TODO

- [x] Add a redirect to the action modal to the action page from the action saga
- [ ] Fix annotations on Token Unlock action page not showing, due to no initiator. **UPDATE** separate issue created #3085.

NOTE: Due to the annotations currently requiring an initiator to act as a type of spam protection (see #2407 Fix: `getAnnotationFromSubgraph address filtering) and the TokenUnlocked() contract event not passing any parameters, it is not currently possible to display annotations on the Unlock Token Action page. I have created a separate issue for this #3085.

Resolves #2817

### DIFF
--- a/src/i18n/en-actions.ts
+++ b/src/i18n/en-actions.ts
@@ -9,6 +9,7 @@ const actionsMessageDescriptors = {
       ${ColonyMotions.PaymentMotion} {Pay {recipient} {amount} {tokenSymbol}}
       ${ColonyActions.MoveFunds} {Move {amount} {tokenSymbol} from {fromDomain} to {toDomain}}
       ${ColonyMotions.MoveFundsMotion} {Move {amount} {tokenSymbol} from {fromDomain} to {toDomain}}
+      ${ColonyActions.UnlockToken} {Unlock native token {tokenSymbol}}
       ${ColonyActions.MintTokens} {Mint {amount} {tokenSymbol}}
       ${ColonyMotions.MintTokensMotion} {Mint {amount} {tokenSymbol}}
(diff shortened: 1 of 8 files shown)


chinins (approved):
Good, working as expected!

danbr (approved):
All good. Unlock Token Action page works for me.

Screenshot 2022-01-10 at 15 50 33

PR #3083, opened 2022-01-03
Fix token activation confirm check when using max button

### LINKED ISSUE #3064: When clicking Max on staking tokens, confirm gets disabled.
## Steps to reproduce

1. Mint tokens.
2. Pay yourself 7777
3. Try to activate tokens, clicking button Max 
4. You should see as in the screenshot: 7699.22999
5. Confirm button should be disabled
6. Removing last decimal nine, makes confirm get reenabled
 
image

## Expected behaviour
* When you click max, the correct amount should be inputted, so you can confirm just after that.

## Actual behaviour

* Clicking Max, should not disable the confirm button (Im guessing some gas calculcations or something are going awry with the decimals, but not sure)

chinins:
## Description

This PR fixes the situation in token activation popover when we press max button. The confirm button was disabled although the value was correct. This happens because the `toFinite` method rounds up the value in the check and the value appears to be larger than it is and so fails the disabled check.

In the current fix I have removed the `toFinite` from the check (it doesn't seem appropriate to use it here). I tried to think/remember what problem it was solving or try to find bugs/problems with my solution but couldn't. If somebody knows why `toFinite` method was added, I would then know which problem I should fix if we can't use it.

The reason why `toFinite` is not working:
the `amount` comes as a string with lots of numbers after decimals and in this case `toFinite` cuts it only to 2 decimal places (often rounding up). There is no option in this method to cut to the specific number of places.

Resolves #3064

### DIFF
--- a/src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
+++ b/src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
@@ -247,7 +247,7 @@ const ChangeTokenStateForm = ({
               disabled={
                 !isValid ||
                 values.amount === undefined ||
-                new Decimal(unformattedTokenBalance).lt(values.amount)
+                new Decimal(unformattedTokenBalance).lt(values.amount || '0')
               }
             />
           </div>
--- a/src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
+++ b/src/modules/users/components/TokenActivation/TokenActivationContent/ChangeTokenStateForm.tsx
@@ -247,7 +247,7 @@ const ChangeTokenStateForm = ({
               disabled={
                 !isValid ||
                 values.amount === undefined ||
-                new Decimal(unformattedTokenBalance).lt(toFinite(values.amount))
+                new Decimal(unformattedTokenBalance).lt(values.amount)
               }
             />
           </div>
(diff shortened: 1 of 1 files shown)


arrenv (changes requested):
This fixes the max button, however reintroduces a previous issue that does not allow a decimal to be interpreted as a number, see https://github.com/JoinColony/colonyDapp/pull/3023, or simply add a "." into the input after removing toFinite.


I think the bug is actually caused by the same issue regarding the decimals I was talking about. The `unformattedTokenBalance` function uses `import moveDecimal from 'move-decimal-point'` which, when removed and tested with a more manual divide, calculates the max without the incorrect decimals and without causing the button to be disabled.

Here is a demo showing the incorrect decimals:
after-fix-max-button

chinins:
> This fixes the max button, however reintroduces a previous issue that does not allow a decimal to be interpreted as a number, see #3023, or simply add a "." into the input after removing toFinite.
> 
> I think the bug is actually caused by the same issue regarding the decimals I was talking about. The `unformattedTokenBalance` function uses `import moveDecimal from 'move-decimal-point'` which, when removed and tested with a more manual divide, calculates the max without the incorrect decimals and without causing the button to be disabled.

Thank you for the link to the PR! I now understand what was the initial problem I need to solve.
I thought that it was moveDecimal initially but it is not. The problem is in `toFinite` rounding up and then the value being more than it actually is (as I mentioned in my PR description). So the disabled state is triggered by it although the value is correct.

I will look into how to solve it differently... Thank you again for pointing to the initial PR that introduced `toFinite`

arrenv:
@chinins you're welcome. Also, thank you for clarifying.

chinins:
@arrenv, I've added some updates. Can you please have a look and test and tell me what you think? Thank you!

arrenv (approved):
This is now working on my end without any issues.

I tried a few different options previously as well, however I like this one the best so far.

Issue #3086, opened 2022-01-04
Unlock tokens & mint tokens don't appear in the manage funds dialog menu if the user doesn't have permissions


chinins:
## Description

These items should appear when it's possible to unlock the token and mint tokens in general (disregarding if the user has permissions or not). They should just appear with a warning sign and be disabled if the user doesn't have necessary permissions & the voting extensions in not enabled.

It is especially important to see them when voting extension is enabled. Otherwise only users with permissions can raise motions (which shouldn't be the case)

Current Manage funds (voting extension enabled, user doesn't have permissions):

<img width="715" alt="Screenshot 2022-01-04 at 15 30 12" src="https://user-images.githubusercontent.com/34057551/148120043-52612f23-0469-4f3c-b7ae-c40d987cd5fc.png">

chinins:
Duplicate.

PR #3087, opened 2022-01-05
Add motion support in unlock token modal & root motion saga

### LINKED ISSUE #2818: Update Unlock Token Modal to support creating a Motion
### * SEE BRIEF NEXT COMMENT DOWN

Outline:
- check for extension installed
- add forced switch
- reputation check and errors
- refactor transformer if needed
- logic to send different redux action (if motion)

### LINKED ISSUE #2819: Add Unlock Token Motion Saga
_This is still in progress of being spec'd out._

Outline:
- add unlock token motion saga redux actions
- add unlock token motion saga redix actions types
- add unlock token motion saga

chinins:
## Description

This PR adds motion support to Unlock token modal similar to other dialogs.

Right now there is a bug that doesn't allow mint tokens or unlock token item to be seen in the Manage funds dialog menu even if voting extension is installed. So to test you need to remove the filtered part from the `ManageFunds` dialog. I've already raised an issue and am trying to fix this bug: #3086 

The testing should be done with a user that has necessary permissions & the one who doesn't with voting extension enabled and without.

Resolves #2818 
resolves #2819

### DIFF
--- a/src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenDialog.tsx
+++ b/src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenDialog.tsx
@@ -1,53 +1,66 @@
 import { FormikProps } from 'formik';
-import React, { useCallback } from 'react';
+import React, { useCallback, useState } from 'react';
 import { useHistory } from 'react-router-dom';
+import * as yup from 'yup';
 
-import Dialog, { DialogProps } from '~core/Dialog';
+import Dialog, { ActionDialogProps, DialogProps } from '~core/Dialog';
 import { ActionForm } from '~core/Fields';
 import { ActionTypes } from '~redux/index';
-import {
-  pipe,
-  mergePayload,
-  withMeta,
-  withKey,
-  mapPayload,
-} from '~utils/actions';
-import { Colony } from '~data/index';
+import { RootMotionOperationNames } from '~redux/types/actions';
+import { pipe, withMeta, withKey, mapPayload } from '~utils/actions';
 import { WizardDialogType } from '~utils/hooks';
 
 import UnlockTokenForm from './UnlockTokenForm';
 
-interface CustomWizardDialogProps {
-  prevStep?: string;
-  colony: Colony;
+export interface FormValues {
+  forceAction: boolean;
+  annotation: string;
 }
 
 type Props = DialogProps &
   Partial<WizardDialogType<object>> &
-  CustomWizardDialogProps;
+  ActionDialogProps;
 
 const displayName = 'dashboard.UnlockTokenDialog';
+
+const validationSchema = yup.object().shape({
+  forceAction: yup.bool(),
+  annotation: yup.string().max(4000),
+});
 
 const UnlockTokenDialog = ({
   colony: { colonyAddress, colonyName },
   colony,
+  isVotingExtensionEnabled,
   cancel,
   close,
   callStep,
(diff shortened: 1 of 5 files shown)


arrenv (changes requested):
Nice one @chinins. There are just a couple of things to add.

- Add a `NotEnoughReputation` reputation check.
- Update the copy used on the annotation description "Explain why you're unlocking the native token (optional)", which was recently approved by product.

chinins:
> Nice one @chinins. There are just a couple of things to add.
> 
> * Add a `NotEnoughReputation` reputation check.
> * Update the copy used on the annotation description "Explain why you're unlocking the native token (optional)", which was recently approved by product.

Good catch - thank you! All done now.

src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenDialog.tsx
```diff
@@ -55,23 +69,37 @@ const UnlockTokenDialog = ({
...
-          <UnlockTokenForm
-            {...formValues}
-            colony={colony}
-            back={prevStep && callStep ? () => callStep(prevStep) : undefined}
-          />
-        </Dialog>
-      )}
+      {(formValues: FormikProps<FormValues>) => {
+        if (formValues.values.forceAction !== isForce) {
+          setIsForce(formValues.values.forceAction);
```
arrenv, line 89: This is an error that seems to be with all `force` toggle switches. It probably needs to be wrapped in an effect. Do you think this is worth creating another issue for?

formik-setState-error

arrenv (approved):
Working great, nice one!!

**Forcing**

force-unlock-token

**Without reputation and with reputation**

with-without-reputation-unlock-token-motion

src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenDialog.tsx
chinins: It's not an error, it's just a warning and it's present in other dialogs as well (as per example) as they are all done in a similar way
Screenshot 2022-01-08 at 12 26 44
 
It might be worth creating an issue but with a low priority as it's only a warning.

arrenv: Thanks @chinins

src/modules/dashboard/components/UnlockTokenDialog/UnlockTokenForm.tsx
```diff
@@ -6,18 +6,24 @@ import { ColonyRole } from '@colony/colony-js';
...
+import { useLoggedInUser } from '~data/index';
+import { useTransformer } from '~utils/hooks';
+import { useDialogActionPermissions } from '~utils/hooks/useDialogActionPermissions';
 
 import { getAllUserRoles } from '../../../transformers';
 import { hasRoot } from '../../../users/checks';
 
 import styles from './UnlockTokenForm.css';
 import { Annotations } from '~core/Fields';
+import { ActionDialogProps } from '~core/Dialog';
```
ArmandoGraterol, line 26: Nitpick but this import should be grouped with the other relative imports :)

ArmandoGraterol (approved):
Good to go! About the bug you mentioned, #2985 should fix that. 

I haven't been able to merge because I would like another review. It's really tempting to merge it now considering that it's now affecting this feature.

PR #3088, opened 2022-01-05
Allow mintTokens & unlockToken motions without permissions

### LINKED ISSUE #3086: Unlock tokens & mint tokens don't appear in the manage funds dialog menu if the user doesn't have permissions

chinins:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

resolves #3086

### DIFF
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -405,14 +405,23 @@ export const colonyResolvers = ({
       }
     },
     async canMintNativeToken({ colonyAddress }) {
-      const colonyClient = await colonyManager.getClient(
-        ClientType.ColonyClient,
-        colonyAddress,
+      // const colonyClient = await colonyManager.getClient(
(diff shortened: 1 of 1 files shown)


chinins:
@ArmandoGraterol is already fixing it in a different PR.

PR #3090, opened 2022-01-06
Add token icon to details widget in DefaultAction & DefaultMotion

### LINKED ISSUE #2823: Add token's icon to token value field on the action/motion page
Currently, when we show a token value in the details widget (Like for example, when we make a payment, mint tokens, etc) we only show the amount with the short name of the token:

FireShot Capture 458 - Colony - localhost

While according to the specs we have, we should also show the token's icon/avatar:

FireShot Capture 459 - Colony Simplified – Figma - www figma com

In order to do this, you could add the `HookedTokenIcon` component inside the `DefaultAction` and `DefaultMotion`, to be more specific, you have to add it to the `actionAndEventValues` object that both pages have and make sure the `DetailsWidget` component renders it correctly.

**NOTE:** Both the action _and_ motion pages should have this change so after testing that it looks good on the action page, make sure that you have reputation on the colony or domain you want to test in, install the `Motions & Disputes` extension, and create a motion.

For an example of `HookedTokenIcon` being used, check out the `TokenInfo` component.

The specs in question: https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=2343%3A29658

danbr:
## Description

This PR adds token icon in detailsWidget when amount is shown.

**Changes** 🏗

- Update TokenInfoQuery
- Update DetailsWidget, TokenIcon, DefaultAction, DefaultMotion components.

resolves #2823 

Using Figma spec with an icon size of 16px. (looks rather small)
Screenshot 2022-01-06 at 17 45 36

### DIFF
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -1701,7 +1701,7 @@ export type TokenInfoQueryVariables = Exact<{
 }>;
 
 
-export type TokenInfoQuery = { tokenInfo: Pick<TokenInfo, 'decimals' | 'name' | 'symbol' | 'iconHash'> };
+export type TokenInfoQuery = { tokenInfo: Pick<TokenInfo, 'decimals' | 'name' | 'symbol' | 'iconHash' | 'address'> };
 
 export type UserNotificationsQueryVariables = Exact<{
   address: Scalars['String'];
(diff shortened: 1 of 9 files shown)


chinins (approved):
Looks good, сongrats on your first official PR!

arrenv (approved):
Nice addition. Looks good!

actions-page

motions-page

PR #3093, opened 2022-01-11
Filter duplicated actions for passed motions

### LINKED ISSUE #3070: Motions within team are being presented as Forced
## Steps to reproduce

1. Submit an expenditure as a motion from within a team
2. Finalize the motion
3. View the All Teams Action view

## Expected behaviour

* The expenditure motion is shown as a successfully executed motion

## Actual behaviour

* The expenditure motion appears at the root level
* There's an additional forced expenditure at the team level

image

https://xdai.colony.io/colony/test-kh

---

Shapeshift first reported this. In their situation, there are only two users that have the permissions to force an expenditure (willy & diggy smalls). Here's their report:

> We're also noticing that recently passed motions are showing up as forced in the Actions list.
> 
> Example of passed motion being shown as forced: https://xdai.colony.io/colony/shapeshift/tx/0xfb5c700dace17487cfd2da5d5b5b558a4fcdc7efe04552e3f18b41c221aaf20f
> 
> Link to motion: https://xdai.colony.io/colony/shapeshift/tx/0x944f762154e19fa30ba971c4d976ca24bcd84e8d93b07581c8c9081f6c645733

arrenv:
## Description

This PR is intended to remove the duplicate actions appearing for a passed motion and the incorrectly displayed 'Forced' tag on a passed motion action.

However, due to non-persistence of an uninstalled extension, the forced tag fix will only work for the currently installed Motions & Disputes extension.

The filter will only take effect once the motions have loaded into the actions list.

**These are the circumstances that have been tested and the expected results:**

- Motions & Disputes extension is enabled - _It should correctly only show the passed motion action_
- Motions & Disputes extension is depreciated - _It should correctly only show the passed motion action_
- Motions & Disputes extension is depreciated then re-enabled - _It should correctly only show the passed motion action_

- Motions & Disputes extension is uninstalled - _It will show the action with no tag_

- Motions & Disputes extension is re-installed - _It will show previous extension's motions as the singular action with a forced tag and correctly only show the passed motion action for the newly installed extension_

- Motions & Disputes extension is re-installed then depreciated - _It will show previous extension motions as the action with no tag and only show the passed motion action for the most recently depreciated extension_

**Changes** 🏗

* Updated the actions filter to also filter out the actions initiated by the Motions & Disputes extension

### Testing

- Install the Motions & Disputes extension
- Add Reputation to the colony by making a forced payment
- Refresh the page to ensure reputation is loaded
- Unlock the received funds, ready for staking against motion
- Create a motion and stake for it
- Speed up the chain by at least the time left in the motion. You can do this using this request in your console (which is a little over 3 days)
`curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[260000],"id": 1}' localhost:8545 && curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[]}' localhost:8545`
- Check your 'Actions' list to ensure that there is no duplicate for the same action.

### Before
repeated-event-incorrectly-forced

### After
correctly-filtered-motion-action

Resolves #3070

### DIFF
--- a/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
+++ b/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
@@ -24,6 +24,7 @@ import {
 import { SortOptions, SortSelectOptions } from '../shared/sortOptions';
 import { getActionsListData } from '../../transformers';
 import { useTransformer } from '~utils/hooks';
+import { createAddress } from '~utils/web3';
 import {
   ColonyActions as ColonyActionTypes,
   FormattedAction,
@@ -31,7 +32,6 @@ import {
 } from '~types/index';
 
 import styles from './ColonyActions.css';
-import { createAddress } from '~utils/web3';
 
 const MSG = defineMessages({
   labelFilter: {
@@ -175,8 +175,7 @@ const ColonyActions = ({
 
   /* Needs to be tested when all action types are wirde up & reflected in the list */
   const filteredActions = useMemo(() => {
-    let filterActions = actions;
-    filterActions = !ethDomainId
+    const filterActions = !ethDomainId
       ? actions
       : actions.filter(
           (action) =>
@@ -195,7 +194,7 @@ const ColonyActions = ({
           .filter((motion) => motion.extensionAddress)
           .map((motion) => createAddress(motion.extensionAddress)) || [];
       if (motionExtensionAddresses.length > 0) {
-        filterActions = actions.filter((action) => {
+        return filterActions.filter((action) => {
           return !motionExtensionAddresses.includes(
             createAddress(action.initiator),
           );
--- a/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
+++ b/src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
@@ -31,6 +31,7 @@ import {
 } from '~types/index';
 
 import styles from './ColonyActions.css';
+import { createAddress } from '~utils/web3';
 
 const MSG = defineMessages({
   labelFilter: {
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -31,6 +31,7 @@ import {
 } from '~types/index';
 
 import styles from './ColonyActions.css';
+import { createAddress } from '~utils/web3';
```
ArmandoGraterol, line 34: A bit of a nitpick but this import should be grouped with the other relative imports

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -173,21 +174,37 @@ const ColonyActions = ({
...
-              (ethDomainId === 1 && action.fromDomain === undefined) ||
-              /* when transfering funds the list shows both sender & recipient */
-              (action.actionType === ColonyActionTypes.MoveFunds &&
-                Number(action.toDomain) === ethDomainId),
-          ),
-    [ethDomainId, actions],
-  );
+  const filteredActions = useMemo(() => {
+    let filterActions = actions;
+    filterActions = !ethDomainId
```
ArmandoGraterol, line 179: Here when `!ethDomainId` is `true` we already assign `actions` to `filterActions`, so I think the assignation above is not needed

ArmandoGraterol (approved):
Tested it and it seems to be working properly :ok_hand:

Pointed out some small things I noticed in the code but I'm approving it anyway

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -173,21 +174,37 @@ const ColonyActions = ({
...
-              /* when no specific domain in the action it is displayed in Root */
-              (ethDomainId === 1 && action.fromDomain === undefined) ||
-              /* when transfering funds the list shows both sender & recipient */
-              (action.actionType === ColonyActionTypes.MoveFunds &&
-                Number(action.toDomain) === ethDomainId),
-          ),
-    [ethDomainId, actions],
-  );
+  const filteredActions = useMemo(() => {
+    let filterActions = actions;
```
alicjakujawa, line 178: I would avoid let whenever possible, just do const filterActions = [and condition here] with ternary operator

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
```diff
@@ -173,21 +174,37 @@ const ColonyActions = ({
...
+      const motionExtensionAddresses =
+        motions.motions
+          .filter((motion) => motion.extensionAddress)
+          .map((motion) => createAddress(motion.extensionAddress)) || [];
+      if (motionExtensionAddresses.length > 0) {
+        filterActions = actions.filter((action) => {
+          return !motionExtensionAddresses.includes(
+            createAddress(action.initiator),
+          );
+        });
```
alicjakujawa, line 202: so when you get rid of let, you can just put return here in if

alicjakujawa (changes requested):
Small comments

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
arrenv: Good pickup, a symptom of using auto-complete on function calls.

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
arrenv: Yeah, I agree, I will update.

src/modules/dashboard/components/ColonyActions/ColonyActions.tsx
arrenv: Thanks you, that helps!

chinins (approved):
Good to go! I tested it extensively with uninstalling extensions etc, and it works well.

PR #3096, opened 2022-01-12
Handle long strings & URLs in comments

### LINKED ISSUE #3048: Better way to handle long strings in comments
## Description

To fix the broken comment boxes, due to large strings, comments have been wrapped which currently breaks words within comments.

## Expected behaviour

* Whole words should be visible.
* Long strings should be accomodated.

## Actual behaviour

* Words are currently broken within the comments.

### Current

broken-words-2

### Previously

broken-words

## Suggestion

* We could break by whole words and break strings up that are over a certain number of characters

danbr:
**Description**

- This PR handles the wrapping of long words and URLs within comments. As `Comment` core component is updated, the fix should be dapp wide. The fix maintains the page layout.

Resolves #3048 

**After Fix**

CoinMachine
Screenshot 2022-01-12 at 17 44 20

Screenshot 2022-01-12 at 17 43 54

From Action Page
Screenshot 2022-01-12 at 17 54 39

Screenshot 2022-01-12 at 17 56 53

**Before Fix**

Screenshot 2022-01-12 at 14 14 40

Screenshot 2022-01-12 at 14 14 33

A bug was discovered in the delete comment dialog, with URLs overflowing.

CoinMachine 
Screenshot 2022-01-12 at 14 05 21

From Action Page
Screenshot 2022-01-12 at 17 58 14

### DIFF
--- a/src/modules/core/components/Comment/Comment.css
+++ b/src/modules/core/components/Comment/Comment.css
@@ -5,6 +5,7 @@
   padding: 11px 0 11px 0;
   width: 100%;
   border-radius: var(--radius-large);
+  word-break: break-word;
 }
 
 .main:not(.stateDisableHover):hover, .main.stateActiveActions:not(.stateDisableHover) {


danbr:
Is there a reason why we dont (attempt to) make URLs in comments hyperlinks? Clickable? 
Should it be a feature?

alicjakujawa (approved):

arrenv:
> Is there a reason why we dont (attempt to) make URLs in comments hyperlinks? Clickable? Should it be a feature?

Personally, I think having hyperlinks away from a token sales page is not desirable.

arrenv (approved):
Simple and works well!

chinins (approved):
It seems to be working well in all cases. However, there might be some problem that I don't recall so it would be good to have @rdig approval before merging.

rdig (approved):
Looks good to me. At least for this change, there isn't anything else, as you covered all the cases _(including comments in modals)_

Nice work

PR #3102, opened 2022-01-13
Vertically align user avatar in notifications

### LINKED ISSUE #3098: Improve alignment of avatar in notifications popover
## Steps to reproduce

1. Create a first transaction in the dapp
2. Check the notifications popover
3. Notice the misalignment of the icon in the notifications

## Expected behaviour

* Icon should be vertically centered with the text

See spec here - https://www.figma.com/file/OD53CoZpRKVqDdKsMYGldko8/Inbox?node-id=194%3A0

## Actual behaviour

* Icon is aligned below the inline text.

alignment-of-avatar-notification

dit7ya:
## Description

Fixed the vertical alignment of the user avatar in the notifications popup. The centering is with respect to `cap-height` (height of the first `Y` here) of the text. I can manually force it to center according to the perceived height but I personally don't like that.

**Changes** 🏗

Screen Shot 2022-02-07 at 13 33 21

Resolves  #3098.

### DIFF
--- a/src/modules/users/components/Inbox/InboxItem/InboxItem.css
+++ b/src/modules/users/components/Inbox/InboxItem/InboxItem.css
@@ -29,11 +29,6 @@
 }
 
 .inboxAction {
-  /*
-   * @NOTE Netative margins used to better align
-   * the event's copy with the user's avatar
-   */
-  margin-top: -5px;
   overflow: hidden;
   font-weight: 400;
   text-overflow: ellipsis;
@@ -115,4 +110,5 @@
 .avatarWrapper {
   display: inline-block;
   margin-right: 10px;
+  vertical-align: 8%;
 }


arrenv (changes requested):
@dit7ya the icon is looking better. I seem to have a misalignment on the username now though. Do you get the same?

### Before
cc43342a-ce25-4c31-896a-cd468fce66a3

### After

download (12)

dit7ya:
Thanks for pointing this out @arrenv. Seems like this behavior is specific to Chrome and that's why I missed it. I should really check on Chrome before submitting PRs from now on. I decided to manually adjust the avatar icon's `vertical-align` instead cause otherwise too many things need to be changed. Let me know what you think.

arrenv (approved):
@dit7ya the manual vertical alignment is unnatural to me as it can only be done by eye, however, it does look better to me.

src/modules/users/components/Inbox/InboxItem/InboxItem.css
```diff
@@ -115,4 +110,5 @@
 .avatarWrapper {
   display: inline-block;
   margin-right: 10px;
+  vertical-align: 8%;
```
rdig, line 113: Why is this percentage based ? The spacing between top and left margins will always be the same no matter the height of the notification, or am I missing something

rdig (approved):
Left a review note, other than that I'm good with it.

Also, a note for the future, we only support Chrome, so please try to do your development and testing on it, even if your main browser is Firefox.

src/modules/users/components/Inbox/InboxItem/InboxItem.css
dit7ya: No special reason, I just wanted to avoid writing a fractional `px`. Hope that's okay.

PR #3106, opened 2022-01-14
Add dialogs directory to dashboard & move all Dialog components

### LINKED ISSUE #2776: Add `Dialog` folder to `dashboard` & move all dialog components there - low priority
Raul's suggestion:
"This is not something we should do right now, but at some point, I suggest we move all dialog components under something like: ~dashboard/Dialogs/... and then we can remove the Dialog part from the component's name."

danbr:
## Description

- This PR Adds a dialog directory to dashboard.
- Dialog components moved to dialogs directory.
- Imports updated to reflect the new paths.

- Whitelist directory was not included in change.

**Testing**

I ran colonyDapp, opened many dialogs. All functioned as expected.

Resolves #2776 

Screenshot 2022-01-14 at 15 46 01


src/modules/dashboard/components/dialogs/AdvancedDialog/AdvancedDialog.tsx
```diff
@@ -8,12 +8,12 @@ import IndexModal from '~core/IndexModal';
...
 import { useLoggedInUser, Colony } from '~data/index';
 
-import { getAllUserRoles } from '../../../transformers';
+import { getAllUserRoles } from '../../../../transformers';
 import {
   canEnterRecoveryMode,
   hasRoot,
   canArchitect,
-} from '../../../users/checks';
+} from '../../../../users/checks';
```
alicjakujawa, line 16: Wonder if we could also in this PR create aliases for paths, to avoid any mistake in "../../.." i know its not part of scope in this issue, but it could reduce risk of bad imports, what you think?

alicjakujawa (approved):
Approving, but added one comment to consider

arrenv (approved):
All seems to be working well for me.

Was there any discussions regarding updating the name of the components to remove "Dialog", as of the raised issue?

src/modules/dashboard/components/dialogs/AdvancedDialog/AdvancedDialog.tsx
danbr: It's a good idea. Think it will be cleaner. I will add.

danbr:
> I will check with @rdig, if 'Dialog' should be dropped from the names of the components. Just to be sure!!

ArmandoGraterol (commented):
Hmm, I think the dialogs folder should be inside the `/components` folder and capitalized like the other components, what do you guys think?

Also, if they are inside a `Dialogs` folder, the components inside it might not need to have the `...Dialog` at the end of every component's name, like the issue's description mentions, so you could remove it.

src/modules/dashboard/components/dialogs/AdvancedDialog/AdvancedDialog.tsx
danbr: I have added 3 new aliases for dialogs, checks, and transformers.

danbr:
> Hmm, I think the `dialogs` folder should be inside the `/components` folder and capitalized like the other components, what do you guys think?
> 
> Also, if they are inside a `Dialogs` folder, the components inside it might not need to have the `...Dialog` at the end of every component's name, like the issue's description mentions, so you could remove it.

I spoke with @rdig about both issues.
For filename: It could get a little confusing without 'Dialog' in the component name, so we should keep.
For directory name: It should be capitalised.

I have added an alias for Dialogs, and another 'modules' to help shorten & simplify relative paths for some imports, as suggested by @alicjakujawa.

src/modules/dashboard/components/dialogs/AdvancedDialog/AdvancedDialog.tsx
danbr: I have added an alias for Dialogs, and another 'modules' to help shorten & simplify relative paths for some imports, as suggested by you.

ArmandoGraterol (approved):
Nice changes! Working as expected

src/modules/core/components/Comment/Comment.tsx
```diff
@@ -13,8 +13,8 @@ import HookedUserAvatar from '~users/HookedUserAvatar';
 import { AnyUser, Colony, useLoggedInUser } from '~data/index';
 import TextDecorator from '~lib/TextDecorator';
 
-import { userHasRole } from '../../../users/checks';
-import { getUserRolesForDomain } from '../../../transformers';
+import { userHasRole } from '~modules/users/checks';
```
chinins, line 16: You should remove the empty line between `~lib` import and your 2 modules imports.
Please, do this for all the files. We follow the following logic generally:
- external libraries
- our modules that have aliases
- relative imports
Each of this import section is separated by an empty line. Additionally, within relative imports different levels are also separated by an empty line.

This way it will be really clean :)

chinins (changes requested):
It all looks good and works well but I would like you to remove/add empty lines where necessary as per my separate comment.

Other than that - nice work, must be quite tedious making changes in so many files.

chinins (approved):
Looks better - good to go!

danbr:
> Hmm, I think the `dialogs` folder should be inside the `/components` folder and capitalized like the other components, what do you guys think?
> 
> Also, if they are inside a `Dialogs` folder, the components inside it might not need to have the `...Dialog` at the end of every component's name, like the issue's description mentions, so you could remove it.

I spoke with @rdig about both issues.
For filename: It could get a little confusing without 'Dialog' in the component name, so we will keep.
For directory name: It should be capitalised.

I have added an alias for Dialogs, and another 'modules' to help  shorten & simplify relative paths for some imports, as suggested by @alicjakujawa.

PR #3111, opened 2022-01-17
Align and resize GasStation left caret icon

### LINKED ISSUE #3103: Fix alignment and size of left caret in GasStation
## Steps to reproduce

1. Create an action and wait for the GasStation to appear (as of the provided screenshots).
2. Take note of the alignment of the left caret size and alignment relative to the inline text.

## Expected behaviour

* The left caret should match the spec, the latest spec can be found via this Figma:

https://www.figma.com/file/a0dTYBGRu5hoBAKF6wGZxpUH/Gas-Station?node-id=1547%3A526

left-caret-alignment-spec

## Actual behaviour

* The left caret icon does not match the spec as of the provided Figma

left-caret-alignment

dit7ya:
## Description

Fixed the alignment and size of the left caret icon in GasStation.

Used a CSS override to avoid messing with `Icon`'s default `vertical-alignment`.

**Changes** 🏗

Screenshot 2022-01-17 at 18-11-07 Colony

Resolves #3103.

### DIFF
--- a/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css
+++ b/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css
@@ -17,3 +17,11 @@
   color: var(--text);
   cursor: pointer;
 }
+
+.returnToSummary svg {
+  /*
+   * NOTE Using this to override Icon.css's vertical-align property
+   * which might be being used elsewhere.
+   */
+  vertical-align: -5px;
+}
--- a/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.tsx
+++ b/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.tsx
@@ -20,7 +20,7 @@ const displayName = 'users.GasStation.TransactionBackToList';
 const TransactionBackToList = ({ onClose }: Props) => (
   <button type="button" className={styles.returnToSummary} onClick={onClose}>
     <Icon
-      appearance={{ size: 'small' }}
+      appearance={{ size: 'normal' }}
       name="caret-left"
       title={MSG.returnToSummary}
     />


src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css
```diff
@@ -17,3 +17,11 @@
   color: var(--text);
   cursor: pointer;
 }
+
+.returnToSummary svg {
+  /*
+   * NOTE Using this to override Icon.css's vertical-align property
+   * which might be being used elsewhere.
+   */
+  vertical-align: -5px;
```
arrenv, line 26: My preference here would be to instead use flex to align this, using the button as the container, then center aligning items.

arrenv (approved):
Looks better, just left a comment regarding my preferred, less manual approach. Let me know your thoughts.

ArmandoGraterol (approved):
Looks good to me :ok_hand:

src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css
chinins: I agree with Arren, would be good to use flex.

chinins (commented):
Please, try to implement Arren's suggestion. I will then have a look again.

dit7ya:
@arrenv suggested the following for this

On
https://github.com/JoinColony/colonyDapp/blob/2bc075cc34e5d2bdb2ce758051d28bf794321c73/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css#L18

```diff
   color: var(--text);
   cursor: pointer;
+  composes: flexContainerRow flexAlignCenter from "~styles/layout.css";
 }
```
 
 However that does not create the desired effect on my system (NixOS Linux, Wayland, either Firefox or Brave) while it works on Arren's system.
 
Result
 
 
 I also tried the following initially which did not work either.
 
 ```diff
   color: var(--text);
   cursor: pointer;
+  display: flex;
+  align-items: center;
 }
```

rdig:
Just a note that we don't officially support Firefox, just Chrome _(and it's derivatives)_

chinins:
> @arrenv suggested the following for this
> 
> On
> 
> https://github.com/JoinColony/colonyDapp/blob/2bc075cc34e5d2bdb2ce758051d28bf794321c73/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css#L18
> 
> ```diff
>    color: var(--text);
>    cursor: pointer;
> +  composes: flexContainerRow flexAlignCenter from "~styles/layout.css";
>  }
> ```
> 
> However that does not create the desired effect on my system (NixOS Linux, Wayland, either Firefox or Brave) while it works on Arren's system.
> 
> Result
> 
> I also tried the following initially which did not work either.
> 
> ```diff
>   color: var(--text);
>   cursor: pointer;
> +  display: flex;
> +  align-items: center;
> }
> ```

How does it look on google chrome on your system? As Raul said the most important is Chrom. Have you tried to add line-height, sometimes that fixes the problem for cross-system compatibility.

dit7ya:
> > @arrenv suggested the following for this
> > On
> > https://github.com/JoinColony/colonyDapp/blob/2bc075cc34e5d2bdb2ce758051d28bf794321c73/src/modules/users/components/GasStation/TransactionDetails/TransactionBackToList.css#L18
> > 
> > ```diff
> >    color: var(--text);
> >    cursor: pointer;
> > +  composes: flexContainerRow flexAlignCenter from "~styles/layout.css";
> >  }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > However that does not create the desired effect on my system (NixOS Linux, Wayland, either Firefox or Brave) while it works on Arren's system.
> > Result
> > I also tried the following initially which did not work either.
> > ```diff
> >   color: var(--text);
> >   cursor: pointer;
> > +  display: flex;
> > +  align-items: center;
> > }
> > ```
> 
> How does it look on google chrome on your system? As Raul said the most important is Chrom. Have you tried to add line-height, sometimes that fixes the problem for cross-system compatibility.

Just checked on a fresh install of Chrome (and Chromium as well) but still the same on my system.

danbr (approved):

arrenv:
> Struggled for last three hours with this to have a clean solution but couldn't find anything other than the manual approach I tried first.

@ArmandoGraterol is also on Linux, maybe it would be good to get him to do a quick check. However, if it is showing correctly with your original fix. Probably just go with that.

ArmandoGraterol (approved):
For me, the caret is still correctly aligned so from my side this is good to go

chinins:
> > Struggled for last three hours with this to have a clean solution but couldn't find anything other than the manual approach I tried first.
> 
> @ArmandoGraterol is also on Linux, maybe it would be good to get him to do a quick check. However, if it is showing correctly with your original fix. Probably just go with that.

agree with Arren

chinins (approved):

PR #3115, opened 2022-01-18
Fix claiming tokens when coin machine is installed and token is unlocked

### LINKED ISSUE #3033: Can't activate tokens when coin machine is installed & token is locked
## Steps to reproduce

1. Install coin Machine without unlocking token
2. Try to activate tokens

## Expected behaviour

Tokens should be activated

## Actual behaviour

The is an estimation error:

## Suggestions (optional)

I've talked to Alex about it and here is what he said:

```
And I'm saying that the problem is the authority deployed when coinmachine is installed
specifically, the addresses passed that are allowed to transfer
which, on deployment of the token in the colony and colony creation flow will include the token locking contract
But in the authority that's deployed with the coinmachine it's omitted
The same authority contract is deployed both times
The parameters are under the control of the dapp interface
```

chinins:
## Description

This PR fixes the bug where you can't activate tokens when coin machine is installed and token is still locked.

The fix is to add `tokenLockingAddress` to the `deployTokenAuthority` transaction when you enable coin machine.

Testing:

- create a colony
- mint tokens
- install & enable coin machine
- send tokens to coin machine
- buy tokens
- create a payment
- activate tokens

All of this operations should work without any problems. Before the fix the activate tokens part wouldn't work.

**Changes** 🏗

- update `extensionEnable` saga

Resolves #3033

### DIFF
--- a/src/modules/dashboard/sagas/extensions/colonyExtensionEnable.ts
+++ b/src/modules/dashboard/sagas/extensions/colonyExtensionEnable.ts
@@ -182,11 +182,14 @@ function* colonyExtensionEnable({
       }
 
       if (shouldSetNewTokenAuthority) {
+        const { networkClient } = colonyManager;
+        const tokenLockingAddress = yield networkClient.getTokenLocking();
+
         yield createGroupTransaction(deployTokenAuthority, {
           context: ClientType.ColonyClient,
           methodName: 'deployTokenAuthority',
           identifier: colonyAddress,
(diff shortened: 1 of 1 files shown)


danbr (approved):
Works for me.

Screenshot 2022-01-19 at 20 22 34

ArmandoGraterol (approved):
What a deal, short and sweet! Good job

arrenv (approved):
Nice! Works great for me.

PR #3125, opened 2022-01-23
Standardize numbers display

### LINKED ISSUE #3110: Standardized Numbers Display
This issue is here to track refactoring the core `Numeral` component to be able to provide a proper standardized numbers display across the whole dapp.

Decimals will be displayed with 5 relevant precision points _(eg: `0.00002`, `0.0002`, `0.002`, `0.02`, `0.2`, `0`)_

While the integer part of the number will have a human readable cutoff after the million mark:
```js
100
5,000
20,000.002
800,231
999,999.03
1 M
```

Once the million mark hits, the decimals will no longer be displayed: _(as the numbers will be tough of as being rounded)_
```js
700,000.004
800,000.004
900,000.004
1 M // this is actually 1.000,000.004
```

Larger numbers numbers that millions should also follow the above example: billion, trillion, etc

See the list of [Standard dictionary numbers](https://en.wikipedia.org/wiki/Names_of_large_numbers) for the complete list _(bonus points if you are crazy enough to implement the G -- Googol)_

Note that you can use an external library if you don't want to implement the large number display logic by hand _(just make sure you properly research it first -- with an emphases on being production ready)_

The new core component needs to support
- use number, string or bigNumber as the input value
- displaying numbers in the above scheme
- displaying both short-hand version as well as the long form of number display _(M vs. Million)_
- Be able to display a token symbol alongside the number notation name
- be able to display the full number amount, without the notion short-hand _(off by default)_ -- note that for this, after certain point the scientific notation of JS will kick in, that is to be expected
- be able to display the full decimal number amount without being truncated after 5 relevant decimal points _(off by default)_ -- note that for this, after certain point the scientific notation of JS will kick in, that is to be expected

After the new component is done, you'll need to replace all instances of number display in the dapp to use this component _(some of them currently do, but most parts just use raw number displays coming from bigNumbers)_

To get started on this feature:
- set up a github project for this: https://github.com/JoinColony/colonyDapp/projects
- set up a feature branch / draft PR for this _(assign the whole dapp team)_
- split it up in as many issues you think it's relevant so that more team members can work on it

danbr:
## Description

This is a feature branch PR to refactor the core Numeral component, providing a standardized numbers display across the whole dapp.

Resolves #3110


danbr:
This Feature branch has been in QA for sometime (rebased with Master). 
I recently fixed a bug, but otherwise no changes.

arrenv (approved):
Very nice!! I tried out as many cases as with numbers of all sizes and shape and it looks to be handling it well :)

ArmandoGraterol (approved):
Approving so that you can merge it when it's ready to :)

PR #3130, opened 2022-01-25
Show full colony name while creating colony

### LINKED ISSUE #3128: Token selector dialogs should show colony name instead of colony url
## Steps to reproduce

1. Create new colony with different name and URL
2. Check the next pages

## Expected behaviour

* Should show the full colony name 

## Actual behaviour

* Shows the short url instead

Screenshot 2022-01-24 at 16-29-23 Colony
Screenshot 2022-01-24 at 16-29-39 Colony
Screenshot 2022-01-24 at 16-29-57 Colony
Screenshot 2022-01-24 at 16-30-09 Colony

First pointed out by @arrenv here - https://github.com/JoinColony/colonyDapp/pull/3113#issuecomment-1018542738

dit7ya:
## Description

Shows full colony name instead of ENS name in the Create Colony Wizard.

Screenshot 2022-01-25 at 13-34-57 Colony
Screenshot 2022-01-25 at 13-35-10 Colony
Screenshot 2022-01-25 at 13-35-16 Colony

**Note to reviewers**
- Was the ENS normalization useful for something I am totally missing?

Resolves #3128.

### DIFF
--- a/src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.css
+++ b/src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.css
@@ -6,6 +6,10 @@
   margin-top: 180px;
 }
 
+.titleSection span {
+  word-break: break-word;
+}
+
 .inputFields {
   margin-top: 40px;
 }
--- a/src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.tsx
+++ b/src/modules/dashboard/components/CreateColonyWizard/StepCreateToken.tsx
@@ -7,7 +7,6 @@ import { Form, FormStatus, Input } from '~core/Fields';
 import Heading from '~core/Heading';
 import Button from '~core/Button';
 import { multiLineTextEllipsis } from '~utils/strings';
-import ENS from '~lib/ENS';
 import styles from './StepCreateToken.css';
 
 const MSG = defineMessages({
@@ -82,6 +81,7 @@ type FormValues = {
   tokenAddress: string;
   colonyName: string;
   tokenChoice: string;
+  displayName: string;
 };
 
 type Props = WizardProps<FormValues>;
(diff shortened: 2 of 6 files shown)


arrenv (commented):
@dit7ya looks much better, nice one.

One question, what are your thoughts on the character limits? I feel that they could be increased to have a max of 2 lines. You could also add in a `word-break: break-word;` for long strings.

download (4)

download (5)

download (6)

ArmandoGraterol (commented):
Looks to be working as expected!

@arrenv raised a valid point, perhaps it could be beneficial to bring this up to Product to see what they think about increasing the limit, leaving it as it is, or breaking the words/sentence.

And about the normalization function, it seems like it comes from quite old code so I would consult with @rdig first to know what does it do, and if it's still needed before removing it

dit7ya:
@arrenv @ArmandoGraterol I have no opinion on truncation length. However, I feel like if it is a very long name, there should be some visual indication (like bold font) for the colony name.

Adding word-break seems like a good idea irrespective of the above.

alicjakujawa (commented):
Are we sure that displayName will always exist? Shouldn't we fallback to colonyName?

dit7ya:
@alicjakujawa I cannot think of a scenario of creating the colony without a full name. Does not the following make it required?
https://github.com/joincolony/colonyDapp/blob/f508fccad7f958707db356bb4262d560943401c7/src/modules/dashboard/components/CreateColonyWizard/StepColonyName.tsx#L82

If you want I can make it fallback to colonyName nonetheless.

dit7ya:
Added word break and made `120` characters the limit for ellipsifying (chosen arbitrarily).

chinins (approved):
Good to go!

ArmandoGraterol (approved):
LGTM :+1:

PR #3132, opened 2022-01-25
Refactor enabling sagas for extensions

### LINKED ISSUE #2807: Refactor extension enabling saga logic
Currently, the saga has a lot of if/else statements to accommodate various extensions, and in the future it could have even more.

We need to have a way to have independent sagas for each extension so that we can have custom extension logic separated.

ArmandoGraterol:
Refactored `colonyEnableExtension.ts`, extracted some code from it and created `removeOldExtensionClients`, `modifyParams` and `setupEnablingGroupTransactions`. 

The idea is to remove all of the annoying if/else statement that we have in the original saga for the coin machine and whitelist extension, and future-proof ourselves in case more extension need to have a "specialized" extension enabling.

I tried to avoid code repetition as reasonable and possible as it was, if you have improvements and/or ideas please share them!

Resolves #2807


arrenv (approved):
Reading through makes sense to me, based on my understanding of the sagas. Tested installing all extensions with different scenarios and all works as expected. Nice one!

src/modules/dashboard/components/Extensions/utils.ts
```diff
@@ -28,3 +30,19 @@ export const createExtensionDefaultValues = (
...
+  extensionId: string,
+) => {
+  const actionEnd = actionType === 'SUBMIT' ? '' : `_${actionType}`;
+  let actionBeginning = 'COLONY_EXTENSION';
+
+  if (extensionId === Extension.CoinMachine) {
+    actionBeginning = 'COIN_MACHINE';
+  } else if (extensionId === Extension.Whitelist) {
+    actionBeginning = 'WHITELIST';
+  }
```
arrenv, line 45: Maybe this could be a switch, to account for more extensions in future.

Or, better yet, do we have to manually assign a string here? Can we use the string from `Extension.CoinMachine` and use that in ActionTypes as well? I'm sure there is a reason for this, just checking though.

ArmandoGraterol: Yeah, this could be a good place for a switch

About using `Extension.CoinMachine` for the ActionType, It's not possible to use directly because the format of the ActionTypes is like `COIN_MACHINE_START_THE_PARTY`, so we use snake case + uppercase, while the format that comes with `Extension.` is camel case with capitalization (`CoinMachine`). We would need to use 1 or 2 functions first to format them like we format our ActionTypes, which may not be worth it

src/modules/dashboard/sagas/utils/enableExtension.ts
```diff
@@ -0,0 +1,76 @@
+import { fork } from 'redux-saga/effects';
```
chinins, line 1: The file name was a bit confusing to me. I would suggest something like `extensionEnableFunctions` or `extensionEnableHelpers`, but it's a nitpick - up to you.

chinins (approved):
Very nice work! It looks much better and the code is more readable and clear.

src/modules/dashboard/sagas/utils/enableExtension.ts
ArmandoGraterol: I agree with you! It could have a better name

PR #3133, opened 2022-01-26
Disable action submit button on submit

### LINKED ISSUE #3124: Disable all forms while submitting an aciton/motion
Right now when you submit an action via form (e.g. from a dialog) the form is not disabled and you can still change things inside. If you do it won't affect the action inputs but the gas station will disappear and it doesn't forward you to the next screen.

I think that this is poor UX and the forms should be disabled when submitting actions/motions. Ideally, we would implement it in one place, probably `ActionForm` and just add light changes where it's used if necessary.

form-action

ArmandoGraterol:
All the submit buttons which had a loading state when submitting should now be disabled instead of showing the primary color.

This includes all the submit buttons in the UAC, and some others like the one in the banning user dialog (So long as a button has a loading state, they are included).

Resolves #3124

### DIFF
--- a/src/modules/core/components/Button/Button.css
+++ b/src/modules/core/components/Button/Button.css
@@ -68,7 +68,7 @@
  * Disabled styles via `disabled` attribute
  */
 
-.main[disabled]:not([aria-busy="true"]) {
+.main[disabled] {
   border-color: disabledBackground;
   background-color: disabledBackground;
   color: disabledColor;
--- a/src/modules/core/components/Comment/Dialogs/BanUser/BanUser.tsx
+++ b/src/modules/core/components/Comment/Dialogs/BanUser/BanUser.tsx
@@ -170,7 +170,9 @@ const BanUser = ({ colonyAddress, cancel, close, isBanning = true }: Props) => {
                 name="userAddress"
                 filter={filterUserSelection}
                 renderAvatar={supRenderAvatar}
-                disabled={loadingBannedUsers || loadingBanAction}
+                disabled={
+                  loadingBannedUsers || loadingBanAction || isSubmitting
+                }
               />
             </div>
           </DialogSection>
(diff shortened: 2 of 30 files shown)


arrenv:
@ArmandoGraterol is this one ready for review?

ArmandoGraterol:
> @ArmandoGraterol is this one ready for review?

@arrenv Oops, forgot to add reviewers! Thank you Arren

arrenv (approved):
Looks better to me!

### Before
button-loader-before

### After
button-loader-after

ArmandoGraterol:
@dit7ya You are right, we should also disable the fields to avoid any confusing behavior.

Tbh I don't think the users would expect that the transaction can be edited while submitting, it also sounds like a pain in the ass to implement hahaha I think so long as we disable the fields while submitting then it should be good

chinins (changes requested):
Yeah, the inputs in the form need to be disabled - this was more the issue...

chinins (approved):
Nice work, looks like it's very comprehensive :)

arrenv (commented):
@ArmandoGraterol works great!

Do you think there needs to be handling of rejected transactions? For e.g.

handle-rejected-transaction


Given you have gone to all the effort of disabling inputs, would there be any benefit in disabling the force toggle, in your opinion?

It does not seem to cause any issues with the transaction, it just keeps the dialog open.

toggle-force

ArmandoGraterol:
@arrenv 

> Do you think there needs to be handling of rejected transactions? For e.g.

I think that this happens whenever an error occurs while making a transaction, not just when you reject transactions, well, at least I've seen this happens not just when you reject them. But yeah, I think we need to handle this.

> Given you have gone to all the effort of disabling inputs, would there be any benefit in disabling the force toggle, in your opinion?
> 
> It does not seem to cause any issues with the transaction, it just keeps the dialog open.

Oh nice, yeah I think it should also be disabled! I'm gonna change that

Issue #3095, opened 2022-01-12
Minting large number of tokens results into display of wrong amount


dit7ya:
## Steps to reproduce

1. Create a colony
2. Mint `9007199254740991000000` native tokens (this is `Number.MAX_SAFE_INTEGER * 100000` in JavaScript)

![Screenshot 2022-01-12 at 16-09-46 Colony](https://user-images.githubusercontent.com/14034137/149124993-b973b274-04af-44f7-9745-282e9f01f36d.png)

3. Check All Teams actions view 

## Expected behaviour

Should display the correct amount minted

## Actual behaviour

Shows `9` token minted

![Screenshot 2022-01-12 at 16-09-03 Colony](https://user-images.githubusercontent.com/14034137/149124874-6484be4a-cc5e-48c9-b745-f6fac7c5223d.png)

dit7ya:
This seems to be due to https://github.com/JoinColony/colonyDapp/blob/f09f84ca4c54cc4ae59e00fd0089bdb439230bf0/src/utils/numbers/formatTokenValue.ts#L49

The `formatNumber` function expects a number but `parseFloat` loses precision if the number is too big. 

I will investigate more tomorrow and report here.

rdig:
I think we can use `BigDecimal` to solve this. Please investigate and let me know

arrenv:
This appears to be getting resolved by the `Standardized Numbers Display` project - https://github.com/JoinColony/colonyDapp/projects/22

PR #3137, opened 2022-01-27
New standarized number formatter

### LINKED ISSUE #3126: Refactor Numeral component to use external package, add formatting features
Incorporate external package into `Numeral` component to standardise number display.

danbr:
## Description

This PR implements a new number formatter based on NumbroJS (https://numbrojs.com/) that will provide a consistent display of token values.

- New formatted added
- Old formatter removed
- Dapp src updated to use

This is the first development in this area. Product will need to provide feedback, to finalise the standard display format.

This is the first (albeit large) step towards standarising number display. Therefore, all comments & suggestions are welcomed to promote further improvements.

A new interface could be added to make use of the functionality provided by NumbroJS. Conversely, perhaps a restricted interface should be used to support consistant display of numbers, such as:

- getShortDisplayNumber()
- getFullDisplayNumber()

etc.....

Resolves #3126


ArmandoGraterol (commented):
Question, is the purpose of this PR _only_ to introduce the new formatting function? 

I want to ask that because I'm seeing that there are a lot of places that are not using this new format, but then that might be outside of the scope of this PR, depending on how you planned the issues.

Leaving that aside, are waiting for Product to provide a new way to format small numbers? I saw that this piece of code was deleted:

```
  if (decimalValue.lt(0.00001) && decimalValue.gt(0)) {
    return SMALL_TOKEN_AMOUNT_FORMAT;
  }
```

And because of that numbers who are less than `0.00001` look like this:
FireShot Capture 530 - Colony - localhost

Finally, I don't know if it's just me, but I feel like the `std` in `getStdFormatted...` and `stdNumberFormatter` might be a bit redundant as the purpose of this branch is to have 1 way of formatting the token numbers in the whole app anyway, so I feel like it would be fair to assume that it's standard without having it in the name.

Lemme know what you think!

danbr:
> Question, is the purpose of this PR _only_ to introduce the new formatting function?
> 
> I want to ask that because I'm seeing that there are a lot of places that are not using this new format, but then that might be outside of the scope of this PR, depending on how you planned the issues.
> 
> Leaving that aside, are waiting for Product to provide a new way to format small numbers? I saw that this piece of code was deleted:
> 
> ```
>   if (decimalValue.lt(0.00001) && decimalValue.gt(0)) {
>     return SMALL_TOKEN_AMOUNT_FORMAT;
>   }
> ```
> 
> And because of that numbers who are less than `0.00001` look like this: FireShot Capture 530 - Colony - localhost
> 
> Finally, I don't know if it's just me, but I feel like the `std` in `getStdFormatted...` and `stdNumberFormatter` might be a bit redundant as the purpose of this branch is to have 1 way of formatting the token numbers in the whole app anyway, so I feel like it would be fair to assume that it's standard without having it in the name.
> 
> Lemme know what you think!

Hey - thanks for your comments. To answer you in order:

Yes perhaps I should have made this more clear - this PR is to give us a starting point to work from and not the finished article. An opportunity for everyone to take a look at Numbro and decide if it fulfils our requirements for providing a standardized number display, and provide some input.

This PR was not attempting to provide a complete and finished solution.

Many of the more prominent displays & dialogs were updated to use the new number formatter, but indeed there are more places that need to be changed. This will be included in future issues, once we're agreed on the formatter.

We can make use of Numbro to handle formatting small ( & large) numbers rather than writing and maintaining our own implementations, or certainly, it is the direction that we should be heading. 

I agree, std is excessive in the name and can be dropped. It was kept for development while removing the existing routines. We can drop the 'std' or perhaps there is a better-suited name?

ArmandoGraterol:
> Yes perhaps I should have made this more clear - this PR is to give us a starting point to work from and not the finished article. An opportunity for everyone to take a look at Numbro and decide if it fulfils our requirements for providing a standardized number display, and provide some input.
> 
> This PR was not attempting to provide a complete and finished solution.
> 
> Many of the more prominent displays & dialogs were updated to use the new number formatter, but indeed there are more places that need to be changed. This will be included in future issues, once we're agreed on the formatted.

That's understandable! Thank you for the clarification Dan

> We can make use of Numbro to handle formatting small ( & large) numbers rather than writing and maintaining our own implementations, or certainly, it is the direction that we should be heading. But your comment about SMALL_TOKEN_AMOUNT_FORMAT is a great example of where it would be good to get some help. Especially situations and edge cases that are not covered. I can add SMALL_TOKEN_AMOUNT_FORMAT into the new function.

If `Numbro` has a way of formatting really small numbers similar to how we formatted them before then that could be nice, other than that, we are going to need to use something like what we had with `SMALL_TOKEN_AMOUNT_FORMAT` before.

> I agree, std is excessive in the name and can be dropped. It was kept for development while removing the existing routines. We can drop the 'std' or perhaps there is a better-suited name?

For now, I can't think about a better name other than dropping the `std`, perhaps another reviewer can come up with something better.

About the formatter itself, right now the only thing that I've noticed is how the less than `.00001` amounts are formatted as `0`, which is an edge case that should be handled as well. I'm gonna take another look at the formats now taking into account the new info provided by your comment, thank you Dan!

danbr:
> > Yes perhaps I should have made this more clear - this PR is to give us a starting point to work from and not the finished article. An opportunity for everyone to take a look at Numbro and decide if it fulfils our requirements for providing a standardized number display, and provide some input.
> > This PR was not attempting to provide a complete and finished solution.
> > Many of the more prominent displays & dialogs were updated to use the new number formatter, but indeed there are more places that need to be changed. This will be included in future issues, once we're agreed on the formatted.
> 
> That's understandable! Thank you for the clarification Dan
> 
> > We can make use of Numbro to handle formatting small ( & large) numbers rather than writing and maintaining our own implementations, or certainly, it is the direction that we should be heading. But your comment about SMALL_TOKEN_AMOUNT_FORMAT is a great example of where it would be good to get some help. Especially situations and edge cases that are not covered. I can add SMALL_TOKEN_AMOUNT_FORMAT into the new function.
> 
> If `Numbro` has a way of formatting really small numbers similar to how we formatted them before then that could be nice, other than that, we are going to need to use something like what we had with `SMALL_TOKEN_AMOUNT_FORMAT` before.
> 
> > I agree, std is excessive in the name and can be dropped. It was kept for development while removing the existing routines. We can drop the 'std' or perhaps there is a better-suited name?
> 
> For now, I can't think about a better name other than dropping the `std`, perhaps another reviewer can come up with something better.
> 
> About the formatter itself, right now the only thing that I've noticed is how the less than `.00001` amounts are formatted as `0`, which is one of those edge cases that you mention. I'm gonna take another look at the formats now taking into account the new info provided by your comment, thank you Dan!

I have removed `std` from file & function names, and played with SMALL_TOKEN_AMOUNT_FORMAT vs Numbro formatting for small numbers.

As a comparison currently in master:
Screenshot 2022-02-02 at 18 13 16

Screenshot 2022-02-02 at 18 13 08

Verses small number formatted using Numbro, 

Screenshot 2022-02-02 at 18 40 36

Example of handling long numbers, short (averaged) Vs long

In the Actions listview, the number is 'averaged'
Screenshot 2022-02-02 at 18 44 53

detailed view
Screenshot 2022-02-02 at 18 44 39

Question:
In your example:
https://user-images.githubusercontent.com/18473896/152051496-f899c175-f38f-4d75-bac0-4e18036663e6.png
Ideally, what should be the output? for numbers < 0.00001 ? I assume scientific notation is not our goal.

I would assume on the Actions page, a general answer, and on the page for each action - a more detailed answer should be displayed (best effort of course, we cant show everything)

src/modules/dashboard/components/ActionsPageFeed/ActionsPageEvent/ActionsPageEvent.tsx
```diff
@@ -47,13 +47,13 @@ const MSG = defineMessages({
   rolesTooltip: {
     id: 'dashboard.ActionsPageFeed.ActionsPageEvent.rolesTooltip',
     defaultMessage: `{icon} {role, select,
-      0 {This permission allows users to put the colony in and out of 
+      0 {This permission allows users to put the colony in and out of
```
danbr, line 50: VSCode decided that it should automatically remove some redundant spaces. I've checked the tool tips, and this change did not cause any side effects.

chinins (approved):
I think it looks good and numbro will solve the most cases for us.

ArmandoGraterol:
> Question: In your example: https://user-images.githubusercontent.com/18473896/152051496-f899c175-f38f-4d75-bac0-4e18036663e6.png Ideally, what should be the output? for numbers < 0.00001 ? I assume scientific notation is not our goal.

When I worked in the old `getFormattedTokenValue` the consensus that we reached with Product was that the numbers `< 0.00001`  would need to be shown as `0.00000...` **always** because the idea was to _only_ show up to 5 decimals and no more. So perhaps you can use that same standard or consult with Product again if they would like to change that to something else.

> I would assume on the Actions page, a general answer, and on the page for each action - a more detailed answer should be displayed (best effort of course, we cant show everything)

That sounds like a good idea! I like it

arrenv (commented):
@danbr looks good so far based on my tests. I had a looked a the library, but couldn't see it anything about being able to customise the formatting of the 'billion', 'trillion' symbol for example. Do you know if it is possible? If you see my screen shot below, it shows a lowercase 't' and no space, just curious if that can be customised.

Like Armando has said, I think having a defined list of number formats would be good. Something like the following:

number-formatting


I also checked it with this issue - https://github.com/JoinColony/colonyDapp/issues/3095 - and it seemed to resolve it.

75e71d62-8c88-41ad-906d-35a329629f2f

danbr:
> @danbr looks good so far based on my tests. I had a looked a the library, but couldn't see it anything about being able to customise the formatting of the 'billion', 'trillion' symbol for example. Do you know if it is possible? If you see my screen shot below, it shows a lowercase 't' and no space, just curious if that can be customised.

@arrenv Thanks.
Customizing the abreviations should be possible. Generally, Numbro bases this, and some other behaviours upon the locale. There is no reason why we can simply create our own: ( i was surprised for most English lang the abrev are lowercase)

 abbreviations: {
        thousand: "k",
        million: "m",
        billion: "b",
        trillion: "t"
    },

here are the various locales, for comparison:
https://github.com/BenjaminVanRyseghem/numbro/tree/develop/languages

As for the space between the number and abrev, there is a parameter that controls its presence. Easy to change.

I assume everyone is (generally) happy with Numbro and proceed.

arrenv:
> I assume everyone is (generally) happy with Numbro and we can proceed.

@danbr it looks like it will do what we need. The usage, active support and history of it seem to be good signs. Out of curiosity, what was your preference for picking Numbro over other options?

danbr:
> > I assume everyone is (generally) happy with Numbro and we can proceed.
> 
> @danbr it looks like it will do what we need. The usage, active support and history of it seem to be good signs. Out of curiosity, what was your preference for picking Numbro over other options?

@arrenv Its an active project, supports typescript, and offers lots of flexibility in terms of formatting.

arrenv (approved):
Looks to be a good option, nice one :)

PR #3138, opened 2022-01-27
Cancel team selector disable when no reputation in Smite

### LINKED ISSUE #3135: Domain selector is disabled in the Smite dialog if there is no reputation
It doesn't allow to change domain where there can be reputation. 
Screenshot 2022-01-27 at 16 09 50

chinins:
## Description

This PR removes disabled state from team selector when voting extension is enabled and there is no reputation in a team. This way we can choose a different team where potentially there is reputation and a motion can be created.

**Changes** 🏗

- update `SmiteDialogForm`

resolves #3135 
Screenshot 2022-01-27 at 17 39 08


danbr (approved):

arrenv (approved):
@chinins good pickup, this fixes it!

arbitration-selector

ArmandoGraterol (approved):
Looks good to me :+1:

PR #3139, opened 2022-01-27
Remove caret `onClick` when input disabled in SingleUserPicker

### LINKED ISSUE #3136: Can select user clicking on caret even when input is disabled
In `SingleUserPicker` you can click on caret and change user selection even when the input field is disabled
Screenshot 2022-01-27 at 17 19 08

chinins:
## Description

When `SingleUserPicker` input is disabled the `caret` was still working and you could change user by clicking on it. I have removed the `onClick` prop from `caret` when input is disabled.

**Changes** 🏗

- update `SingleUserPicker`

resolves #3136

### DIFF
--- a/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
+++ b/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
@@ -230,7 +230,7 @@ const SingleUserPicker = ({
             </div>
             {(!value || (value && !isResettable)) && (
               <Icon
-                onClick={openOmniPicker}
+                {...(disabled ? {} : { onClick: openOmniPicker })}
                 className={classnames(styles.arrowIcon, {
                   [styles.arrowIconActive]: omniPickerIsOpen,
                 })}


alicjakujawa (commented):
Do we want to actually have onClick event on Icon which is <i> tag? Maybe we should add button here? and Icon inside button

arrenv (commented):
@chinins I tested with a button as @alicjakujawa suggested. It adds an extra step for keyboard usage given onFocus on the input triggers the options.

However, it would allow you to more easily update the colour of the caret, when disabled. So that it also appears in a disabled state.

chinins:
@arrenv, @alicjakujawa, I had almost finished implementing the request when I realised that this is the way it's implemented across the dapp and it will be inconsistent to change just this one.

I think that this PR shouldn't do that. We can raise an issue to change it everywhere, but to be honest with you I don't think it's really worth it...

arrenv (approved):
@chinins makes sense, nice work double checking!

ArmandoGraterol (approved):
Looks good to me! Yeah, if this is how it is all over the dapp then it's better to address it in a dedicated issue

Issue #3134, opened 2022-01-27
The arbitration permission is shown as `disabled`


ArmandoGraterol:
In QA I encountered this:
![FireShot Capture 527 - Colony - qaxdai colony io](https://user-images.githubusercontent.com/18473896/151403902-cc0444ff-252e-4d27-86c0-0c80a8c4b3ee.png)

The description makes the permission look like it should be disabled, but it's actually not.

This may be a leftover from when the permission was disabled and smite wasn't even implemented.

While in my dev env it looks like this:
![FireShot Capture 528 - Colony - localhost](https://user-images.githubusercontent.com/18473896/151405850-a7b5f07f-6477-4e1d-ae3d-43e608ed5e08.png)

Now that we implemented smite, it should be expected that the permission needed to smite users is available for assignment

## Steps to reproduce

1. Go to the permissions dialog
2. Take a look at the arbitration permission

## Expected behavior

* Should show the same dark color as the other's description
* Should be interactable

## Actual behavior

* In my colony in QA it shows a grey color, the color that is used for the `disabled` state. Even though the checkbox is not really disabled, you can interact with it.
* In my dev env, it's completely disabled.

PR #3140, opened 2022-01-28
Fix arbitration permission text color in Permission Management dialog

chinins:
## Description

This PR makes the arbitration permission description in the Permission Management dialog the same color as all other permissions.

**Changes** 🏗

- update `PermissionManagementCheckbox`

**Deletions** ⚰️

- delete leftover css class

resovles #3134

### DIFF
--- a/src/modules/dashboard/components/Dialogs/PermissionManagementDialog/PermissionManagementCheckbox.css
+++ b/src/modules/dashboard/components/Dialogs/PermissionManagementDialog/PermissionManagementCheckbox.css
@@ -19,7 +19,3 @@
 .popoverContent {
   padding: 10px;
 }
-
-.permissionNotAvailable {
-  color: var(--text-disabled);
-}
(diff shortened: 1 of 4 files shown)


arrenv (approved):
@chinins showing as expected. So I understand, why it was originally added only for arbitration, what has changed?

ArmandoGraterol (dismissed):
There's still something related to the leftover code that is disabling the permission from being assigned or removed when you *don't* have the voting extension enabled. For example, I can't assign the `Arbitration` permission to a member that recently joined my colony:

FireShot Capture 529 - Colony - localhost

In `PermissionManagementForm` this line is disabling the checkbox: `!canRoleBeSet(role)` because inside that function we have this leftover:
```
      switch (role) {
        // Can't set arbitration at all yet
        case ColonyRole.Arbitration:
          return false;
```

arrenv:
@chinins I tested it without the Motions extension installed as well and I am getting the same issue as @ArmandoGraterol.

chinins:
@ArmandoGraterol, @arrenv, good catch guys!  have updated the PR to enable Arbitration permission without the voting extension but I am still waiting for a confirmation from Raul that we should enable it for all instances. You can re-review meanwhile.

rdig (approved):
Good to go

PR #3141, opened 2022-01-28
Make hover state on select items non-sticky

### LINKED ISSUE #3131: Hover state on select items are sticky
## Steps to reproduce

There many instances of it since it is the CSS of the SelectOption component. One example - 

1. Create colony and create an extra team.
2. Try to transfer funds between them.
3. Hover over a team from the dropdown but then move mouse away. 

## Expected behaviour

The hover background color should go away when not hovering.

## Actual behaviour

Background color remains on the last hovered option.

recording

dit7ya:
## Description

**Changes** 🏗

CSS selector changed from `[aria-selected="true"]` to `:hover` on SelectOption component. 

Resolves #3131.
Colony

### DIFF
--- a/src/modules/core/components/Fields/SelectOption/SelectOption.css
+++ b/src/modules/core/components/Fields/SelectOption/SelectOption.css
@@ -8,7 +8,7 @@
   cursor: pointer;
 }
 
-.main[aria-selected="true"] {
+.main:hover {
   background-color: var(--grey-blue-0);
   outline: none;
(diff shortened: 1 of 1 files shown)


arrenv (approved):
@dit7ya working well for me.

alicjakujawa (commented):
I remember it was introduced for better accessibility, so worth to make sure that we want get rid of this with @rdig

PR #3144, opened 2022-01-28
Header reputation and network info styling

### LINKED ISSUE #3129: Update Reputation styling, hover state, and add Tooltip in header
## Description

To ensure the Dapp matches the design, the styling of the reputation value in the header needs to be updated.

There also needs to be a tooltip added to help provide more context to what the value represents.

1. You will need to ensure the Motions & Reputation extension is installed.
2. You will need to double check the appearance with and without a value for reputation.

## Expected behaviour

* The design should match the spec
* There should be a hover state
* There should be a tooltip with the copy "You have {REPUTATION_VALUE}% of the reputation in this colony."

[DESIGN] - https://www.figma.com/file/DjVCPq7LpkjEMEfQZtKgBdC1/Style-Guide?node-id=2905%3A0

reputation-star-hover-state

## Actual behaviour

* The star is in the wrong location, there is no hover state, and no tooltip.

reputation-star

dit7ya:
## Description

**New stuff** ✨

Added hover state and tooltip to the current user reputation in the nav bar.
Colony

Colony (2)

Also added hover state to the network info in the nav bar since that was missing too.

Colony (1)

**Changes** 🏗

Switched the star icon to the right in MemberReputation component. This affects all places where the MemberReputation component is.

**Notes**

Used a slightly different copy than the spec to avoid introducing a new variable in the MemberReputation component. (Displaying the reputation percentage in the tooltip is complicated since the value of that is computed in the MemberReputation component, which is a child of the Tooltip component.

Resolves  #3129.

### DIFF
--- a/src/modules/core/components/Icon/Icon.tsx
+++ b/src/modules/core/components/Icon/Icon.tsx
@@ -38,7 +38,7 @@ interface Props extends Omit<HTMLAttributes<HTMLElement>, 'title'> {
   name: string;
 
   /** Html title for the icon element */
-  title: string | MessageDescriptor;
+  title?: string | MessageDescriptor;
 
   /** Values for html title (react-intl interpolation) */
   titleValues?: SimpleMessageValues;
@@ -77,7 +77,7 @@ const Icon = ({
     typeof title === 'object' ? formatMessage(title, titleValues) : title;
   return (
     <i
-      title={iconTitle}
+      title={title ? iconTitle : undefined}
       className={
         className || getMainClasses(multiColorAppearance || appearance, styles)
       }
--- a/src/modules/core/components/MemberReputation/MemberReputation.css
+++ b/src/modules/core/components/MemberReputation/MemberReputation.css
@@ -9,7 +9,7 @@
 
 .icon {
   display: inline-block;
-  margin-right: 3px;
+  margin-left: 3px;
   height: 11px;
   width: 11px;
 }
--- a/src/modules/core/components/MemberReputation/MemberReputation.tsx
+++ b/src/modules/core/components/MemberReputation/MemberReputation.tsx
@@ -64,19 +64,6 @@ const MemberReputation = ({
 
   return (
     <div>
-      <Icon
-        name="star"
-        appearance={{ size: 'extraTiny' }}
-        className={styles.icon}
-        title={
-          userPercentageReputation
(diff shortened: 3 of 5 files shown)


ArmandoGraterol (commented):
Nice! Looks good right now but I think it would look better if the tooltip was lower, so that it doesn't overlaps the reputation.

The `Tooltip` component can receive a prop to move the tooltip around like this:
```
popperProps={{
  modifiers: [
    {
      name: 'offset',
      options: {
        offset: [-8, 8],
      },
    },
  ],
}}
```

dit7ya:
@ArmandoGraterol Thank you for pointing it out. I have added `popperProps` (but slightly changed it since I also changed the placement of the tooltip to be `bottom-start`.

ArmandoGraterol (approved):
Nice work! It looks good to me

chinins (approved):
Looks good!

arrenv (changes requested):
@dit7ya Are you able to take a look at the alignment of the text in the box, to vertically align it better. I am creating another issue regarding the alignment and font sizes difference in the other elments.

header-text-alignment


Also, the hover seems a little temperamental.

tempermental-on-hover

dit7ya:
Good point @arrenv. After some tinkering, I think the visible rerender of the tooltip is due to the existing title property of the `Icon` component inside it. Can you please confirm if you observe the same behavior in the Lock icon in the Events page? I guess the question mark tooltip does not have this problem because it is circular. I don't really know though. Removing the Icon from the component makes the rendering much more predictable.

If that is the case, the only solution I can think of is re-writing the Icon component to make the title optional. What do you think?

dit7ya:
I can't seem to get around the shaky tooltip behavior even after removing the icon or the text. Any suggestions @arrenv ?

arrenv (changes requested):
@dit7ya I think that is not really a big issue.

However, I did notice the title tag overlapping the popover. I think that it is worth adding another component prop to disable the title tag to account for this.

title-tag-overlay
.

dit7ya:
@arrenv Made the title property optional in the Icon component and removed it from the tooltiped component. Also, I don't think this bug is related to this PR but please take a look at the console when you hover over the reputation component. It is creating  failing network requests which might be a separate issue somewhere else.

PR #3145, opened 2022-01-29
Make Toggle component work when scrolled

### LINKED ISSUE #3143: Force switch not working on scroll
## Steps to reproduce

1. Ensure you have the Motions & Disputes extension installed
2. Go to a long action dialog, such as 'Permissions'
3. Scroll and try to force the action

## Expected behaviour

* Force switch should work if visible

## Actual behaviour

* Force switch not working due to the position of the input remaining relative to the dialog box

force-not-working-on-scroll

## Suggestions

* Add a position relative to the switch container

dit7ya:
## Description

Fixed toggle component invisibly shifting when scrolled.

Colony

**Changes** 🏗

Completed the existing checkbox hack implementation in the component. Had to make some compromises to make the linter happy.

Resolves #3143.

### DIFF
--- a/src/modules/core/components/Fields/Toggle/Toggle.css
+++ b/src/modules/core/components/Fields/Toggle/Toggle.css
@@ -3,6 +3,7 @@
 
 .container {
   display: flex;
+  position: relative;
 }
 
 .container label {


ArmandoGraterol (approved):
Looks good to me :+1:

arrenv (commented):
@dit7ya Was it not possible to just add `position: relative;` to the main class of the Toggle component? Or, does it have an impact elsewhere?

dit7ya:
> @dit7ya Was it not possible to just add `position: relative;` to the main class of the Toggle component? Or, does it have an impact elsewhere?

I don't think it will work that way. The idea in the existing component is to use the so called [Checkbox Hack](https://css-tricks.com/the-checkbox-hack/) to bind the label and input so that when you click on the label the button works - while you do your custom styling + logic on a separate element that looks like a toggle.

Please let me know what you think - I can give it another try.

arrenv:
@dit7ya Sure, I have just tested and `position: relative` on the `main` class seems to work.

This is with this PR's changes, the input is still positioned below the visible toggle:

toggle-position


This is with `position: relative;`:

toggle-position-2

dit7ya:
Did you mean just adding that line without any changes from this PR? I just tested that and it seems the original scroll issue still remains in that case.

arrenv:
@dit7ya I am on `master`, and only added `position: relative;` to the main `.container` in `Toggle.css`. Does this not work for you though?

This is what I changed:

Screenshot 2022-02-01 212705

The result for me is this:

hover-on-toggle

chinins (commented):
I don't really like the current implementation, because there are potentially 2 labels for the input. I tried what Arren suggested and it's working for me.

dit7ya:
Just tested on master again and @arrenv's suggestion is working fine. I am pretty sure though it did not work when I first worked on this PR - don't know what happened. Anyway, pushing soon with the simple change.

arrenv (approved):
Working well for me!

PR #3151, opened 2022-01-31
Enhance the management of 'confirm' button state in components that use TokenSelector & update copy

### LINKED ISSUE #3121: Fix button state while loading and the error message in "Manage tokens" dialog
## Steps to reproduce

1. Go to the Funds page of a Colony.
2. Click on Manage tokens.
3. Add a token contract address that is not in your local dev e.g. 0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48
4. You are able to 'Confirm' the token while it is loading.
5. Look at the console when browsing the dapp, there will now always be an error and no way to remove the added token.

**Additionally**
4b. Do it again with another token address (e.g. 0xdac17f958d2ee523a2206206994597c13d831ec7). Once it is not able to find the token it disables the button and provides an error and a prompt which includes "Please type in token details". However no options exist.

## Expected behaviour

* You shouldn't be able to click 'Confirm' to add a token while the token address check is loading.
* There should be no prompt to add in the token details. Message can be changed to: "Token data not found. Please check the token contract address."

## Actual behaviour

* You are able to add a token while the token address check is loading.
* You are prompted to enter in token details, with no way of doing so.

can-add-token-while-loading

shows-disabled-confirm-button

danbr:
## Description

This PR adds corrects the behavior of `TokenEditDialog`, and updates the copy to read "Token data not found. Please check the token contract address.".

A new callback prop to `TokenSelector` was added to inform the caller of when an address is being searched.

Initially, `TokenEditDialog` was enhanced to guard the 'confirm' button when performing verification & validation. But 2 other components (`UserTokenEditDialogForm` & `StepSelectToken`) that use `TokenSelector` also had the same issues, so they were also updated.

Before updates: 

Screenshot 2022-01-30 at 17 43 34

After updates:

Screenshot 2022-01-30 at 17 46 54

Screenshot 2022-01-30 at 17 41 01

Screenshot 2022-01-30 at 17 47 08

Screenshot 2022-01-30 at 17 47 18

Resolves #3121

### DIFF
--- a/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
+++ b/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
@@ -101,6 +101,14 @@ const StepSelectToken = ({
 }: Props) => {
   const [tokenData, setTokenData] = useState<OneToken | undefined>();
   const { formatMessage } = useIntl();
+  const [isCheckingAddress, setIsCheckingAddress] = useState<boolean>(false);
+  const [tokenSelectorHasError, setTokenSelectorHasError] = useState<boolean>(
+    true,
+  );
+
+  const handleCheckingAddress = (isChecking: boolean) => {
+    setIsCheckingAddress(isChecking);
+  };
 
   const handleTokenSelect = (token: OneToken, setFieldValue: SetFieldValue) => {
     setTokenData(token);
@@ -108,6 +116,10 @@ const StepSelectToken = ({
       setFieldValue('tokenName', token.name);
       setFieldValue('tokenSymbol', token.symbol);
     }
+  };
+
+  const handleTokenSelectError = (hasError: boolean) => {
+    setTokenSelectorHasError(hasError);
   };
 
   const goToTokenCreate = useCallback(() => {
@@ -161,6 +173,8 @@ const StepSelectToken = ({
               onTokenSelect={(token: OneToken) =>
                 handleTokenSelect(token, setFieldValue)
               }
+              onTokenSelectError={handleTokenSelectError}
+              onCheckingAddress={handleCheckingAddress}
               tokenData={tokenData}
               extra={
                 <button
@@ -198,8 +212,13 @@ const StepSelectToken = ({
               <Button
                 appearance={{ theme: 'primary', size: 'large' }}
                 type="submit"
-                disabled={!isValid || (!dirty && !stepCompleted)}
                 text={MSG.continue}
+                disabled={
+                  tokenSelectorHasError ||
+                  !isValid ||
+                  (!dirty && !stepCompleted) ||
+                  isCheckingAddress
+                }
               />
             </div>
           </div>
--- a/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
+++ b/src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
@@ -106,16 +106,16 @@ const StepSelectToken = ({
     true,
   );
 
-  const handleCheckingAddress = (isChecking: boolean) => {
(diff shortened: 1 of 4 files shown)


src/modules/dashboard/components/CreateColonyWizard/TokenSelector.tsx
```diff
@@ -107,19 +112,21 @@ const TokenSelector = ({
...
       }
     },
-    [onTokenSelect, onTokenSelectError],
+    [onTokenSelect, onTokenSelectError, onCheckingAddress],
   );
 
   const handleGetTokenError = useCallback(
     (error: Error) => {
       setLoading(false);
+      onCheckingAddress(false);
```
alicjakujawa, line 121: you can see here and in other places that "onCheckingAddress" has the same meaning as "loading" so we actually don't need this new method.

alicjakujawa (changes requested):
I would get rid of this repeated code.  I would make use of "onTokenSelect" callback, as if it was fired && it has token data - it means that its not "loading" anymore (or is not "checking" anymore)

src/modules/dashboard/components/CreateColonyWizard/TokenSelector.tsx
danbr: Agreed:  "onCheckingAddress" has the same meaning as "loading"  - I refactored, and also moved the isCheckingAddress functionality into "onTokenSelect".

arrenv (approved):
Looks to be working great.

One minor suggestion, we could probably change this status indicator to be red on the "Token data not found..." error message.

error-message

src/modules/dashboard/components/CreateColonyWizard/TokenSelector.tsx
```diff
@@ -93,16 +102,20 @@ const TokenSelector = ({
...
     (token: OneToken) => {
-      const { name, symbol } = token;
+      const { name, symbol } = token || {};
       setLoading(false);
+
       if (!name || !symbol) {
-        onTokenSelect(null);
+        onTokenSelect(false, null);
+        setHasError(true);
+        onTokenSelectError(true);
```
chinins, line 115: I think you can eliminate the `hasError` state in this component and just pass it as a prop.
Please, check React best practices/pros & cons of doing so.

chinins (commented):
Looks good, I have just one doubt about maintaining same states (`hasError`) in 2 components (as per my comment directly in the file)

src/modules/dashboard/components/CreateColonyWizard/StepSelectToken.tsx
```diff
@@ -101,13 +101,25 @@ const StepSelectToken = ({
...
+  ) => {
     setTokenData(token);
-    if (token) {
-      setFieldValue('tokenName', token.name);
-      setFieldValue('tokenSymbol', token.symbol);
-    }
+    setIsCheckingAddress(checkingAddress);
+
+    setFieldValue('tokenName', token ? token.name : '');
+    setFieldValue('tokenSymbol', token ? token.symbol : '');
```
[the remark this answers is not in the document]
danbr: Thank you. I will include your suggestion.

src/modules/dashboard/components/CreateColonyWizard/TokenSelector.tsx
danbr: I have refactored the state management and moved the states of `Error` and `isCheckingAddress` to the parent.

chinins (approved):
I think this looks better now, with single source of truth for both the parent & the child.

ArmandoGraterol (approved):
Nice! I quite like these changes :ok_hand:

PR #3154, opened 2022-01-31
Add Stakes tab to Token Activation Popover

### LINKED ISSUE #3152: View and Claim All Stakes Tab
### Story

If you stake on a motion in a Colony and don't keep track of the motions, it can be quite time consuming to find those motions to claim back your stake. Especially if there are a lot of actions performed in the colony. The Dapp should help make it easier to claim their stakes from finalized motions.

### Description

By including a 'Stakes' tab in the token activation popover, we are able to provide users with an easy way to claim all finalized motions and view a list of completed motions. This feature involves the following (the claim all button and motions list will be raised with other issues):

- The tab functionality has already partly been built into the `TokenActivationContent.tsx` tab, with this feature it will need to be reviewed, uncommented, and updated to match the design spec and to allow for the other functionality to be included.

- You will also need to query the relevant available motions, so, that the can be used by the by the list and claim all button.

- With this feature, you will also need to create a project, assigning the dapps team, that includes this initial tab PR.

NOTE: The notification badge can be excluded for the time being if it is too costly.

### Design

[Figma link](https://www.figma.com/file/m96yJW1PPXezFl2mkuYha6/Token-Activation?node-id=2411%3A7599)

stakes-claim-all

- [x] Add the 'Stakes' tab to the Token Activation Popover
- [x] Create the "Claiming Stakes Tab" project
- [x] Include a query that can be used by the claim all stakes button and staked motions list

arrenv:
## Description

This PR adds the a 'Stakes' tab into the token activation popover and provides a query of claimable `motionId`'s that can use used by the 'Claim All' stakes button and the staked motions list.

stakes-tab

NOTE: I went around in circles on this with creating a custom query, using existing queries, using resolvers... I ended up just using the existing `useSubgraphUserMotionTokenEventsQuery` query and using a filter for simplicity. I would love to hear any suggestions for alternative / better ways to get the list of claimable motions.

Or, any thoughts on handling requests greater then 1,000 (given the query only gets staked events and won't remove them from the query if claimed). Although, I figure this is not important and by sorting query date by descending we can capture majority of cases (unless they are staking a ridiculous amount and never claiming).

**New stuff** ✨

* Added/Re-enabled ClaimsTab in `TokenActivationContent.tsx`
* Added `walletAddress` to UserTokenActivationButton and  TokenActivationContent
* Added query for claimable motions
* Added resolver `claimableStakedMotions` and query `ClaimableStakedMotions`

## Testing
* You will need to have the Motions & Disputes extension installed.
* Create some motions and finalize, but don't claim (if you do a lot, try claiming some). 
Reminder, you can speed up the chain 3 days using command: `curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[260000],"id": 1}' localhost:8545 && curl -H "Content-Type: application/json" -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[]}' localhost:8545`
* Navigate to the `TokenActivation` popover and click on the 'Stakes' tab.
* You should see a loader while waiting for the query, then the relevant messages 'There are no stakes to claim.' or '[Claimable stakes appear here]' (this is just a temporary message).
* Console log `unclaimedMotions` to see array of claimable motionIds.

Resolves #3152


danbr (approved):
Looks good - and its working!

src/modules/users/components/TokenActivation/TokenActivationContent/ClaimsTab.tsx
```diff
@@ -1,19 +1,91 @@
...
+  } = useSubgraphUserMotionTokenEventsQuery({
+    variables: {
+      colonyAddress: colonyAddress?.toLowerCase(),
+      walletAddress: walletAddress?.toLowerCase(),
+      extensionAddress: votingReputationExtension?.address?.toLowerCase() || '',
+      sortDirection: 'desc',
+    },
+  });
+
+  const unclaimedMotions = useMemo(() => {
```
chinins, line 49: I think that using existing query and filtering in these events is good. It doesn't seem that we really have any other option.

However, I would move this logic into a resolver, so create a new resolver & a query that uses this resolver. Then you can use inside that resolver the existing `SubgraphUserMotionTokenEvents` query. For example in `user` resolver file there is 
 a `getUserStakedBalance` function where we do similar processing (we just get the total user balance instead of the array of unclaimed motions).

src/modules/users/components/TokenActivation/TokenActivationContent/ClaimsTab.tsx
```diff
@@ -1,19 +1,91 @@
...
   },
 });
 
-const ClaimsTab = () => {
+export interface ClaimsTabProps {
+  colonyAddress: Address;
+  walletAddress: Address;
+}
+
+const ClaimsTab = ({ colonyAddress, walletAddress }: ClaimsTabProps) => {
```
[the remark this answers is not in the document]
chinins: I tend to agree with Dan here, I think that it's less confusing to have Stakes in the file name as it's what the tab is called.

chinins (commented):
It's all working properly. But as per my comment I would move the filtering logic into a resolver. Let me know what you think.

src/modules/users/components/TokenActivation/TokenActivationContent/ClaimsTab.tsx
arrenv: @chinins thank you, I have refactored to use a resolver. I ended up putting it in the `stakes.ts` resolver, as the user one was already quite big and I felt it made sense to have it in either.

chinins (approved):
Looks much better now! Great job :)

PR #3158, opened 2022-02-01
Claiming Stakes Tab

arrenv:
### Story

If you stake on a motion in a Colony and don't keep track of the motions, it can be quite time consuming to find those motions to claim back your stake. Especially if there are a lot of actions performed in the colony. The Dapp should help make it easier for users to claim their stakes from finalized motions.

### Description

By including a 'Stakes' tab in the token activation popover, we are able to provide users with an easy way to claim all finalized motions and view a list and navigate to a list of a users finalized motions. 

e95e8282-219e-4992-8580-e1188aec401d

### Testing

* You will need to have the Motions & Disputes extension installed. I suggest having the `Staking Phase Duration` set to a low number, so that you don't have to wait for the default 72 hours to test.
* Motions will need to be staked, then finalized or failed for them to appear in the tab.
* Navigate to the `TokenActivation` popover, by clicking on the native token balance in the top right.
* Click on the 'Stakes' tab to view the list of claimable staked motions.
* You should see a loader while waiting for the query, then either the 'There are no stakes to claim.' message or the stakes list.
* Try navigating to the motions from the list.
* Try clicking the 'Claim all' button

### Scenarios to Test

* Try testing with unfinalized motions, to ensure they don't appear in the list.
* Try testing with failed motions, to ensure that they can still be claimed.
* Try testing with several motions, so that you can see how it works with a lot of claims at once.
* Any other scenarios you can think of.


danbr (approved):
Looks good. 
I was also able to claim both, a stake from a motion that had failed, and a stake from a motion that had passed and that was finalised. 

Nice to see it all the work come together.

PR #3160, opened 2022-02-02
Remove `force` toggle for unauthorised users from Mint Token dialog

### LINKED ISSUE #3147: Mint token dialog has `force` toggle even when the user doesn't have permissions
## Steps to reproduce

1. Create a colony
2. Install & enable voting extension
3. Log in as another user and join the colony (this user shouldn't have any permissions)
4. Go to `Mint token` dialog

Expected  behaviour:
- There shouldn't be any `force` toggle & user shouldn't be able to use

Actual behaviour:
- There is a force toggle and user can initiate an action
- The action will result in error on the contracts side as the user doesn't have necessary permissions

The user doesn't have funding permissions
Screenshot 2022-01-29 at 17 43 13

danbr:
## Description

This PR updates the Mint token dialog. It removes the `force` toggle only for users that do not have permissions. 

After fix, unauthorised user:

Screenshot 2022-02-01 at 17 58 17

Authorised user:

Screenshot 2022-02-01 at 18 02 50

Resolves #3147

### DIFF
--- a/src/modules/dashboard/components/Dialogs/TokenMintDialog/TokenMintForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/TokenMintDialog/TokenMintForm.tsx
@@ -65,9 +65,8 @@ const TokenMintForm = ({
   const { walletAddress } = useLoggedInUser();
   const allUserRoles = useTransformer(getAllUserRoles, [colony, walletAddress]);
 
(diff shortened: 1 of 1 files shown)


chinins (approved):
Good to go!

ArmandoGraterol (approved):
Looking good Mr. Dan :sunglasses:

PR #3166, opened 2022-02-05
Add component to display Staked motions list

### LINKED ISSUE #3156: Add component to display Staked motions list
### Story

As a user, you want to be able to easily view a list of and navigate to the motions that you have staked on. To make it easier to claim your stakes or view the status of the motion.

### Description

By including a list of motions in the 'Stakes' tab, users will be able to easily view and navigate to motions they have staked on. This feature involves the following:

- Creating a component to list the user's motions they have staked on and are ready to claim and allow them to navigate to the motion.
- Adding the component to the 'Stakes' tab

### Design

[Figma link](https://www.figma.com/file/m96yJW1PPXezFl2mkuYha6/Token-Activation?node-id=2411%3A7599)

stakes-claim-all

- [ ] Add list of users claimable motions
- [ ] Add component to 'Stakes' tab

danbr:
## Description

This PR creates a component to list staked motions, that can be claimed via a single button.

Please note, this doesn't include a working link in the listItem to the motions details page. Retrieving the transaction hash was moved to a separate issue. https://github.com/JoinColony/colonyDapp/issues/3196

Resolves #3156 

Screenshot 2022-02-20 at 12 00 58

Screenshot 2022-02-20 at 12 01 44


src/modules/users/components/TokenActivation/StakesTab/StakesListItem.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { defineMessages } from 'react-intl';
+
+import Link from '~core/Link';
+
+import styles from './StakesTab.css';
+
+const MSG = defineMessages({
+  motionUrl: {
+    id: 'users.TokenActivation.TokenActivationContent.StakesListItem.motionUrl',
```
ArmandoGraterol, line 10: Is this the correct id? Shouldn't it be `TokenActivation.StakesTab.StakesListItem`?

src/modules/users/components/TokenActivation/StakesTab/StakesTab.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { FullColonyFragment } from '~data/index';
+
+import ClaimAllButton from './ClaimAllButton';
+import StakesListItem from './StakesListItem';
+
+import styles from './StakesTab.css';
+
+const MSG = defineMessages({
+  yourStakes: {
+    id: 'users.TokenActivation.TokenActivationContent.StakesTab.yourStakes',
```
ArmandoGraterol, line 18: Same here, it's `TokenActivation` instead of `TokenActivationContent` right?

ArmandoGraterol (changes requested):
Nice job! One small change that you could do so that the items look closer to the ones in the design is to remove the `border` and add a `margin-bottom: 1px`. This is how they look with that change:

FireShot Capture 542 - Colony - localhost

src/modules/users/components/TokenActivation/TokenActivationPopover.tsx
```diff
@@ -4,9 +4,8 @@ import Popover, { PopoverChildFn } from '~core/Popover';
 
 import { removeValueUnits } from '~utils/css';
 
-import TokenActivationContent, {
-  TokensTabProps,
-} from './TokenActivationContent/index';
+import TokenActivationContent from './TokenActivationContent/index';
+import { TokensTabProps } from './TokensTab/TokensTab';
```
ArmandoGraterol, line 8: You can use shorter imports here:
```
import TokenActivationContent from './TokenActivationContent';
import { TokensTabProps } from './TokensTab';
```

For the `TokensTabProps`, you would need to add the export to the index file of the `TokensTab`:
```
export { default, TokensTabProps } from './TokensTab';
```

ArmandoGraterol (changes requested):

src/modules/users/components/TokenActivation/StakesTab/StakesTab.tsx
danbr: Yep - 100%. I refactored into another dir and didnt update the ids. I'll add the update.

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
```diff
@@ -2,25 +2,40 @@ import React, { useState } from 'react';
 import { FormattedMessage, defineMessages } from 'react-intl';
 
 import { Tab, Tabs, TabList, TabPanel } from '~core/Tabs';
+import { useClaimableStakedMotionsQuery } from '~data/generated';
+
+import TokensTab, { TokensTabProps } from '../TokensTab';
+import StakesTab from '../StakesTab/StakesTab';
```
ArmandoGraterol, line 8: This import can be written as `../StakesTab`

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
```diff
@@ -2,25 +2,40 @@ import React, { useState } from 'react';
...
 
-import TokensTab, { TokensTabProps } from './TokensTab';
-import StakesTab from './StakesTab';
 import styles from './TokenActivationContent.css';
 
 const MSG = defineMessages({
   yourTokens: {
-    id: 'users.TokenActivation.TokenActivationContent.yourTokens',
+    id: 'TokenActivation.TokenActivationContent.yourTokens',
```
ArmandoGraterol, line 14: My bad, I think I explained myself badly. I wasn't saying that the ids shouldn't have the `users.`, just that the `TokenActivationContent` should've been removed from the ids in  `StakesTab` and `StakesListItem` (And any other component that is not used exclusively in the `TokenActivationContent`), but the `users.` in the ids is correct so long as they are inside the `users` module

src/modules/users/components/TokenActivation/TokenActivationPopover.tsx
```diff
@@ -4,9 +4,8 @@ import Popover, { PopoverChildFn } from '~core/Popover';
 
 import { removeValueUnits } from '~utils/css';
 
-import TokenActivationContent, {
-  TokensTabProps,
-} from './TokenActivationContent/index';
+import TokenActivationContent from './TokenActivationContent/index';
```
ArmandoGraterol, line 7: This `.../index` can be removed so that it's just `./TokenActivationContent`

ArmandoGraterol (approved):
Looking good! There are some small changes left but aside from that, I'm approving

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
danbr: I'll clean this up. thx

src/modules/users/components/TokenActivation/TokenActivationPopover.tsx
danbr: I'll clean this up. thx

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
danbr: @ArmandoGraterol Whats the guidance on MSG ids?
I checked other instances of ids and the string.paths used. Of course they need to be unique, and follow the dir hierarchy. (And should go back far enough for uniqueness)
But different people go back differing amounts. @rdig goes back to `core` but most others start from the first dir after `components`.

Is there a reason to go back as far as `users`? Other than personal choice? (just for me to understand and decide on a rule that i can apply consistently.)

arrenv (changes requested):
@danbr It is looking good. Some things I noticed:

- I think we need a max height on the tab, with a scroll bar, as in cases were you would have a lot of stakes to claim, it will be very long.

- The whole element has a hover state and pointer cursor, but it is not clickable.

- The hover state does not have a border in the design.

staked-items

src/modules/users/components/TokenActivation/TokenActivationContent/TokenActivationContent.tsx
ArmandoGraterol: Generally speaking, the criteria that are normally used (or at least the one that I use always) is to see where is the component going to be used.

So, if you take a look at the components that we have in the `core` module, most of them start with just the name of the component, they don't go as far as to include `core.` into the id and that's because the core components normally are **NOT** used, at least exclusively, inside the `core` module, they are used all over the dapp. 

Now, if you go to look at some components at the `dashboard` module, they most likely should begin with a `dashboard.` and that's because those components are exclusively used inside that module, they don't go outside it.

In summary, the reason why I would expect these ids you added/updated, to begin with, `users.` is because those components are just being used inside the `users` module and the id should reflect that. If they don't, it's assumed that these components are being used in other modules other than `users`.

danbr:
> @danbr It is looking good. Some things I noticed:
> 
> * I think we need a max height on the tab, with a scroll bar, as in cases were you would have a lot of stakes to claim, it will be very long.
> * The whole element has a hover state and pointer cursor, but it is not clickable.
> * The hover state does not have a border in the design.
> 
> staked-items

@arrenv Thanks. As discussed, i've updated 2 out of 3 issues that you identified (support scrolling & remove border). For the   'clickable' issue, i need to change the behaviour and deal with the URLs so i need the txHashes. So it would be better to tackle that issue once i have merged #3212 into this branch.

chinins (approved):
Well done!

arrenv:
@danbr regarding what you raised, I have played around a little, but I am going to try something tomorrow to see if I can get the `Claim All` button to clear out the list. Otherwise, I think it should be fine.

arrenv (changes requested):
@danbr it is working well!



The notification bubble looks like the font size of the number needs to be smaller, when compared to the design and other instances of it.

7fcfef6c-8e6f-4c07-b9aa-91fbf801c29d

Other instances: 
download (16)
)


Having the scrolling is definitely better.

d5ad41cd-f3af-456e-b677-df5a01efa87f


Lastly, regarding the claimable motions not disappearing, I think we can get rid of `TokenActivation` popover when the claim all starts, as it overlaps it anyway. This seems to fix the issue.

danbr:
> @danbr it is working well!
> 
> The notification bubble looks like the font size of the number needs to be smaller, when compared to the design and other instances of it.
> 
> 7fcfef6c-8e6f-4c07-b9aa-91fbf801c29d
> 
> Other instances: download (16) )
> 
> Having the scrolling is definitely better.
> 
> d5ad41cd-f3af-456e-b677-df5a01efa87f
> 
> Lastly, regarding the claimable motions not disappearing, I think we can get rid of `TokenActivation` popover when the claim all starts, as it overlaps it anyway. This seems to fix the issue.

@arrenv reduced font size slightly
Screenshot 2022-03-02 at 12 58 02

Now closing the `TokenActivation` popover when Claim all button is pressed. Just wondering if you tested using the same implementation?

arrenv:
@danbr This is how it is appearing on my screen:

4b8e2260-9421-4369-9de2-7c2f103c430d
.

danbr:
> @danbr This is how it is appearing on my screen, i.e. the number in the notification is not centered:
> 
> 4b8e2260-9421-4369-9de2-7c2f103c430d .

@arrenv 
I've reduced the size slightly more to match other dots, and improved the positioning (at least on my system)

Screenshot 2022-03-04 at 11 35 35

arrenv (approved):
@danbr Great, looking good for me now.

48b72b77-6451-4682-bc6a-a659ac94686b

PR #3173, opened 2022-02-08
Fix alignment of max button on smite dialog

ArmandoGraterol:
It seems like at some point, don't know why, something in the styles of the max button changed and it resulted in the button being badly aligned in `master`:

Before:
FireShot Capture 535 - Colony - localhost

After:
FireShot Capture 534 - Colony - localhost

### DIFF
--- a/src/modules/dashboard/components/Dialogs/SmiteDialog/SmiteDialogForm.css
+++ b/src/modules/dashboard/components/Dialogs/SmiteDialog/SmiteDialogForm.css
@@ -27,7 +27,7 @@
 }
 
 .inputContainer button {
-  right: auto;
+  right: initial;
   left: -6px;
 }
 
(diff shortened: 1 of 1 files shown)


dit7ya (commented):
Slightly related. It looks like the `Amount of reputation...` section is not aligned (slightly toward to left) with the other two sections.

dit7ya:
@ArmandoGraterol Some flavor of Linux (NixOS) + either of Chrome and Firefox.

ArmandoGraterol:
> @ArmandoGraterol Some flavor of Linux (NixOS) + either of Chrome and Firefox.

@dit7ya and this screenshot is from Chrome? It's weird that you are seeing that because for me the distance between the border of the modal and the fields is created using padding and all of the sections have a `20px` padding

arrenv (approved):
Looks okay to me, although, I did not have an issue before on Windows with Chrome. As left is overriding.

Although, my approach to this would be to use `right: initial;` what do you think @ArmandoGraterol ?

max-alignment

dit7ya (approved):
Sorry my bad, was something wrong on my end I guess. Looks fine now.

PR #3188, opened 2022-02-14
Make shabang in provision script more compatible

dit7ya:
In some systems `bash` is not located at `/bin/bash` - for example in NixOS it is symlinked at `/run/current-system/sw/bin/bash`. `/usr/bin/env bash` searches `PATH` for the bash executable.

### DIFF
--- a/scripts/provision_submodules.sh
+++ b/scripts/provision_submodules.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 while [ $# -gt 0 ]; do
   case "$1" in


arrenv:
@dit7ya I have looked into this, it works, however it apparently can cause security risks. So, I don't think it should committed to master. Have you looked into the potential risks? What do you think?

dit7ya:
What kind of security risk can there be do you think?

arrenv:
@dit7ya here is some context - https://unix.stackexchange.com/questions/206350/what-is-the-difference-if-i-start-bash-with-bin-bash-or-usr-bin-env-bash

Although, I am not really familiar with it, and may not even be relevant.

rdig:
This should stay as `#!/bin/bash` both for the reasons Arren mentioned, as well as for compatibility with MacOS systems.

The codebase should be as general as possible, and it's the developer's environment that should adapt to the codebase, not the other way around, because as many developers, as many different, custom, environments.

If you run something else, that doesn't have standard locations expected by the codebase, just create symlinks locally to adapt it. Again, the machine / OS of you as Colony developer, should adapt to use the various Colony codebases.

_(this does not however block any future changes, that can benefit developers at large, but in this case, it does not)_

We can't proceed further with this. Please close it.

PR #3189, opened 2022-02-14
Add verified icon to MembersListItem

### LINKED ISSUE #3183: [UI] Add tick with tooltip to whitelisted addresses on Members page
## Specification
* Add tick with popover to item on members page
* It should be displayed if user is whitelisted [Logic will be implemented in different tasks, so just add bool]

https://www.figma.com/file/Phby7kEYihhvGgQn69A314/Verified-Recipients?node-id=79%3A15782

dit7ya:
## Description

**New stuff** ✨

* Adds an optional check-mark of primary color in the MembersListItem component.

**Changes** 🏗

* Changed `CopyableAddress` to `InvisibleCopyableAddress` in the same component.

**Testing the PR**

Since the logic for whitelisting will be implemented in different PR, try the following hack to test it locally.

On https://github.com/JoinColony/colonyDapp/blob/ae47419bae3068bdd79d78f3d615776ce4803104/src/modules/core/components/MembersList/MembersListItem.tsx#L56

change 

```diff
-    isWhiteListed,
+    isWhiteListed = true,
```

and visit the Members page on your colony.

Resolves #3183.

### DIFF
--- a/src/modules/core/components/MembersList/MembersListItem.tsx
+++ b/src/modules/core/components/MembersList/MembersListItem.tsx
@@ -7,10 +7,10 @@ import React, {
 } from 'react';
 import { AddressZero } from 'ethers/constants';
 
+import { defineMessages } from 'react-intl';
 import { createAddress } from '~utils/web3';
 import UserMention from '~core/UserMention';
 import { ListGroupItem } from '~core/ListGroup';
-import CopyableAddress from '~core/CopyableAddress';
 import { AnyUser, Colony, useUser } from '~data/index';
 import { ENTER } from '~types/index';
 import HookedUserAvatar from '~users/HookedUserAvatar';
@@ -18,6 +18,9 @@ import { getMainClasses } from '~utils/css';
 import MemberReputation from '~core/MemberReputation';
 
 import styles from './MembersListItem.css';
+import InvisibleCopyableAddress from '~core/InvisibleCopyableAddress';
+import MaskedAddress from '~core/MaskedAddress';
+import IconTooltip from '~core/IconTooltip';
 
 interface Props<U> {
   extraItemContent?: (user: U) => ReactNode;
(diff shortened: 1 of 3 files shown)


danbr (approved):
Looks good (who doesn't love a ✅)

arrenv (changes requested):
@dit7ya It looks like you have included some extras in there, you may need to check your rebase.

Also, the IconTooltip provides a prop `tooltipClassName` that allows you to override the width of the tooltip. So, you don't have the empty space.

dit7ya:
@arrenv Fixed that in the last commit. Astonishing that I totally missed the extra space - it is right there even in the screenshot.

arrenv (approved):
Looks good now!

whitelisted-tick

src/modules/core/components/MembersList/MembersListItem.tsx
```diff
@@ -127,7 +139,22 @@ const MembersListItem = <U extends AnyUser = AnyUser>(props: Props<U>) => {
...
           <div className={styles.address}>
-            <CopyableAddress>{walletAddress}</CopyableAddress>
+            <InvisibleCopyableAddress address={walletAddress}>
+              <MaskedAddress address={walletAddress} />
+            </InvisibleCopyableAddress>
+            {isWhiteListed && (
+              <IconTooltip
+                icon="check-mark"
+                tooltipText={MSG.whitelistedTooltip}
+                iconSize="13px"
```
chinins, line 149: I am not a big fan of passing `13px` like this. It's not about your PR, it's coming form `IconTooltip`, I think it would be good to create an issue to add `appearance` prop with iconSize key and have several standard options there - similar to what is done in other core components...
@arrenv @ArmandoGraterol what do you think?

chinins (approved):
Nice work!

If Armando & Arren agree with me about `IconTooltip` can you please raise an issue int he Maintenance project for that? Thank you.

src/modules/core/components/MembersList/MembersListItem.tsx
arrenv: @chinins I agree, good thinking!

dit7ya: Right. On a related note, I also want to point out that the Tooltip/ Popover component we are using is a bit hacky. For example, you cannot right click and Inspect Element on a Tooltip-ed component - try the colony address for instance.

PR #3193, opened 2022-02-15
Show the correct token in the events list

### LINKED ISSUE #3148: Wrong token displayed in Events log after Claiming
## Steps to reproduce

1. Send CLNY (in this case) to a colony (note: CLNY had not been added through the Manage tokens flow yet)
2. Claim on Events page
3. The output in the Events log displays that the native token for that colony is what got claimed

## Expected behaviour

"<user> claimed <amount>CLNY for colony" displayed in Events log

## Actual behaviour

<user> claimed <amount><nativeColonyToken> for colony" displayed in Events log

arrenv:
## Description
This PR includes a token info query in the event list to ensure the correct token is displayed. Especially when the token as not been added to the colony.

**New stuff** ✨
* Adds token query to `ColonyEventsListItem`

## Testing

* Create two colonies with different tokens.
* Send tokens from one colony to the other (without adding the second colonies tokens to the first colony's token list).
* Go to events and claim the token transfer.
* You should now see the correct token symbol on the event.

## Before

claiming-tokens-not-in-manage-before

## After

claiming-tokens-not-in-manage-after

Resolves #3148

### DIFF
--- a/src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.tsx
+++ b/src/modules/dashboard/components/ColonyEvents/ColonyEventsListItem.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useEffect } from 'react';
 import {
   FormattedDateParts,
   FormattedMessage,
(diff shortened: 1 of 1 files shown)


chinins (approved):
All good here. But the other PR was also good, I think that the network-only fetch policy was a good addition. I just thought that the filtering was unnecessary. You could have just removed it from that PR. I am sorry if I wasn't clear and caused all of the extra PR opening-closing...

ArmandoGraterol (approved):
Looks good to me :ok_hand:

arrenv:
@chinins no problem at all, just thought this would be a lot cleaner.

Issue #3186, opened 2022-02-11
Add ‘confusable’ characters validation to Payment modal and to user display name creation


alicjakujawa:
## Specification
* Use created in separate ticket validation method in Payment modal 
* Display warning if user name dont pass validation
* Use this validation method also on user display name creation

### Design (optional, might not be applicable)
https://www.figma.com/file/Phby7kEYihhvGgQn69A314/Verified-Recipients?node-id=79%3A11161

alicjakujawa:
Blocked by https://github.com/JoinColony/colonyDapp/issues/3185

PR #3194, opened 2022-02-15
Add "confusable characters present" warning

### LINKED ISSUE #3185: Create validation method for  'confusable' characters
## Specification
* Create validation method for  'confusable' characters - see example implementation here: https://github.com/MetaMask/metamask-extension/pull/9187
* It will be used in Payment modal AND while creating user display name
* UI Part and usage will be handle in separate ticket

https://www.figma.com/file/Phby7kEYihhvGgQn69A314/Verified-Recipients?node-id=79%3A11161

ArmandoGraterol:
Added `unicode-confusables` package which comes with a `isConfusing` function that detects if the string has any confusable letters/characters, now a warning should be shown in the payment modal when the user has a confusable character in their name.

An example of a name with a confusable character: `fоо`

FireShot Capture 538 - Colony - localhost

Resolves #3185 and #3186

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -78,6 +78,7 @@
         "reselect": "^4.0.0",
         "scroll-into-view-if-needed": "^2.2.16",
         "subscriptions-transport-ws": "^0.9.18",
+        "unicode-confusables": "^0.1.1",
         "userflow.js": "^2.2.0",
         "utility-types": "^3.7.0",
         "web3-eth-abi": "1.3.0",
@@ -54958,6 +54959,11 @@
         "node": ">=4"
       }
     },
+    "node_modules/unicode-confusables": {
+      "version": "0.1.1",
+      "resolved": "https://registry.npmjs.org/unicode-confusables/-/unicode-confusables-0.1.1.tgz",
+      "integrity": "sha512-XTPBWmT88BDpXz9NycWk4KxDn+/AJmJYYaYBwuIH9119sopwk2E9GxU9azc+JNbhEsfiPul78DGocEihCp6MFQ=="
+    },
     "node_modules/unicode-match-property-ecmascript": {
       "version": "1.0.4",
       "resolved": "https://registry.npmjs.org/unicode-match-property-ecmascript/-/unicode-match-property-ecmascript-1.0.4.tgz",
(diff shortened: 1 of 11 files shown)


dit7ya (commented):
Looks good at the first glance. 



I am pondering if it is feasible and a good idea to restrict creation of usernames to only Latin lowercase letters and numbers.

Would also be a good idea to show the punycode representation of the rogue username and a Learn More link for people unfamiliar with homoglyph attacks.

arrenv (commented):
@ArmandoGraterol It looks to be working really well, nice work.

It is not mentioned in the issues, however, this should also be applied to the main username creation. Which seems to pass as long as the first character is not confusable.

Do you want to include it in this PR, or do you want it as a separate issue?

For reference:

confusable-characters

ArmandoGraterol:
> I am pondering if it is feasible and a good idea to restrict creation of usernames to only Latin lowercase letters and numbers.

I personally, don't think it's a good idea, the users should be able to create their username in their respective native language without restrictions.

> Would also be a good idea to show the punycode representation of the rogue username and a Learn More link for people unfamiliar with homoglyph attacks.

I think the current warning is enough for the average user tbh, not everybody is curious enough to want to know the details, and the curious ones might as well see the "This username has confusable characters" and check out for themselves what are confusable characters and start from there. Then again, I'm not a UX person so if you feel strongly about adding this in then it could be useful to bring it to Product to see what they think.

ArmandoGraterol:
> It looks good - nice work. The only thing that I would potentially change is the copy when it's shown for your own username. I would be confused if I saw the warning as it is when I try to change my name. Maybe it's worth confirming with product again...

Yeah, I see how it could be confusing! I'm gonna confirm this with Product then

ArmandoGraterol:
> Do you want to include it in this PR, or do you want it as a separate issue?

I agree! That is also a place that should have a warning. I would prefer for that to be included in this PR, seeing that is _really_ closely related

chinins (approved):
Looks good!

arrenv (approved):
@ArmandoGraterol nice work, it is working great!

confusable-character-warning

confusable-character-warning-2

dit7ya (approved):
LGTM!

PR #3209, opened 2022-02-22
Fix max button calculation in `BuyTokens` widget

### LINKED ISSUE #3046: Max button does not work if max value is a decimal
## Steps to reproduce

1. Ensure CoinMachine is enabled and funded.
2. Ensure your wallet balance is low enough so that you are buying less then 1 token.
3. Click the 'Max' button under the amount to buy input.

## Expected behaviour

* Max button should allow decimal values.

## Actual behaviour

* Clicking max causes an error when the maximum purchasable is a decimal value.

max-button-decimals

chinins:
## Description

This PR fixes the bug when the max button doesn't work when max value is decimal.

**Changes** 🏗

- update `BuyTokens`

resolves #3046

### DIFF
--- a/src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
+++ b/src/modules/dashboard/components/CoinMachine/BuyTokens/BuyTokens.tsx
@@ -258,17 +258,17 @@ const BuyTokens = ({
         : bigNumberify(1);
 
       const maxUserBalancePurchase = userTokenBalance
-        .div(currentPrice)
         /*
-        when we divide by a number with moved decimal our final number
-        gets smaller by the '10 * the number of 0 that are added'
-        so we need to counteract it by multiplying it back
-         */
+      when we divide by a number with moved decimal our final number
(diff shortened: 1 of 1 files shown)


arrenv (approved):
@chinins looks like a good fix and makes sense. Nice one!

One unrelated issue I noticed is that the balance is not updating after a buy. In the gif you will be able to see the Balance remains at 93. Let me know if it is doing the same for you and I will raise a separate issue.

working-decimals-max

ArmandoGraterol (approved):
Looks good to me :+1: 

@arrenv I can confirm that the balance in the buy widget doesn't update after buying tokens. The balance displayed in the gas station does but in the widget it doesn't

chinins:
> @chinins looks like a good fix and makes sense. Nice one!
> 
> One unrelated issue I noticed is that the balance is not updating after a buy. In the gif you will be able to see the Balance remains at 93. Let me know if it is doing the same for you and I will raise a separate issue.
> 

@arrenv, I confirm that the balance is updated after a refresh only.

PR #3210, opened 2022-02-23
Setup Playwright and some minor tests

dit7ya:
This is just an experiment as of now. Don't worry about this until (and if) it is ready for review.

Reasons to prefer Playwright

- I have found Playwright to be more robust, easy-to-use, and featureful.
- Playwright is generally way faster than Cypress. This page has a speed benchmark - https://blog.checklyhq.com/cypress-vs-selenium-vs-playwright-vs-puppeteer-speed-comparison/. You can also write parallel tests in the same test file.
- This is subjective, but I find Playwright's async await syntax cleaner than Cypress's chaining.
- Playwright supports both headless and headed tests in multiple browsers in parallel, on your local machine, without subscription to anything. Cypress has a premium plan for some features.
- Playwright supports **multiple tabs and domains** - This will be useful when we will be testing something like sending funds from one colony to another.
- **Writing arbitrary wait times (which will only be much aggravated because of the dApp) is frustrating**. Playwright just automatically waits and has saner and cleaner choices around it.
- In my small anecdata of a couple of friends, everybody who switched to Playwright told me they like it way more than Cypress and are never going back. 
- Playwright is developed by Microsoft, has 34k stars on Github (Cypress has 36k but it is an older project), and generally has momentum as a dev tool.

A longer quick comparison - https://cathalmacdonnacha.com/cypress-vs-playwright-which-is-best-for-e2e-testing

Some extra notes I wrote somewhere else.

1. Playwright has its own set of debugging tools (including GUIs) - https://playwright.dev/docs/debug
It has a GUI trace viewer - https://playwright.dev/docs/trace-viewer and a GUI Inspector - https://playwright.dev/docs/inspector

2. It also has a very easy to use GUI Codegen tool which lets you `walkthough` the tests and automatically generate the test code - https://playwright.dev/docs/codegen

3. Some more comparison between the two - https://github.com/microsoft/playwright/discussions/11201 suggesting Playwright can achieve nearly all the things cypress can & more due to it being a fully scriptable browser.

4. One reason why Playwright is generally faster is that it is easily parallelizable. (Imagine us having 20 different E2E tests and how much difference a 4x execution time makes.) 

**I just want to remind everyone that though it might seem like I am pushing too hard for this, there is nothing personal about it and I will be happy to work with whatever the team finally agrees to use.**

Instructions for running the tests

- Start dev server but do nothing else
- You might have to run `npx playwright install` to install the browser binaries

Try both of the following
- `npx playwright test --headed playwright-tests/basic.spec.ts` for the headed tests (this is more visual)
- `npx playwright test playwright-tests/basic.spec.ts` for the headless tests (this is how it will run in CI)

- The other testfiles are not ready yet and will just fail - so don't worry about them.

 Contributes towards #3162.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [ main, master ]
+  pull_request:
+    branches: [ main, master ]
+jobs:
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2
+      with:
+        node-version: '14.x'
+    - name: Install dependencies
+      run: npm ci
+    - name: Install Playwright
+      run: npx playwright install --with-deps
+    - name: Run Playwright tests
+      run: npx playwright test
+    - uses: actions/upload-artifact@v2
+      if: always()
+      with:
+        name: playwright-report
+        path: playwright-report/
+        retention-days: 30
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,5 @@ ganache-accounts.json
 /dist
 /styleguide
 .dappPids.json
+test-results/
+playwright-report/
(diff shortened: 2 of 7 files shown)


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
dit7ya, line 1: Please ignore this Action file. This needs some more work.

PR #3211, opened 2022-02-23
Improve Icon's alignment

### LINKED ISSUE #2825: Fix question mark icon alignment
The question mark icon alignments is different in several places in the app. This is caused by different people implementing parts of the app - we have different OS and it seems that the misalignment is caused by this.

This is a link to the initial notion QA issue:

https://www.notion.so/colony/QA-008ecbf3acb84c6281ef8b619f5fcb34?p=19571a27bcd14959a6952659a3541033

Both Raul and I had a go at trying to fix it in this PR (just to see the code of some attempts):
https://github.com/JoinColony/colonyDapp/pull/2484

ArmandoGraterol:
An attempt to improve the alignment of the many icons that we use in the dapp, with an emphasis on the `QuestionMarkTooltip` icon.

I removed a CSS property from the core `Icon` component that, as expected, affects all of the icons in the dapp so make sure to watch out for any misaligned icon.

FireShot Capture 546 - Colony - localhost

FireShot Capture 544 - Colony - localhost

Resolves #2825

### DIFF
--- a/src/modules/core/components/DecisionHub/DecisionOption.css
+++ b/src/modules/core/components/DecisionHub/DecisionOption.css
@@ -36,6 +36,7 @@
 
 .rowIcon {
   flex: 0 0 50px;
+  height: 16px;
   text-align: center;
 }
 
--- a/src/modules/core/components/Fields/InputLabel/InputLabel.css
+++ b/src/modules/core/components/Fields/InputLabel/InputLabel.css
@@ -1,5 +1,7 @@
 .main {
-  display: block;
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
   margin-bottom: 6px;
   width: 100%;
   position: relative;
@@ -58,12 +60,6 @@
   display: flex;
   justify-content: space-between;
   align-items: baseline;
-}
-
-.extra {
-  position: absolute;
-  top: 0;
-  right: 0;
 }
 
 .stateScreenReaderOnly {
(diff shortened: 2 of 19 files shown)


arrenv (approved):
@ArmandoGraterol Nice one, it all looks good on my Windows.

icon-alignment-1

icon-alignment-2

icon-alignment-3

icon-alignment-4

chinins (approved):
Oh, finally!!! Looking great and the code is much better! It's amazing that you managed to fix it!

dit7ya (changes requested):
A great effort, so happy about this PR!

I found a couple of places where the Icons are not vertically centered (at least on Linux).




Also, should the following be aligned with the text `Buy <tokenName>`?





Also, I noticed something bizarre which is not relevant to this PR but to the Confusable characters PR - plain English character `m` is being treated as a Confusable.

I know `m` is commonly confused with `rn` but is that a strong enough reason? And if that is the case, maybe we should show which character is problematic to the user.

I will create a separate issue if you guys think this is a bug.

Screenshot 2022-02-28 at 11-59-41 Colony
Screenshot 2022-02-28 at 12-00-00 Colony

danbr (approved):
Nice work. These types of issues are a constant annoyance and often require a lot of effort, time, and patience to debug. Well done for resolving.

ArmandoGraterol:
> I found a couple of places where the Icons are not vertically centered (at least on Linux).
> 
> 

Those icons are correctly aligned according to the designs in Figma, they aren't supposed to be vertically centered.

> Also, should the following be aligned with the text `Buy <tokenName>`?
> 
> 

Nope, they should not be aligned with the text, according to the specs that we have, so they are also good :)

dit7ya (approved):
@ArmandoGraterol Cool! Then I think this is good to go.

PR #3227, opened 2022-03-07
Add test for extension management

### LINKED ISSUE #3208: Add testing for extensions
Extension actions:
- install
- enable
- deprecate
- reactivate
- uninstall

ArmandoGraterol:
Added tests related to the comment functionalities.

They should be testing the functionality related to extension management (Installing, enabling, deprecating, re-enabling, and uninstalling).

Resolves #3208


arrenv (approved):
@ArmandoGraterol  The test make sense and are working for me.

Are you only running the Whitelist extension test intentionally, because it has a variance in the `testExtensionManagementFlow` function? Or, should all tests be running here?

ArmandoGraterol:
=> Are you only running the Whitelist extension test intentionally, because it has a variance in the `testExtensionManagementFlow` function? Or, should all tests be running here?

Oh shiii, I forgot to remove the `.only` after I was done testing. Thank you for pointing it out Arren

chinins (commented):
I see linting errors in your test files. Can you please check and fix them? As per my comment about the other PR.

So I start seeing this at some point while running tests... not sure why this is happening.
Screenshot 2022-03-10 at 17 32 39

cypress/e2e/4-extensions.js
```diff
@@ -0,0 +1,76 @@
...
+  );
+
+  // Can enable extension
+  cy.getBySel('close-gas-station-button').click();
+  cy.getBySel('enable-extension-button').click();
+  if (extensionId === Extension.Whitelist) {
+    cy.getBySel('policy-selector').eq(1).click({ force: true });
+  }
+  cy.getBySel('setup-extension-confirm-button').click().wait(5000);;
+  cy.getBySel('extension-status-tag').should(
```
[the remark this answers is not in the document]
ArmandoGraterol: For all the waits you pointed out, it would not be possible to use timeout as by the way the selector is implemented there's **always** going to be an `extension-status-tag` available for selecting, so if we don't use wait then we risk grabbing the wrong one because the UI will not update fast enough.

With that being said, I have an idea that could possibly allow the use of timeouts instead of wait by changing the selectors, so I'll try that

src/modules/dashboard/components/Extensions/ExtensionStatus.tsx
```diff
@@ -62,9 +63,19 @@ const ExtensionStatus = ({
...
+          text={status}
+          data-test={`${kebabCase(status.defaultMessage)}-status-tag`}
+        />
+      ) : null}
       {installedExtension && installedExtension.details.deprecated ? (
-        <Tag appearance={{ theme: 'danger' }} text={MSG.deprecated} />
+        <Tag
+          appearance={{ theme: 'danger' }}
+          text={MSG.deprecated}
+          data-test="deprecated-status-tag"
```
chinins, line 77: Any reason why you are going for the kebab case? I chose camel case and I also noticed that  both camel case and kebab case were used for the data-test in the dapp before (you can see leftover ones in some components). I don't mind either way, I think we should just agree on one and try to use it consistently across the dapp.
@arrenv @ArmandoGraterol  what do you think?

chinins (commented):
I like it very much how you got rid of the wait! 👍

I have a problem when running tests. It quits abruptly at some point. I thought it was caused by selecting a policy in the whitelist. I noticed that on my machine (not sure if this is the case on other machines) when you select a policy on the whitelist extension it unselects it so you need to select twice.
https://user-images.githubusercontent.com/34057551/158182262-22f9a78e-0cd8-4ce3-8c05-ace7614d7aee.mp4

I skipped the whitelist tests and I still get the same quitting abruptly
Mar-14-2022 16-41-24

Any ideas what might be causing it?
@danbr @dit7ya can you please try running tests on your machine to check if you have the same problem?

src/modules/dashboard/components/Extensions/ExtensionStatus.tsx
arrenv: @chinins I have so far only noticed CamelCase being used already in the Dapp, so, I have gone with that. I also prefer it, as with CamelCase you can double click to select the text vs. having to drag to select the text.

ArmandoGraterol:
> I skipped the whitelist tests and I still get the same quitting abruptly Mar-14-2022 16-41-24
> 
> Any ideas what might be causing it? @danbr @dit7ya can you please try running tests on your machine to check if you have the same problem?

In my experience, this happens whenever I have a lot of things opened and I'm running the tests. I think this is the way Chrome tells you that it crashed because it doesn't have enough ram available to continue. That has been my experience with this error screen tho, it could be something else.

chinins (commented):
The coinmachine test is passing, the voting reputation and whitelist tests are failing:
Screenshot 2022-03-16 at 16 03 48

chinins (approved):
I've just run all the tests in the headless mode and they passed. So I think that there is something with my environment/machine that prevents them to pass in the browser.

I don't want to torture you any more. Nice work!

ArmandoGraterol:
> I've just run all the tests in the headless mode and they passed. So I think that there is something with my environment/machine that prevents them to pass in the browser.
> 
> I don't want to torture you any more. Nice work!

Thank you @chinins! I'm gonna merge this but I'm still gonna take a look at that error. It's a weird error, as it looks like it _should_ not be related to the tests but then again, it seems like it's happening only when running them

PR #3233, opened 2022-03-09
Add LoadingTemplate delayed/failed updates

### LINKED ISSUE #3229: Add dapp loading delaying and failed updates
#### dApp taking some time to load 

If after 15 seconds the DApp hasn’t loaded: 

- Show a message to the user with the following copy: 
“As you can see, it’s taking a while to connect to the service.”
“Please hold tight while we keep trying. Sorry, we know this is boring.”

- Only add the copy, there is no need to change the existing loading animation

Design - [Figma](https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=269%3A8094)

#### dApp not loading in long time 

If after 30 - 60 seconds the DApp hasn’t loaded, whichever is most appropriate (we can have a variable for this that can be attuned later for different kinds of deployment scenarios) -

- Show a message to the user with the following copy: 
“Oh noes! We tried and tried but couldn’t get any response from the service.”
”We’re probably already trying to fix this, so please try again later.”

Design - [Figma](https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=269%3A8265)

dit7ya:
## Description

Adds status updates in the LoadingTemplate components - as per the specification from the issue. The timers are of 15 seconds and 30 seconds for delaying and failing respectively.

Note: The timer starts once the main script (or this component)) is loaded.

Since there are just two issues in this project I have just left `master` as the base branch. If you think there should be a feature branch, please let me know.

Resolves #3229.


arrenv (approved):
While hard to test accurately, unless you have some ideas. It looks to be working nice, good stuff.

loader-error-state

ArmandoGraterol (approved):
LGTM :+1: It was indeed difficult to test, I had to manually change the conditional to always show the loading state in order to actually see the different components

src/img/naked-mole-without-bg.svg
```diff
@@ -0,0 +1,19 @@
+<svg width="223" height="116" viewBox="0 0 223 116" fill="none" xmlns="http://www.w3.org/2000/svg">
```
chinins, line 1: I am no expert about svgs but I know that we always try to keep viewbox to `0 0 30 30` (please, see other svg examples in the repo). Is there any reason we can't have viewbox with these values?

chinins (commented):
It's looking good, just have a question about svgs.

dit7ya:
@arrenv and @ArmandoGraterol, there is a Network Throttle setting in the browser dev tools which can be used to test these kind of things (You need to enable the throttle after letting the page load once and refresh though, otherwise the initial SPA will take forever to load). However, while developing I followed what Armando mentioned for quick iterations.

https://developer.chrome.com/docs/devtools/network/#throttle

@chinins I just copied the svg from Figma. The other `naked-mole.svg` in the same folder has `0 0 600 150` as viewbox and neither do I have much idea about the implications of this. What do you suggest as the values? I am going to prettify that file btw.

src/img/naked-mole-without-bg.svg
ArmandoGraterol: I think the `0 0 30 30` here is not needed as this isn't an icon, so therefore it doesn't follow the 30x30 size standard that we have for icons in the dApp

chinins: I thought we have it for all svgs, but the other naked mole doesn't have it indeed. Thank you for clarification.

chinins (approved):
Nice work!

PR #3240, opened 2022-03-14
Separate number formatting from display, update Standard Number display format rules

### LINKED ISSUE #3236: Update formatting rules for Standard Number Display
**Description**

As a result of QA testing and discussions, update formatting rules for Standard Number Display to implement the agreements.

danbr:
## Description

This PR implements the outcome of discussions on standard number display. The notable changes include a max length = 6, and handle very large numbers >= 1000000000000000.

https://www.notion.so/colony/Standardized-Number-Display-Format-334f22de4c824d5dbb49fdcdc730d1b7
https://stackblitz.com/edit/number-format-playground-yfjmhc?file=numberFormatter.ts

Screenshot 2022-03-11 at 14 55 13

Screenshot 2022-03-11 at 15 55 37

## Explanation
When formatting very large numbers engineering (standard scientific) notation will be used.  The exponent will be superscripted.... (this small requirement caused many changes)

Previously, getFormattedTokenValue or the numberFormatter returned a string with the (decimal) formatted value. This output could be displayed directly (like in most situations) or passed to <Numeral> to be displayed.

When handling very large numbers, the returned formatted number string will contain HTML tags that superscript the exponent. Therefore, the string needs to be interpreted, and cannot simply be displayed.

This prompted the separation of responsibilities into 2 aspects.

1. Accessing (formatted) token value
2. Displaying numbers using Standard Number Display formatting rules (via <Numeral>)

I've tried to provide the same underlining functionality of being able to access the value in standard form. Where the number undergoes minimal formatting. 
Whenever a number is to be displayed, it should be passed to <Numeral> that will format the number (depending on large or small), average, reduce mantissa, etc..... and finally add pre/suf-fixes. 

To handle the potential that the formatted number string might contain HTML, <Numeral> then adds the output to a reference of an html object useRef<HTMLSpanElement>.

Almost all numerical displays now <Numeral>. There are several edge cases that need consideration, like:

-Coin machine - Numbers & prices. Some values are in ETH, so are less than < than 0.00000.... 
-Slider for motions  - i saw the issue from Shapeshift that suggest that some 'rounding' may be causing their 99% problems.. <Numeral> is not used in the slider.

When testing with very large numbers (325.363×10^39) protection must also be given to protect fields that have a 'max' button.

##Observations
Numbro: Complexities of competing formatting rules can cause unexpected outputs.

Setting `totalLength`:  seems to effect the formatting, with the number being averaged, despite being less than 6 digits, and average=false.
example:
```
 { num: 18785.68144 },
 { num: 235255 },
 { num: 400000 },

average:false
totalLength:  6

//output
18.7857K ETH
235.255K ETH
400K ETH

Without 
average:false
//totalLength:  6,

//output
18,785.68144 ETH
235,255 ETH
400,000 ETH
```

##Testing
This PR touches many parts of the Dapp, so should be tested rigorously. 
It is also a good candidate for making use of Cypress and our new testing infrastructure. This will certainly be beneficial today, and also in the future, should we need to refine the formatting rules further.

For testing and to identify fields as being Numeral'ified, i used:

Numeral.tsx: Line 63
``outputRef.current.innerHTML  =  `NUMERAL ${prefix  ?  `${prefix} `  :  ''}``

Resolves #3236

### DIFF
--- a/src/modules/core/components/Numeral/Numeral.tsx
+++ b/src/modules/core/components/Numeral/Numeral.tsx
@@ -1,8 +1,8 @@
-import React, { HTMLAttributes } from 'react';
+import React, { HTMLAttributes, useEffect, useRef } from 'react';
 import { BigNumber } from 'ethers/utils';
 
 import { getMainClasses } from '~utils/css';
-import { numberFormatter } from '~utils/numbers';
+import { numberDisplayFormatter } from '~utils/numbers';
 
 import styles from './Numeral.css';
 
@@ -11,7 +11,6 @@ interface Appearance {
   size?: 'medium' | 'large' | 'small';
   weight?: 'medium';
 }
-
 export interface Props extends HTMLAttributes<HTMLSpanElement> {
   /** Appearance object */
   appearance?: Appearance;
@@ -25,17 +24,17 @@ export interface Props extends HTMLAttributes<HTMLSpanElement> {
   /** Suffix the value with this string */
   suffix?: string;
 
-  /** Number of decimals to show after comma */
-  truncate?: number;
+  /** Number of mantissa digits to show */
+  mantissa?: number;
+
+  /** Total length of number to show */
+  totalLength?: number;
 
   /** Number of decimals to format the number with, or unit from which to determine this (ether, gwei, etc.) */
   unit?: number | string;
 
   /** Actual value */
-  value: number | string | BigNumber;
-
-  /** Should large number be truncate to 5 figures */
-  reducedOutput?: boolean;
+  value: string | BigNumber | number;
 }
 
 const Numeral = ({
@@ -43,28 +42,36 @@ const Numeral = ({
   className,
   prefix,
   suffix,
-  truncate,
+  mantissa,
+  totalLength,
   unit,
   value,
-  reducedOutput = true,
   ...props
 }: Props) => {
-  const formattedNumber = numberFormatter({
+  // formattedNumber could contain HTML,
+  // use outputRef to reference as an html object
+  const outputRef = useRef<HTMLSpanElement>(null);
+  const formattedNumber = numberDisplayFormatter({
     unit,
     value,
-    prefix,
-    suffix,
-    truncate,
-    reducedOutput,
+    mantissa,
+    totalLength,
   });
+
+  useEffect(() => {
+    if (outputRef.current) {
+      outputRef.current.innerHTML = `${prefix ? `${prefix} ` : ''}
+        ${formattedNumber}
+        ${suffix ? ` ${suffix}` : ''}`;
+    }
+  }, [outputRef, formattedNumber, prefix, suffix]);
 
   return (
     <span
       className={className || getMainClasses(appearance, styles)}
       {...props}
-    >
-      {formattedNumber}
-    </span>
+      ref={outputRef}
+    />
   );
 };
 
--- a/src/modules/core/components/EthUsd/EthUsd.tsx
+++ b/src/modules/core/components/EthUsd/EthUsd.tsx
@@ -30,9 +30,6 @@ interface Props extends NumeralProps {
   /** Should the suffix be visible? */
   showSuffix?: boolean;
 
-  /** Number of decimals to show */
-  truncate?: number;
-
   /** Ether unit the number is notated in (e.g. 'ether' = 10^18 wei) */
   unit?: string;
 
@@ -46,7 +43,6 @@ const EthUsd = ({
   appearance,
   showPrefix = true,
   showSuffix = true,
-  truncate = 2,
   unit = 'ether',
   value,
   ...rest
@@ -101,7 +97,6 @@ const EthUsd = ({
       appearance={appearance}
       prefix={showPrefix && valueUsd ? '~ ' : ''}
       suffix={showSuffix ? ` ${suffixText}` : ''}
-      truncate={truncate}
       value={valueUsd || '-'}
       {...rest}
     />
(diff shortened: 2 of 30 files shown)


arrenv (commented):
@danbr I have tested quite a bit and have not been able to notice any issues.

I believe the 1 million should still be the cut off for abbreviation. I think it just may have been overlooked in the Stackblitz.

There was only one place I noticed that didn't have the formatting, which you mention, however would formatting these numbers impact the rounding of the slider?

1d8fde9a-3d8c-4708-8912-7d797fd280f8

danbr:
> I believe the 1 million should still be the cut off for abbreviation. I think it just may have been overlooked in the Stackblitz.

Numbro implements this slightly differently. When totalLength is set to non zero, it has to use averages & abbreviation. Although i dont see why 400,000 is formatted to 400k.
But to use the average & abbreviation for 18785.68144 to 18.7857K makes sense.

To circumvent this behaviour, totalLength is set (0) when the number <= 6, otherwise totalLength:6.
You can play with different inputs here :
https://stackblitz.com/edit/number-format-playground-yncb7k?file=numberFormatter.ts

> There was only one place I noticed that didn't have the formatting, which you mention, however would formatting these numbers impact the rounding of the slider?

I made the updates to the `StakingSlider` so it now uses `<Numeral>`. It should not cause any issues, the slider doesnt use the 'display' numbers. I did some testing, and all fields displayed appropriate values.

"&&&&" is used to show where `<Numeral>` is being used.
Screenshot 2022-03-17 at 13 11 09

Screenshot 2022-03-17 at 13 11 59

Screenshot 2022-03-17 at 13 12 21


> -Coin machine - Numbers & prices. Some values are in ETH, so are less than < than 0.00000....

ETH price value -  I presume that we will leave this field unformatted by `<Numeral>`??

arrenv:
Ignore my recent activity, it seems GitHub was down for a little bit and was messing with me.

danbr:
> The dapp is breaking for me on the motions page now with the following error:

@arrenv 
This has nothing to do with `<Numeral>` or the formatting, other than the std number display project has opened the door to the Dapp dealing with very large number issues..........

The errror is now evident because you are testing with very large numbers, too large!! It's an overflow error thrown from BigNumber `toNumber()` when trying to convert. I can consistently reproduce.

```errors.js:76 Uncaught Error: overflow (operation="setValue", fault="overflow", details="Number can only safely store up to 53 bits", version=4.0.47)```

```
Number.MAX_SAFE_INTEGER = 9007199254740991; // Math.pow(2, 53) - 1;
```
Guess we need to give this some thought and come up with an approach that is universal.
We could wrap the conversion in try/catch - but a decision needs to be taken on how to fail gracefully.

> Also, there seems to be some alignment issues with the superscript maybe the <sup> can be positioned absolute? Although, that would mean that there would need to be forced padding after numbers only when <sup> is needed. The superscript seems to be a pain, however, I'm yet to think of an alternative solution.

Indeed superscript has added a lot of extra work. So on the issue of superscript affecting formatting & layout, i've made some updates that have fixed things. I've made the exponent slightly smaller, etc..... seems to be typical. Its easy to change if you feel it looks better with a different format.

Padding: I'm not sure i see the issue here. I think it look ok. Below are 2 images with & without a white space after the superscript. But if you think it needs attention I can work on an alternative approach.

Screenshot 2022-03-18 at 14 33 36

Screenshot 2022-03-18 at 14 33 19

This is not related to superscript (or this PR), but for me I always felt there was too much space between the number and symbol, on the main page total. There is a margin=13px.

Screenshot 2022-03-18 at 14 41 54


Placement: This should be universally fixed.

Screenshot 2022-03-18 at 14 16 30

Screenshot 2022-03-18 at 14 14 58


> This is in my notes as a point from the Product Sync, the main colony balance likely does not need to abbreviated. As it is a key figure, and has plenty of room, what do you think?

I partially agree!!!
I think it's important to show to the user the true (non-truncated) token balance somewhere. There is a lot of space on the main page. But as with everything in life.... there are limits, and the values we've been tested with would wrap to many lines. 
So i think a new "main balance display format" should be agreed!!!! Perhaps that means that no averages nor abbreviations are applied, and then once > 1000000000000000 - scientific notation is used.
I happy to do this, but let's create a new issue for the task.

File: src/modules/core/components/Numeral/Numeral.tsx [hunk 1]
```diff
@@ -25,47 +24,52 @@ export interface Props extends HTMLAttributes<HTMLSpanElement> {
...
-      className={className || getMainClasses(appearance, styles)}
-      {...props}
-    >
-      {formattedNumber}
-    </span>
-  );
+  useEffect(() => {
+    if (outputRef.current) {
+      outputRef.current.innerHTML = `${prefix ? `${prefix} ` : ''}
+        ${formattedNumber}
```
chinins, line 64: I really like how you cleaned  up adding the spaces and moved it all centrally here.

chinins (approved):
That's a big and complicated PR - well done! It works very well.

The scientific notation looks very nice
Screenshot 2022-03-18 at 15 16 12

I've just noticed that you are missing numeral in smite/award reputation action/motion  page:

Screenshot 2022-03-18 at 15 24 57
Screenshot 2022-03-18 at 15 26 43

I've also caught a bug but it's not related to your PR - I'll raise an issue for that
reputation

As for the "immense" numbers & "main balance display format", I agree that these should be separate issues. This PR is already very big.

ArmandoGraterol (approved):
Solid work! I know that dealing with number formatting can be complicated and sometimes annoying, from first-hand experience, so kudos to you :ok_hand:

arrenv (approved):
Great work on this one @danbr

chinins (approved):
The latest changes are all good!

danbr:
@chinins @arrenv @ArmandoGraterol 
I noticed a minor CSS issue (`StakingSlider` amount on wrong side & font not formatted), which i corrected with the latest commit. 
Sorry to ask you to review again!!

Before:
Screenshot 2022-03-23 at 11 01 05

After:
Screenshot 2022-03-23 at 11 00 26

Issue #3203, opened 2022-02-18
Add tests for user: updates, token management


chinins:
Testing for all actions with the user:

- [x] update the profile
- [x] token management
- [x] activation/deactivaiton of tokens
- [x] stakes claiming

PR #3241, opened 2022-03-15
Update User Profile & Tokens Tests

arrenv:
## Description

This PR adds testing for the Update User Profile settings and tokens. You will be able to run `7-updateUser.js` to see the tests in action.

Note: The add token tests requires that the colony's native token has not been added to the user's wallet. It would be possible to check this prior to add with some refactoring, or hard coding in a different token contract address. Happy to get some opinions on this.

**New stuff** ✨

* Test for updating a user's profile settings
* Test for adding and removing user avatar
* Add new token to users wallet
* Remove new token from users wallet

Contributes to #3203


cypress/e2e/7-update-profile.js
```diff
@@ -0,0 +1,43 @@
...
+  () => {
+    const userSettingsName = 'ColonyFan';
+    const userSettingsBio = 'I really like Colony';
+    const userSettingsWebsite = 'http://colony.io';
+    const userSettingsLocation = 'Earth';
+    const profilePicturePath = 'src/img/favicon.png';
+
+    beforeEach(() => {
+      cy.login();
+      cy.visit(`/landing`);
```
ArmandoGraterol, line 12: I think you don't need this here as `.login()` already visits `/landing`

cypress/e2e/7-update-profile.js
```diff
@@ -0,0 +1,43 @@
...
+      cy.getBySel('userSettingsLocation').should('have.value', userSettingsLocation);
+    });
+
+    it('User with permission can upload and remove profile picture', () => {
+      cy.getBySel('avatarUploaderDrop', { timeout: 40000 }).selectFile(profilePicturePath, { action: 'drag-drop' });
+      cy.getBySel('avatarUploaderDrop').trigger('drop', { force: true });
+      cy.getBySel('avatarUploaderRemove', { timeout: 40000 }).click();
+    });
+  },
+);
```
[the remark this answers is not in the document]
arrenv: Thank you, how did you get your linter to work?

cypress/e2e/7-update-profile.js
arrenv: That is good to know, thank you, updated!

chinins (changes requested):
This is not ready. Several points:
- linter is still not working 
- the input fields are not filled in properly - you just add to already existing ones, so the tests don't pass. You need to empty each field before you type something, otherwise it will just add up
- the avatar drag and drop is not working (I had the same problem), you can use `cy.getBySel('fileUpload').selectFile(filePath, { action: 'drag-drop' });` to fix it (you can look at my colony tests PR for a reference)
- for the user you also need to add edit tokens (add, remove) tests

Screenshot 2022-03-24 at 23 09 57

arrenv:
@chinins thank you for the review. That is a big help and clarified a few things for me. I have now updated the tests as well as added tests for user's wallet token management.

cypress/e2e/7-userSettings.js
```diff
@@ -0,0 +1,95 @@
...
+    cy.getBySel('userWallet').click();
+    cy.getBySel('editUserTokens').click();
+    cy.get('@colonyTokenAddress').then((colonyTokenAddress) => {
+      cy.getBySel('tokenSelectorInput')
+        .click()
+        .clear()
+        .type(colonyTokenAddress);
+    });
+    cy.getBySel('tokenEditSubmit').click();
+  });
```
chinins, line 82: I think that we need a test here to check if the token was added to the list.

cypress/e2e/7-userSettings.js
```diff
@@ -0,0 +1,95 @@
...
+    cy.getBySel('gasStationPopover', { timeout: 40000 }).click();
+    cy.getBySel('userWallet').click();
+    cy.getBySel('editUserTokens').click();
+    cy.getBySel('tokenEditItem')
+      .eq(1)
+      .get('label')
+      .get('[type="checkbox"]')
+      .uncheck({ force: true });
+    cy.getBySel('tokenEditSubmit').click();
+  });
```
chinins, line 94: And same here, we need a test to check if the token has been removed.

chinins (changes requested):
Looking great - very nice work! I think we just miss 2 test assertions at the end of add/remove token to check if the action was successful.

cypress/e2e/7-userSettings.js
arrenv: Good point, I have updated now with the check.

cypress/e2e/7-userSettings.js
arrenv: This has now been added.

cypress/e2e/7-update-profile.js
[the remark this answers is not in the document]
ArmandoGraterol: I didn't need the linter, GitHub already had a warning that a new line was needed while reviewing them.

However, if you haven't figured out how to use the dev env linter, as some time has passed ever since you wrote this comment, you can go to `package.json` and change the `lint` command from `eslint --fix 'src/**/*.{ts,tsx}` to `eslint --fix 'cypress/**/*.js`, that's what I do :)

cypress/e2e/7-userSettings.js
```diff
@@ -0,0 +1,118 @@
...
+    cy.getBySel('tokenEditSubmit').click();
+    // Check the number of tokens has increased by one
+    cy.get('@tokenCardCount').then((tokenCardCount) => {
+      cy.getBySel('userTokenCards')
+        .find(`[data-test="tokenCardItem"]`)
+        .should('have.length', tokenCardCount + 1);
+    });
+  });
+
+  it.only('User can remove token type from their wallet', () => {
```
ArmandoGraterol, line 97: It seems like this wild `.only` has escaped its enclosure

ArmandoGraterol (approved):
Looks good to me! Excellent work, just make sure to remove any `.only` before merging

cypress/e2e/7-userSettings.js
arrenv: Oh, the elusive `.only`. Thank you @ArmandoGraterol :)

src/modules/core/components/FileUpload/__tests__/__snapshots__/FileUpload.test.tsx.snap
```diff
@@ -155,7 +155,6 @@ exports[`FileUpload component Renders initial component 1`] = `
               </InputLabel>
               <div
                 className=""
-                data-test="fileUpload"
```
chinins, line 158: Why are you removing this data-test? I am surprised that you can push with it. I had to specifically add  it after adding `data-test` prop to the FileUpload, otherwise it wouldn't pass the push tests.

chinins (approved):
Very good now - nice work. I just have one question about the `data-test` in the `Fileupload.test.tsx.snap` file

src/modules/core/components/FileUpload/__tests__/__snapshots__/FileUpload.test.tsx.snap
arrenv: In my PR I set dataTest to a dynamic value `data-test={dataTest}`. Then when regenerating the snapshot it removed it for this reason. I did a search in the `cypress` directory to see if `fileUpload` was being used but could not find it used anywhere.

arrenv: @chinins I'm happy to change it to a static data-test attribute value in FuleUpload. What are your thoughts?

chinins: you are right, I actually removed it as I changed to using a different selector. Keep it like this - it's all good.

PR #3242, opened 2022-03-15
Add new Tooltip to token balance of the TokenActivation component 

### LINKED ISSUE #3231: Add new Tooltip to the token balance of the TokenActivation component
### Story

We want to provide an explanation of what the additional functionality is when clicking on the token balance of the TokenActivation  component.

### Description

The tooltip will need to be activated on hover, while still maintaining the functionality of the onclick popover.

The copy should be: "View and activate tokens for staking or claim any unclaimed stakes."

### Design

[Figma link](https://www.figma.com/file/m96yJW1PPXezFl2mkuYha6/Token-Activation?node-id=2645%3A7546)

Screenshot 2022-03-09 092720

danbr:
### Description

This PR adds a tooltip to the TokenActivationButton that provides an explanation of the functionality of the TokenActivation popover component.

To add clarity, the tooltip is displayed on hover, but is disabled when the TokenActivation popover component is displayed.

Screenshot 2022-03-15 at 18 17 27

Screenshot 2022-03-15 at 18 17 40

Resolves #3231

### DIFF
--- a/src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
+++ b/src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
@@ -15,7 +15,7 @@ const displayName = 'users.UserTokenActivationButton';
 
 const MSG = defineMessages({
   tooltip: {
-    id: 'pages.NavigationWrapper.UserNavigation.tooltip',
+    id: 'users.UserTokenActivationButton.tooltip',
     defaultMessage:
       'View and activate tokens for staking or claim any unclaimed stakes.',
   },
--- a/src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
+++ b/src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
@@ -20,7 +20,6 @@ const MSG = defineMessages({
       'View and activate tokens for staking or claim any unclaimed stakes.',
   },
 });
-
 interface Props {
   userLock: UserLock;
   nativeToken: UserToken;
@@ -77,6 +76,16 @@ const UserTokenActivationButton = ({
                   <FormattedMessage {...MSG.tooltip} />
                 </div>
               }
+              popperProps={{
+                modifiers: [
+                  {
+                    name: 'offset',
+                    options: {
+                      offset: [120, 10],
+                    },
+                  },
+                ],
+              }}
             >
               <div>
                 <span
--- a/src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
+++ b/src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
@@ -1,17 +1,25 @@
 import React from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
 import { bigNumberify } from 'ethers/utils';
 
 import { TokenActivationPopover } from '~users/TokenActivation';
-
 import { getFormattedTokenValue } from '~utils/tokens';
+import { Tooltip } from '~core/Popover';
 import Numeral from '~core/Numeral';
-
 import { FullColonyFragment, UserLock, UserToken } from '~data/index';
 import { Address } from '~types/index';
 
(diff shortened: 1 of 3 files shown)


src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
```diff
@@ -48,24 +56,38 @@ const UserTokenActivationButton = ({
...
-              }`}
-            />
-            <Numeral
-              suffix={` ${nativeToken?.symbol} `}
-              value={formattedTotalBalance}
-            />
+            <Tooltip
+              appearance={{ theme: 'dark', size: 'medium' }}
+              placement="bottom"
+              trigger={!isOpen ? 'hover' : 'disabled'}
```
dit7ya, line 70: Why not `isOpen ? 'disabled' : 'hover'` ?

dit7ya (approved):
Looks great! Works as expected.

src/modules/users/components/UserTokenActivationButton/UserTokenActivationButton.tsx
danbr: You say 'pOtatOes' i say 'pAhtAtoes'.... 

No real reason, either works. 
Although i tend to think of the 1st option as the primary flow, and 2nd as well..... secondary. So then it comes down to the meaning/logic of the flag vs meaning/logic of the action, and i was focused on the 'hover'.

danbr:
> As of previous comment.

@arrenv Well i felt talented getting it all working, but then felt like an imbecile when trying to change the placement of the tooltip!! I was banging my head against the wall for quite some time.

I've arrived at a solution, although it wasn't what i was originally planned. I think there might be a bug in our wrapper, or popper??? Playing with the props for arrow padding did nothing, or it's possible that i was doing something wrong.

The Tooltip correctly places itself relative to the reference element, but the arrow is not correctly positioned and i couldn't change its location.

Perfectly positioned, apart from arrow.
Screenshot 2022-03-16 at 12 28 38

I think this issue has already been discovered on reflection. BTW i was not able to trigger this tooltip on my system (screen capture from your pic)
Screenshot 2022-03-16 at 12 56 59


So I settled on this placement. I tested with very short/long token number (button width) and it worked well.
Screenshot 2022-03-16 at 12 35 38

I originally followed the Figma specs for the font, but now changed.

arrenv:
> I've arrived at a solution, although it wasn't what i was originally planned. I think there might be a bug in our wrapper, or popper??? Playing with the props for arrow padding did nothing, or it's possible that i was doing something wrong.

I have also had the same problems, the Popper does not play very nice. I have thought about looking into an upgrade, for example `https://atomiks.github.io/tippyjs/`, which uses the same library but adds onto it. Do you think it is worth looking into also after your experience?

> I originally followed the Figma specs for the font, but now changed.

Yeah, there is a difference in the Figma and what we use in the DApp because of rendering difference. It has happened a couple of times.

danbr:
> > I've arrived at a solution, although it wasn't what i was originally planned. I think there might be a bug in our wrapper, or popper??? Playing with the props for arrow padding did nothing, or it's possible that i was doing something wrong.
> 
> I have also had the same problems, the Popper does not play very nice. I have thought about looking into an upgrade, for example `https://atomiks.github.io/tippyjs/`, which uses the same library but adds onto it. Do you think it is worth looking into also after your experience?

@arrenv tippy looks good. 
Lots of ways to specialise the tooltips & popovers, very flexible & configurable. 
On the downside, it doubles the bundle size, compared to Popper (https://bundlephobia.com/package/tippy.js@6.3.7). It may not necessarily fix the arrow placement issue (to be investigated), and it would require a lot of effort to add to the Dapp. 

> 
> > I originally followed the Figma specs for the font, but now changed.
> 
> Yeah, there is a difference in the Figma and what we use in the DApp because of rendering difference. It has happened a couple of times.

Are there any other issues that need addressing in this PR? (a change is requested)

arrenv (approved):
@danbr good points. Nope, all looks good to me :)

PR #3243, opened 2022-03-15
Add tests for colony actions

### LINKED ISSUE #3205: Add a set of tests for actions - part 1
Tests for actions. Actions that should be covered:

Create payment, create/edit team,  transfer funds, award/smite reputation

ArmandoGraterol:
Added tests for the following actions:

- Payment
- Create team
- Edit team
- Transfer funds
- Smite
- Award

I would recommend that in order to test them, execute them in the order that they are in the file (Also make sure to have minted tokens). In the case of the smite action, make sure that the user being smitten has reputation beforehand.

Resolves #3205

### DIFF
--- a/cypress/e2e/3-actions.js
+++ b/cypress/e2e/3-actions.js
@@ -1,7 +1,12 @@
+import Decimal from 'decimal.js';
 import { bigNumberify } from 'ethers/utils';
+import ganacheAccounts from '~lib/colonyNetwork/ganache-accounts.json';
 
-describe('Colony can mint tokens via action', () => {
-  it('mint native tokens', () => {
+import { splitAddress } from '~utils/strings';
+import { createAddress } from '~utils/web3';
+
+describe('User can create actions via UAC', () => {
+  it('Can mint native tokens', () => {
     const amountToMint = 10;
     const annotationText = 'Test annotation';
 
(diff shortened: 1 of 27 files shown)


src/modules/core/components/Fields/Input/Input.tsx
```diff
@@ -135,6 +137,7 @@ const Input = ({
     disabled,
     maxLength,
     maxButtonParams,
+    'data-test': dataTest,
```
arrenv, line 140: I did this slightly differently, which is your preference? 

https://github.com/JoinColony/colonyDapp/pull/3241/files#diff-cda2cd04917ac9d136b544add03a27d755e47db79353775e5a0c060e7c6191d4R117

cypress/e2e/3-actions.js
```diff
@@ -53,4 +58,287 @@ describe('Colony can mint tokens via action', () => {
...
+    cy.login();
+
+    cy.visit(`/colony/${Cypress.config().colony.name}`);
+
+    cy.getBySel('colonyTotalFunds', { timeout: 30000 }).then(($text) => {
+      const amount = $text.text().split(',').join('');
+      prevColonyFunds = amount;
+    });
+
+    cy.getBySel('new-action-button', { timeout: 60000 }).click();
```
arrenv, line 82: All selectors should be camel case, right?

arrenv (changes requested):
All tests ran and passed for me with some timeout tweaks. This was a big one, nice work!

chinins (commented):
Very nice work!  There are just a couple of small points:
- on my machine the create payment action is not going through because you need to either click address or hit enter after you type it in
Screenshot 2022-03-25 at 15 31 20

- I thought we agreed to change the selectors to the camel case. Is there any specific reason for the kebab case here?

src/modules/core/components/Fields/Input/Input.tsx
ArmandoGraterol: Yours look cleaner :+1: I'm gonna check your change out

ArmandoGraterol:
> * on my machine the create payment action is not going through because you need to either click address or hit enter after you type it in

Are you sure about this? The address should be clicked by the test already, and in the screenshot, you are showing it looks like the colony doesn't have tokens to pay with.

> * I thought we agreed to change the selectors to the camel case. Is there any specific reason for the kebab case here?

 Nope, I just forgot to update those selectors after our agreement, my bad!

chinins (commented):
It's working perfectly. You are right about the payment action - my mind probably had a blackout at the end of the week 🙈

The only item is about selectors for the modal menu items. Please, have a look at my specific comment. Holding my approval until then.

cypress/e2e/3-actions.js
```diff
@@ -53,4 +58,287 @@ describe('Colony can mint tokens via action', () => {
...
+
+    cy.visit(`/colony/${Cypress.config().colony.name}`);
+
+    cy.getBySel('colonyTotalFunds', { timeout: 30000 }).then(($text) => {
+      const amount = $text.text().split(',').join('');
+      prevColonyFunds = amount;
+    });
+
+    cy.getBySel('newActionButton', { timeout: 60000 }).click();
+    cy.getBySel('indexModalItem').eq(0).click();
```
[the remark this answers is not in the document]
ArmandoGraterol: I can agree with that, because as you mentioned, in the future the order might change. Gonna wait a bit to see what the others think about it before implementing it or not.

src/modules/dashboard/components/ColonyHome/ColonyFunding/TokenItem.tsx
```diff
@@ -29,7 +29,10 @@ const TokenItem = ({
 
   return typeof balance === 'undefined' ? null : (
     <div className={styles.tokenItem}>
-      <span className={styles.tokenValue}>
+      <span
+        className={styles.tokenValue}
+        data-test={isTokenNative ? 'colonyFundingNativeTokenValue' : null}
+      >
```
chinins, line 35: I really like how you solved the problem here.

chinins (approved):
Very nice! It's a big one as well.

PR #3246, opened 2022-03-16
Fix dapp sometimes crashing on extension tests

ArmandoGraterol:
Added optional chaining operator to `details` prop.

Sometimes the prop can be `null/undefined` by the time the code tries to access it, causing the dapp to crash, notably when running the extension tests. I guess that would be because the dapp when you run the test runs slower than in the dev env, for example?

**To test**: Run the extension tests and check that the dapp doesn't crash during the tests or a bit after finishing the tests due to the `initialized` prop being inaccessible


arrenv (approved):
I was not getting any DApp crashing running the extension tests a few time. So, looks good to me.

chinins (approved):
Working great! Thank you for fixing!

PR #3220, opened 2022-03-01
Add address tab in ManageWhitelist modal

danbr:
## Description

-This pr adds content to the whitelisted tab, it allows addresses to be verified and submitted.
-It refactors `UploadAddressesWidget` into many parts to be more generic and allow use in its current location `Whitelist`, and support the new `ManageWhitelist` dialog.

-As the supporting logic is not ready, I use the same saga `updateWhitelist` for submitting.
-To allow addresses to be added, whitelist extension needs to be installed.

New stuff ✨

-`UploadAddresses` component
-`WhitelistedValidation` schemas

Single address input:
Screenshot 2022-03-10 at 15 38 08
Screenshot 2022-03-10 at 15 38 39

File input:
Screenshot 2022-03-10 at 15 38 22
Screenshot 2022-03-10 at 15 39 05

Resolves #3177

PR #3248, opened 2022-03-17
Whitelist UI dialog styles fix

### LINKED ISSUE #3245: Manage whitelist dialog UI updates
## Things to fix

1. Tabs border alignment 
Screenshot 2022-03-16 at 11 52 47

1. Modal content (we should be able to scroll content)
Screenshot 2022-03-16 at 11 52 58

dit7ya:
## Description

Made some styling changes.

For the context of the first screenshot see here - https://github.com/JoinColony/colonyDapp/pull/3220#issuecomment-1068964332

Resolves #3245.

### DIFF
--- a/src/modules/core/components/Dialog/Dialog.css
+++ b/src/modules/core/components/Dialog/Dialog.css
@@ -33,5 +33,5 @@
 }
 
 .overflowEnabled {
-  overflow: initial;
+  overflow: auto;
 }
--- a/src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialogForm.css
+++ b/src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialogForm.css
@@ -18,6 +18,8 @@
   display: flex;
   height: 100%;
   width: 100%;
+  position: relative;
+  bottom: -1px; /* HACK to overlap the tabs' bottom borders */
 }
 
 .toggleContainer {


alicjakujawa (changes requested):

arrenv (approved):
Looks good on my end. I think you just need to confirm Alicja's feedback.

05b5cabe-0a44-49ff-a436-343754133488

ArmandoGraterol (approved):
LGTM :ok_hand:

I agree with Alicja, if the change is only needed in 1 dialog, it's best to avoid changing the core component just in case

dit7ya:
@alicjakujawa I actually forgot to write about that in the PR. I wanted to ask @ArmandoGraterol about that change since that CSS was introduced by them just a month ago and is part of this project only. I thought changing that in the component will save us from having similar bugs when we use this component in other places in future. Of course I might be wrong. What do you think @ArmandoGraterol ?

ArmandoGraterol:
> @alicjakujawa I actually forgot to write about that in the PR. I wanted to ask @ArmandoGraterol about that change since that CSS was introduced by them just a month ago and is part of this project only. I thought changing that in the component will save us from having similar bugs when we use this component in other places in future. Of course I might be wrong. What do you think @ArmandoGraterol ?

I wrote that when I implemented the tooltip in the `Whitelisted` section because the tooltip would get cut off by the Dialogs always hiding the overflow. 

Currently, this change you made should only affect the manage whitelist dialog as far as I remember. Now it seems like the tooltip is not getting cut off anymore with or without your change **and** my change, so with that being said, I tested changing the `auto` back to `initial` and resized my screen to a smaller screen and it seems like your change is solving the modal not having a scroll in the smaller screens so I'm good with it :+1:

alicjakujawa (approved):

PR #3254, opened 2022-03-21
Prevent extra margin in DefaultMotion tags

### LINKED ISSUE #3253: Improve placement of tag in fully staked Motion message
## Steps to reproduce

1. Fully stake a Motion

The `Objection` tag has an unnecessary margin:10.

Perhaps there are other instances of tags that are poorly placed.

With margin:

Screenshot 2022-03-19 at 11 45 25

Without margin:

Screenshot 2022-03-19 at 11 48 17

dit7ya:
## Description

I the cause of the issue is https://github.com/joincolony/colonyDapp/blob/f09568540a45c9a29aea6fb49061ca796344ddc7/src/modules/core/components/Tag/Tag.css#L11

- the sibling selector kicks in when there are two tags in the same line - even if there is text in between them.

In DefaultMotion wrapping the Tag component in a `span` prevents that from happening.

Screenshot 2022-03-21 at 11-10-59 Pay halum 199 TOK Motion Colony - Notun

Resolves  #3253

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
@@ -134,7 +134,9 @@ const DefaultMotion = ({
     return Object.values(MOTION_TAG_MAP).reduce((acc, object) => {
       const { theme, colorSchema } = object as TagAppearance;
       acc[object.tagName] = (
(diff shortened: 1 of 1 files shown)


arrenv:
@dit7ya Looks good, the fix looks to be working for me.

### Before
download (24)

### After

download (23)

arrenv (approved):

danbr (approved):
LGTM!!

PR #3259, opened 2022-03-23
Server status indicator

### LINKED ISSUE #3230: Add server status in the header
The header status will handle two states.

#### Network health is Poor

Visible if IPFS, Reputation Oracle or SubGraph is unresponsive/behind.

- Show loader just before showing the status [Figma Link](https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=258%3A8700)

- Use the existing icon `triangle-warning`

- Learn more link should direct to: https://colony.gitbook.io/colony/get-started/understanding-the-interface/network-health (This is the correct link, however, the page may be in draft)

Design - [Figma](https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=240%3A7099)

#### Network health is Critical

Visible if RPC is unresponsive.

- Show loader just before showing the status [Figma Link](https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=258%3A8700)

- Use the existing icon `triangle-warning`
- Learn more link should direct to: https://colony.gitbook.io/colony/get-started/understanding-the-interface/network-health (This is the correct link, however, the page may be in draft)

Design - [Figma](https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=240%3A7099)

dit7ya:
## Description

This PR adds a server status (healthy | poor | critical) indicator to the user navigation bar.

**Notes to reviewers**

After halfway through this PR I realized this is a complicated one - and merits to be split into a couple of separate ones. It is NOT DONE yet but it would be very helpful if you point out if I am doing something horribly wrong.

Also, It is not clear to me yet how to best test this without manually editing stuff. But I will think about that and write here. Meanwhile, for checking out the UI component you can manually edit the initial state of the network to be `poor` and `critical` here

https://github.com/joincolony/colonyDapp/blob/db5ce6ab889116e857547b3c877c2f40cb6373f9/src/modules/pages/components/RouteLayouts/UserNavigation.tsx#L149-L151

Many thanks to Raul and Arren for helpful pointers.

NOTE: The IPFS and Reputation Oracle being behind checks will be added in a separate PR.
**New stuff** ✨

- Server status bar

Resolves #3230.


dit7ya:
@rdig There are still some things that need to be done on this. But this is ready for review as of now so that I can have some comments/ feedback before moving on with the rest of it. I will change the commit messages with a refactor and rebase.

rdig:
Great, I'll review it then

src/modules/pages/components/RouteLayouts/UserNavigation.tsx
```diff
@@ -115,8 +145,177 @@ const UserNavigation = () => {
     }
   }, [userDataLoading, userLock, dispatch, ethereal]);
 
+  type NetworkHealthType = 'healthy' | 'poor' | 'critical';
+  const [networkHealth, setNetworkHealth] = useState<NetworkHealthType>(
+    'healthy',
+  );
+
+  const networkCheckInterval = 30 * 1000; // 15 seconds
```
rdig, line 153: Interval should be higher, at least 1 to 2 minutes, having it every 30 seconds is very expensive.

You need to strike a balance between usefulness _(of the feature)_, and load on the servers.

Also, the comment is wrong

src/modules/pages/components/RouteLayouts/UserNavigation.tsx
```diff
@@ -115,8 +145,177 @@ const UserNavigation = () => {
...
+          rpcBlockResponsePromise,
+          latestBlockFromSubgraphPromise,
+          serverResponsePromise,
+          reputationOracleResponsePromise,
+        ]);
+
+        if (rpcBlockResponse.status !== 200) {
+          setNetworkHealth('critical');
+          return;
+        }
```
rdig, line 226: A lot of services might take more than 10 seconds to respond _(due to other network factors)_, so I think this will trigger even for services that are running.

src/modules/pages/components/RouteLayouts/UserNavigation.tsx
```diff
@@ -115,8 +145,177 @@ const UserNavigation = () => {
...
+      }
+    }, networkCheckInterval);
+    return () => clearInterval(networkCheck);
+  }, []);
+
   return (
     <div className={styles.main}>
+      {networkHealth !== 'healthy' && (
+        <div>
+          <MiniSpinnerLoaderWrapper milliseconds={networkHealthLoadingTime}>
```
rdig, line 256: Why is this decoupled in terms of logic?

Why not hide it when everything is set up properly here ?

rdig (commented):
These are just some notes I observed at a first pass.

There's stuff you should change, but the general idea you are going for is pretty good.

src/modules/core/components/MiniSpinnerLoaderWrapper/MiniSpinnerLoaderWrapper.tsx
```diff
@@ -0,0 +1,27 @@
+import React, { ReactNode, useEffect, useState } from 'react';
+
+import { MiniSpinnerLoader } from '~core/Preloaders';
+
+interface Props {
+  children: ReactNode;
+  milliseconds: number;
+}
+
+const MiniSpinnerLoaderWrapper = ({ children, milliseconds }: Props) => {
```
danbr, line 10: I think you could choose a better name for the component. Being specific about what it has/does more than `MiniSpinnerLoader` (not more general).

Also `milliseconds` is a wasted opportunity of a variable name. It doesnt tell me whats it being used for, or give a context. 
A unit type as a name could be confusing.

dit7ya: Good points. I will change `milliseconds` to something like `loadingDuration`. Any suggestion for a better component name?

src/modules/pages/components/RouteLayouts/UserNavigation.css
```diff
@@ -190,3 +190,48 @@
...
+  font-size: var(--size-small);
+  font-weight: var(--weight-bold);
+  color: var(--action-secondary);
+}
+
+.networkHealthIcon {
+  margin-right: 8px;
+}
+
+.networkHealthIcon-critical svg {
```
danbr, line 229: I'm interested why you opted for kebab-case? I thought we should use only camelcase? (even though the *.css.d.ts generates camelcase)

src/modules/core/components/MiniSpinnerLoaderWrapper/MiniSpinnerLoaderWrapper.tsx
danbr: > Good points. I will change milliseconds to something like loadingDuration. Any suggestion for a better component name?

I'm trying to improve my own naming convention thought process.  I need to daydream for a while...... My default is usually to be very literal. 

Probably `MiniSpinnerLoader` isnt a great name (Mini) to start with, as the appearance can easily be changed with a prop. But now you have to build upon it!!! (or start changing lots of things)

Do we even need another `Mini-SpinnerLoader-Wrapper` component? We have lots and they are all very similar. 

I think that having a timeout is a useful feature in a spinner. 
So you should not make another component but instead, add the timeout to the 'base' component SpinnerLoader. Add the `loadingDuration` prop (default=0) then everyone can make use of it, or not.

..... and for the extra work, you get to avoid the problem of deciding on an expressive name ;)

src/modules/pages/components/RouteLayouts/UserNavigation.css
dit7ya: I saw other instances of kebab-case in the codebase. For example - https://github.com/joincolony/colonyDapp/blob/41f9ed47667ca73822740cf4e04b88ed0f4b00f5/src/modules/core/components/OmniPicker/OmniPicker.tsx#L227 .

danbr: 👍 I'll try to use it if an appropriate situation arises.

arrenv (commented):
My thoughts are that it would be good to separate all this logic out of `UserNavigation` into it's own component. What do you think?

In terms of design, while the loader colour is different, I think the grey is fine to be consistent with the rest of the loaders. However, it needs some right margin.

13a3e0a6-6800-417c-887a-9be61822fa52

Also, the icons, look a little bold, you can make the stroke transparent to improve `stroke: transparent;`

579ffc91-00b4-4f48-a465-254f48db97f1

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,176 @@
...
+                    <span>
+                      {formatMessage(MSG.networkHealthHeader, {
+                        networkHealth,
+                      })}
+                    </span>
+                  </div>
+                  <span className={styles.networkHealthDescription}>
+                    {formatMessage(MSG.networkHealthDescription, {
+                      networkHealth,
+                    })}
```
rdig, line 135: Same as above

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,176 @@
...
+                    name: 'offset',
+                    options: { offset: [0, 9] },
+                  },
+                ],
+              }}
+            >
+              <div
+                className={`${styles.elementWrapper}
+                            ${styles.networkInfo}
+                            ${styles.networkHealthHeading}`}
```
rdig, line 157: What triggered this code styling here? Why not have them inlined ?

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,176 @@
...
+              >
+                <span
+                  className={cx(
+                    styles[`networkHealthIcon-${networkHealth}`],
+                    styles.networkHealthIcon,
+                  )}
+                >
+                  <Icon name="triangle-warning" />
+                </span>
+                Network
```
rdig, line 167: This needs to be a message descriptor

rdig (changes requested):
You're heading in the right direct, however there's still a lot of changes needed before this is ready.

I've left notes on what you should change

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,176 @@
...
+  const [showLoading, setShowLoading] = useState(true);
+  // 1 second for the spinner to show
+  const loadingDuration = 1 * 1000;
+
+  useEffect(() => {
+    const loadingTimer = setTimeout(() => {
+      setShowLoading(false);
+    }, loadingDuration);
+    return () => clearTimeout(loadingTimer);
+  }, [loadingDuration]);
```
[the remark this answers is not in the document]
dit7ya: This is the intended behavior per the spec. The network status indicator is only visible when it is the status is not healthy, and the 1 second loading spinner is just cosmetic.

https://www.figma.com/file/zGguqh4iZSIdzBZZMylKYn/Improve-Network-Health?node-id=258%3A8700

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
dit7ya: Do you mean why not put these three lines in a single line? It exceeds the 80 character line length limit enforced by the linter.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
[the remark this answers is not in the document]
rdig: That spec doesn't make sense. 

What's "start to boil?" Are we to detect if the server is "about to go down", that's really not possible with the current infrastructure.

So either the spec needs to change, or it needs to be clarified.

But at any rate, that 1 second spinner, with no actual logic tied to it, is not something we want to add.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,185 @@
...
+            latestSubgraphBlock.latestSubgraphBlock)
+      ) {
+        setNetworkHealth(NetworkHealthType.Poor);
+      } else {
+        // If everything is okay, set health to healthy (to correct for the previous state)
+        setNetworkHealth(NetworkHealthType.Healthy);
+      }
+    }, networkCheckInterval);
+    return () => clearInterval(networkCheck);
+  });
```
rdig, line 97: Running these every second incurs a very serious performance penalty. Both in app memory, but especially in infrastrcture resources.

See demo: 
demo-1m-requests-per-second

This is not necessarily a problem with your implementation as you just followed the specs, and more of a problem of specs not looking at the big picture.

For this we need to take a step back and come up with some sort of deffering or debouncing mechanism _(although debouncing is not really suitable here)_

We need to have a call with the Product team and come up with alternatives as I'm not comfortable pushing this into production.

Make sure to schedule a call w/ you, @arrenv and me to hash this out.

rdig (changes requested):
This is mostly fine, however, I think the specs need to be amended as they aren't production worthy the way they are right now.

We need to get to the bottom of two things:
- that spinner logic
- the excessive strain on our infrastructure

Please address the minor notes I've left, but you can consider this blocked until we find a better alternative. 

_(make sure to schedule that call I've mentioned in which we'll discuss these things)_

rdig:
@dit7ya Please don't forget to set up that call w/ Product

src/data/graphql/queries/status.graphql
```diff
@@ -0,0 +1,16 @@
+query LatestRpcBlock {
+  latestRpcBlock @client
+}
+
+
```
arrenv, line 5: Same comment as Raul, an extra space here.

arrenv (approved):
From what I can tell, it looks good! Nice work on this.
I definitely think Raul should do a review of this one again, based off of our discussions. You've pinged him though, so all good.

chinins (approved):
Looks good. I am just curious why you decided to go for 4 queries instead of 1 that combines 4 values (I saw Raul's comment that there is a reason for this).

dit7ya:
@chinins The rationale for splitting the queries is that we might be interested in the individual loading or error states of the requests in this case.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -0,0 +1,169 @@
...
+      critical
+        {You will have trouble performing actions and retrieving
+        information from the chain. We are working to resolve this.}
+        other {Network is healthy.}}`,
+  },
+});
+
+const NetworkStatus = () => {
+  enum NetworkHealthType {
+    Healthy = 'healthy',
```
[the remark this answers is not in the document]
rdig: I think what Olga meant here is to move them outside of the component **file** as well, and just import them, usually inside `src/types`

That way they can be share by the whole dapp, in case there's a need in the future to touch, or integrate, parts of the server health system somewhere else

rdig (approved):
Tested one last time and I can confirm that the specific requests the status component makes are only fired once every 5 minutes.

We do make a shit ton of other requests though, which we'll need to cut down in the future version of CDapp

arrenv:
Does anyone remember the status of this one?

ArmandoGraterol:
@arrenv 
> Does anyone remember the status of this one?

I remember that this feature went out to QA and it was difficult to test because we basically needed to manually sabotage our services in order to see changes in the server statuses.

That's the only I can recall, perhaps someone else remembers more

rdig:
There was still something to fix regarding IPFS if I remember correctly, but I haven't had time to look into it.

PR #3273, opened 2022-03-28
Fix `canColonyMintNativeToken` resover to work with all tokens

rdig:
The current implementation of this resolver assumes that all tokens have the `authority()` function built in their contract, but this is rather of our doing, so only Colony Contract created tokens do have it. Others might not _(most likely will not)_

Due to this, if a colony will have been created using an external token _(in this case the actual `metacolony`)_, the dapp will actually crash as we don't handle this case.

Even worse, tokens might be able to be minted, even if they don't have the `authority` method, as the `mint` function is somewhat standard on most, and the user creating the colony might actually be the owner of said token, so we need to check for that as well.

For this I've chosen to go the "estimate route", by estimating gas on the call to mint, and if it succeeds assume the user can mint, and if it fails, they can't.

This approach should cover most cases

### Testing

This is quite tricky to test _(it's actually easier on QA)_, but I'll try to do my best to explain how to.

**1. Start the `truffle` console**

Inside the `src/lib/colonyNetwork` folder, run `yarn truffle console` _(or `./node_modules/.bin/truffle console`)_

**2. Deploy a new "non-standard" token. We'll use `ERC20PresetMinterPauser` since it's available to us already.**

Inside the truffle console, run:
```
tk = await ERC20PresetMinterPauser.new("Token Name", "TKN")
```
then
```
tk.address
```

**3. Create a new colony via the Dapp and use the above address as an external token _(don't create a new token)_**

Once created, if the colony loads, and doesn't show the "Mint" and "Unlock" menu entries in the "Actions" menu, then this fix worked.

### Changes

- [x] `canColonyMintNativeToken` resolver in the `colony` resolvers file

### DIFF
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -12,8 +12,6 @@ import {
   getHistoricColonyRoles,
   formatColonyRoles,
 } from '@colony/colony-js';
-import { Contract } from 'ethers';
-import { abi as tokenAuthorityABI } from '@colony/colony-js/lib-esm/contracts/deploy/TokenAuthority.json';
 
 import { Color } from '~core/ColorTag';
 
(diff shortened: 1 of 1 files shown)


arrenv (approved):
@rdig I like the approach, I will have to remember this one. It is working well for me.
 
### Tested with newly created token:

download (25)

### Tested with existing token and permission:

download45

### Test with existing token and no permission

download (26)

### Test with existing token and no permission and Motions & Disputes extension

download (26)

download (27)

ArmandoGraterol (approved):
Didn't know that you could also do this to check if the colony can mint! TIL

PR #3283, opened 2022-04-01
Add motion tests - part 2

### LINKED ISSUE #3251: Add a set of tests for motions - part 2
Splitting the initial issue #3206 so that it's more manageable.

- unlock token
- manage permissions
- edit colony details
- manage tokens

chinins:
## Description

This is part 2 of motions tests PR.

You need to make sure that the reputation miner is on before running the tests.

To test I would recommend to run the actions tests first in a headless mode. If you don't have the colony with the given name already installed then you can change the cypress variable skipInitTests to false and run the create colony tests before as well. This way a subdomain will be created, funds transferred, payments made and there will be reputation in the root and necessary subdomain as well.

You also need to have at least 2 members in the colony where the tests are running or skip the permissions test in actions file.

resolves #3251

### DIFF
--- a/cypress/support/commands.js
+++ b/cypress/support/commands.js
@@ -84,6 +84,7 @@ Cypress.Commands.add('claimNewUserName', (numberFromList) => {
 });
 
 Cypress.Commands.add('createColony', (colony, useNewToken) => {
+  cy.visit('/landing');
   cy.getBySel('createColony').click();
 
   cy.get('input').first().click().type(colony.name);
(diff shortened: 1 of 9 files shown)


ArmandoGraterol (approved):
Looking good! As always, working with these tests can be quite annoying so I must congratulate you for your patience so far hahaha :)

cypress/support/commands.js
```diff
@@ -450,3 +469,105 @@ Cypress.Commands.add('smiteUser', (amountToSmite, isMotion) => {
...
+  cy.getBySel('permission').eq(2).click({ force: true });
+  cy.getBySel('permission').eq(3).click({ force: true });
+  cy.getBySel('permissionAnnotation').click().type(annotationText);
+
+  cy.getBySel('permissionConfirmButton').click();
+
+  cy.getBySel('actionHeading', { timeout: 100000 }).contains(
+    isMotion
+      ? 'Remove the administration, architecture, funding permissions'
+      : `Assign the administration, funding, architecture permissions in Root to`,
```
arrenv, line 564: Would it be better to pull these from the language files?

arrenv (approved):
Nice work, tests make sense and passed for me. Albeit, not all in one run.

cypress/support/commands.js
chinins: Yeah, I think we shouldn't really make the tests even more compliated

PR #3284, opened 2022-04-03
Saving settings confirmation message

### LINKED ISSUE #3265: Saving settings confirmation message
### Story

As a user saving my settings, I don’t know if the details have been saved or not, as there is no confirmation message that appears.

### Description

We want to add a confirmation message when a user updates their settings and avatar. Showing either a success or failed message.

This will need to function like a toast element. So, that is closable, but also closes automatically after a period of 5 secs. E.g.  (For e.g. https://www.npmjs.com/package/react-toastify)

Consideration will need to be made if an external library should be introduced to achieve this.
Consideration should be made for the Beta Warning message that currently sits in the bottom left of the screen.

The copy for each case is as follows:

**Updating Avatar**

- Failed: “Profile picture was not able to be updated. Try again.”
- Success: “Profile picture has been updated.”

**Updating Settings**

- Failed: “Profile settings were not able to be updated. Try again.”
- Success: “Profile settings have been updated.”

### Design

[Figma link](https://www.figma.com/file/UzfpoAvGUS6c14du7Y39sCYm/User-Settings?node-id=2016%3A2791)

0eef2878-0ce9-4b07-b383-4d42b7ab8e70

danbr:
## Description

This PR adds a confirmation message when a user updates their settings and avatar. Showing either a success or failed message.

- The toast auto closes after 5 secs (I think 3 secs is less annoying)
- Progress bar is used as countdown to auto close
- Less width is used (than suggested in figma)

- These details can be easily changed, have a play......

So perhaps reviewers have an opinion???

- The appropriate place to locate the Toast wrapper source? (its not strictly UI components, but they will produce nevertheless)
- The Toast wrapper, exports only a function, so i opted for lowercase name & dir. But internally has some react components, so it has .tsx suffix - That was my thinking, do you (dis)agree? i couldn't find a comparable function/component in the codebase to follow.

How it looks:
Screenshot 2022-04-05 at 11 46 15

Screenshot 2022-04-05 at 11 45 47

Wider toast (suggested in figma)
Screenshot 2022-04-05 at 11 15 32

Resolves #3265


src/modules/users/components/UserProfileEdit/UserProfileEdit.tsx
```diff
@@ -52,6 +54,14 @@ const MSG = defineMessages({
...
     defaultMessage: 'Location',
   },
+  toastSuccess: {
+    id: 'users.UserProfileEdit.toastSuccess',
+    defaultMessage: 'Profile settings have been updated.',
+  },
+  toastError: {
+    id: 'users.UserProfileEdit.toastError',
+    defaultMessage: 'Profile settings were not able to be updated. Try again.',
+  },
```
dit7ya, line 64: These can be combined using `select`.

dit7ya, line 64: These can be combined using `select`.

dit7ya (commented):
Adding some initial thoughts.


1. Clicking save when there has been no change also creates the same success toast. Is this the expected behavior?

2. The green circle is not vertically centered.
3. My preference is the width suggested in the spec.



4. There is just so much negative space on the right side of this page. Why not show the notifications there?

src/modules/core/components/toast/toast.css
```diff
@@ -0,0 +1,51 @@
+:root {
+  --toastify-color-success: rgba(104, 210, 210, 1) !important;
```
arrenv, line 2: Can we pull these colours from `src/styles/variables.css` in all cases.

arrenv (changes requested):
After playing around, I think I like the design width, the font seems less squishy. As well as the `5 secs` as it seems less rushed.

danbr:
> Adding some initial thoughts.
> 
> 1. Clicking save when there has been no change also creates the same success toast. Is this the expected behavior?
> 2. The green circle is not vertically centered.
> 3. My preference is the width suggested in the spec.
> 4. There is just so much negative space on the right side of this page. Why not show the notifications there?

1. I agree we should guard against this behaviour, not only for toast, but for all forms/api calls. 
In this instance, it should be implemented within the form, checking for any 'dirty' inputs, and then disabling the 'action' button. I see this as outside the scope of this PR. If there is agreement, I would be happy to create an issue?

2. I presumed you mean horizontally? After a lot of trouble with CSS i have updated to match figma.  
    
3.  Noted

4.  I bow-down to product!

src/modules/core/components/toast/toast.css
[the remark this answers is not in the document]
danbr: Sadly it is necessary (i would also prefer not to use) but i had a lot of trouble trying to override existing css classes & variables. 
Our use of CSS modules makes it difficult. The only way i found was to use global (i.e. :global(.Toastify__toast-icon)). 
I would be happy to hear of another option.

src/modules/core/components/toast/toast.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+const SuccessDot = () => <div className={styles.successDot} />;
+const ErrorDot = () => <div className={styles.errorDot} />;
+
+const toast = {
+  success(msg, options = {}) {
+    return reactToast.success(handleMsg(msg), {
+      ...options,
+      hideProgressBar: false,
+      autoClose: 5000,
```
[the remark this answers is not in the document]
danbr: I did previously consider. Now default values are extracted.

src/modules/users/components/UserProfileEdit/UserProfileEdit.tsx
danbr: I prefer these individual msgs being kept separate as they are serving different outcomes.

src/modules/users/sagas/user.ts
```diff
@@ -42,6 +44,21 @@ import {
...
+  toastAvatarUpdateSuccess: {
+    id: 'users.sagas.user.toastAvatarRemoveSuccess',
+    defaultMessage: 'Profile picture has been updated.',
+  },
+  toastAvatarError: {
+    id: 'users.UserProfileEdit.toastError',
+    defaultMessage: 'Profile picture was not able to be updated. Try again.',
+  },
+});
+
```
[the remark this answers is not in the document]
danbr: The success msgs are good candidates to combine. Done.

src/modules/core/components/toast/toast.css
danbr: 👍

rdig:
Screenshot from 2022-04-21 13-57-47

@danbr please don't add `WIP` commits to draft / working PR _(or any PRs for that matter)_, it's a given that any work done in a PR up to the point it's ready for review is a work in progress.

Also, it looks like these two commits are duplicated somehow...

danbr:
> Screenshot from 2022-04-21 13-57-47
> 
> @danbr please don't add `WIP` commits to draft / working PR _(or any PRs for that matter)_, it's a given that any work done in a PR up to the point it's ready for review is a work in progress.
> 
> Also, it looks like these two commits are duplicated somehow...

@rdig 
Sure NP. I thought it was ok to sparingly use `WIP`. I will refrain from using in the future. 

The two commits have duplicated message: I can squish and rename (and you'll never see WIP again!!)

arrenv:
@danbr After looking into this more, and noticing the size of the Toastify dependency, it would be better to just have a notice appear underneath the save button.

- Please see spec here - https://www.figma.com/file/UzfpoAvGUS6c14du7Y39sCYm/User-Settings?node-id=2557%3A2821
- This will also probably work better with a 3 sec timer as it won't need to be closable.
- Have the notice animate in and out.

danbr:
"After looking into this more, and noticing the size of the Toastify dependency, it would be better to just have a notice appear underneath the save button."
The original issue was resolved in a different PR

PR #3288, opened 2022-04-04
Fix old member rep being weird on motion page

### LINKED ISSUE #3249: User reputation on motion page is acting funny
## Steps to reproduce

Create a motion & then add/deduct reputation for yourself. 

## Expected behaviour

Reputation should never be more than 100%

## Actual behaviour

Reputation is higher than 100%

reputation

ArmandoGraterol:
Fixed MemberReputation on the motion page not using the motion's root hash for the total domain reputation, and not using a dynamic domain id (It was always using the root domain id to get the rep).

Now the reputation shown on the motion page should be consistent with the rep that the users had at the time of the motion creation and not change percentages when the user gets more or less reputation.

Resolves #3249

### DIFF
--- a/src/modules/core/components/MemberReputation/MemberReputation.tsx
+++ b/src/modules/core/components/MemberReputation/MemberReputation.tsx
@@ -51,6 +51,7 @@ const MemberReputation = ({
       address: AddressZero,
       colonyAddress,
       domainId,
+      rootHash,
     },
     fetchPolicy: 'cache-and-network',
   });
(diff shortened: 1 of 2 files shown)


arrenv (approved):
Nice fix, works for me!

### Before
c1c3474f-139b-4749-99ab-8d67586d311c


### Before after refresh
download (28)


### After on navigate and after refresh
download (29)

chinins (approved):
Works well!

dit7ya (approved):
Works great! I was so baffled by this - turns out the solution is elegant and simple.

| Before | After |
|---------|-------|
|Screenshot 2022-04-05 at 16-36-04 Pay halum 54 NCC Motion Colony - Notun| Screenshot 2022-04-05 at 16-36-04 Pay halum 54 NCC Motion Colony - Notun|

PR #3294, opened 2022-04-07
Add tests for the whole process of a motion

### LINKED ISSUE #3207: Add testing to go through the whole process of a motion
Pick a motion and add tests to go through the motion process.

chinins:
## Description

This is a very intense test. So I would recommend to run it in a headless mode. Make sure that the colony is created and there is reputation. Then you can run only the tests that install voting extension, activate tokens and go through the whole motion.

**Important**
I have changed staking slider step from 0.01 to 1. This way the tests and the user need to do only 100 right arrow clicks instead of 10000 to get to the max stake. I've tested and it seems to be working well. I think it might also help fix us occasional bugs with motions being not fully staked. I think that 0.01 step on the scale from 0 to 100 is too small but let me know what you think.

resolves #3207


cypress/support/commands.js
```diff
@@ -189,21 +201,14 @@ Cypress.Commands.add('mintTokens', (amountToMint, isMotion) => {
...
   cy.getBySel('mintTokensDialogItem').click();
 
-  cy.get('input')
-    .last()
-    .click()
-    .type(amountToMint)
-    .get('textarea')
-    .click()
-    .type(annotationText);
+  cy.get('input').last().click().type(amountToMint);
```
ArmandoGraterol, line 204: Does this mean that we won't check for annotations in the mint actions/motions?

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
```diff
@@ -161,7 +161,7 @@ const StakingSlider = ({
           name="amount"
           value={values.amount}
           limit={userStakeLimitPercentage}
-          step={0.01}
+          step={1}
```
ArmandoGraterol, line 164: I agree that this may take care of those edge cases where you try to stake 100% and you stake 99% instead. However, I think this comes with the tradeoff that it offers less flexibility or specificity when staking.

I'm not sure whether it's worth it in comparison with this

ArmandoGraterol (commented):
Looking good! I just have a couple of questions that I want to address before approving

cypress/e2e/8-motions.js
```diff
@@ -117,6 +118,66 @@ describe('User can create motions via UAC', () => {
...
   });
 
+  it('Can claim tokens', () => {
+    cy.login();
+    cy.visit(`/colony/${Cypress.config().colony.name}`);
+    // // activate tokens
+    cy.getBySel('tokenActivationButton', { timeout: 120000 }).click();
+    cy.getBySel('inputMaxButton').click();
+    // eslint-disable-next-line cypress/no-unnecessary-waiting
+    cy.getBySel('tokenActivationConfirm').click().wait(10000);
```
arrenv, line 128: I think that this should be checking that the token activation has worked. As in my runs, it passes even without the tokens being activated.

src/modules/dashboard/components/ActionsPage/StakingWidget/StakingSlider.tsx
ArmandoGraterol: Addressed through Discord

ArmandoGraterol (approved):

arrenv (approved):
Working for me, very cool!

PR #3296, opened 2022-04-11
Fetch whitelisted addresses and filter out users dropdown in Payment modal

### LINKED ISSUE #3182: Fetch whitelisted addresses and filter out users dropdown in Payment modal
## Specification
*Fetch from IPFS whitelisted addresses
*Filter out user addresses in dropdown [Payment modal]

https://www.figma.com/file/Phby7kEYihhvGgQn69A314/Verified-Recipients?node-id=79%3A12158

danbr:
## Description

This PR updates the payment modal. Whitelisted addresses are retrieved from IPFS and used to filter recipient addresses from the dropdown.

- If the whitelist is active - only show addresses that are both in colonyMembers subscribedUsers & in verified recipients list
- If the whitelist is inactive - show colonyMembers subscribedUsers (as before update)

If an address is manually entered and whitelist is active, then a warning will be shown:
-`Warning. You are about to make a payment to an address not on the whitelist. Are you sure the address is correct?`

Screenshot 2022-04-19 at 11 10 35

Resolves #3182

### DIFF
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { FormikProps } from 'formik';
 import * as yup from 'yup';
 import { ROOT_DOMAIN_ID } from '@colony/colony-js';
@@ -11,7 +11,12 @@ import { ActionForm } from '~core/Fields';
 
 import { Address } from '~types/index';
 import { ActionTypes } from '~redux/index';
-import { AnyUser, useMembersSubscription } from '~data/index';
+import {
+  useVerifiedUsersQuery,
+  useColonyFromNameQuery,
+  AnyUser,
+  useMembersSubscription,
+} from '~data/index';
 import { getTokenDecimalsWithFallback } from '~utils/tokens';
 import { pipe, withMeta, mapPayload } from '~utils/actions';
 import { WizardDialogType } from '~utils/hooks';
@@ -93,6 +98,42 @@ const CreatePaymentDialog = ({
     variables: { colonyAddress },
   });
 
+  const { data: colonyData } = useColonyFromNameQuery({
+    variables: { name: colonyName, address: colonyAddress },
+  });
+  const isWhiteListActive = colonyData?.processedColony?.isWhitelistActivated;
+
+  const { data } = useVerifiedUsersQuery({
+    variables: {
+      verifiedAddresses:
+        colonyData?.processedColony?.whitelistedAddresses || [],
+    },
+    fetchPolicy: 'network-only',
+  });
+
+  const storedVerifiedRecipients = useMemo(
+    () =>
+      (data?.verifiedUsers || []).map((user) => user?.profile.walletAddress),
+    [data],
+  );
+
+  const filteredVerifiedRecipients = isWhiteListActive
+    ? (colonyMembers?.subscribedUsers || []).filter((member) =>
+        storedVerifiedRecipients.some(
+          (el) => el.toLowerCase() === member.id.toLowerCase(),
+        ),
+      )
+    : colonyMembers?.subscribedUsers || [];
+
+  const getIsUnverifiedRecipient = (walletAddress) => {
+    if (!walletAddress) return false;
+    return isWhiteListActive
+      ? !storedVerifiedRecipients.some(
+          (el) => el.toLowerCase() === walletAddress.toLowerCase(),
+        )
+      : false;
+  };
+
   const transform = useCallback(
     pipe(
       mapPayload((payload) => {
@@ -161,6 +202,7 @@ const CreatePaymentDialog = ({
         if (formValues.values.forceAction !== isForce) {
           setIsForce(formValues.values.forceAction);
         }
+
         return (
           <Dialog cancel={cancel}>
             <DialogForm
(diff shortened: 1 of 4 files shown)


src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -93,6 +98,42 @@ const CreatePaymentDialog = ({
     variables: { colonyAddress },
   });
 
+  const { data: colonyData } = useColonyFromNameQuery({
+    variables: { name: colonyName, address: colonyAddress },
+  });
+  const isWhiteListActive = colonyData?.processedColony?.isWhitelistActivated;
```
rdig, line 104: You already have this colony object passed as a prop to this modal, why are you calling the query for it again ?

This is super expensive. _(in terms of requests made to the server)_

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -161,15 +202,19 @@ const CreatePaymentDialog = ({
...
             <DialogForm
               {...formValues}
               colony={colony}
               isVotingExtensionEnabled={isVotingExtensionEnabled}
               back={() => callStep(prevStep)}
-              subscribedUsers={colonyMembers?.subscribedUsers || []}
+              subscribedUsers={filteredVerifiedRecipients}
               ethDomainId={ethDomainId}
+              showWhitelistWarning={getIsUnverifiedRecipient(
+                formValues.values?.recipient?.profile?.walletAddress,
```
rdig, line 216: I'm not totally sure about this, as I have a suspicion the `formValue` you get here is actually the previous state _(value)_, so this might screw with the warning display logic.

Please double check it _(with multiple values)_ to make sure it works properly

rdig (changes requested):
Mostly good with this, I've just highlighted some stuff that needs to be addressed.

alicjakujawa (approved):
Seems to be working fine, leaving small comments. The only thing i would ask product team, if we should include in payment modal dropdown addresses which are in whitelist but they did not join colony

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -93,6 +98,42 @@ const CreatePaymentDialog = ({
...
+  });
+  const isWhiteListActive = colonyData?.processedColony?.isWhitelistActivated;
+
+  const { data } = useVerifiedUsersQuery({
+    variables: {
+      verifiedAddresses:
+        colonyData?.processedColony?.whitelistedAddresses || [],
+    },
+    fetchPolicy: 'network-only',
+  });
```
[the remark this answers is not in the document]
danbr: I have already removed, in the latest version.

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -93,6 +98,42 @@ const CreatePaymentDialog = ({
...
+  const filteredVerifiedRecipients = isWhiteListActive
+    ? (colonyMembers?.subscribedUsers || []).filter((member) =>
+        storedVerifiedRecipients.some(
+          (el) => el.toLowerCase() === member.id.toLowerCase(),
+        ),
+      )
+    : colonyMembers?.subscribedUsers || [];
+
+  const getIsUnverifiedRecipient = (walletAddress) => {
+    if (!walletAddress) return false;
```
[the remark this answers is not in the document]
danbr: I agree this could be confusing when viewed by fresh eyes. I have changed the func name to `showWarningForAddress`, then the return values make more sense.

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
danbr: Well spotted. You're correct. I removed the extra calls using `useColonyFromNameQuery` & `useVerifiedUsersQuery`.

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
rdig: Just making sure. I trust when you say it's working as intended :+1:

danbr:
@alicjakujawa @rdig 
Through testing this PR, i noticed a situation that needs to be considered. I think it should be addressed in a different PR.

Issue:
There is a conflicting link between whitelist addresses & Whitelist activated toggle (isWhitelistActivated).
If the whitelist addresses is empty, what should be the value of isWhitelistActivated?
Currently if an address is added to the whitelist, and then removed, isWhitelistActivated === true.
(I have not checked for the value of isWhitelistActivated when a colony is created (before address added to whitelist for the first time))

The current approach doesn't seem logical:

1. You cannot toggle isWhitelistActivated unless there is an address in the whitelist. So you would have to add an address, toggle, then remove address.
2. When filtering addresses in the payment modal, if isWhitelistActivated === true but the whitelist is empty, then none of the subscribed members will be listed, and a warning will be displayed for an address manually added.
More importantly from the users perspective, it could be confusing. 
If they were to check the whitelist via manage whitelist dialog, it would appear to be empty, and one would assume that it is not active, when in actual fact it is. Currently, this only effects filtering in the payment dialog.

Screenshot 2022-04-20 at 11 44 22


I would expect the isWhitelistActivated === false when the whitelist is empty.

chinins (approved):
Looks good!

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -101,12 +101,17 @@ const MSG = defineMessages({
     payment.`,
   },
   userPickerPlaceholder: {
-    id: 'SingleUserPicker.userPickerPlaceholder',
+    id: `dashboard.CreatePaymentDialog.CreatePaymentDialogForm.userPickerPlaceholder`,
     defaultMessage: 'Search for a user or paste wallet address',
   },
+  warningText: {
+    id: `dashboard.CreatePaymentDialog.CreatePaymentDialogForm.warningText`,
+    defaultMessage: `<span>Warning.</span> You are about to make a payment to an address not on the whitelist. Are you sure the address is correct?`,
```
rdig, line 109: Not a big of putting the tags in the message descriptors text, but I guess it's fine since you're overwriting the chunk in the values

rdig (approved):
All good from my side. Nice work!

PR #3303, opened 2022-04-12
Remove new line symbols from `span` in `Numeral` component

chinins:
The way the innerHtml of the span element was created in `Numeral` component, it was adding `\n` around the number.

This was not seen because span is a single line element. Without this change if you change span to div on another element that can be multi line you will see that the number and token is split in several lines.

resolves #3303


ArmandoGraterol (approved):
LGTM

danbr (approved):
All good. 

Not caused by your change, but nevertheless GH is complaining about an issue:

Screenshot 2022-04-13 at 16 03 49

PR #3304, opened 2022-04-12
Fix search filter excluding addresses with no username

### LINKED ISSUE #3300: Fix search functionality in ManageWhitelist modal
## Steps to reproduce

1. Try to use search for whitelisted addresses in ManageWhitelist modal

## Expected behaviour

* Addresses are correctly filtered (by address or username)

## Actual behaviour

* Addresses are not filtered

ArmandoGraterol:
Self-explanatory title. This should make the search functionality in the manage whitelist dialog actually search by usernames AND wallet addresses

Resolves #3300

### DIFF
--- a/src/modules/core/components/SingleUserPicker/utils.ts
+++ b/src/modules/core/components/SingleUserPicker/utils.ts
@@ -1,28 +1,26 @@
+import { AnyUser } from '~data/index';
+
 /*
 Extracts the required values to be used in the SingleUserPicker
 on selection
 */
-export const filterUserSelection = (data, filterValue) => {
+export const filterUserSelection = (data: AnyUser[], filterValue: string) => {
   if (!filterValue) {
     return data;
   }
 
-  const filtered = data.filter(
-    (user) =>
-      user &&
-      user.profile &&
-      user.profile.username &&
-      filterValue &&
-      (user.profile.username
-        .toLowerCase()
-        .includes(filterValue.toLowerCase()) ||
-        user.profile.walletAddress
-          .toLowerCase()
-          .includes(filterValue.toLowerCase()) ||
-        user.id.toLowerCase().includes(filterValue.toLowerCase())),
-  );
+  const filteredUsers = data.filter((user) => {
+    const username = user.profile?.username || '';
+    const walletAddress = user.profile?.walletAddress || '';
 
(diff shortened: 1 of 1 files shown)


src/modules/core/components/SingleUserPicker/utils.ts
```diff
@@ -1,34 +1,32 @@
...
-        .toLowerCase()
-        .includes(filterValue.toLowerCase()) ||
-        user.profile.walletAddress
-          .toLowerCase()
-          .includes(filterValue.toLowerCase()) ||
-        user.id.toLowerCase().includes(filterValue.toLowerCase())),
-  );
+  const filteredUsers = data.filter((user) => {
+    const username = user.profile?.username || '';
+    const walletAddress = user.profile?.walletAddress || '';
```
rdig, line 14: I mean... the wallet address must always exist, but in this case we're also not excluding the entry if there isn't one.

I think we should filter out entries which don't have a wallet address _(but you are correct in allowing entries without a username)_

Granted, what I just explained above shouldn't really happen in real life

rdig (approved):
All good here code-wise, but I would filter out entries without a wallet address

chinins (approved):
I get where Raul is coming from with his comment. I guess it's better to be overly cautious even though the situation should not happen in real life.
Other than that very nice!

src/modules/core/components/SingleUserPicker/utils.ts
ArmandoGraterol: Yeah. now that you mentioned it, it _shouldn't_ happen that a user doesn't have a wallet address. 

With that being said, I'll add an early return in the case of the object not having a wallet address, just for safeguarding.

alicjakujawa:
@ArmandoGraterol is this ready for re-review? did you fix the problem i mentioned on discord?

ArmandoGraterol:
> @ArmandoGraterol is this ready for re-review? did you fix the problem i mentioned on discord?

@alicjakujawa Except for that 1 time I encountered the bug, I haven't been able to reproduce it, as of now I don't know why it's happening just for that 1 time it happened to me. I will continue to try to see if I can actually reproduce it again to fix it.

alicjakujawa (approved):

PR #3305, opened 2022-04-13
Verified recipients tick logic & Manage whitelist trigger

### LINKED ISSUE #3301: Display "whitelisted" tick on colony members page base on whitelisted addresses
## Specification

* Fetch whitelisted addresses on colony members page and display green "tick" for whitelisted members (see design)
* Frontend (tick) already exist  (https://github.com/JoinColony/colonyDapp/pull/3189)

https://www.figma.com/file/Phby7kEYihhvGgQn69A314/Verified-Recipients?node-id=79%3A15782

arrenv:
## Description

This adds the logic to show the tick for verified recipients on the Members page. I also took the opportunity to add the `Manage whitelist` dialog toggle on the members page.

**New stuff** ✨

* Verified recipient tick on members
* `Manage whitelist` dialog toggle

87e28c78-e754-4852-bc78-a1bd9bcefcb2

### Testing
* Add some members to the whitelist and ensure it is activated
* Ensure the tick only appears only on the relevant addresses
* Try adding members to the whitelist by using the `Manage whitelist` link on the members page

Resolves #3301

### DIFF
--- a/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
+++ b/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useEffect, MouseEventHandler } from 'react';
+import React, { useCallback, useEffect } from 'react';
 import { useParams, Redirect } from 'react-router-dom';
 import { defineMessages } from 'react-intl';
 import { ColonyVersion, Extension } from '@colony/colony-js';
@@ -115,22 +115,12 @@ const ColonyMembers = () => {
 
   const openToggleManageWhitelistDialog = useDialog(ManageWhitelistDialog);
 
-  const handleToggleWhitelistDialog = useCallback<
-    MouseEventHandler<HTMLButtonElement>
-  >(
-    (evt) => {
-      evt.stopPropagation();
-      /*
-       * We don't have, and can't inject all the required props that the component
-       * is expecting when using it in a wizard
-       */
-      // @ts-ignore
-      return openToggleManageWhitelistDialog({
-        colony: colonyData?.processedColony as Colony,
-      });
-    },
-    [openToggleManageWhitelistDialog, colonyData],
-  );
+  const handleToggleWhitelistDialog = useCallback(() => {
+    // @ts-ignore
+    return openToggleManageWhitelistDialog({
+      colony: colonyData?.processedColony as Colony,
+    });
+  }, [openToggleManageWhitelistDialog, colonyData]);
 
   // eslint-disable-next-line max-len
   const oneTxPaymentExtension = colonyExtensions?.processedColony?.installedExtensions.find(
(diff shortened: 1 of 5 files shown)


src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -108,6 +113,25 @@ const ColonyMembers = () => {
...
+  const openToggleManageWhitelistDialog = useDialog(ManageWhitelistDialog);
+
+  const handleToggleWhitelistDialog = useCallback<
+    MouseEventHandler<HTMLButtonElement>
+  >(
+    (evt) => {
+      evt.stopPropagation();
+      /*
+       * We don't have, and can't inject all the required props that the component
+       * is expecting when using it in a wizard
```
alicjakujawa, line 125: So maybe they are not required actually? if we can have both cases

src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialogForm.tsx
```diff
@@ -181,7 +181,12 @@ const ManageWhitelistDialogForm = ({
         <Button
           appearance={{ theme: 'secondary', size: 'large' }}
           onClick={back}
-          text={{ id: 'button.back' }}
+          text={{
+            id:
+              back.toString() === `() => callStep(prevStep)`
+                ? 'button.back'
+                : 'button.cancel',
```
alicjakujawa, line 188: Maybe it will be better to do something like that:
'whateverButtonText' - with default to 'button.back'
and then just put 'button.cancel' for this case

I dont really like this approach of converting to string and comparing if this is that function

alicjakujawa (approved):
Works as excepted. Leaving small comments. Also wondering if we should show tick when whitelist is not activated? Question to product team... 🤣

src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialogForm.tsx
arrenv: Haha, yeah, I don't like it either. I will try it as you suggested.

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
arrenv: Good point, I will give it another try.

arrenv:
> Works as excepted. Leaving small comments. Also wondering if we should show tick when whitelist is not activated? Question to product team... 🤣

Yes, I agree. Ticks should not be visible when it is deactivated. I will account for it.

ArmandoGraterol (approved):
Putting aside the comments Alicja made and my question regarding the empty brackets, this is looking good :+1:

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -216,6 +241,17 @@ const ColonyMembers = () => {
                   </li>
                 </>
               )}
+              {canManageWhitelist && (
+                <>
```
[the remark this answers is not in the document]
arrenv: You're right it is not needed, it is a left over. Removing now.

PR #3308, opened 2022-04-13
Fix the finalize motion error condition

### LINKED ISSUE #3307: Unnecessary error that doesn't allow to finalise mint tokens motion
The error appears when the colony funds is less than the amount of tokens to mint. This prevents finalising the motion.

chinins:
## Description

Update the condition that prevents finalising motion.

To test you need to mint an mount that is more than the current colony token balance

Before

Screenshot 2022-04-13 at 17 09 12

After

Screenshot 2022-04-13 at 17 07 50

resolves #3307

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx
@@ -204,7 +204,7 @@ const FinalizeMotionAndClaimWidget = ({
     }
 
     return true;
-  }, [domainBalanceData, motionAmount, fromDomain]);
+  }, [domainBalanceData, motionAmount, fromDomain, actionType]);
 
   const nativeToken = tokens.find(
     ({ address }) => address === nativeTokenAddress,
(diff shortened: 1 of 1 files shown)


src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx
```diff
@@ -194,6 +194,7 @@ const FinalizeMotionAndClaimWidget = ({
     const domainBalance = bigNumberify(domainBalanceData?.domainBalance || '0');
 
     if (
+      actionType !== ColonyMotions.MintTokensMotion &&
```
rdig, line 197: I don't really understand how this fix works. 

We're filtering out the Mint Motion as an action from the _Finalize_ button validation logic, correct ?

That means that we're not really fixing the underlying problem, but just brute forcing it for the Mint Motion, where this problem shows up

I feel like there is something else at play here that is preventing the Motion to be finalized if you're minting more tokens than your current balance.

rdig (changes requested):
Possibly this is the correct approach, but I don't understand it _(or understand something else)_

So this "require changes" is more to help me understand this problem, rather than requiring actual code changes :)

src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx
chinins: It's just not applicable to Min tokens motion. We can finalise motion even if the balance is zero. There is nothing to prevent us from it. In other situations we need the token balance be more than what we need to pay/transfer. Otherwise we'll have a contract error - we can transfer/pay more than we our balance. Again for mint tokens it's not important as we don't need any balance to "transfer from".
Let me know if it's still not clear

ArmandoGraterol (approved):
Working as expected! Nice work

rdig (approved):
Still can't quite wrap my head around it, but I don't really want to keep blocking it.

PR #3261, opened 2022-03-24
Testing token activation and deactivation & claiming stakes

arrenv:
## Description

This PR adds testing for Token Activation and Deactivation. You will be able to run `9-tokenActivation.js` and `8-motions.js` specifically the `Claiming Stakes` test to see the tests in action.

**New stuff** ✨

* Test for activating a user's tokens
* Test for deactivating a user's tokens
* Tests for claiming stakes

**Testing**

- This requires the user to already have tokens

Contributes to #3203

Issue #3252, opened 2022-03-18
Add tests for numbers across the dapp


chinins:
Based on our refactoring of the numbers across the dapp.

Blocked until colonyDapp #3125 is merged into master.

rdig:
What's the state of this? Is it still blocked ?

chinins:
> What's the state of this? Is it still blocked ?

Good point. I've removed the blocked.

PR #3310, opened 2022-04-14
Fix & consolidate the whole set of cypress tests

### LINKED ISSUE #3299: Check that all of the tests are running from start to finish without any additional inputs
Once #3298 & #3261 are done.

chinins:
## Description

This consolidates all the tests. To run it properly, you need to do it first thing after you start the dev environment.

Remember to have the reputation monitor on.

This needs to be run in a headless mode with `npm run cy:run` command. You also  need to change a cypress variable `skipInitTests` to false.

Sometimes annotations don't get through and this causes a test to fail although it all worked fine. Not sure how to deal with it yet...

## TODO

Refactor restructuring of the config variables where they are used.

resolves #3299

### DIFF
--- a/cypress/e2e/2-createColony.js
+++ b/cypress/e2e/2-createColony.js
@@ -3,9 +3,10 @@ describe('Create a new colony', () => {
     colony,
     colony: { name: colonyName, nativeToken },
     baseUrl,
+    skipInitTests,
   } = Cypress.config();
 
-  if (!Cypress.config().skipInitTests) {
+  if (!skipInitTests) {
     it('creates a new colony with new token', () => {
       cy.login();
 
--- a/cypress/e2e/2-createColony.js
+++ b/cypress/e2e/2-createColony.js
@@ -36,7 +36,7 @@ describe('Create a new colony', () => {
     });
 
     it('creates a new colony with existing token', () => {
-      const newColony = { name: 'feola' };
+      const newColony = { name: 'maya' };
 
       cy.login();
       cy.getColonyTokenAddress(colonyName);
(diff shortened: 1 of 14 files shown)


arrenv (commented):
We are going to have a merge conflict with this and https://github.com/JoinColony/colonyDapp/pull/3261, I refactored based on Raul's feedback. I will likely need to undo my updates it seems.

cypress/e2e/2-createColony.js
```diff
@@ -1,21 +1,25 @@
 describe('Create a new colony', () => {
   if (!Cypress.config().skipInitTests) {
     it('creates a new colony with new token', () => {
-      const { name, nativeToken } = Cypress.config().colony;
+      const { colony, baseUrl } = Cypress.config();
```
rdig, line 4: Does the config need to be called every time, or can we do this somewhere at the start of the file, somewhere inside the `describe` declaration, but outside of any individual `it` declarations

Not major, just a thought

rdig (approved):
Looks good to me. Nice work!

cypress/e2e/2-createColony.js
chinins: Yeah, I already thought about it. But it seems that Arren has already refactored it everywhere. I need to wait until he merges his PR first so that I can make the final pass at all the tests. And then I'll make sure that it's consistent across all the files.

ArmandoGraterol (approved):
After much pain and adjustments, managed to pass all the tests from start to finish! Nice work

dit7ya:
I tried to run the tests but this is what happened in the second test and every other test failed after that.

chinins:
> I tried to run this so that I can start working on #3252 but I am getting the following on the second test - and every other test after that just fails. Any clue what might be the reason?
> 


You also  need to change a cypress variable `skipInitTests` to false.

PR #3315, opened 2022-04-18
Refactor Tooltips

### LINKED ISSUE #3297: Refactor tooltips in the colonyDapp
### Description 

In https://github.com/JoinColony/colonyDapp/pull/3282 we added new lib to handle tooltips. This lib seems to be working/and looking better than existing implementation. We want be consist across the colonyDapp to use same component for handling tooltips.

### TODO
* Refactor tooltips in colonyDapp to use new lib
* Remove old/not used code/lib after all of this will be refactored

dit7ya:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #3297.


src/modules/core/components/Popover/Tooltip.tsx
```diff
@@ -37,32 +56,58 @@ interface Props {
...
+      <div ref={setTriggerRef}>
+        {typeof children === 'function'
+          ? children({
+              id: nanoid(),
+              ref: () => {},
+              open: () => {},
+              close: () => {},
+              isOpen: visible,
+              toggle: () => {},
+            })
```
dit7ya, line 94: What exactly is happening here? What is the alternative?

src/modules/core/components/Fields/Checkbox/Checkbox.tsx
```diff
@@ -135,7 +136,7 @@ const Checkbox = ({
           appearance={{ theme: 'dark' }}
           content={tooltipText}
           placement="bottom"
-          popperProps={tooltipPopperProps}
```
dit7ya, line 138: Either call everything `popperOptions` or `popperProps`.

rdig (changes requested):
For so many files being touched this went off without a hitch. Congratulations, this is some top quality work.

The changes I want you to make, are more or less not really related to this PR, but while you're at it, might as well get to them.

Because the previous Tooltips were broken, we had places where the Tooltips were displayed without a arrow, since that would render either off center or just plain wonky. Here's some examples:

Screenshot from 2022-04-20 12-43-03
Screenshot from 2022-04-20 12-43-14

Please look after those examples and re-add the arrows for those Tooltips, assuming they can be added. _(make a note if they can't)_

The second thing I want to address, again not related to this PR, but in this case related to your general git commit style.

Please stop using `WIP` commits: 

Screenshot from 2022-04-20 12-03-16
Screenshot from 2022-04-20 12-03-37

It's a given that commits made during the PR is in a _draft_ state, or just considered as being _in progress_ are work in progress. 

I really don't want us to push commits marked `WIP` into `master`, just use normal work commits _(other commits you add are fine)_, and just keep adding them as additions, even if you don't know the full outcome of that PR's scope from the start.

Also, I've noticed you do all your pushes as force pushes:

Screenshot from 2022-04-20 12-03-47
Screenshot from 2022-04-20 12-03-51

Why is that? is this related with what I mentioned earlier that you keep overwriting commits? It's better to avoid doing that as much as possible, only re-write them if special cases, not as a normal work flow.

rdig (changes requested):
So apparently some of  the arrows work as expected _(eg: the actions page)_ but not the ones for the events list.

Screenshot from 2022-04-26 18-20-52

Revert the changes for the events list, but just those, leave everything else as it is. Once that's done I'll approve this

ArmandoGraterol (changes requested):
Overall, it's looking good for such a general use component

I found that the tooltips in the motion page now have a bigger padding than the one that they should have:

**NEW**
FireShot Capture 583 - Mint 1000 PRR - Motion - Colony - wonker - localhost

**OLD**
FireShot Capture 582 - Mint 1000 PRR - Motion - Colony - wonker - localhost

This is happening for every single QuestionMarkTooltip in the motion page.

dit7ya:
@rdig @ArmandoGraterol Fixed both of your concerns. Only last two commits are new (after the rebase) to review.

ArmandoGraterol (approved):
Nice work! Correct padding and the tooltips in the new event log are looking good :ok_hand:

rdig (approved):
I still found two instances  where the tooltips need to be adjusted.

On the action / motion page:

Screenshot from 2022-04-28 16-56-11

This should look like this, but with an arrow:

Screenshot from 2022-04-28 17-02-15

The second thing I found is that the tooltips in the user navigation have different text sizes _(while this is not technically something you introduced, please adjust it)_

different-tooltip-text-sizes

I'm pre-approving, but please fix the above issues before merging it

chinins (approved):
Nice work!

arrenv (changes requested):
@dit7ya Great work on this, it looks like a great improvement. I have noticed a couple of things.

The spec rendering of tooltips comes out wrong at 12px, I have spoken to Karol about this and he has confirmed that the font sizing should be 11px, which is what is currently in `master`.

This is what `master` is using around the text content in tooltips:
```
    padding: 5px 10px;
    font-size: var(--size-tiny);
    line-height: 16px;
    letter-spacing: 0.5px;
```

This shows an example of the difference, this is from this PR:

2e8e4484-492b-4750-8f13-baff303a0427


This is currently what is in `master`:

download (42)


I feel as these in the TokenActivationPopover could be aligned over the word, which is where you need to hover for them to appear:

2a54dde4-a089-4708-9ba5-a124120453b6

arrenv (approved):
The tooltips look much better, nice one. I am still getting the alignment issues in the `TokenActivationPopover`. But it seems local to me, so approving.

PR #3319, opened 2022-04-19
Fix the `UserCheckbox` default values

### LINKED ISSUE #3309: Whitelisted addresses all show as red when new address is added
## Steps to reproduce

1. Add some addresses to the whitelist.
2. Check the `Whitelisted` tab.
3. You should notice that the addresses are in red state, indicating they are unchecked, when they are not.

## Expected behaviour

* Addresses should only show red if they are unchecked.

## Actual behaviour

* Whitelisted addresses all show as red when new address is added.

addresses-labelled-unchecked-after-update

chinins:
## Description

This PR fixes removes the red background every time the whitelisted addresses list in the dialog is updated (removed or added). The red background is caused by the `isChecked` boolean value of `false`. When we update the list, the value used to default to false, hence the red background  on any updates.

Resolves #3309

### DIFF
--- a/src/modules/core/components/UserCheckbox/UserCheckbox.tsx
+++ b/src/modules/core/components/UserCheckbox/UserCheckbox.tsx
@@ -34,7 +34,7 @@ const UserCheckbox = ({
   showDisplayName = true,
 }: Props) => {
   const userProfile = useUser(createAddress(walletAddress || AddressZero));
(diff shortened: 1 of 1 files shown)


rdig (approved):
LGTM! :+1:

ArmandoGraterol (approved):
:100:

danbr (approved):
All good.......(as expected!!!)

PR #3323, opened 2022-04-21
Add incoming funds aside in home layout

### LINKED ISSUE #3316: Implement new "Incoming funds" aside
### Description

With the Events page being moved, claiming funds will also be moved to 2 locations location, this issue addresses one of those locations, the right-hand aside

When we enter the colony home page/action page, we want to see a new aside that shows the incoming funds of the colony (replacement for the current incoming funds on the events page).

- The aside should be only visible when there are incoming funds to claim.
- The title of the section in the aside will link to the funds page.
- The claim button in the aside will trigger the claim for all tokens. (This saga is going to be updated in a separate issue)
- The aside will only show one of the incoming fund tokens, and then the tally underneath as of the design.

### Design

[Figma link](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=6452%3A82853)

163577652-ab62f572-d410-4ec3-8a66-d7124ddb0d9f

chinins:
## Description

This PR removes the claim funds from events page and adds a different version of it - Incoming funds to the right side in the home layout.

The `Claim` button doesn't claim all the funds as there is a separate issue to update saga to do just that.

I would have `Claim all` button for better understanding for the user but the specs said just `Claim` so that's what I implemented. Happy to update if you also think that `Claim all` is clearer.

I removed the arrow from the tooltip because it looks wacky when the screen is a bit more narrow. Not investing too much time here as I know that there is an issue that rewrites the tooltips.

Screenshot 2022-04-21 at 14 00 49

resolves #3316


ArmandoGraterol (changes requested):
Excellent work! I noticed some small things that could be changed before this is ready to go

arrenv (changes requested):
Looking good, in addition to Armando's comments I noticed these UI issues. In regards to the `Claim all` I think that it is more suitable as well.


Tooltip is not consistent with the rest of the DApp, the padding and the font size (font size in Figma is wrong, I will remind Karol to have it updated across the DApp).

Untitled-2

No hover state on the title.

Untitled-123

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+                disabled={!isNetworkAllowed || !hasRegisteredProfile}
+              />
+            </Tooltip>
+          </li>
+          <li>
+            <div className={styles.tokenItem}>
+              +{data?.processedColony.unclaimedTransfers.length - 1}
+              <FormattedMessage {...MSG.more} />
+            </div>
+          </li>
```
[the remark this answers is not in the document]
rdig: Screenshot from 2022-04-26 18-26-07

Agree with Armando, we shouldn't be showing the additional +0 tokens line, in this case _(only if that becomes +1, or +2, ...etc)_

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+            >
+              <ActionButton
+                text={MSG.claimButton}
+                className={styles.button}
+                submit={ActionTypes.COLONY_CLAIM_TOKEN}
+                error={ActionTypes.COLONY_CLAIM_TOKEN_ERROR}
+                success={ActionTypes.COLONY_CLAIM_TOKEN_SUCCESS}
+                transform={transform}
+                disabled={!isNetworkAllowed || !hasRegisteredProfile}
+              />
```
rdig, line 152: Screenshot from 2022-04-26 18-29-05

The claim button styles are a bit weird. Should it have a underline ?

Also the tooltip is:
- missing the proper padding and text alignment
- white color for text, I think normally we have it a bit more gray / faded out _(I could be wrong though, please check)_
- no arrow

Maybe wait with this until Sayan merges his popover refactor PR #3315 and then you'll have a easier time to update the tooltip's styles

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+      <MiniSpinnerLoader
+        className={styles.main}
+        loadingText={MSG.loadingData}
+      />
+    );
+  }
+
+  const token = tokenData?.token;
+
+  return data && data.processedColony.unclaimedTransfers.length ? (
```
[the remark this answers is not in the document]
rdig: Agree

```suggestion
  return data?.processedColony?.unclaimedTransfers?.length ? (
```

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+import Tooltip from '~core/Popover';
+
+import { ActionTypes } from '~redux/index';
+import { mergePayload } from '~utils/actions';
+import { checkIfNetworkIsAllowed } from '~utils/networks';
+import { getTokenDecimalsWithFallback } from '~utils/tokens';
+
+import styles from './ColonyUnclaimedTransfers.css';
+
+const displayName = 'dashboard.ColonyUnclaimedTransfers';
```
rdig, line 25: Since this is nested under `ColonyHome`, I think this would make it so that the name contains it as well:

```suggestion
const displayName = 'dashboard.ColonyHome.ColonyUnclaimedTransfers';
```

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+interface Props {
+  colony: Colony;
+}
+
+const MSG = defineMessages({
+  title: {
+    id: 'dashboard.ColonyUnclaimedTransfers.title',
```
rdig, line 33: The above then also applies to message descriptor id's

```suggestion
    id: 'dashboard.ColonyHome.ColonyUnclaimedTransfers.title',
```

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
```diff
@@ -0,0 +1,169 @@
...
+                transform={transform}
+                disabled={!isNetworkAllowed || !hasRegisteredProfile}
+              />
+            </Tooltip>
+          </li>
+          <li>
+            <div className={styles.tokenItem}>
+              +{data?.processedColony.unclaimedTransfers.length - 1}
+              <FormattedMessage {...MSG.more} />
+            </div>
```
rdig, line 159: Please refactor this to make use of the message descriptor's value prop.

So something like this:
```suggestion
            <div className={styles.tokenItem}>
              <FormattedMessage {...MSG.more} values={{ extraClaims: data?.processedColony?.unclaimedTransfers?.length - 1 }} />
            </div>
```
Maybe even, extract the extra claim value into a separate variable to make it even more readable:
```js
const extraClaims = data?.processedColony?.unclaimedTransfers?.length - 1;
```

And then, inside the message descriptor, you do something like this:

```jsx
  more: {
    id: 'dashboard.ColonyHome.ColonyUnclaimedTransfers.more',
    defaultMessage: `+{extraClaims} more`,
  },
```

rdig (changes requested):
More or less the same thing as the others found. Left notes where appropriate.

src/modules/dashboard/components/ColonyHome/ColonyUnclaimedTransfers/ColonyUnclaimedTransfers.tsx
chinins: > Maybe wait with this until Sayan merges his popover refactor PR #3315 and then you'll have a easier time to update the tooltip's styles

Indeed, I was waiting for it

arrenv (changes requested):
Not sure if something weird is happening, but I am still seeing an inconsistency with the Tooltip.

I am seeing text for 2 more tokens, however, there is only one more token. Can he hide the token while it is still loading, instead of showing the Unknown token text?

claim-funds

arrenv (approved):
Looks good to me. Very nice!

ArmandoGraterol (approved):
Excellent work, Agent Olga!

PR #3327, opened 2022-04-22
Fix number overflow on motion page

### LINKED ISSUE #3262: Protect against overflow errors when converting very large numbers
## Description
When testing with very large numbers (scientific number territory) for the std number display project, it exposed an overflow issue when trying to convert the large number. 
The error is thrown from BigNumber toNumber() when trying to convert. errors.js:76 Uncaught Error: overflow (operation="setValue", fault="overflow", details="Number can only safely store up to 53 bits", version=4.0.47) 

Search out potential vulnerabilities in the Dapp, and implement a defensive strategy (inc. try/catch) to handle such overflows, depending on each context.

## Steps to reproduce

1. Mint a very large number of tokens (in the order of 50x10^15)
2. Try creating a motion to make a payment

To locate potential vulnerabilities, search in the code for `BigNumber toNumber()`

Example of the issue:
image

ArmandoGraterol:
Fixed BigNumbers overflowing on `.toNumber()` in the motion page. I also checked all of the other places where we use `.toNumber()` and it may overflow, however, the only places that gave problems was this one in the motion page.

In order to test:

- Mint a lot of tokens (for example, `100x10^18` or `^20` or even more).
- Make a large payment to yourself with the rep miner ON, with a similar amount as the one minted so that a lot of reputation points are generated.
- With rep, now install the voting extension and make any motion.
- If the dapp doesn't crash when on the motion page, the fix works.

Resolves #3262

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
@@ -243,27 +243,23 @@ const DefaultMotion = ({
 
   const threshold = bigNumberify(
     votingStateData?.votingState?.thresholdValue || 0,
-  )
-    .div(bigNumberify(10).pow(18))
-    .toNumber();
+  ).div(bigNumberify(10).pow(18));
 
   const totalVotedReputationValue = bigNumberify(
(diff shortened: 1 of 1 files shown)


rdig (approved):
Good to go!

PR #3341, opened 2022-04-26
Add liveness check for reputation oracle and IPFS

### LINKED ISSUE #3313: IPFS and Reputation Miner status
The current server-status PR #3259 does not include IPFS and Reputation Miner being behind checks.

Those should be included.

It has to be considered how the reputation miner being behind should be checked.

Blocked by #3259.

ArmandoGraterol:
Added check for IPFS and updated the reputation oracle check.

The IPFS check is only going to work on QA and prod, unfortunately, but the code thankfully is straight forward so it's expected that it should work. 

Resolves #3313

### DIFF
--- a/src/data/resolvers/status.ts
+++ b/src/data/resolvers/status.ts
@@ -1,9 +1,13 @@
 import { Resolvers } from '@apollo/client';
+import { ClientType, ROOT_DOMAIN_ID } from '@colony/colony-js';
+import { AddressZero } from 'ethers/constants';
 
 import { Context } from '~context/index';
 
 import { getLatestSubgraphBlock } from './colony';
 import { NETWORK_DATA, DEFAULT_NETWORK } from '~constants';
+import { PINATA_GATEWAY } from '~lib/pinata/constants';
+import { isDev } from '~utils/debug';
 
 const serverEndpoint = process.env.SERVER_ENDPOINT;
 
@@ -21,6 +25,7 @@ const networkNameForReputationOracle =
 
 export const statusResolvers = ({
   colonyManager: { networkClient },
+  colonyManager,
   apolloClient,
 }: Required<Context>): Resolvers => ({
   Query: {
@@ -51,13 +56,31 @@ export const statusResolvers = ({
         return null;
       }
     },
-
     async isReputationOracleAlive() {
       try {
+        const latestRootHash = await networkClient.getReputationRootHash();
+        const metaColonyAddress = await networkClient.getMetaColony();
+        const colonyClient = await colonyManager.getClient(
+          ClientType.ColonyClient,
+          metaColonyAddress,
+        );
+        const { skillId } = await colonyClient.getDomain(ROOT_DOMAIN_ID);
         const res = await fetch(
-          `${reputationOracleEndpoint}/${networkNameForReputationOracle}`,
+          `${reputationOracleEndpoint}/${networkNameForReputationOracle}/${latestRootHash}/${metaColonyAddress}/${skillId}/${AddressZero}/noProof`,
         );
         return res.status === 200;
+      } catch (error) {
+        console.error(error);
+        return null;
+      }
+    },
+    async isIPFSAlive() {
+      try {
+        if (!(isDev || process.env.DEV)) {
+          const res = await fetch(PINATA_GATEWAY);
+          return res.status === 200;
+        }
+        return true;
       } catch (error) {
         console.error(error);
         return null;
--- a/src/data/resolvers/status.ts
+++ b/src/data/resolvers/status.ts
@@ -23,6 +23,9 @@ const networkNameForReputationOracle =
         process.env.NETWORK || DEFAULT_NETWORK
       ].shortName.toLowerCase(); // NOTE: when adding new networks make sure this will work for it
 
+const initialReputationRootHash =
+  '0x0000000000000000000000000000000000000000000000000000000000000000';
+
 export const statusResolvers = ({
   colonyManager: { networkClient },
   colonyManager,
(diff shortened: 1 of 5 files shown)


rdig (approved):
Seems strait forward.

We'll see how this actually performs when it hits QA :)

dit7ya (changes requested):
The specs originally asked for comparing the latest block synced by the reputation oracle with the latest RPC block as well (and set state to poor if it is behind).

rdig:
> synced by the reputation oracle with the latest RPC block as well

Actually not sure if that's possible. I think we only track submissions for reputation, but better to check with Alex to find this out for sure

ArmandoGraterol:
@dit7ya @rdig Just talked with Alex about this. 

There is no endpoint from which we can know the latest block synced by the oracle, also, so long as we check that the reputation is up-to-date, then it doesn't matter if the rep oracle is a few blocks behind.

rdig:
Works for me

src/data/resolvers/status.ts
```diff
@@ -51,17 +56,35 @@ export const statusResolvers = ({
...
+        const latestRootHash = await networkClient.getReputationRootHash();
+        const metaColonyAddress = await networkClient.getMetaColony();
+        const colonyClient = await colonyManager.getClient(
+          ClientType.ColonyClient,
+          metaColonyAddress,
+        );
+        const { skillId } = await colonyClient.getDomain(ROOT_DOMAIN_ID);
         const res = await fetch(
-          `${reputationOracleEndpoint}/${networkNameForReputationOracle}`,
+          `${reputationOracleEndpoint}/${networkNameForReputationOracle}/${latestRootHash}/${metaColonyAddress}/${skillId}/${AddressZero}/noProof`,
```
dit7ya, line 69: This fails with 400 when there is no reputation in the colony (for example when the Motions Extension has not been enabled). That makes the Network Status change to Poor.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
```diff
@@ -90,7 +96,16 @@ const NetworkStatus = () => {
...
+  }, [
+    latestRpcBlockError,
+    isReputationOracleAlive,
+    isColonyServerAlive,
+    ipfsLivenessData,
+    latestRpcBlock,
+    latestSubgraphBlock,
+    latestSubgraphBlockError,
+    networkCheckInterval,
+  ]);
```
dit7ya, line 108: Are you sure about this dependency array? IIUC, it means if one of those requests take some time and changes one of these things, the setInterval starts afresh. This means in practice there are way more frequent requests to the server than the 5 minutes interval (which puts extra pressure on the server).

src/data/resolvers/status.ts
```diff
@@ -51,17 +56,35 @@ export const statusResolvers = ({
...
         );
         return res.status === 200;
       } catch (error) {
         console.error(error);
         return null;
       }
     },
+    async isIPFSAlive() {
+      try {
+        if (!(isDev || process.env.DEV)) {
```
dit7ya, line 79: This is just a stylistic preference but I would do this check in the component rather than here (I think that avoids some extra computation, I am not super sure though.)

dit7ya (changes requested):
I have some comments and questions but otherwise looks good. :+1:

src/data/resolvers/status.ts
ArmandoGraterol: Realistically, I feel like this case shouldn't happen but yeah, while in the dev it could be misleading.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
ArmandoGraterol: Yes, I'm sure. This `useEffect` is not making any requests to the networks, it's only modifying the local state of `networkHealth` via `setNetworkHealth` so no requests are made. 

Also now that you mention this, I took a closer look at the interval and I believe that it can be safely removed now that we have a dependency array, it should also be better in the sense that before without the dependency array and the interval the useEffect would not know of any updated values

src/data/resolvers/status.ts
ArmandoGraterol: Whether we move this check to the component or we keep it here, we would need to do some extra computation, yes?  For example, if we moved this check to the component, due to the rule of hooks, we couldn't just do something like, unfortunately:

```
if (!(isDev || process.env.DEV)) {
    const { data: isReputationOracleAlive } = useReputationOracleLivenessQuery({
      pollInterval: networkCheckInterval,
    });
}
```

We would have to use something like a lazy query, and then conditionally request the data in a useEffect or something similar, so the difference in performance between these 2 methods may be marginal.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
dit7ya: Ummm, I am still not convinced about this. So, are you guaranteeing that the checks will demonstrably fire at every 5 minutes **only**? Last time I checked they were firing much more frequently than that (due to the reason I mentioned above. Maybe I am missing something. If you are confident then let me know and I will approve this.

dit7ya (commented):
Looks all good except the comment I left.

src/modules/core/components/NetworkStatus/NetworkStatus.tsx
ArmandoGraterol: If you are so unsure about it you don't need to rely on my words to know if this works like I'm saying it does or not, we can _actually_ test it:

Placed some `console.log()` inside this `useEffect` and each of the checks we do:
Screenshot from 2022-05-04 12-21-31

This is the initial load, component renders, `useEffect` and queries are executed for the first time. After some of the queries return the data the `useEffect` is executed once more because the values were updated by the queries.

Again a few seconds later the useEffect ran again because some of the above queries probable finished returning data:
Screenshot from 2022-05-04 12-18-19

Finally, nothing more happens until _5 minutes_ later when the queries ran again due to the polling interval that they have: 
Screenshot from 2022-05-04 12-23-41

BUT the `useEffect` didn't run again because the values of the queries didn't change since the last time it was queried.

So, `useEffect` running 6 times in total as of now doesn't affect the number of queries because the `useEffect` itself is not executing any queries, the `useEffect` is only doing state management (setting network health).

Hope this clears any doubts.

dit7ya: I missed the polling logic. And with the removal of the interval I think it works as expected. Great work!

dit7ya (approved):

PR #3353, opened 2022-05-02
Saving settings: Simple confirmation message

### LINKED ISSUE #3265: Saving settings confirmation message

danbr:
## Description

- This PR is a follow-on from https://github.com/JoinColony/colonyDapp/pull/3284. 
- The requirements were updated to include:

> After looking into this more, and noticing the size of the Toastify dependency, it would be better to just have a notice appear underneath the save button.
> 
> Please see spec here - https://www.figma.com/file/UzfpoAvGUS6c14du7Y39sCYm/User-Settings?node-id=2557%3A2821
> This will also probably work better with a 3 sec timer as it won't need to be closable.
> Have the notice animate in and out.

-This PR displays a confirmation message for 3 secs when a user updates their settings (un)successfully.

Screenshot 2022-05-02 at 23 18 04

Resolves #3265


arrenv (approved):
Looks great to me! And the reusability is good :)

saving-settings

ArmandoGraterol (approved):
Looks much better than before! Great job

PR #3358, opened 2022-05-04
Standardise `learn more` text & link behaviour

### LINKED ISSUE #3356: Inconsistent onhover behaviour of links
##Description

Across the Dapp, the onhover behaviour of links varies, even within the same dialog.
Screenshot 2022-05-03 at 18 36 25
Screenshot 2022-05-03 at 18 48 21
Screenshot 2022-05-03 at 18 52 28
Screenshot 2022-05-03 at 18 52 45

Search & destroy these inconsistencies.

The expected onhover behaviour (as shown above in first image):

1. underline link
2. link colour switches to dark blue

danbr:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #3356


arrenv (approved):
Looks more consistent to me!

chinins (commented):
I think it's great to have more consistency but I see lots of repeating code. I think it would be good to add a prop to the external link (theme, appearance - similar to how it's done with other components) so that we can eliminate all of that repeating css.

dit7ya (changes requested):
More consistent styling, yay! I found some other instances of learn more links having different styling.

https://github.com/joincolony/colonyDapp/blob/f07457b1b5af1854782c93033a1ce096cbd69d24/src/modules/dashboard/components/CoinMachine/CoinMachine.css#L26



https://github.com/joincolony/colonyDapp/blob/f07457b1b5af1854782c93033a1ce096cbd69d24/src/modules/core/components/BetaCautionAlert/BetaCautionAlert.css#L34

https://github.com/joincolony/colonyDapp/blob/f07457b1b5af1854782c93033a1ce096cbd69d24/src/modules/dashboard/components/ColonyHome/ColonyUpgrade/ColonyUpgrade.tsx#L114

danbr:
> I think it's great to have more consistency but I see lots of repeating code. I think it would be good to add a prop to the external link (theme, appearance - similar to how it's done with other components) so that we can eliminate all of that repeating css.

On review, i agree the implementation of the goal of standardisation can be optimised further.
Most of the styling being passed to `ExternalLink` is repeated, and therefore can be included in the styling of the component to reduce unnecessary repeating of css. 
There is already a warning comment in the code for the className prop not to ABUSE:

- "Allows for link style customization (Eg: we need to disquise the link as a button) 
- Don't abuse it!"

But I dont think implementing themes & appearance is the appropriate approach. 
If we want consistency, external links should all generally look & behave the same. It should have default styling, that is seldom overriden. Not a range of different themes. 

If we did use themes in `ExternalLink`- there would be only 1 (not a range of different themes), which is the default styling.

Also to implement themes, it require a lot of work and attention to remove css styling that is already being passed, and then checking that the appropriate styling is being applied and no bugs or side effects added.

On a different tack, as discussed with @arren, to standardise internal links, we should create an `InternalLink` component with standardised styling & behaviour.

danbr:
Sorry guys but i made a more drastic change, so perhaps its worth re-reviewing?

I became braver (or more stupid) and decided to refactor to reduce duplicate CSS. This should promote more consistency, and give a clear direction for future styling.
With these changes, `ExternalLink` has default css which can be added to or overridden ONLY when necessary. 

I tried to not break anything and maintain the same UI. I didn't make `blanket changes` to everything. If i was too hawkish/dovish in applying the updates, then it should be easy to compare, identify and apply the same approach.

chinins (approved):
Very good! Nice job!

dit7ya (approved):
Very nice!

PR #3359, opened 2022-05-04
Update Contracts and Events

rdig:
This PR updates `colonyJS` to support the newly released contract versions, as well as update all the related submodules that are relevant to this _(`colonyNetwork` and `subgraph`)_:
- Colony `v9`
- OneTx `v3`
- Coin Machine `v4` and `v5` _(defaults to the latest)_
- Voting Reputation `v4`
- Whitelist `v2`

This also updates the `SubgraphEventsThatAreActions` subscription so that it tracks the both the old `TokenUnlocked()` and the new `TokenUnlocked(address)` events that were added to the contract

I've also updated the `subgraph` repo separately so that it tracks _(and also supports)_ the new events.

Note that the `rc` release of `colonyJS` is temporary, and once I confirm it works on QA, will release a mainline version and will update this PR with it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "ISC",
       "dependencies": {
         "@apollo/client": "^3.0.2",
-        "@colony/colony-js": "^4.1.2",
+        "@colony/colony-js": "^4.1.3-rc.0",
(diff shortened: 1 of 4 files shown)


chinins (approved):
The most time it took for the review was actually to rebuild the environment :)

PR #3362, opened 2022-05-04
Add validation for toDomain value

### LINKED ISSUE #3344: Incorrect permissions in UI for transferring funds to a parent domain
## Steps to reproduce

1. Ensure you have a subdomain setup and with funds in it.
2. Have a user set up with the Funding permission in that subdomain.
3. Try to transfer funds from the subdomain to a parent domain or other subdomains.
4. It will be permissible in the UI, however, the contract will reject it.
5. This is also the case when you have the Motions and Disputes Extension enabled, and have reputation in the child domain. You have permission to force in the UI, but not on the contract.

## Expected behaviour

* The UI should not show the permission to transfer funds from a subdomain to a parent domain and other subdomains.

## Actual behaviour

* The UI shows incorrectly that you have permission to transfer funds from a subdomain to a parent domain and other subdomains.

download (39)
unknown
unknown23

ArmandoGraterol:
Added missing validation for the toDomain balue in the transfer funds dialog, now when creating an action an user without permission in any of the 2 domains will not be able to submit.

FireShot Capture 588 - Actions - Colony - wonker - localhost

Resolves #3344

### DIFF
--- a/src/modules/dashboard/components/Dialogs/TransferFundsDialog/TransferFundsDialogForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/TransferFundsDialog/TransferFundsDialogForm.tsx
@@ -125,7 +125,9 @@ const TransferFundsDialogForm = ({
   const toDomainId = values.toDomain
     ? parseInt(values.toDomain, 10)
     : undefined;
-
+  const toDomain = domains.find(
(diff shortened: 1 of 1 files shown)


rdig (approved):
Seems to be working fine

arrenv (approved):
Working for me, it shows required permissions or creates a motion. Nice one!

PR #3363, opened 2022-05-05
Plug network fees into CreatePaymentDialog validation

### LINKED ISSUE #3349: Trying to pay all funds results into VM error
## Steps to reproduce

1. Mint some tokens/ have some tokens
2. Enter that amount in a Create Payment action

## Expected behavior
It should not let you take that action, given that the network counting the network fee, it exceeds the total available funds.

## Actual behavior
Payment action is allowed but fails with VM error.

## Suggestions 

This is introduced as a side-effect of #3275. The available funds check needs to incorporate the network fees too.
Also related, maybe there can be a view (tooltip on hover on a info icon) that shows the total amount to be spent including  fees. This needs to be checked with product, what do you think @arrenv ?

dit7ya:
User is prevented from paying all available funds in a Create Payment.
Resolves #3349


ArmandoGraterol (approved):
It's working as expected :ok_hand:

rdig (approved):
Looks good! :+1:

PR #3389, opened 2022-05-11
Resolves confusing characters & warnings 

### LINKED ISSUE #3268: [verified recipents] `m` is being treated as a confusable character
## Steps to reproduce

While creating your username, use any word that has the English letter `m` in it. 

Though `m` does look like `rn`, I don't think that is enough reason to shun it.

Screenshot 2022-03-28 at 10-07-56 Colony
Screenshot 2022-03-28 at 10-08-22 Colony

danbr:
## Description

This PR updates colonyDapp to use `@colony/unicode-confusables-noascii`. As the name suggests, the new package does not include ascii characters (numbers are still considered) in its ruleset.
Therefore, 'm' & 'l' will not be considered worthy of a warning.

Screenshot 2022-05-10 at 19 36 27

Screenshot 2022-05-10 at 19 36 33

A more detailed explanation, and the hidden part of this PR:

The package unicode-confusable has a script that at build time, takes the official unicode definitions ([https://unicode.org/Public/security/10.0.0/confusables.txt](https://unicode.org/Public/security/10.0.0/confusables.txt "https://unicode.org/Public/security/10.0.0/confusables.txt")) and extracts to confusables.json which is then distributed within the NPM and used as the source of its decisions of confusability. 

Steps followed:
- Forked unicode-confusable (currently to my github - need to move to colony) 
- Updated the script to ignore ascii characters from being included in confusables.json. 
- All other routines are untouched and remain the same as original package. 
- Published NPM '@colony/unicode-confusables-noascii' 
- Updated colonyDapp to use '@colony/unicode-confusables-noascii' 

This approach should be more inline with Metamask and other users of confusable, and hopefully requires less maintenance in the future. Part of which can be 'piggybacked' from unicode-confusable

Resolves #3268


chinins (approved):
Nice work!

ArmandoGraterol (approved):
From what I could see, it's working great! Surely all the Marks and Ians in the World would appreciate this update

PR #3392, opened 2022-05-12
Update members list's items in the Members page

### LINKED ISSUE #3385: Updates to member items on members page
### Description

Update the members item with the following:
- Move reputation value from the start to the end as of the design
- Add the value of reputation points (Numeral formatted) to `MemberReputation` component. Have this as a conditional element based on newly added boolean prop.
- Add the kebab menu item and menu with actions as of the design
    - The "Add to Address Book" would open the dialog box for the "Address book" with the users address pre-filled. https://github.com/JoinColony/colonyDapp/issues/3371

The kebab menu should allow for

### Design
[Figma link: Members items](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=6736%3A93430)
[Figma link: Kebab menu](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=6739%3A94092)

Screenshot 2022-05-10 210035

Screenshot 2022-05-10 210014

ArmandoGraterol:
Updated styles of mainly `MembersListItem` and added new things to it like reputation points display along with the percentage and a members action popover that can be used to ban, view the user on blockscout, and add them to the address book. The logic related to the sorting row will be done in a separate PR.

FireShot Capture 629 - Members - Colony - wonker - localhost

#### TODO

- [x] Connect `Add to address book` member action to the verified recipient's dialog

Resolves #3385

### DIFF
--- a/src/modules/core/components/MemberReputation/MemberReputation.tsx
+++ b/src/modules/core/components/MemberReputation/MemberReputation.tsx
@@ -4,6 +4,7 @@ import { ROOT_DOMAIN_ID } from '@colony/colony-js';
 import { AddressZero } from 'ethers/constants';
 import Decimal from 'decimal.js';
 
+import { DEFAULT_TOKEN_DECIMALS } from '~constants';
 import { useUserReputationQuery } from '~data/index';
 import { Address } from '~types/index';
 import Numeral from '~core/Numeral';
@@ -32,6 +33,7 @@ interface Props {
   onReputationLoaded?: (reputationLoaded: boolean) => void;
   showIconTitle?: boolean;
   showReputationPoints?: boolean;
+  nativeTokenDecimals?: number;
 }
 
 const displayName = 'MemberReputation';
@@ -44,6 +46,7 @@ const MemberReputation = ({
   onReputationLoaded = () => null,
   showIconTitle = true,
   showReputationPoints = false,
+  nativeTokenDecimals = DEFAULT_TOKEN_DECIMALS,
 }: Props) => {
   const { data: userReputationData } = useUserReputationQuery({
     variables: { address: walletAddress, colonyAddress, domainId, rootHash },
@@ -66,7 +69,7 @@ const MemberReputation = ({
   );
   const formattedReputationPoints = getFormattedTokenValue(
     new Decimal(userReputationData?.userReputation || 0).toString(),
-    18,
+    nativeTokenDecimals,
   );
 
   useEffect(() => {
--- a/src/modules/core/components/MemberReputation/MemberReputation.tsx
+++ b/src/modules/core/components/MemberReputation/MemberReputation.tsx
@@ -2,16 +2,16 @@ import React, { useEffect } from 'react';
 import { defineMessages } from 'react-intl';
 import { ROOT_DOMAIN_ID } from '@colony/colony-js';
 import { AddressZero } from 'ethers/constants';
-
 import Decimal from 'decimal.js';
+
 import { useUserReputationQuery } from '~data/index';
 import { Address } from '~types/index';
 import Numeral from '~core/Numeral';
 import Icon from '~core/Icon';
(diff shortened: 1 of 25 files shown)


src/modules/core/components/MemberReputation/MemberReputation.tsx
```diff
@@ -60,6 +64,10 @@ const MemberReputation = ({
     userReputationData?.userReputation,
     totalReputation?.userReputation,
   );
+  const formattedReputationPoints = getFormattedTokenValue(
+    new Decimal(userReputationData?.userReputation || 0).toString(),
+    18,
```
rdig, line 69: Reputation should have the decimals of the native token, as far as I remember.

Maybe check with Alex, but I think this should change

src/modules/core/components/MembersList/Actions/MemberActionsPopover.tsx
```diff
@@ -0,0 +1,124 @@
...
+    () =>
+      openBanUserDialog({
+        colonyAddress: colony.colonyAddress,
+      }),
+    [openBanUserDialog, colony],
+  );
+  const openManageWhitelistDialog = useDialog(ManageWhitelistDialog);
+  const handleManageWhitelist = useCallback(
+    () =>
+      // @ts-ignore
```
rdig, line 65: Why is this ignored ?

src/modules/core/components/MembersList/Actions/MemberActionsPopover.tsx
```diff
@@ -0,0 +1,124 @@
...
+      openManageWhitelistDialog({ userAddress, colony }),
+    [openManageWhitelistDialog, userAddress, colony],
+  );
+
+  const BLOCKSCOUT_URL = `https://blockscout.com/xdai/mainnet/address/${userAddress}/transactions`;
```
rdig, line 70: Please add the url constant in `src/modules/externalUrls.ts` file

rdig (changes requested):
If you set a user's display name to a very long string, it will break the table / push the UI outside the viewport

Screenshot from 2022-06-14 21-20-07
Screenshot from 2022-06-14 21-20-26

Also, found these as well, although not really sure if they're related to this PR not. I'm still going to point them out though.

1. Even though user `@b` has reputation, they are still listed under the "watchers" section, rather then "contributors"
Screenshot from 2022-06-14 20-53-49
Screenshot from 2022-06-14 20-54-02
Screenshot from 2022-06-14 20-54-07

2. Search icon is too large
Screenshot from 2022-06-14 20-53-54

willm30:
FYI Search icon has been fixed: https://github.com/JoinColony/colonyDapp/pull/3470

Will merge today.

src/modules/core/components/MemberReputation/MemberReputation.tsx
ArmandoGraterol: You are right! Forgot to change this

src/modules/core/components/MembersList/Actions/MemberActionsPopover.tsx
ArmandoGraterol: Fixed. I was using the dialog as I saw it in other places, using it like this returned some confusing ts error, but I took a closer look and it seems like with a small adjustment this ignore won't be needed. So, I'll remove it here and others places where I saw it

chinins (dismissed):

ArmandoGraterol:
@rdig The bug related to contributors appearing as watchers will be handled with #3507, so this is ready for a re-review :)

rdig (approved):
Other than the fix from #3507 this should be good to go

Screenshot from 2022-06-17 14-51-47

chinins (approved):
Very nice work! I'm merging the fix for #3507 now, so you'll have it after the rebase.

Issue #3390, opened 2022-05-12
Expenditures: As a User I can change top parameters


arrenv:
[Provided by Michał via Contractors project management platform]

### Description
User should be able to change the top parameters like Expenditure type, team and owner and also check the balance.

### Acceptance Criteria

- an option to change the Expenditure type
- an option to change team
- an option to check the balance
- an option to change the owner (autocomplete component)

### Design Link
https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2416%3A320414
![23df800cf2ef1bd8a391bb21fa1d755b](https://user-images.githubusercontent.com/33682027/168041511-ce451b41-ea27-4e98-9b6c-ccce54593407.png)

![23d9be6df14b201cc201c255186b35b8](https://user-images.githubusercontent.com/33682027/168041515-6813c513-6b7e-4c8d-8f84-40b59dc432b3.png)

![23e2b18ddd8c094a791613c26952ad7c](https://user-images.githubusercontent.com/33682027/168041516-77212afa-bfe5-41b1-ade3-44c0b491730f.png)

![23ddc5f4aad7a6599a50512bb8cedc28](https://user-images.githubusercontent.com/33682027/168041517-14660d15-b2ca-4f1f-b634-3d681c6c90ba.png)

PR #3393, opened 2022-05-13
Add top parameters (expenditure type, team, owner and balance) on the expenditure creation page

agnieszkajarosikloj:
This PR adds the first part of the form to create new expenditure. 
[Figma](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2416%3A320414)

Resolves  https://github.com/JoinColony/colonyDapp/issues/3390


jdakowicz (approved):

src/modules/core/components/Fields/InputStatus/InputStatus.tsx
```diff
@@ -14,6 +14,7 @@ export interface Appearance {
   statusShema?: 'info';
   size?: 'small' | 'medium';
   textSpace?: 'wrap';
+  align?: 'right';
```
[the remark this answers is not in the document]
chinins: I  actually don't think that more values are needed here, similar to other props: direction, textSpace. It's mote for the optional use of this css.

chinins (changes requested):
This PR will take some time to review so here are several things to potentially start fixing:
- In the expenditure type we should have 3 selectable options but there is only 1 - Advanced
Screenshot 2022-05-13 at 22 20 44

- I assume the cursor should be a pointer on all of the selectable elements. It is on all but one - the owner.
- the avatar (the owner field) is not clickable, only a small caret next to it. It should be clickable to open the list of owners.
- when I try to search for the users in the owners' list it gives me error and the app crashes
Screenshot 2022-05-13 at 22 14 00

src/modules/core/components/Fields/InputStatus/InputStatus.tsx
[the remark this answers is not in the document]
agnieszkajarosikloj: I wrote "align" this way, similar like textSpace and statusSchema were written before, because I add this class using getMainClasses function: 
```
<Element
      className={getMainClasses(appearance, styles, {
        error: !!error,
        hidden: !text || (!!error && !isNil(touched) && !touched),
      })}
    >
```
With getMainClasses function I don't have to write condition to add this class.

arrenv:
> In the expenditure type we should have 3 selectable options but there is only 1 - Advanced

@agnieszkajarosikloj @chinins We only have the one on initial launch, so having only the 1 - Advanced is fine for now. Providing the selection is still working.

> the avatar (the owner field) is not clickable, only a small caret next to it. It should be clickable to open the list of owners.

@agnieszkajarosikloj @chinins There was a last minute change on the contracts side. There is no need to have a caret or selection on the Owner field.

chinins:
> arrenv

arrenv:
@chinins for the moment, it should be there, just not changeable. It is auto added based on who creates the Expenditure.

src/modules/dashboard/components/ExpenditurePage/TopParameters/TopParameters.tsx
```diff
@@ -0,0 +1,166 @@
...
+import TeamDropdownItem from '~dashboard/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/TeamDropdownItem';
+import XDAIIcon from '../../../../../img/tokens/xDAI.svg';
+
+import { balanceData, colonyAddress, domains } from './consts';
+import { Appearance } from '~core/Fields/Select/types';
+import { Appearance as DialogAppearance } from '~core/Dialog/DialogSection';
+
+const MSG = defineMessages({
+  defaultExpenditureTypeLabel: {
+    id: 'TopParameters.defaultExpenditureTypeLabel',
```
arrenv, line 20: Can we use more descriptive id's for these, `expenditures.settings.type.label`

arrenv (changes requested):
This is a big one @agnieszkajarosikloj, nice.

1. We have a token component, so, you can remove the two SVG icons, please take a look at the `HookedTokenIcon` component `src/modules/dashboard/components/HookedTokenIcon/HookedTokenIcon.tsx`

2. There appear to be a lot of differences in the padding and alignment between the design and the PR. See the side by side below, I understand Figma can make it difficult to get exact paddings, but we should match by appearance as best as possible.


Untitled-1w
 

3. Can we please rename `TopParameters` to `ExpenditureSettings`

arrenv (changes requested):
In terms of UI I noticed these two things.

fc3e1974-f6a0-46c8-82d6-3f9dbe406e6b

download (5)

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureSettings.css
```diff
@@ -39,13 +40,15 @@
   justify-content: flex-end;
   align-items: center;
   padding: 0;
+  font-size: 14px;
```
arrenv, line 43: We have a variable for this font size `--size-normal`. In this file `src/styles/variables.css`

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureSettings.css
```diff
@@ -39,13 +40,15 @@
...
   font-weight: var(--weight-normal);
   color: var(--dark);
 }
 
 .option {
   margin-bottom: 12px;
   padding: 16px 20px 0 0;
+  font-size: 14px;
```
arrenv, line 51: Save with this on, we can use the variable `--size-normal`.

arrenv (commented):
@agnieszkajarosikloj It is looking much better nice one, just a couple of comments.

src/modules/core/components/Dialog/DialogSection.tsx
```diff
@@ -4,10 +4,11 @@ import { getMainClasses } from '~utils/css';
 
 import styles from './DialogSection.css';
 
-interface Appearance {
+export interface Appearance {
   theme?: 'heading' | 'sidePadding' | 'footer';
   align?: 'center' | 'right';
   border?: 'top' | 'bottom' | 'none';
+  margins?: 'small' | 'large';
```
chinins, line 11: Please, remove the `margins` prop & create another core component `InfoSection` ( or something like this) where you reuse `DialogSection`. There you can just have the margin as you need it. I would recommend just have it without the prop even. Just give a default value for now (the one that you need).

src/modules/core/components/Fields/Select/Select.css
```diff
@@ -156,3 +189,22 @@
 .baseSelect.sizeLarge {
   font-size: var(--size-large);
 }
+
+.activeOptionColorBlue {
```
chinins, line 193: We would prefer that you don't modify our core components and just use 2 options instead:
- target the change with the css from the component where you use the core component
- create a wrapper component around where you would add the necessary changes.

The second option is least preferable and should be used only when there are several places where it's used and the code would be repeated.

In this case I feel that the active option color can easily be done via css without adding any extra props so please remove and do it via css.

src/modules/core/components/Fields/InputLabel/InputLabel.css
```diff
@@ -38,6 +38,10 @@
   color: var(--dark);
 }
 
+.colorSchemaLightGrey .labelText {
```
chinins, line 41: Please, remove this extra variant and just use the dark colour schema. I think Arren has updated the specs.
Just generally, when you find that there are different colours to our core components, check with Arren please for him to confirm which ones of the existing colours to use.

src/modules/core/components/Fields/Select/Select.css
```diff
@@ -17,12 +17,35 @@
   composes: field from '~styles/fields.css';
 }
 
+.containerHorizontal {
```
chinins, line 20: For this please create a horizontal Select wrapper around the Select component

chinins (dismissed):
Nice work. There are some major changes we would like you to do. Please, see my comments.

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureSettings.tsx
```diff
@@ -0,0 +1,247 @@
...
+      return domain ? domain.color : defaultColor;
+    },
+    [colonyData],
+  );
+
+  const renderActiveOption = useCallback<
+    (option: SelectOption | undefined, label: string) => ReactNode
+  >(
+    (option, label) => {
+      const value = option ? option.value : undefined;
```
rdig, line 89: Why not this ?

```suggestion
      const value = option?.value;
```

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureSettings.tsx
```diff
@@ -0,0 +1,247 @@
...
+      const color = getDomainColor(value);
+      return (
+        <div className={styles.activeItem}>
+          <ColorTag color={color} />{' '}
+          <div className={styles.activeItemLabel}>{label}</div>
```
rdig, line 94: The space that the string occupies could have been more easily achieved via css (a margin or a padding)

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureSettings.tsx
```diff
@@ -0,0 +1,247 @@
...
+      );
+    },
+    [getDomainColor],
+  );
+
+  const filterDomains = useCallback((optionDomain) => {
+    const optionDomainId = parseInt(optionDomain.value, 10);
+    if (optionDomainId === 0) {
+      return false;
+    }
```
rdig, line 105: We have a constant for this `COLONY_TOTAL_BALANCE_DOMAIN_ID` _(although it has a bit of an awkward name)_

rdig (approved):
Looks good, minor nitpicks here and there

arrenv (approved):
Very nice @agnieszkajarosikloj, looking good!

PR #3394, opened 2022-05-13
Add "Total reputation in team" section

### LINKED ISSUE #3384: Team Reputation points balance on members page
### Description

Show the current balance of points in a Team on the Members aside

### Design

[Figma link](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=6736%3A93430)

Screenshot 2022-05-10 205310

ArmandoGraterol:
Added section in the sidebar of the colony members page that shows the total reputation points in the domain selected. The amount should change depending on what team is selected by the current domain selector in the page's header.

FireShot Capture 591 - Members - Colony - wonker - localhost
FireShot Capture 592 - Members - Colony - wonker - localhost

Resolves #3384

### DIFF
--- a/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
+++ b/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
@@ -2,9 +2,9 @@ import React, { useCallback, useEffect, useState } from 'react';
 import { useParams, Redirect } from 'react-router-dom';
 import { defineMessages, FormattedMessage } from 'react-intl';
 import { ColonyVersion, Extension } from '@colony/colony-js';
-
 import Decimal from 'decimal.js';
 import { AddressZero } from 'ethers/constants';
+
 import Button from '~core/Button';
 import { useDialog } from '~core/Dialog';
 import { BanUserDialog } from '~core/Comment';
--- a/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
+++ b/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
@@ -1,6 +1,6 @@
 import React, { useCallback, useEffect, useState } from 'react';
 import { useParams, Redirect } from 'react-router-dom';
-import { defineMessages, FormattedMessage } from 'react-intl';
+import { defineMessages } from 'react-intl';
 import { ColonyVersion, Extension } from '@colony/colony-js';
 import Decimal from 'decimal.js';
 import { AddressZero } from 'ethers/constants';
@@ -9,6 +9,7 @@ import Button from '~core/Button';
 import { useDialog } from '~core/Dialog';
 import { BanUserDialog } from '~core/Comment';
 import Heading from '~core/Heading';
+import Numeral from '~core/Numeral';
 import {
   COLONY_TOTAL_BALANCE_DOMAIN_ID,
   DEFAULT_TOKEN_DECIMALS,
@@ -61,10 +62,6 @@ const MSG = defineMessages({
   totalReputationTitle: {
     id: 'dashboard.ColonyMembers.totalReputationTitle',
     defaultMessage: 'Total reputation in team',
-  },
-  reputationPoints: {
-    id: 'dashboard.ColonyMembers.reputationPoints',
-    defaultMessage: '{teamReputationPoints} reputation points',
   },
 });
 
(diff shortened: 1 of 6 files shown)


dit7ya (commented):
I am encountering a strange bug which I cannot deterministically reproduce but it occurs when going back and forth to the members page using browser history controls. The page is blank and this is what I see in the console.



Otherwise looks good to me mostly. Also, do you think the total reputation for the All Teams view should be called 'Total reputation in colony' instead of 'Total reputation in team''?

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -108,6 +152,19 @@ const ColonyMembers = () => {
     });
   }, [openPermissionManagementDialog, colonyData, isVotingExtensionEnabled]);
 
+  const selectedDomain = colonyData?.processedColony.domains.find(
+    ({ ethDomainId }) => ethDomainId === selectedDomainId,
+  );
```
arrenv, line 157: I think we need to add a check to `processedColony`. When navigating to the Members page I am getting null for the value, crashing the DApp.

Screenshot 2022-05-16 114323

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -108,6 +152,19 @@ const ColonyMembers = () => {
...
+
+  const nativeToken = colonyData?.processedColony.tokens.find(
```
arrenv, line 159: Here as well.

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -108,6 +152,19 @@ const ColonyMembers = () => {
...
+    ({ address }) => address === colonyData?.processedColony.nativeTokenAddress,
+  );
+
+  const formattedTotalDomainRep = getFormattedTokenValue(
+    new Decimal(totalReputation?.userReputation || '0').abs().toString(),
```
arrenv, line 164: I think we should use Numeral formatting on this.

arrenv (changes requested):
Looking good, a couple of comments.

rdig (approved):
Looks good, but the error Sayan noticed needs to be properly investigated, as I think it's some intermittent loading error _(race condition?)_

And if it shows up locally, it's going to go nuts on QA or Production

arrenv (approved):
Nice one! I love it when a plan comes together.

67e476f1-f757-4c38-8516-1765d2334503

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
```diff
@@ -108,6 +149,20 @@ const ColonyMembers = () => {
...
+  );
+
+  const nativeToken = colonyData?.processedColony?.tokens.find(
+    ({ address }) =>
+      address === colonyData?.processedColony?.nativeTokenAddress,
+  );
+
+  const formattedTotalDomainRep = getFormattedTokenValue(
+    new Decimal(totalReputation?.userReputation || '0').abs().toString(),
+    nativeToken?.decimals || DEFAULT_TOKEN_DECIMALS,
```
chinins, line 163: The number of decimals in the reputation, does it really depend on which token we use? That doesn't seem logical to me... It's not a token value, shouldn't it use the same decimals across all colonies? What am I missing?

chinins (commented):
Looking good, I just have a question about the reputation - as per my comment.

src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
ArmandoGraterol: As a disclaimer, I'm following the example of places where Mr.Dan used the `getFormattedTokenValue` to format reputation.

With that being said, this makes sense to me because, from what I understand, when a user is paid a certain amount of tokens, an equal amount of token is generated as reputation points, right?

So If we pay someone `25`, which without formatting it could be `25x10^18`,  `25x10^15`. or  `25x10^23` just to give some examples, we can expect that the generated rep points will be the same as the paid amount.

I could be wrong tho, as this is just my understanding from seeing how the reputation oracle generates the exact amount of tokens that the user is paid.

chinins (approved):
It's all good! Could you, please, just extract the total reputation UI and associated logic into a separate component. I think it will be better for the code structure. Our components are so huge and it's better to split them up a bit. In my PR I'm extracting ColonyControls from the `ColonyMembers` to keep them separate.
Thank you!

dit7ya:
Is it just me? I am still encountering the error I wrote about earlier.

ArmandoGraterol:
@dit7ya Are you 100% sure that your branch is up-to-date? Just re-tested and there were times where `processedColony` came up as `null` due to the data being loaded and it didn't crash

PR #3422, opened 2022-06-01
Fix watchers invite link tooltip

### LINKED ISSUE #3421: Fix Tooltip on Invite link
## Steps to reproduce

1. Use the Invite link on the new Watchers section of the Contributors and Watchers feature.

## Expected behaviour

* Tooltip should be appearing correctly and not broken.

## Actual behaviour

* Tooltip is not working as expected, including appearing in the wrong position and breaking. Refer to gif below.

invite-link-popover

dit7ya:
## Description

Fixes weird behavior of the tooltip, also matches the spec more closely for the tooltip width.

Resolves #3421.


rdig (approved):
Good to go

arrenv (approved):
Working better now, great!

PR #3429, opened 2022-06-07
Add logs of the expenditure

### LINKED ISSUE #3416: Expenditures: As a user I can check logs of the expenditure
[Provided by Michał via Contractors project management platform]

## Description
User is able to add comments and also check logs that will show after taken actions across the app

## Acceptance Criteria

- an to option to add comments
- an option to check logs
- an option to display log after taken actions
- possible logs:
  - staked
  - creating draft
  - canceling draft
  - reclaimed the stake
  - locking the expenditure
  - changing the expenditure
  - creating a motion
  - changing owner
  - adding founds
  - ending the expenditure
  - realising the founds
  - claiming founds

## Design
- https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2420%3A1080905
- https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2420%3A1087531
- https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2424%3A323596

cc815eec9ecb38a7ba6b5b29295f5151

cc79c1d6b55f4b385c2ced755d144d0f

cc8347a3f023651e1d6d516f3345d41b

agnieszkajarosikloj:
## Description

Screenshot 2022-06-07 at 12 23 25
UI part of section where you can check logs of the expenditure and add a comment.
Logs and comments data is temporary.
CommentInput component was used to add a comment, but it has to be checked if the endpoint to submit a comment is correct.

[Figma design:](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2420%3A1080905)

## TODO

- add backend data
- check a logic to submit a comment and add a logic to  display comments

Resolves #3416


arrenv (changes requested):
@agnieszkajarosikloj There appears to be issues with the font size, padding and alignment. Which, I think would be resolved if you instead used the same implementation as how `ActionsPageFeed`, `<ActionsPageEvent` work. Which, you can see a full example of with system messages included, here:

https://github.com/JoinColony/colonyDapp/blob/e091ab95a23be07591bd56417c075af7f0a6050d/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx#L536

Feel free to message me if you need help with this.

src/modules/dashboard/components/ExpenditurePage/LogsSection/LogsSection.css
```diff
@@ -0,0 +1,105 @@
...
+  background-color: inputBackground;
+  font-size: var(--size-smallish);
+  font-weight: var(--weight-normal);
+  line-height: var(--size-medium-l);
+  box-shadow: shadow;
+}
+
+.disabledComment input::placeholder {
+  font-style: normal;
+  color: var(--text-disabled);
```
arrenv, line 46: This is not needed anymore, right? As you are using the `CommentInput` component.

src/modules/dashboard/components/ExpenditurePage/LogsSection/LogsSection.tsx
```diff
@@ -0,0 +1,247 @@
...
+        //   {isFormEditable ? (
+        //     <div className={styles.disabledComment}>
+        //       <Input
+        //         name="disabled"
+        //         placeholder={MSG.commentPlaceholder}
+        //         elementOnly
+        //         disabled
+        //       />
+        //     </div>
+        //   ) : (
```
arrenv, line 231: Is this needed here? Comments should still be usable in locked state.

arrenv (approved):
Looking good, just one comment regarding unneeded elements.

willm30 (commented):
I notice there's an error in the DOM resulting from the `CommentInput` form element being nested within the form that spans the whole page.

form-in-form

And also notice the bottom border / horizontal rule in the spec isn't in this branch. I think it breaks up the space quite nicely.
_From the spec_:
hr

src/modules/pages/components/ExpenditurePage/ExpenditurePage.tsx
```diff
@@ -82,10 +103,43 @@ const ExpenditurePage = () => {
         <div className={styles.mainContainer}>
           <main className={styles.mainContent}>
             <TitleDescriptionSection isEditable />
+            {loading ? (
```
chinins, line 106: These 2 sections are almost identical. It's not good to repeat them. You can extract the repeating code to a variable or to a function (probably the latter, as it seems `isEditable` value will be different) and reuse it without repeating the code.
Also you need to change the html structure a but so as not to have `CommentInput` as a descendant of the `Form` component`, then you won't have the warning that Will mentioned.

chinins (changes requested):
Nice work but still needs changes (as per my comments). Most important is in the Expenditure page.

src/modules/pages/components/ExpenditurePage/ExpenditurePage.tsx
agnieszkajarosikloj: Thanks for comments. 
I removed `isEditable` variable and repetitions from the return statement.
To be honest, I don't know how can I change html structure to not have `CommentInput` as a descendant of the `Form` component. I need access to the form in all sections. Instead, I copied the `CommentInput` component and created a new one without the `form` tag in order not to nest them.

chinins: There is no need in `SubformCommentInput` component. We can use the `CommentInput`. What you need to do is move `CommentInput` from the `LogSection` file to the `ExpenditurePage` and move/organise HTML structure so that it's not a descendant of the form. The inputs of your Form can be separate from the `CommentInput` form, and we treat comments differently. So it should all work.
Please, let me know if you still have any questions - happy to clarify.

chinins (changes requested):
Very nice work but please, remove the new comment component, we don't need it. I've left a separate comment on how to extract `CommentInput` not to be a child of the Form in the ExpenditurePage.

Thank you!

src/modules/pages/components/ExpenditurePage/ExpenditurePage.tsx
```diff
@@ -58,15 +61,31 @@ const validationSchema = yup.object().shape({
...
+
   const sidebarRef = useRef<HTMLElement>(null);
   const submit = useCallback((values) => {
     // eslint-disable-next-line no-console
     console.log({ values });
   }, []);
 
   return (
-    <Form
+    <Formik
```
chinins, line 88: Why do you need `Formik` here around the whole block of code? It doesn't seem that it's necessary but maybe I'm missing something... You are using `Form` in `ExpenditureForm` separately and I would just use `Form` around or in `TitleDescription`.

chinins (commented):
It all looks much better. I just have a question/potential correction about the use of `Formik` in `ExpenditurePage` (please see my comment).

src/modules/pages/components/ExpenditurePage/ExpenditurePage.tsx
agnieszkajarosikloj: I use Formik in all sections because there is a button in the Stages component that triggers validation and sending the form.
```
const Stages = ({ states, activeStateId }: Props) => {
  const { values, handleSubmit, validateForm, resetForm } = useFormikContext<ValuesType>();

  const openDeleteDraftDialog = useDialog(DeleteDraftDialog);
  const openDraftConfirmDialog = useDialog(StakeExpenditureDialog);

  const handleSaveDraft = useCallback(async () => {
    const errors = await validateForm(values);
    const hasErrors = Object.keys(errors)?.length;

    return (
      !hasErrors &&
      openDraftConfirmDialog({
        onClick: () => {
          handleSubmit(values);
        },
      })
    );
  }, [handleSubmit, openDraftConfirmDialog, validateForm, values]);

  const handleDeleteDraft = () =>
    openDeleteDraftDialog({
      onClick: () => {
        resetForm?.();
    });
```

chinins: ok, thank you.

PR #3440, opened 2022-06-08
Remove default grouped transactions count from TransactionStatus

### LINKED ISSUE #3435: Wrong counter dot showing up when executing transacitons
bug-verified-recipients-tx-icon

That dot element shouldn't show up in individual transactions lines at all, it's a top level element that should just be visible on the group.

This is only tested on the #3190 branch, so most likely it's affected by something in it

dit7ya:
Seems like this was introduced in https://github.com/JoinColony/colonyDapp/commit/d45701d460f466f2ad5a0b48dbaafb0ee2d4ddc8 but I couldn't find the rationale behind it. So paging @ArmandoGraterol for especially reviewing this. Please let me know if I missed something. 

Fixes #3435.

### DIFF
--- a/src/modules/users/components/GasStation/TransactionCard/TransactionStatus.tsx
+++ b/src/modules/users/components/GasStation/TransactionCard/TransactionStatus.tsx
@@ -47,7 +47,7 @@ const displayName = 'users.GasStation.TransactionStatus';
 const TransactionStatus = ({
   hash,
   status,
-  groupCount = 4,
+  groupCount,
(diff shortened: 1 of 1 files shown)


ArmandoGraterol (approved):
My bad, this was a leftover from when I was testing the alignment of the icons in the dapp.

chinins (approved):
Short and sweet!

rdig (approved):
Seems to be working fine

PR #3453, opened 2022-06-09
Add error message when fewer than 3 characters

### LINKED ISSUE #3426: No error when entering less then 3 characters
## Steps to reproduce

1. Go to a action page and try to enter in a comment with 2 or less characters.

## Expected behaviour

* There should be a validation error that the user has entered less then the minimum characters.
* Show validation error "You must enter at least 3 characters to comment".
* Or, if there is no real reason for the minimum character length, remove the requirement.

## Actual behaviour

* There is no feedback when a user enters a comment with less then 3 characters.

38667aa1-4899-43af-9be5-346643fbd8b0

willm30:
## Description

When you try to add a comment in the Actions comments that contains fewer than 3 characters, the form will not submit. However, currently there is no feedback in the UI informing the user why. This PR fixes that by displaying a message to the user that their comment should contain at least 3 characters. 

**New stuff** ✨

* -[x] Display message: "You must enter at least 3 characters to comment" when input value contains more than 0 and fewer than 3 characters.

__Note: This is the message element I've added:__

```
<div
  className={
    !isValid && message.length > 0
      ? styles.minCharactersWarning
      : styles.minCharactersWarningFadeOut
  }
  >
  <FormattedMessage
    {...MSG.minCharactersWarning}
  />
</div>
```

- [x] Wrap both messages in a flex container so they format nicely
- [x] Set the "Return to Send" message position to absolute so it doesn't overlap

**Changes** 🏗

* None

**Deletions** ⚰️

* None

## Screenshot

**_The warning message is displayed after the user has entered a character and fades once the user has entered 3._**

message-demo

## TODO

None

Resolves #3426

### DIFF
--- a/src/modules/core/components/Comment/Input/CommentInput.css
+++ b/src/modules/core/components/Comment/Input/CommentInput.css
@@ -33,6 +33,8 @@
 }
 
 .sendInstructions {
+  position: absolute;
+  right: 0;
   margin-top: 5px;
   font-size: var(--size-tiny);
   font-weight: var(--weight-bold);
@@ -54,6 +56,26 @@
 .sendInstructionsFadeIn {
   composes: sendInstructions;
   opacity: 1;
+}
+
+.minCharactersWarningFadeOut {
+  margin-top: 5px;
+  font-size: var(--size-tiny);
+  font-weight: var(--weight-bold);
+  text-align: left;
+  color: var(--temp-grey-0);
+  opacity: 0;
+  white-space: nowrap;
+  transition: opacity 0.2s linear;
+}
+
+.minCharactersWarning {
+  composes: minCharactersWarningFadeOut;
+  opacity: 1;
+}
+
+.inputMessage {
+  display: flex;
 }
 
 .submitting {
--- a/src/modules/core/components/Comment/Input/CommentInput.css
+++ b/src/modules/core/components/Comment/Input/CommentInput.css
@@ -33,9 +33,9 @@
 }
 
 .sendInstructions {
+  margin-top: 5px;
   position: absolute;
   right: 0;
-  margin-top: 5px;
   font-size: var(--size-tiny);
   font-weight: var(--weight-bold);
   text-align: right;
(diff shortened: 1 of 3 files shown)


rdig (approved):
Works as expected. Nice work

rdig:
@willm30 what's with these commits ?

Screenshot from 2022-06-09 16-24-01

Which mostly seem to be dealing with changing the quotes styles. 

Please try to keep your commits readable and more consistent.

Either remove the un-needed commits in this case, or just merge them into one, but it either case, don't merge them like this

willm30:
> @willm30 what's with these commits ?
> 
> Screenshot from 2022-06-09 16-24-01
> 
> Which mostly seem to be dealing with changing the quotes styles.
> 
> Please try to keep your commits readable and more consistent.
> 
> Either remove the un-needed commits in this case, or just merge them into one, but it either case, don't merge them like this

@rdig Apologies, I didn't have the stylelint extension installed on my VS Code, so every time I commited I got a new style error in the CI test that I had to fix (and this happened 3/4 times because it exits at the first error it finds and therefore doesn't show you all the errors up front). Shouldn't happen again now I've installed the extn.

rdig:
> @rdig Apologies, I didn't have the stylelint extension installed on my VS Code, so every time I commited I got a new style error in the CI test that I had to fix (and this happened 3/4 times because it exits at the first error it finds and therefore doesn't show you all the errors up front). Shouldn't happen again now I've installed the extn.

Sure, don't worry. Let me know if you need help cleaning up your branch.

src/modules/core/components/Comment/Input/CommentInput.css
```diff
@@ -18,7 +18,7 @@
   box-shadow: shadow;
 }
 
-.commentBox textarea[aria-invalid="true"] {
+.commentBox textarea[aria-invalid='true'] {
```
chinins, line 21: You should correct back to double quotes as everywhere else in the dapp. I guess it was changed during the ordeal with your style linter.

chinins (approved):
Nice work! I like your RP description very much - very easy to review after that.

src/modules/core/components/Comment/Input/CommentInput.css
willm30: Have done, thanks. Have also updated my prettier config so won't happen again!

PR #3455, opened 2022-06-09
Fix error message overflow

### LINKED ISSUE #3430: Error message overflows container
## Steps to reproduce

This error was caused by something very similar to [this issue](https://github.com/JoinColony/colonyDapp/issues/3041). 

1. Install the Motions and Disputes extension in a new colony for the first time and enable it. It will enable, but will not update the UI to show it's been enabled 
2. Go to the extensions menu, and click on the Motions and Disputes extension, which will appear to not be enabled.
3. Click the "enable" button again.
4. Results in the following error, because it has already been enabled. The error message overflows its container.

error-msg

## Expected behaviour

* The error message should not overflow its container.

## Actual behaviour

* The message: "Estimation error: VM Exception while processing transaction revert voting-rep-already-initialised"  results in the row overflowing its container.

willm30:
## Description

Currently, long error messages in the actions popover are overflowing their container. This PR fixes this.

**New stuff** ✨

- [x] Add min height of error message at 40px

**Changes** 🏗

- [x] Set height of error message to auto

**Deletions** ⚰️

- [x] Remove negative margins on the error message container (because it was interfering with padding of ancestor li)
- [x] Remove comment explaining the use of negative margins (now redundant)

## Screenshot

**_The error message no longer overflows its container, and maintains correct padding between ancestor li element._**
**_Updated with line height of 1.6._**

line-height

## TODO

none

Resolves #3430

### DIFF
--- a/src/modules/users/components/GasStation/TransactionCard/GroupedTransactionCard.css
+++ b/src/modules/users/components/GasStation/TransactionCard/GroupedTransactionCard.css
@@ -83,5 +83,6 @@
   margin-left: 14px;
   font-size: var(--size-unreadable);
   font-weight: var(--weight-bold);
+  line-height: 1.6;
   color: var(--danger);
 }
--- a/src/modules/users/components/GasStation/TransactionCard/GroupedTransactionCard.css
+++ b/src/modules/users/components/GasStation/TransactionCard/GroupedTransactionCard.css
@@ -2,7 +2,8 @@
   display: flex;
   align-items: center;
   padding: 12px 15px 12px 20px;
-  height: 40px;
+  height: auto;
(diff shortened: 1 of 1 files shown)


arrenv (changes requested):
Nice fix, I think we should add a `line-height: 1.6;` here to reduce the height a little and make it look a little neater.

rdig (approved):
Good from a code perspective, but I agree with Arren that the space between lines needs to be smaller

willm30:
I've updated the line height on the error message to 1.6.

arrenv (approved):
Looks better now! Thanks.

PR #3458, opened 2022-06-09
Fix Numeral default size

### LINKED ISSUE #3419: Fix Numeral font sizing overrides
## Steps to reproduce

1. Check the font sizing of elements where Numeral is used.

## Expected behaviour

* The Numeral font size should not override font styling of elements where it is used.

## Actual behaviour

* Font styling is being overridden by the Numeral component

2f331da1-241e-4136-b2be-5089afe72c62

willm30:
## Description

This PR makes the Numeral main class inherit its font size instead of setting it to 14px by default and thereby overwriting the size of its container.

**New stuff** ✨

* none

**Changes** 🏗

- [x] Set font-size to "inherit" on Numeral.main 

**Deletions** ⚰️

* none

## Screenshot 

**_The numeral font is the same as its container, 13px_**

default-font

## TODO

* none

Resolves #3419

### DIFF
--- a/src/modules/core/components/Numeral/Numeral.css
+++ b/src/modules/core/components/Numeral/Numeral.css
@@ -1,5 +1,5 @@
 .main {
-  font-size: var(--size-normal);
+  font-size: inherit;
 }
 
 /*


arrenv (approved):
Nice simple fix, works as expected everywhere I checked.

rdig (approved):
LGTM

PR #3460, opened 2022-06-10
Address book tx redirects

### LINKED ISSUE #3436: Update Address Book Annotation not resetting after the transaction has been sent
bug-verified-recipients-modal-not-resetting

The annotation content remains there after the transaction is executed, as well as after you change the tab. In both cases, it needs to reset:
- for the first because we need a clean slate after the transaction went through
- for the second because it's a different transaction (from the user's perspective anyway), so it needs a new/different message

Also, after the transaction has been sent, it needs to redirect to it's corresponding action page

dit7ya:
This PR makes address book transactions redirect to the transactions details page. It also makes the annotations box reset on tab change in that action's modal.

Fixes #3436.

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialog.tsx
@@ -44,6 +44,7 @@ const displayName = 'dashboard.ManageWhitelistDialog';
 
 const ManageWhitelistDialog = ({
   cancel,
+  close,
   callStep,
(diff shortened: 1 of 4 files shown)


rdig (approved):
Good to go.

This branch does not contain the notification dot fix, right ?

As I'm still experiencing it:
Screenshot from 2022-06-10 13-25-03

dit7ya:
No it does not.

arrenv (approved):
Working as expected!

chinins (approved):
Good to go!

danbr (approved):
Very good. Working as expected. Well done.

PR #3464, opened 2022-06-10
Update the way manage whitelist dialog refreshes

### LINKED ISSUE #3454: `Manage address book` dialog not updating properly when opened from the members page
`Manage address book` dialog when opened from the members page is not updating the verified users list properly (you need to reload the page to have the changes). When you open it via new action button it works properly.

If verified recipients branch is not merged to the master branch, you need to branch off the former.

chinins:
## Description

This PR fixes the refresh of the manage whitelist dialog when opened from Members page

**Changes** 🏗

- add polling to `colonyFromNameQuery`

resolves #3454

### DIFF
--- a/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
+++ b/src/modules/dashboard/components/ColonyMembers/ColonyMembers.tsx
@@ -74,6 +74,7 @@ const ColonyMembers = () => {
 
   const { data: colonyData, error, loading } = useColonyFromNameQuery({
     variables: { name: colonyName, address: '' },
+    pollInterval: 5000,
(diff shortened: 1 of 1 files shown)


dit7ya (approved):
I could not successfully reproduce the issue but the change makes sense.

arrenv (approved):
Same for me, could not replicate it, but looks good.

danbr (approved):
Very good. Its working well. Problem resolved. (I could reproduce the issue)

I'm not sure why there is a problem when accessing the dialog one-way vs another? Polling seems to resolve the issue, but is probably not the most efficient solution.

PR #3469, opened 2022-06-10
Address Book filter on UserPicker is disabled if there are no addresses

### LINKED ISSUE #3466: UserPicker shows no users and you cannot disable Address Book
## Steps to reproduce

1. Add a user to the Address book, it will auto activate it.
2. Remove the user from the Address book.
3. Go to a UserPicker and try to select a user, they won't be there.
4. You will not be able to deactivate the Address book.

## Expected behaviour

* UserPicker should not filter anyone if no addresses are in the Address book.

## Actual behaviour

* UserPicker filters out all addresses when the Address book was previously activated but all addresses were removed.

list-filtered

41abe021-8775-4fc3-beb6-1c283db27fd8

arrenv:
## Description

This adds a check to the `whitelistedAddresses` user filter, so, that the filter does not work if there are no addresses in the Address Book.

**New stuff** ✨

* Adds check for 0 length of `whitelistedAddresses`

## Testing

1. Add a user to the Address book, it will auto activate it.
2. Remove the user from the Address book.
3. Go to a UserPicker and try to select a user, it should allow you to select any user again.

Resolves #3466

### DIFF
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
@@ -117,7 +117,7 @@ const CreatePaymentDialog = ({
   const filteredVerifiedRecipients = useMemo(() => {
     return isWhitelistActivated &&
       colonyData?.processedColony?.whitelistedAddresses &&
-      colonyData?.processedColony?.whitelistedAddresses?.length > 0
+      colonyData.processedColony.whitelistedAddresses.length > 0
       ? (colonyMembers?.subscribedUsers || []).filter((member) =>
           colonyData?.processedColony?.whitelistedAddresses.some(
             (el) => el.toLowerCase() === member.id.toLowerCase(),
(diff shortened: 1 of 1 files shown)


chinins (approved):
All good, makes sense!

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
```diff
@@ -115,7 +115,9 @@ const CreatePaymentDialog = ({
     colonyData?.processedColony?.isWhitelistActivated;
 
   const filteredVerifiedRecipients = useMemo(() => {
-    return isWhitelistActivated
+    return isWhitelistActivated &&
+      colonyData?.processedColony?.whitelistedAddresses &&
+      colonyData?.processedColony?.whitelistedAddresses?.length > 0
```
rdig, line 120: Just a nitpick, but if `colonyData?.processedColony?.whitelistedAddresses` evaluates, then you no longer need the check on the length, since you know it exists already.

So that would make it something like:
```suggestion
    return isWhitelistActivated &&
      colonyData?.processedColony?.whitelistedAddresses &&
      colonyData.processedColony.whitelistedAddresses.length > 0
```

rdig (approved):
Good to go!

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
arrenv: Good point, removing the redundancy.

PR #3470, opened 2022-06-10
Fix/3461-magnify-icon

### LINKED ISSUE #3461: Search icon size is incorrect
## Steps to reproduce

1. Go to Members page on Watchers and Contributors

## Expected behaviour

* Icon should match design - https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=6736%3A93022

## Actual behaviour

* Size of icon is incorrect

download (13)

willm30:
## Description

This PR fixes the oversized Search Icon in the Watchers and Contributors feature.

**Changes** 🏗

- [x] Set icon height and width to *20px* as per [figma](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=6736%3A90099)

## Screenshot

**_Search icon is the correct height_**

search-icon

search-icon-20

Resolves #3461

### DIFF
--- a/src/modules/dashboard/components/Members/MembersTitle.css
+++ b/src/modules/dashboard/components/Members/MembersTitle.css
@@ -49,10 +49,10 @@
 }
 
 .icon {
-  height: 30px;
-  width: 30px;
+  height: 20px;
+  width: 20px;
   position: absolute;
-  top: 50%;
+  top: 40%;
   right: 2px;
   z-index: 1;
   transform: translateY(-50%);


src/modules/dashboard/components/Members/MembersTitle.css
```diff
@@ -49,10 +49,10 @@
 }
 
 .icon {
-  height: 30px;
-  width: 30px;
+  height: 13.33px;
```
chinins, line 52: The icon should actually be 20 px as per figma file. The 13.3 px is only if you select the circle only.
Generally, we try to never use decimals for pixels as well.
Screenshot 2022-06-10 at 18 27 06

chinins (changes requested):
The height of the icon is not correct as per my comment.

willm30:
> The height of the icon is not correct as per my comment.

Thanks for pointing out, have now fixed.

rdig (approved):
Seems fine now

dit7ya (approved):
LGTM.

chinins (approved):
All good now!

PR #3472, opened 2022-06-13
Fix: include @ symbol in user search

### LINKED ISSUE #3462: Account for "@" in User searches
This has been reported by a user as a confusing aspect of our search.

## Steps to reproduce

1. Use the UserPicker to search for a user starting with `@`

## Expected behaviour

* Including `@` at the start of the search should still work, as this is how all users are displayed in the DApp and is common on other platforms as well.
* This should also be addressed in the Members page search.

## Actual behaviour

* When starting a search with `@`, it does not match any users.

at-in-search

at-in-search-members

willm30:
## Description

This PR includes the @ symbol in user searches, so you can search for a user like "@ will" as well as simply "will". 

**Changes** 🏗

- [x] Add '@' symbol to user selection filter

## Screenshot

**_Previously, starting a search with the @ symbol would not return the selected user._**

at-symbol-user-search

Resolves #3462


src/modules/core/components/SingleUserPicker/utils.ts
```diff
@@ -7,20 +7,20 @@ export const filterUserSelection = (data, filterValue) => {
     return data;
   }
 
-  const filtered = data.filter(
-    (user) =>
+  const filtered = data.filter((user) => {
+    const { username } = user.profile;
```
rdig, line 11: The user object might not exist ?

```suggestion
    const { username } = user?.profile;
```

src/modules/core/components/SingleUserPicker/utils.ts
```diff
@@ -7,20 +7,20 @@ export const filterUserSelection = (data, filterValue) => {
...
+    return (
       user &&
       user.profile &&
```
rdig, line 14: Aren't you already checking for it's existence above? 

You should just check for `username` in this instance, which should be enough.

rdig (changes requested):
The `@` filtering logic is fine, however, the `user` object existence checks are not

willm30:
> The `@` filtering logic is fine, however, the `user` object existence checks are not

Should be better now.

src/modules/core/components/SingleUserPicker/utils.ts
```diff
@@ -7,20 +7,17 @@ export const filterUserSelection = (data, filterValue) => {
...
-  );
+  const filtered = data.filter((user) => {
+    const { username, walletAddress } = user?.profile;
+
+    return (
+      username &&
+      walletAddress &&
+      (`@${username}`.toLowerCase().includes(filterValue.toLowerCase()) ||
+        walletAddress.toLowerCase().includes(filterValue.toLowerCase()) ||
+        user.id.toLowerCase().includes(filterValue.toLowerCase()))
```
rdig, line 18: I'm going to nitpick some more :P

Why call `filterValue.toLowerCase()` three times ? It won't change, so why not assign it to a variable, and use that for checks ?

```suggestion
      (`@${username}`.toLowerCase().includes(valueToLoweCase) ||
        walletAddress.toLowerCase().includes(valueToLoweCase) ||
        user.id.toLowerCase().includes(valueToLoweCase))
```

rdig, line 18: `user.id` is the same as `walletAddress`, so there's no point in filtering by it two times. Just remove one of them

rdig (approved):
Good with it now, however I did find more stuff to nitpick about :)

rdig (approved):
Very nice! Much cleaner now.

Sorry for all the back and fourth :)

will:
> a user like "@will"

:(

rdig:
> > a user like "@will"
> 
> :(

Haha. Be careful who you ping @willm30 :stuck_out_tongue:

chinins (approved):
Very nice and clean1

Issue #3441, opened 2022-06-08
Expenditures: As a user I cannot edit values when I locked it


arrenv:
[Provided by Michał via Contractors project management platform]

## Description
This feature adds the locked state to expenditure inputs.

## Acceptance Criteria

- user cannot change values after locking the expenditure

## Design
[Design Link](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2420%3A1087531)

![979bb5069a264467ab2de51a8a4afa71](https://user-images.githubusercontent.com/33682027/172660238-bbfcb9a2-b0f2-43d7-bb3d-7035fedeb35e.png)

PR #3473, opened 2022-06-13
Add locked values

agnieszkajarosikloj:
## Description

It is not possible to edit values of an Expenditure, while it is in a locked state.

Screenshot 2022-06-30 at 07 39 20

[Figma design:](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2420%3A1087531)

## TESTING

1. Ensure you have values in every field.
2. Click the 'Submit draft' button.
3. In the modal click the 'Stake' button.
4. Click the 'Lock values' button.
5. You should then be in the locked state.

## TODO

- add button, which revert the expenditure to an editable state.

Resolves to #3441


src/modules/dashboard/components/ExpenditurePage/Delay/Delay.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  return (
+    <>
+      {amount}
+      {getTimeSymbol()}
+    </>
+  );
+};
+
+export default Delay;
```
arrenv, line 39: After seeing how this needed to be implemented, it is not really necessary to have the shortened version of the time period. We can just have the full time period label for each.

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureSettings.tsx
```diff
@@ -87,6 +92,9 @@ const ExpenditureSettings = () => {
     (option, label) => {
       const value = option?.value;
       const color = getDomainColor(value);
+      // eslint-disable-next-line no-console
+      console.log({ option, label });
+
```
arrenv, line 97: This can be removed.

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/LockedExpenditureSettings.css
```diff
@@ -0,0 +1,120 @@
...
+
+.settingsRow {
+  composes: flexContainerRow flexAlignCenter flexJustifyBetween from '~styles/layout.css';
+  min-height: 41px;
+}
+
+.settingsRow button[aria-expanded="true"] {
+  padding: 2px 0 2px 6px;
+  border-radius: 4px;
+  background-color: rgb(229, 230, 247);
```
arrenv, line 21: Can we please replace all instances of this color with this variable `--grey-blue-1`

src/modules/dashboard/components/ExpenditurePage/Recipient/LockedRecipient.css
```diff
@@ -0,0 +1,150 @@
+.settingsRow {
+  composes: flexContainerRow flexAlignCenter flexJustifyBetween from '~styles/layout.css';
+  min-height: 41px;
+}
+
+.settingsRow button[aria-expanded="true"] {
+  padding: 2px 0 2px 6px;
+  border-radius: 4px;
+  background-color: rgb(229, 230, 247);
```
arrenv, line 9: Can we please replace all instances of this color with this variable `--grey-blue-1`

arrenv (changes requested):
@agnieszkajarosikloj 

1. There are some issues with the validation, for example, the Team, Token and period select options invalidate incorrectly if you don't change them, even if they have their default selections. 

2. The Team field seems to default to the last in the list, when it should be the first in the list.

694701a0-f4d5-4170-ad85-d0960b0ee4e4

3. The 'Claim delay' field is optional and does not need a required validation.

4. The recipient validation can just change the blue icon to red, instead of having the text.

5. There should be validation on the Value field, which should just sit under the input.

6. I think that the token values style and claim delay should be updated to match the style of the balance values, i.e. the same font-weight, size and line-heights.

08b67acb-9cc6-42d0-9fb6-5a25be5db0d5

7. This hover state and title tag are not needed.

Screenshot 2022-07-01 144158

arrenv (changes requested):
@agnieszkajarosikloj, thank you.

There looks to be some alignment issues, as well as a error with the unique key.

7aee2356-ff72-4f88-90ab-bbdfcfc79a2e

In the locked state, in the title, the claim delay would be better with a space between the number and the period.

3843aa65-559f-4750-b3bd-bbd02dc1ccab

arrenv (approved):
Thank you @agnieszkajarosikloj. Please see some comments below.

Can we please have these paddings in the minimised locked state match the minimised state in the non-locked state.

### Locked

download (46)

### Unlocked

download (47)

willm30:
@agnieszkajarosikloj Not suggesting this is a problem with your implementation, but I found the validation in this form too subtle to be useful. I had validation errors after clicking `Submit draft` and had to put `console.logs` in the code to figure out what they were. The pink highlighting of the relevant fields didn't sufficiently draw my attention, and in the case of the `Expenditure Title` field, it tells you _that_ something is wrong but not _what_ (i.e. it requires a title of at least three characters). 

I personally would prefer specific messages under the input fields, as I've become acustomed to in the rest of the app. (e.g. "Enter a title of at least three characters" or similar). 

That aside, I can confirm that form values can't be edited once the values have been locked. 

Also, not sure if it's in scope, but deleting the expenditure doesn't update the UI, which is also confusing because I clicked the button a few times and nothing happened. Refreshing does update the UI, but it would be better to not have to manually refresh the window.

arrenv:
> @agnieszkajarosikloj Not suggesting this is a problem with your implementation, but I found the validation in this form too subtle to be useful. I had validation errors after clicking `Submit draft` and had to put `console.logs` in the code to figure out what they were. The pink highlighting of the relevant fields didn't sufficiently draw my attention, and in the case of the `Expenditure Title` field, it tells you _that_ something is wrong but not _what_ (i.e. it requires a title of at least three characters).
> I personally would prefer specific messages under the input fields, as I've become acustomed to in the rest of the app. (e.g. "Enter a title of at least three characters" or similar).

Good feedback regarding validation, this is a product issue, that we will need to address separately.

> That aside, I can confirm that form values can't be edited once the values have been locked.
> 
> Also, not sure if it's in scope, but deleting the expenditure doesn't update the UI, which is also confusing because I clicked the button a few times and nothing happened. Refreshing does update the UI, but it would be better to not have to manually refresh the window.

This is out of scope of this feature at the moment. This is all UI only.

willm30 (approved):

PR #3498, opened 2022-06-15
fix select dropdown

### LINKED ISSUE #3488: Expenditures: Dropdown is being cut off within the left aside
## Steps to reproduce

1. In the left aside, use any of the select options at a smaller screen width
2. The options will be cut off within the left aside due to overflow hidden being used.

## Expected behaviour

* The dropdown should not be cut off and should display correctly in full.

## Actual behaviour

* The dropdown options are being cut off in the left aside

cb511ec1ddc84e92b78f21faabc9f05a

cb55a58ba4ce00075ff68bc48d3b67f7

agnieszkajarosikloj:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs

Resolves #3488


arrenv (changes requested):
Thanks @agnieszkajarosikloj I know in other cases it has been requested not to modify core components, but the code duplication seems quite high. And the core components could probably benefit from these updates. Thoughts @JoinColony/frontend?

Regarding this, the height of the select item is too small. It should be 40px, it is currently 32px.

a2a6e92f-e3c7-4713-a54e-d3bc9d2e5824

Screenshot 2022-06-27 140148

src/modules/core/components/DomainDropdown/DomainDropdownWithPortal.tsx
```diff
@@ -0,0 +1,139 @@
...
+    colony,
+    currentDomainId,
+    onDomainEdit,
+    showDescription,
+    showAllDomains,
+    filterOptionsFn,
+  ]);
+
+  return (
+    <SelectWithPortalDropdown
```
chinins, line 112: With my suggestions it will be `Select` component with `withDropdownElement` prop set to true. You can just add `withDropdownElement` prop to the `DomainDropdown` core component - similar to my `Select` suggestion.

chinins (changes requested):
I know that we asked to not modify core components but I agree with Arren here - there is too much repetitive code for what it's worth and in this case it's possible to change the core components in a limited way that won't affect the other components.

Please, remove `SelectWithPortalDropdown` and `DomainDropdownWith ` components & add an optional boolean prop with `false` as default value to `Select` & `DomainDropdown`, and modify the code accordingly.

Thank you! Let me know if you have any questions.

arrenv (approved):
Working as expected, thanks @agnieszkajarosikloj

chinins (approved):
All good - very nice work!

PR #3514, opened 2022-06-17
Update `contributorsAndWatchers` query resolver to fix members with reputation list

### LINKED ISSUE #3507: Contributors are being shown as Watchers
## Steps to reproduce

1. Create a colony.
2. Generate reputation for the root user.
3. Join the colony with a fresh user.
4. Generate reputation for the second user.
5. Take a look at the contributors vs watchers section in the aside or the Members page.

## Expected behaviour

* The second, newer member with reputation, should appear along with the first user on the contributor's side.

## Actual behaviour

* The second user appears on the watcher's side even though it has reputation, and this is the case for any other user that doesn't have domain roles but has reputation.

FireShot Capture 627 - Members - Colony - wonker - localhost

## Suggestions

* Tried to find the source of this bug, managed to track it down to the `colonyMembersWithReputation` resolver in the `colony` file. The resolver is _always_ returning an empty array as if there were no members with reputation, including the root user.

chinins:
## Description

This PR fixes the resolver of the `contributorsAndWatchers` query.

I have noticed that `colonyMembersWithReputation` resolver takes `COLONY_TOTAL_BALANCE_DOMAIN_ID` as a default domain. In this case it will always return an empty list. I put root domain id as a default because I assume we want to have all of the members with reputation if no domain id is passed and having always no addresses returned in this case doesn't make sense.

**Changes** 🏗

- update `contributorsAndWatchers` resolver
- update `colonyMembersWithReputation` resolver

resolves #3507

### DIFF
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -330,7 +330,7 @@ export const colonyResolvers = ({
       _,
       {
         colonyAddress,
-        domainId = COLONY_TOTAL_BALANCE_DOMAIN_ID,
+        domainId = ROOT_DOMAIN_ID,
       }: { colonyAddress: Address; domainId: number },
     ) {
       const colonyClient = await colonyManager.getClient(
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -362,13 +362,16 @@ export const colonyResolvers = ({
       });
       const verifiedAddresses = colony?.processedColony.whitelistedAddresses;
 
+      const domainIdchecked =
+        domainId === COLONY_TOTAL_BALANCE_DOMAIN_ID ? ROOT_DOMAIN_ID : domainId;
+
       const domainRoles = getAllUserRolesForDomain(
         colony?.processedColony,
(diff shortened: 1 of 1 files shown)


src/data/resolvers/colony.ts
```diff
@@ -439,17 +442,6 @@ export const colonyResolvers = ({
...
-      //   VerifiedUsersQuery,
-      //   VerifiedUsersQueryVariables
-      // >({
-      //   query: VerifiedUsersDo,
-      //   variables: {
-      //     colonyAddress,
-      //   },
-      //   fetchPolicy: 'network-only',
-      // });
-
```
rdig, line 452: :+1:

rdig (approved):
Good fix!

Does this also fix @ArmandoGraterol 's #3507 by chance ?

chinins:
> Good fix!
> 
> Does this also fix @ArmandoGraterol 's #3507 by chance ?

It is a fix for #3507, I talked to Armando yesterday and suggested to fix it today in a separate PR

dit7ya (approved):
LGTM. :)

Issue #3413, opened 2022-05-27
Restructure and Extract IPFS Metadata Parser functionality


rdig:
The aim of this feature is two fold:
- provide better structure, versioning, typing and robustness to the way we store and read event metadata from IPFS stored JSON objects _(used in metadata events: `ColonyMetadata`, `DomainMetadata`, etc...)_
- extract the above as an external lib so that it can be shared between all our projects _(sdk, dapp, cdapp)_

_NOTE: while _Verified Recipients_ affects this somewhat, this is imagined as to be merged sometime after it, meaning it will have to deal with, and reformat the the data structure it added_

### Current structure

The current data structure is a bit of an afterthought as it was rushed during development.

Currently we have the following structure for handling colony related metadata:
```js
const colonyMetadata = JSON.parse({
  "colonyDisplayName": "...", // just a normal string
  "colonyAvatarHash": "...", // a string, which is (optionally) another IPFS hash, which stores the actual image
  "colonyTokens": ["..."], // an array of token addresses as strings
});
```

Domain related metadata has this structure:
```js
const domainMetadata = JSON.parse({
  "domainName": "...", // just a normal string
  "domainColor": "...", // actually a number from 1 to 16 each representing a color inside an enum
  "domainPurpose": "...", // a normal string
});
```

Annotation related metadata has this structure:
```js
const annotationMetadata = JSON.parse({
  "annotationMessage": "...", // just a normal string
});
```

### Versioning

One of the first improvements will be introducing a prop that will track the current version / iteration of the metadata object.

Something similar to:
```js
const metadata = JSON.parse({
  "version": 2,
  "data": { ... },
});
```

This will track the current data format version, and as soon as something changes: new props, removed props, types, etc, the version number will increment.

The new metadata objects will start counting at `2`, and everything else, that doesn't contain the `version` prop, will be treated as the "old", version "1"

### ID-ing 

Similarly to versioning, we want to introduce ID-ing of metadata objects. This will aid in the event the IPFS hash is not known beforehand, and to have a easier time parsing and decoding it's data.

So we're going to set up a similar `enum` type for the IDs: _(the below example is not the final one, but it's recommended you follow it's structure)_
```typescript
enum MetadataType {
  Colony = 'colony',
  Domain = 'domain',
  Annotation = 'annotation',
};
```

And then use said ID when creating the metadata object:
```typescript
const newMetadata = JSON.stringify({
  version: 2,
  name: MetadataType.Colony,
[...]

PR #3535, opened 2022-06-27
Restructure Colony event metadata in IPFS

danbr:
## Description

This PR restructures Colony event metadata that is stored in IPFS. 

The restructuring happens in ColonyEventMetadataParser (lib) which defines the new format, validation, types and provides routines to (de)serialise.
The Dapp has been updated to use ColonyEventMetadataParser's routines when get/set'ing IPFS.

This is the first unveiling of the enhancements and the interfaces are not fixed, so if any functionality is missing or something can be improved please feel free to make suggestions.

**Version**
The Dapp should support the legacy Metadata format (V1). Therefore, the format version should be checked when parsing, and the latest format should be used when deserialising Metadata objects (to write to IPFS).
There is a helper function in ColonyEventMetadataParser:
`getEventMetadataVersion(ipfsMetadata);`

**Validation**
ColonyEventMetadataConverter uses Yup to define validation through schemas and is used when parsing IPFS data.

**Format**
Below is an overview of the structure, more information can be found in (ColonyEventMetadataParser) types.ts

```
export enum MetadataType {
	Colony = 'colony',
	Domain = 'domain',
	Annotation = 'annotation',
	Misc = 'misc',
}
```

```
interface Metadata {
  version: number;
  name: MetadataType;
  data: ColonyMetadata | DomainMetadata | AnnotationMetadata;
}
```

**Parsing** To retrieve MetaData from IPFS. (String to Object)
```
// Colony MetaData 
const colonyMetadata = getColonyMetadataFromResponse(ipfsDataJSON); 
// Domain MetaData 
const domainMetadata = getDomainMetadataFromResponse(ipfsDataJSON); 
// Annotation Message 
const annotation = getAnnotationMsgFromResponse(ipfsDataJSON);
```

**Deserialising**  Preparing Metadata to send to IPFS. (Object to String)

```
// Colony Metadata packed and converted to string 
const metadataForIPFS = getStringForMetadataColony({ 	
	colonyDisplayName, colonyAvatarHash: 	
	colonyAvatarIpfsHash colonyTokens, 
	verifiedAddresses, isWhitelistActivated, }); 
colonyMetadataIpfsHash = yield call(ipfsUpload, metadataForIPFS); 

// Domain Metadata packed and converted to string 
domainMetadataIpfsHash = yield call( ipfsUpload, 
	getStringForMetadataDomain({ domainName, domainColor, domainPurpose, }), 

// Annotation message packed and converted to string 
ipfsHash = yield call( ipfsUpload, 
	getStringForMetadataAnnotation({ annotationMsg: annotationMessage }), );
```

**Metadata Types**

The new format currently supports 4 Metadata Types 

- 	Colony = 'colony',
- 	Domain = 'domain',
- 	Annotation = 'annotation',
- 	Misc = 'misc',

The first 3 are self explanatory. The final type `Miscellaneous` is used for data such as *avatar image*, *agreement*, and is ready to support *thumbnail avatar image*.

Contributes to #3413


src/data/resolvers/colony.ts
```diff
@@ -220,15 +248,29 @@ export const getProcessedDomain = async (
...
+         */
+        const {
+          domainName: metadataDomainName = null,
+          domainColor = null,
+          domainPurpose = null,
+        } = JSON.parse(ipfsMetadata);
+
+        name = metadataDomainName;
+        color = domainColor;
+        description = domainPurpose;
```
willm30, line 264: Aside from the `JSON.parse(ipfsMetadata)`, couldn't these lines could be written as the lines below? E.g. 

```
const domainMetadata = JSON.parse(ipfsMetadata);
name = domainMetadata?.domainName || null; 

etc...
```

src/modules/dashboard/components/ActionsPageFeed/ActionsPageFeedItemWithIPFS/ActionsPageFeedItemWithIPFS.tsx
```diff
@@ -47,9 +52,21 @@ const ActionsPageFeedItemWithIPFS = ({
...
-    if (annotationObject && annotationObject.annotationMessage) {
-      return annotationObject.annotationMessage;
+
+    const metadataVersion = getEventMetadataVersion(ipfsDataJSON);
+    if (metadataVersion === 1) {
+      /*
+       * original metadata format
+       */
+      const annotationObject = JSON.parse(ipfsDataJSON);
+      if (annotationObject && annotationObject.annotationMessage) {
```
willm30, line 62: I think `annotationObject` here is unnecessary because if `annotationObject.annotationMessage` is defined then `annotationObject` is by definition defined.

rdig: Agree

rdig (approved):
I'm mostly good with this, however I do have some questions about the library helpers / functions implementations _(stuff like `getStringForMetadataAnnotation` vs `getStringForColony` which in my mind should be one and the same)_

I will however look at the library separately and I'll comment stuff related to it there.

Also, I think we need to set up a QA session _(you, me @area)_ to test this properly before releasing it into the wild.

src/data/resolvers/colony.ts
```diff
@@ -119,38 +125,60 @@ export const getProcessedColony = async (
...
+         * new metadata format
+         */
+        const colonyMetadata = getColonyMetadataFromResponse(ipfsMetadata);
+        displayName = colonyMetadata?.colonyDisplayName || '';
+        avatarHash = colonyMetadata?.colonyAvatarHash || '';
+        tokenAddresses = colonyMetadata?.colonyTokens || [];
+        whitelistedAddresses = colonyMetadata?.verifiedAddresses || [];
+        if (colonyMetadata?.isWhitelistActivated) {
+          whitelistActivated = colonyMetadata.isWhitelistActivated;
+        }
```
rdig, line 159: The dapp needs to account for cases where the IPFS data is set to something we don't expect _(number, empty string, empty object, anything else really)_ since that can be set directly, and not through the dapp.

So in this case `getEventMetadataVersion` will return `1` since it can't parse, but our `JSON.parse` I think will fail for something that's not a string. 

I can't remember exactly how the return value looks like, so this might actually not even be a problem, but please investing nonetheless

rdig (commented):
Left some more notes after reading the library's code

ArmandoGraterol (approved):
Looks good to me :+1: Tested several actions and it's working properly

src/data/resolvers/colony.ts
danbr: So I tried not to touch the existing logic that handles Metadata.

When trying to ascertain the MetadataVersion, just as what currently happens, with V1 we make a best effort by assuming thats its V1 (not version tagged), and then assuming the data is consistent & coherent.

With V2 things are improved by using schema validation (Yup) which protects against (most of) the unexpected. So the data should conform and be consistent & coherent.

If i follow your example:
 1. Some 'gobbledegook' is found at the IPFS hash
 2. Try to get the version number, which fails, so V1 is assumed (we have no other choice)
 3. V1 parsing logic is performed, which is what currently happens via JSON.parse in the dapp
 4. Data for property may or may not be found/parsed, so either property has value or null

If we are close to moving to V2 i don't think it is worth the effort to implement schema validation on V1 data or adding logic to ColonyEventMetadataParser to protect V1 dapp logic.

danbr:
> I'm mostly good with this, however I do have some questions about the library helpers / functions implementations _(stuff like `getStringForMetadataAnnotation` vs `getStringForColony` which in my mind should be one and the same)_

@rdig Let me explain my rational.....
There is a slight difference in the interfaces, mainly for convenience. 

AnnotationMetadata & MiscMetadata payloads are singular, containing only 1 field. 
So i didnt think consumers of the data would want the containing Annotation/MiscMetadata object but rather only interested in the data field itself (Annotation or Named-value). Thus:
`const agreement = getNameValueFromMisc('agreement', response)`
or
`const annotation = getAnnotationMsgFromResponse(ipfsDataJSON)`

For the other Metadata objects that have multiple data elements, it makes sense to return the appropriate object.
`const colonyMetadata = getColonyMetadataFromResponse(ipfsDataJSON)`
or
`const domainMetadata = getDomainMetadataFromResponse(ipfsDataJSON)`
then  the `Metadata` object can be passed around and used as desired.

src/data/resolvers/colony.ts
```diff
@@ -47,6 +54,7 @@ import { Address } from '~types/index';
 import { COLONY_TOTAL_BALANCE_DOMAIN_ID } from '~constants';
 
 import { getToken } from './token';
+
```
chinins, line 57: No need for this extra empty line

src/data/resolvers/colony.ts
```diff
@@ -197,7 +225,7 @@ export const getProcessedDomain = async (
     name: domainName,
   } = subgraphDomain;
   let name: string | null = null;
-  let color: string | null = null;
+  let color: number | null = null;
```
chinins, line 228: I think that it should still be string here because it will ben an enum in the library as per Raul's comment
https://github.com/JoinColony/ColonyEventMetadataParser/pull/1#discussion_r920357223

src/modules/dashboard/sagas/actions/editColony.ts
```diff
@@ -16,6 +20,7 @@ import {
   getTxChannel,
 } from '../../../core/sagas';
 import { ipfsUpload } from '../../../core/sagas/ipfs';
+
```
chinins, line 23: no need for this extra empty line

chinins (approved):
I've looked at the library as well - great job!

src/utils/events/index.ts
```diff
@@ -242,11 +247,21 @@ const getPaymentActionValues = async (
   return paymentActionValues;
 };
 
-const getDomainValuesFromIPFS = async (ipfsHash: string) => {
+interface DomainValues {
+  domainName: string | null;
+  domainColor: number | null;
```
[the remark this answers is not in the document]
danbr: domainColor is currently dealt with as a number in the dapp, as Color is a numeric enum.

PR #3537, opened 2022-06-27
Add: Gas Station Popover (mobile)

willm30:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/modules/users/components/GasStation/GasStationContent/GasStationContent.css
```diff
@@ -16,3 +18,16 @@
...
+
+@media screen and query700 {
+  .main {
+    width: calc(100vw - 28px);
+    border-radius: 5px;
+    background-color: var(--grey-blue-2);
+    box-shadow: 0px 2px 11px var(--drop-shadow);
+  }
+
+  .transactionsContainer div[class*="GroupedTransaction_description"] {
```
rdig, line 30: I'm really not comfortable using this class name as the selector, since this is name is from post-processing so either a webpack update or postcss loader might change the format, at which point this will break.

We need to find an alternative for this.

How about just selecting the `nth` class this `div` renders as and just go from there ?

src/modules/users/components/GasStation/GasStationHeader/GasStationHeader.css
```diff
@@ -33,3 +35,55 @@
...
+  }
+
+  .walletAddress span[class*="MaskedAddress"] {
+    font-size: 12px;
+    font-weight: 500;
+  }
+
+  .walletAddress span[class*="copyButton"] {
+    display: none;
+  }
```
rdig, line 77: Again, as above, we need an alternative as these class names might change

rdig (approved):
This is mostly good, but please find a better alternative for the class name selectors, as they won't really work in the long run.

Also, I've noticed that with wider mobile views, the gas station sticks to the right of the screen.

Normal:
Screenshot from 2022-06-29 15-10-31

Right of screen:
Screenshot from 2022-06-29 15-10-36

Anyway, I'm pre-approving, but please address my comments.

src/modules/users/components/GasStation/GasStationHeader/GasStationHeader.tsx
```diff
@@ -40,7 +48,16 @@ const GasStationHeader = ({ close }: Props) => {
           <CopyableAddress>{walletAddress}</CopyableAddress>
         </div>
         <div>
-          <Numeral value={balance} suffix={DEFAULT_NETWORK_TOKEN.symbol} />
+          {isNetworkAllowed && isMobile ? (
```
arrenv, line 51: This is not really needed here. I know the spec has it, annoyingly that is just the version of the `GasStationPopover` design that I dragged in. So, just keep it simply the same as the desktop view.

src/modules/users/components/GasStation/GasStationHeader/GasStationHeader.tsx
```diff
@@ -17,16 +18,23 @@ const MSG = defineMessages({
...
 
 interface Props {
   close?: () => void;
+  isMobile: boolean;
 }
 
 const displayName = 'users.GasStation.GasStationHeader';
+const GasStationHeader = ({ close, isMobile }: Props) => {
+  const { balance, walletAddress, networkId } = useLoggedInUser();
+  const isNetworkAllowed = checkIfNetworkIsAllowed(networkId);
```
arrenv, line 35: So, this is not needed either.

arrenv (changes requested):
Thanks @willm30, my fault for dragging in the only `GasStationPopver` design I could find. However, there are design changes in this that are not required specifically for mobile. We can just use what is already in the DApp. i.e. like this:

1ffc2e9b-5d27-4549-bed7-081589db52a9

src/modules/users/components/GasStation/GasStationHeader/GasStationHeader.css
```diff
@@ -33,3 +35,51 @@
...
+
+  .walletAddress span:first-child {
+    font-size: 12px;
+    font-weight: 500;
+  }
+
+  .walletAddress span:nth-child(2) {
+    display: none;
+  }
```
willm30, line 73: @rdig Is this what we're going for with the class selectors? If so I will use same pattern with other PRs and in general going forward.

rdig: This is not ideal either, but definitely better than using hardcoded class names

willm30:
@rdig Fixed the popover sticking to the side at wider screen sizes, and updated class selectors. 

gas-stn-popover-wide

src/modules/users/components/GasStation/GasStationHeader/GasStationHeader.tsx
willm30: @arrenv Have removed and kept same as desktop.

src/modules/users/components/GasStation/GasStationHeader/GasStationHeader.css
```diff
@@ -33,3 +35,55 @@
...
+  } */
+
+  /* .walletAddress span:nth-child(2) {
+    display: none;
+  } */
+
+  /* .walletHeading h4 {
+    color: black;
+    letter-spacing: 0.25px;
+  } */
```
ArmandoGraterol, line 82: Are those styles going to be used later? Then I would recommend leaving a note explaining that. Otherwise, you might as well delete them

ArmandoGraterol (approved):
Looks great to me :ok_hand: Just left a small comment :)

willm30:
> Looks great to me 👌 Just left a small comment :)

🤦‍♂️ Left those in there by mistake

chinins (approved):
Very nice!

arrenv (changes requested):
I am getting issues with the loader animation being squashed, as of the gif. It seems to be in your screenshot as well.

loader-issue

willm30:
> I am getting issues with the loader animation being squashed, as of the gif. It seems to be in your screenshot as well.

Ah, thought that was because I had frozen the dom using dev tools. Fixed now.

arrenv (approved):
Looks good now, nice one!

PR #3542, opened 2022-06-28
Add "Transfer Funds" interaction variation

### LINKED ISSUE #3459: Add `Control Safe` Dialog's `Transfer Funds` type variation UI
### Story

As a user transferring funds from the safe to a user, I want to see the `Control Safe` variation of the `Transfer Funds` transaction type.

### Description

When the user selects the `Transfer Funds` type using the selector in the dialog's "basic UI" in the `Control Safe` Dialog, the inputs related to the transaction type should appear.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=139%3A17971)

FireShot Capture 619 - Safe Control – Figma - www figma com
FireShot Capture 616 - Safe Control – Figma - www figma com
FireShot Capture 617 - Safe Control – Figma - www figma com
FireShot Capture 618 - Safe Control – Figma - www figma com
FireShot Capture 619 - Safe Control – Figma - www figma com

- [ ] Selecting the "Transfer Funds" in the `Control Safe` dialog makes the related inputs appear.
- [ ] An amount input, reused from the one we use for the `Payment` action, appears.
- [ ] A classic user selector input, reused from the one we use for many actions, appears.

ArmandoGraterol:
Added "Transfer Funds" interaction type variation to `GnosisControlSafeDialog`

**Additional change** 🏗

* Updated a core component called `AmountTokens` that was not being used to reuse the amount input and token selector of the create payment dialog

FireShot Capture 639 - Actions - Colony - wonker - localhost
FireShot Capture 640 - Actions - Colony - wonker - localhost

Resolves #3459


arrenv (changes requested):
@ArmandoGraterol  Looks good to me.

This is not in the spec, but I think that we are going to need a loading state for the whole section in the red box. As the tokens available and the amounts will be based on the Safe that is selected. 

download (25)

Which would mean that the default view would look more like this:

405b3d32-1785-4174-bafc-5d7a234208fc


Do you think it would be better to worry about it when wiring it up?


Also, there is a border on the Gnosis Safe Logo, which is making it looking more fuzzy then it should be.

download (24)

ArmandoGraterol:
@arrenv 

> 405b3d32-1785-4174-bafc-5d7a234208fc

Even the transaction type selector? Why is that? We are not loading data when rendering the selector right? Are the options of the selector going to be different depending on the selected safe?
 
> Do you think it would be better to worry about it when wiring it up?

Yeah, I think it could be more convenient to add the loading state when we start to wire up the dialog :)

arrenv:
> Even the transaction type selector? Why is that? We are not loading data when rendering the selector, right? Or are we? Are the options of the selector going to be different depending on the selected safe?

Yeah, the transaction options will likely be the same, but the tokens and balances will be different. I think it will make for a better user experience to have it all appear after selecting a Safe. Happy to see how it works though.

> Yeah, I think it could be more convenient to add the loading state when we start to wire up the dialog :)

Sounds good to me :)

PR #3555, opened 2022-06-29
Add provider caching for domain balances

rdig:
This PR adds in more overwrites to `ethers`'s provider, using our own extended version of it... `ExtendedJsonRpcProvider`

This time we're using it to cache `eth_call` requests and storing them in the local browser's cache.

This reduces the load and number of requests we make to our production RPC endpoint

_NOTE: This is Alex's own work, I'm just handling the PR/merge/deploy overhead_

### Testing

While this can be tested locally as well, it's more suited for a production environment where more requests are going out.

- Create a colony
- When on the colony home page, open your dev tools to the network tab
- Refresh the page
- Note the number or requests made initially and on subsequent requests
- Browse between pages (home, extensions, funds, etc) to simulate more requests going out

### DIFF
--- a/src/modules/core/sagas/utils/extendedJsonRpcProvider.ts
+++ b/src/modules/core/sagas/utils/extendedJsonRpcProvider.ts
@@ -37,10 +37,17 @@ export class ExtendedJsonRpcProvider extends JsonRpcProvider {
     let useCache = false;
     const tx = await resolveProperties(transaction);
 
-    // Only use cache calls if we're on the actions page / colony home
+    // Use cache calls if we're on the actions page / colony home
     const regex = new RegExp('^/colony/[^/]*$');
     if (regex.test(window.location.pathname)) {
       useCache = true;
+    }
+
+    // Unless we're querying the user's token balances
+    const sig = tx.data.slice(0, 10);
+    // getUserLock(address,address) / getTotalObligation(address,address) / balanceOf(address)
+    if (sig === '0x1cc17c52' || sig === '0x006ad100' || sig === '0x70a08231') {
+      useCache = false;
     }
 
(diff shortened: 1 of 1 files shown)


area:
Update that achieves two things:

1. Never cache calls to `addr(bytes32)`. Caching that request broke the colony creation flow if the user was fast enough (which happened every time locally, tricker to see on production but let's be safe).
2. Includes the blocktag in the cache key. Queries for a specific block can be cached forever, and queries for the same value at two different blocks could return different values and so should be treated separately.

arrenv (changes requested):
Very cool so far!

I ran three tests with the same actions and noticed these request counts.

On `master`:
1. 2160 requests
2. 2227 requests
3. 2165 requests

On this branch:
1. 810 requests
2. 782 requests
3. 796 requests

However, I think we have lost some functionality with Motions. Such as, live updating. This is how it works on `master` but I can not get it to work in the same way on this branch.

motions-master

rdig:
Good catch Arren. We'll investigate!

area:
Okay, the logic here is now: use the cache on the colony home / main actions page, unless we're making queries to do with the user's staked balance. Even if we're not using the cache, we take the opportunity o update the cache in case we go back to the colony home / main actions page.

arrenv (approved):
It has been working really well with all my tests! Very nice optimisation!

ArmandoGraterol (approved):
Excellent work! My numbers were quite similar to the ones Arren has in the branch.

Tested the flow `Home > Funding > Home > Members > Home > Extension > Home > Events`

Noticed a decrease of at least 35% of the requests that I had in `master` with the same flow (My requests in `master` were around 1200~1400), which is quite good!

PR #3559, opened 2022-06-29
Fix: Network fee and Available funds positioning

### LINKED ISSUE #3504: Fix the "Network fee" and "Available Funds" margin under inputs
## Steps to reproduce

1. Open up the New Action dialog (Also known as the UAC or Universal Action Controller).
2. Go to Create Expenditure then to Payment
3. You will notice the "Available Funds" is positioned quite far below the team select.
4. If you enter in a value in the amount field, you will also notice the "Network Fee" is the same.

## Expected behaviour

* The "Network fee" and "Available Funds" should be corrected positioned below their respective inputs, as of the design.
* The reason it is currently the way it is, is due to the inputs error message. So, you will need to factor this into your updates. However, both should not need to be visible at the same time.

[Figma Link](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=3553%3A33889)

Payment

## Actual behaviour

* The "Network fee" and "Available Funds" are positioned quite far below their respective inputs.

9a076f10-255a-4b2e-ac4e-51e4c7870488

Julianahlk:
## Description

This PR fixes the "Network fee" and "Available Funds" margin/positioning under inputs.

**TODO:**

- The "Network fee" and "Available Funds" should be positioned correctly below their respective inputs, as per the design.
- The input error messages and the "Network fee" message do not need to be visible at the same time.
 
**Changes** 🏗

- Updated margin for `domainPotBalance` and `networkFee`.
Screen Shot 2022-06-29 at 10 16 56 PM

- Adjusted the `tokenAmountContainer` gap to match the above changes. Previously the gap was too big as well.
Screen Shot 2022-06-29 at 10 17 49 PM

- The inputs error messages and the "Network fee" message do not need to be visible at the same time.

**Previously**
Screen Shot 2022-06-29 at 3 49 52 PM

Screen Shot 2022-06-29 at 4 57 52 PM

**After**
Screen Shot 2022-06-29 at 10 18 17 PM

Screen Shot 2022-06-29 at 10 40 50 PM

- Added new logic in the `createPaymentDialogForm` that better handles 0 or negative numbers. 

Resolves #3504

### DIFF
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
@@ -14,7 +14,7 @@
   flex-direction: column;
   align-items: flex-start;
   margin-top: 20px;
-  gap: 20px;
+  gap: 5px;
 }
 
 .tokenAmountUsd {
@@ -25,6 +25,7 @@
 }
 
 .domainPotBalance, .networkFee {
+  margin-top: -16px;
   font-size: var(--size-tiny);
   font-weight: var(--weight-bold);
   text-align: right;
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
@@ -25,7 +25,8 @@
 }
 
 .domainPotBalance, .networkFee {
-  margin-top: -16px;
+  position: relative;
+  bottom: 14px;
   font-size: var(--size-tiny);
   font-weight: var(--weight-bold);
   text-align: right;
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
@@ -501,10 +501,9 @@ const CreatePaymentDialogForm = ({
               dataTest="paymentAmountInput"
             />
             {networkFeeInverse &&
+              customAmountError === undefined &&
               values.amount &&
-              values.amount !== '0' &&
-              values.amount !== '0.' &&
-              values.amount !== '.' && (
+              Number(values.amount) > 0 && (
                 <div className={styles.networkFee}>
                   <FormattedMessage
                     {...MSG.fee}


src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
```diff
@@ -501,10 +501,9 @@ const CreatePaymentDialogForm = ({
               dataTest="paymentAmountInput"
             />
             {networkFeeInverse &&
+              customAmountError === undefined &&
               values.amount &&
-              values.amount !== '0' &&
-              values.amount !== '0.' &&
-              values.amount !== '.' && (
+              Number(values.amount) > 0 && (
```
rdig, line 506: I like this change. Just a heads up that Armando is implementing something similar in #3535

So maybe you guys should coordinate to avoid merge conflicts

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
```diff
@@ -25,6 +25,7 @@
 }
 
 .domainPotBalance, .networkFee {
+  margin-top: -16px;
```
rdig, line 28: Please leave a comment explaining the use of negative margins

rdig (approved):
Yeah, this is all good. Nice work.

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
arrenv: Not saying it needs to be changed, although a more robust alternative approach to this could be to use `position: absolute`, with right and bottom positioning and `position: relative` on the fields.

arrenv (approved):
This is looking much better, nice work!

download (27)

download (26)

chinins (approved):
Very nice! :)

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
Julianahlk: Thanks for the heads up Raul. I already talked to Armando and he likes my changes more so I am going to go ahead and merge them.

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
Julianahlk: > Please leave a comment explaining the use of negative margins

Because the `inputStatus` css component has visibility hidden and this is making the error message taking up space so my thought was to use the negative margin to get around that issue.  

The other alternative I was thinking was adding another message field to this component but it seemed that this would be too involved for this change? What do you think?

Julianahlk: > Not saying it needs to be changed, although a more robust alternative approach to this could be to use `position: absolute`, with right and bottom positioning and `position: relative` on the fields.

I could give it a try :) thanks for the suggestion.

rdig: > > Please leave a comment explaining the use of negative margins
> 
> Because the `inputStatus` css component has visibility hidden and this is making the error message taking up space so my thought was to use the negative margin to get around that issue.
> 
> The other alternative I was thinking was adding another message field to this component but it seemed that this would be too involved for this change? What do you think?

Lol. Please leave a comment, in the css file :)

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.tsx
danbr: > I like this change. Just a heads up that Armando is implementing something similar in #3535
> 
> So maybe you guys should coordinate to avoid merge conflicts

@rdig Not sure #3535 is the one.....

src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
Julianahlk: Hey @arrenv I ended up using relative positioning and setting the bottom. It didn't seem that the absolute positioning was necessary but can you please take a look at it again so I can merge this? Thank you!

PR #3562, opened 2022-06-29
Add "Raw transaction" interaction variation

### LINKED ISSUE #3484: Add `Control Safe` Dialog's `Raw transaction` type variation UI
### Story

As a user that wants to make a raw transaction to the safe, I want to see the `Control Safe` variation of the `Raw transaction` transaction type.

### Description

When the user selects the `Raw transaction` type using the selector in the dialog's "basic UI" in the `Control Safe` Dialog, the inputs related to the transaction type should appear.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=116%3A17490)

FireShot Capture 622 - Safe Control – Figma - www figma com

- [ ] Selecting the "Raw transaction" type in the `Control Safe` dialog makes the related inputs appear.
- [ ] A classic user selector input, reused from the one we use for many actions, appears.
- [ ] A "value" input appears which only allows wei amounts.
- [ ] A "data" input appears which allows any type of char. 

ArmandoGraterol:
FireShot Capture 650 - Actions - Colony - wonker - localhost

**NOTE:** There's a bug in the `Select` core component where the placeholder is shown as one of the options in the dialog. This is going to be fixed in another PR.

Resolves #3484

### DIFF
--- a/src/modules/core/components/Fields/Input/Input.tsx
+++ b/src/modules/core/components/Fields/Input/Input.tsx
@@ -5,7 +5,7 @@ import { MessageDescriptor, useIntl } from 'react-intl';
 import { useField } from 'formik';
 import { nanoid } from 'nanoid';
 
-import { SimpleMessageValues } from '~types/index';
+import { ComplexMessageValues, SimpleMessageValues } from '~types/index';
 
 import InputLabel from '../InputLabel';
 import InputStatus from '../InputStatus';
@@ -55,7 +55,7 @@ export interface Props extends Omit<InputComponentProps, 'placeholder'> {
   label?: string | MessageDescriptor;
 
   /** Label text values for intl interpolation */
-  labelValues?: SimpleMessageValues;
+  labelValues?: SimpleMessageValues | ComplexMessageValues;
 
   /** Placeholder text */
   placeholder?: string | MessageDescriptor;
--- a/src/modules/core/components/Fields/InputLabel/InputLabel.tsx
+++ b/src/modules/core/components/Fields/InputLabel/InputLabel.tsx
@@ -1,7 +1,7 @@
 import React, { ReactNode } from 'react';
-import { MessageDescriptor, useIntl } from 'react-intl';
+import { FormattedMessage, MessageDescriptor, useIntl } from 'react-intl';
 
-import { SimpleMessageValues } from '~types/index';
+import { ComplexMessageValues, SimpleMessageValues } from '~types/index';
 import { getMainClasses } from '~utils/css';
 
 import styles from './InputLabel.css';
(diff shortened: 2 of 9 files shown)


Julianahlk:
The input fields Select Recipient, Value and Data bytes are not showing up for me, is there a pre-requisite to be able to see them? 

Screen Shot 2022-07-02 at 6 53 33 PM

chinins (approved):
Very nice!

@Julianahlk, If you slick on select transaction & select raw transaction, then everyhting will appear. I think that this is the bug Armando was referring to in his description. @ArmandoGraterol, am I right?

ArmandoGraterol:
@Julianahlk @chinins 
> @ArmandoGraterol, am I right?

That's correct! It's the bug that I mentioned in the description. Currently, when you enter the dialog, no options are selected even though the Select shows the contrary. I fixed this bug and more in #3570

willm30 (changes requested):
On selecting the All Saints Safe, I received the following error which isn't wrapping:

err-control

Assuming you fixed this error in #3570, the error message should nevertheless wrap.

And, my browser's autofill entered the following number which returned a slightly confusing error. The problem was the commas but I think either the validation should accept numbers in this format or the error message should be clearer:
gt-0

Julianahlk:
> That's correct! It's the bug that I mentioned in the description. Currently, when you enter the dialog, no options are selected even though the Select shows the contrary. I fixed this bug and more in #3570

I don't know why it wasn't showing up for me before but now it works. Other than the wrapping error Will pointed out above, this works as expected.

Julianahlk (approved):

ArmandoGraterol:
@willm30 Both of those bugs should be fixed now!

willm30 (approved):
Errors gone 🙂 Just flagging that, while you can't manually enter a number in the format: 111,111  (i.e. with commas), you still can by using the browser's auto-complete (see below), which may cause problems if the subsequent function are expecting comma-less inputs like "111111".   

commas-control-safe

ArmandoGraterol:
@willm30 
> Errors gone slightly_smiling_face Just flagging that, while you can't manually enter a number in the format: 111,111 (i.e. with commas), you still can by using the browser's auto-complete (see below), which may cause problems if the subsequent functions are expecting comma-less input values like "111111".

Yeah, that's the expected behavior as the amount inputs in the payment and mint tokens dialog work the same. You cannot manually type commas but the commas are added automatically if the number has them (Try typing 1000 in any amount input).

Also, the formatted amounts (amounts with the commas) are only for the user's display, they are not used when submitting an action, or doing further logical or math operations. Just in case you were wondering if the amount with commas is the format that we use when submitting something (We normally use the unformatted WEI amount).

willm30:
@ArmandoGraterol Got it, thanks for explaining!

PR #3566, opened 2022-06-30
Update members page subdomain view

### LINKED ISSUE #3565: Members page item doesn't show details for the selected domain but for the root
Have some members in a colony and at least 1 sub domain. Change from the root to the sub domain. The values will stay the same although they should be different (specific to the subdomain)

chinins:
## Description

This PR fixes the display of contributors while in the sub-domain view.

**Changes** 🏗

- Update the type of the domain id that's passed through the selection.

resolves #3565


danbr (approved):
Well done. LGTM. Approved.......

PR #3569, opened 2022-06-30
Fix: Changed the colonies list scroll to a fixed position

### LINKED ISSUE #3553: Colonies list scrolls with page
Right side colonies list should be static (no matter how many colonies you've subscribed to)

Currently, it scrolls up when scrolling the page. See the demo:

bug-colonies-list-scroll

Julianahlk:
## Description

The colonies list (right side of screen) should be static no matter how many colonies you've subscribed to.

**Changes** 🏗

-  The colonies list was changed to a fixed position

colony

Resolves #3553

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyTitle/ColonyTitle.css
+++ b/src/modules/dashboard/components/ColonyHome/ColonyTitle/ColonyTitle.css
@@ -1,8 +1,7 @@
 .main {
   display: block;
-  margin: 0 -40px 50px;
+  margin: 0 -40px 50px 0px;
   padding-top: 15px;
-  text-align: right;
 }
 
 .wrapper {
--- a/src/modules/pages/components/RouteLayouts/Default/Default.css
+++ b/src/modules/pages/components/RouteLayouts/Default/Default.css
@@ -10,6 +10,7 @@
 .coloniesList {
   flex: 0 0 auto;
   width: 60px;
+  position: fixed;
   border-right: 1px solid var(--temp-grey-13);
 }
 


Julianahlk:
This is an issue because of the negative margin does it make sense to fix this overlapping of the colony name and address here or should we deferred this to the mobile responsiveness tasks?

willm30:
> The above is an issue because of the negative margin. Does it make sense to fix this overlapping of the colony name and address in this PR or should we deferred this to the mobile responsiveness tasks as this only happens in smaller screen sizes?

I wouldn't say it belongs in mobile responsiveness as the screen size you're showing in the screenshot wouldn't be considered mobile. Mobile is currently 768px and under, with plans to ensure the dapp responds nicely at breakpoints 980 and 1200 also.

willm30 (approved):
Works as intended!

willm30:
> The above is an issue because of the negative margin. Does it make sense to fix this overlapping of the colony name and address in this PR or should we deferred this to the mobile responsiveness tasks as this only happens in smaller screen sizes?

Probably worth creating a separate issue for it.

Julianahlk:
> Probably worth creating a separate issue for it.

Okay, I can do that. Thanks for the suggestion.

arrenv:
> Screen Shot 2022-06-30 at 8 41 05 PM
> 
> The above is an issue because of the negative margin. Does it make sense to fix this overlapping of the colony name and address in this PR or should we deferred this to the mobile responsiveness tasks as this only happens in smaller screen sizes?

I think it is worth fixing in this PR, save the extra paper work. Also, because this PR introduces the issue.

arrenv (approved):
Other then the issue with the ColonyTitle, this looks good to me.

Julianahlk:
> Other then the issue with the ColonyTitle, this looks good to me.

Thanks Arrren! I will make the change of the ColonyTitle in this PR as well.

PR #3570, opened 2022-06-30
Add "Contract interaction" transaction variation

### LINKED ISSUE #3483: Add `Control Safe` Dialog's `Contract interaction` type variation UI
### Story

As a user that wants to interact with a safe's contract, I want to see the `Control Safe` variation of the `Contract interaction` transaction type.

### Description

When the user selects the `Contract interaction` type using the selector in the dialog's "basic UI" in the `Control Safe` Dialog, the inputs related to the transaction type should appear.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=115%3A16674)

FireShot Capture 621 - Safe Control – Figma - www figma com

- [x] Selecting the "Contract interaction" type in the `Control Safe` dialog makes the related inputs appear.
- [x] A contract address selector appears, similar to the one we use for users with a help text that indicates if the contract was found.
- [x] A text area appears where the user can copy-paste an ABI/JSON data.
- [x] A interaction selector appears, which would show the available functions of the contract. (Not connected in this PR as this is only UI).

ArmandoGraterol:
As an extra: 

- Fixed a bug in the `Select` core component that displayed the first option of the `options` array as the placeholder.
- Fixed the `Select` component crashing when the `placeholder` prop was a `MessageDescriptor`

FireShot Capture 653 - Actions - Colony - wonker - localhost

Resolves #3483

### DIFF
--- a/src/modules/core/components/Fields/Select/Select.tsx
+++ b/src/modules/core/components/Fields/Select/Select.tsx
@@ -255,11 +255,7 @@ const Select = ({
   }, [handleOutsideClick, isOpen]);
 
   const activeOptionDisplay = useMemo<ReactNode>(() => {
-    /*
-     * @NOTE If the active option is removed by something (ie: filtered out),
-     * fall back to the last entry in the options array
-     */
-    const activeOption = options[checkedOption] || options[options.length - 1];
+    const activeOption = options[checkedOption];
     let activeOptionLabel;
     if (activeOption) {
       if (typeof activeOption.label === 'object') {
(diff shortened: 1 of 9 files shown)


willm30 (approved):
Looks good! Only thought is that "Create Transaction" button should be disabled by default (i.e. until all relevant inputs pass validation)? 
create-trans

chinins (approved):
Very nice! Especially the bug fixes for the Select :)

ArmandoGraterol:
@willm30 
> Looks good! Only thought is that "Create Transaction" button should be disabled by default (i.e. until all relevant inputs pass 

Fixed! There was something weird happening with the validation, thanks

PR #3579, opened 2022-07-05
Add delay to `GasStationPopover`

### LINKED ISSUE #3505: Hide the `GasStationPopover` after all transactions are complete and a delay
### Story

As a user performing transactions within Colony, I often don't realize when the transactions are complete on certain actions as the `GasStationPopover` stays visible.

### Description

This requires adding a timeout to the `GasStationPopover` after all the transactions are complete and a **3 sec delay**.

This is only an issue for certain actions, such as, installing and enabling extensions and claiming incoming funds, so, you can test with these transactions.

gas-station-popover-timeout

Julianahlk:
## Description

This PR uses setTimeout to close the GasStationPopover after a delay only when the GasStationPopover is opened automatically.

To test, this make sure the GasStationPopover closes after a 3 sec delay when opened automatically. The GasStationPopover should remain open when opened from the wallet. 

**Changes** 🏗
`closeAfterDelay` - Created a new prop to tell the popover to close after a 3 sec delay.

**To Test**
- Please make sure you test transactions with MetaMask.

## TODO

- [x]  Add a prop to close the popover after a delay.

Resolves #3505

### DIFF
--- a/src/modules/core/components/Popover/Popover.tsx
+++ b/src/modules/core/components/Popover/Popover.tsx
@@ -134,15 +134,19 @@ const Popover = ({
     [onClose],
   );
 
-  const requestOpen = useCallback(() => {
-    if (isOpen) {
-      return;
-    }
+  useEffect(() => {
     if (closeAfterDelay) {
       closeTimeoutRef.current = setTimeout(() => {
         close();
       }, 3000 + (openDelay || 0));
     }
+  }, [close, openDelay, closeAfterDelay]);
+
+  const requestOpen = useCallback(() => {
+    if (isOpen) {
+      return;
+    }
+
     if (openDelay) {
       openTimeoutRef.current = setTimeout(() => {
         _setIsOpen(true);
@@ -150,7 +154,7 @@ const Popover = ({
       return;
     }
     _setIsOpen(true);
-  }, [isOpen, openDelay, closeAfterDelay]);
+  }, [isOpen, openDelay]);
 
   const handleWrapperFocus = useCallback(() => {
     if (retainRefFocus && referenceElement instanceof HTMLInputElement) {
--- a/src/modules/core/components/Popover/Popover.tsx
+++ b/src/modules/core/components/Popover/Popover.tsx
@@ -54,6 +54,8 @@ interface Props {
   onClose?: (data?: any, modifiers?: { cancelled: boolean }) => void;
   /** Delay opening of popover for `openDelay` ms */
   openDelay?: number;
+  /** Close popover after `closeDelay` ms */
+  closeDelay?: number;
   /** Popover placement */
   placement?: Placement;
   /** Options to pass to the underlying PopperJS element. See here for more: https://popper.js.org/docs/v2/constructors/#options. */
@@ -87,6 +89,7 @@ const Popover = ({
   isOpen: isOpenProp = false,
   onClose,
   openDelay,
+  closeDelay,
   placement: placementProp = 'auto',
   popperOptions = {},
   retainRefFocus,
@@ -102,6 +105,7 @@ const Popover = ({
   const [arrowElement, setArrowElement] = useState<HTMLElement | null>(null);
 
   const openTimeoutRef = useRef<number>();
+  const closeTimeoutRef = useRef<number>();
   const { current: elementId } = useRef<string>(nanoid());
 
   const { attributes, styles, state } = usePopper(
@@ -120,6 +124,11 @@ const Popover = ({
     if (isOpen) {
       return;
     }
+    if (closeDelay && window) {
+      closeTimeoutRef.current = window.setTimeout(() => {
+        _setIsOpen(false);
+      }, closeDelay);
+    }
     if (openDelay && window) {
       openTimeoutRef.current = window.setTimeout(() => {
         _setIsOpen(true);
@@ -133,6 +142,7 @@ const Popover = ({
     (data?: any, modifiers?: { cancelled: boolean }) => {
       if (window) {
         window.clearTimeout(openTimeoutRef.current);
+        window.clearTimeout(closeTimeoutRef.current);
       }
       if (!isOpen) return;
       _setIsOpen(false);
@@ -253,9 +263,13 @@ const Popover = ({
   // Timeouts
   useEffect(() => {
     const currentOpenTimeoutRef = openTimeoutRef && openTimeoutRef.current;
+    const currentCloseTimeoutRef = closeTimeoutRef && closeTimeoutRef.current;
     return () => {
       if (window && currentOpenTimeoutRef) {
         window.clearTimeout(currentOpenTimeoutRef);
+      }
+      if (window && currentCloseTimeoutRef) {
+        window.clearTimeout(currentCloseTimeoutRef);
       }
     };
   }, []);
--- a/src/modules/core/components/Popover/Popover.tsx
+++ b/src/modules/core/components/Popover/Popover.tsx
@@ -150,7 +150,7 @@ const Popover = ({
       return;
     }
     _setIsOpen(true);
-  }, [close, isOpen, openDelay, closeAfterDelay]);
+  }, [isOpen, openDelay, closeAfterDelay]);
 
   const handleWrapperFocus = useCallback(() => {
     if (retainRefFocus && referenceElement instanceof HTMLInputElement) {
@@ -260,20 +260,6 @@ const Popover = ({
       _setIsOpen(!!isOpenProp);
     }
   }, [close, isOpen, isOpenProp, lastIsOpenProp, requestOpen]);
-
-  // Timeouts
-  useEffect(() => {
-    const currentOpenTimeoutRef = openTimeoutRef && openTimeoutRef.current;
-    const currentCloseTimeoutRef = closeTimeoutRef && closeTimeoutRef.current;
-    return () => {
-      if (currentOpenTimeoutRef) {
-        clearTimeout(currentOpenTimeoutRef);
-      }
-      if (currentCloseTimeoutRef) {
-        clearTimeout(currentCloseTimeoutRef);
-      }
-    };
-  }, []);
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


src/modules/core/components/Popover/Popover.tsx
```diff
@@ -116,30 +120,35 @@ const Popover = ({
...
+  );
+
   const requestOpen = useCallback(() => {
     if (isOpen) {
       return;
     }
+    if (closeAfterDelay && window) {
+      closeTimeoutRef.current = window.setTimeout(() => {
+        close();
+      }, 3000 + (openDelay || 0));
```
rdig, line 142: Is there any particular reason why you're using `window.setTimeout` instead of just `setTimeout` ?

src/modules/core/components/Popover/Popover.tsx
```diff
@@ -253,10 +262,14 @@ const Popover = ({
...
   useEffect(() => {
     const currentOpenTimeoutRef = openTimeoutRef && openTimeoutRef.current;
+    const currentCloseTimeoutRef = closeTimeoutRef && closeTimeoutRef.current;
     return () => {
       if (window && currentOpenTimeoutRef) {
         window.clearTimeout(currentOpenTimeoutRef);
       }
+      if (window && currentCloseTimeoutRef) {
+        window.clearTimeout(currentCloseTimeoutRef);
+      }
```
rdig, line 272: Again, I see no reason to use `window.clearTimeout` as opposed to just `clearTimeout`

src/modules/users/components/GasStation/GasStationPopover.tsx
```diff
@@ -38,8 +39,9 @@ const GasStationPopover = ({
     if (prevTxCount != null && txCount > prevTxCount) {
       setOpen(true);
       setTxNeedsSigning(true);
+      setUseCloseDelay(true);
```
rdig, line 42: Not good enough. If the user creates a new transactions and then delays to sign it / approve it, the gas station will close.

See example:
bug-gas-station-closes-prematurely

You need a new check that takes into account the `txCount` and the `prevTxCount` while also checking if the transactions have succeeded, and only then close the gas station

rdig: It will also auto-close if the transaction takes too long to mine.

rdig (changes requested):
This works fine for the "happy path" where the user signs the transaction in a timely fashion, and the transaction mines early.

demo-gas-station-autoclose

But it does not work properly if the user delays the signature, of if the transactions takes too long to mine _(which will happen in production)_

So the solution for this is to also check if the transactions _(all transactions)_ has succeeded, otherwise disable the auto-close logic.

src/modules/core/components/Popover/Popover.tsx
rdig: Great :+1:

Julianahlk: @rdig Ive done some investigation and I am not sure how I can avoid using Window when using setTimeout inside a useCallback. I saw that in StepMetaMask.tsx that the same technique is used where the timeout is assigned to a variable with the type useRef<number>(). Is there someway to not use window inside a useCallback or should I restructure this somehow to do the setTimeout outside of the useCallback?

One option is to change the variable, const openTimeoutRef = useRef<any>(null); and that will allow using setTimeout without window but I am not sure using any is the best solution or is that ok for this case?

willm30: @Julianahlk You can use the Node `setTimeout` and cast the return value as a `number` for use with `clearTimeout`. These appeases the ts compiler and means you can dispense with the references to `window`. 

See: https://stackoverflow.com/questions/45802988/typescript-use-correct-version-of-settimeout-node-vs-window

I believe this would resolve your problem?

Julianahlk: Thanks Will! It did solve the problem.

ArmandoGraterol (commented):
WHat's the state of this? I've been testing with metamask and the gas station is still closing itself when the transaction is in progress

src/modules/core/components/Popover/Popover.tsx
```diff
@@ -116,30 +120,37 @@ const Popover = ({
...
-      if (window) {
-        window.clearTimeout(openTimeoutRef.current);
-      }
-      if (!isOpen) return;
-      _setIsOpen(false);
-      if (typeof onClose == 'function') onClose(data, modifiers);
-    },
-    [isOpen, onClose],
-  );
+  }, [isOpen, openDelay, closeAfterDelay]);
```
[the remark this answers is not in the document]
Julianahlk: > What's the state of this? I've been testing with metamask and the gas station is still closing itself when the transaction is in progress

@ArmandoGraterol I got stuck on figuring out how to determine when all transactions have been completed successfully when connected to Metamask, so I started working on the Gnosis and the Decisions tasks which both seemed to be higher priority features 😅. I can resume work as soon as I am done with the Add Decision Preview Page task. However, if you have any suggestions for how to make this work, please let me know.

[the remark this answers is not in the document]
willm30: @Julianahlk I took a look into this to see if I could offer a solution, and I've made a commit to this branch with what I found.  It checks if all the transactions in the most recent group have succeeded, and if so, will close the popover after the delay. Let me know your thoughts 🙂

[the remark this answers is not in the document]
Julianahlk: @ArmandoGraterol I made some more updates to get this working so that the popover only closes when the transactions have succeeded but it works in most cases and sometimes it closes before all transactions have succeeded. Could you take a look?, I could use some help figuring out what the issue is. Thanks.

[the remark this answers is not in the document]
Julianahlk: @willm30 I just noticed your comment from earlier now, so weird that I didn't see it 😅 Anyway, I also didn't see your commit here in Github. Are you sure you pushed the commit?.

Julianahlk:
Hey @arren, My last updates for this issue have it working so that the popover only closes when the transactions have succeeded. However, the popover delay only works in some cases and sometimes it closes before all transactions have succeeded. I could not figure out a good solution for this issue and even a few months ago when I talked to Raul about it, he did not know a good solution for it as well but he asked me to follow up with Product (you) to decide if this issue is something worth to continue pursuing or what should the next steps be as we are supposed to start working on tasks for the CDapp? 

  Please feel free to test the popover and let me know your thoughts. Thank you!

arrenv (approved):
@Julianahlk This appears to be working well based on my tests.

It closed correctly after 3 seconds for transactions such as extension management, staking, activating tokens, etc.

I made sure to test with Metamask as well and it didn't auto close when leaving it for awhile, rejecting the second transaction, retrying, etc.

Screenshot 2022-10-28 183445



testing-popover-close

willm30 (approved):
Working for me. Have you been able to reproduce the cases where it closes improperly? If so, perhaps open a separate issue to tackle those separately.

chinins (approved):
Very good!

Julianahlk:
> Working for me. Have you been able to reproduce the cases where it closes improperly? If so, perhaps open a separate issue to tackle those separately.

Actually, last time I did some testing my dev env was acting weird and I thought I was having issues with the voting section when logged in with Metamask but after spending more time testing this again and again with a fresh new start to my dev env, even the voting process works well when logged in with Metamask and since no one found issues as well I will go ahead and merge this PR. 

Thank you all for testing it!

src/modules/core/components/Popover/Popover.tsx
```diff
@@ -116,31 +120,42 @@ const Popover = ({
...
 
+  const close = useCallback(
+    (data?: any, modifiers?: { cancelled: boolean }) => {
+      if (openTimeoutRef.current) {
+        clearTimeout(openTimeoutRef.current);
+      }
+      if (closeTimeoutRef.current) {
+        clearTimeout(closeTimeoutRef.current);
+      }
+      _setIsOpen(false);
```
rdig, line 131: I'm not really a fan of prepending the variable name with an underscore. 

Any particular reason why you did that ?

src/modules/core/components/Popover/Popover.tsx
```diff
@@ -116,31 +120,42 @@ const Popover = ({
...
+      if (typeof onClose == 'function') onClose(data, modifiers);
```
rdig, line 132: Please use brackets. It makes the code much more readable

```suggestion
      if (typeof onClose == 'function') {
        onClose(data, modifiers)
      };
```

src/modules/core/components/Popover/Popover.tsx
```diff
@@ -116,31 +120,42 @@ const Popover = ({
...
+    },
+    [onClose],
+  );
+
+  useEffect(() => {
+    if (closeAfterDelay) {
+      closeTimeoutRef.current = setTimeout(() => {
+        close();
+      }, 3000 + (openDelay || 0));
```
rdig, line 141: Why is this value hardcoded ? This should be a prop!

rdig (changes requested):
This is mostly good. The only thing I rejected this for was because you hardcoded the close delay.

src/modules/core/components/Popover/Popover.tsx
rdig: Please do. I see no reason for it to be like this

src/modules/core/components/Popover/Popover.tsx
Julianahlk: oops, I added a prop for this now.

src/modules/core/components/Popover/Popover.tsx
Julianahlk: All fixed! Please check again when possible. Thanks!

rdig (approved):
I'm all good with this, except for that awful variable name :)

Please find a better name for it, then you can merge it.

PR #3585, opened 2022-07-06
Fix: Rendering error for threshold and alignment of help icons

### LINKED ISSUE #3572: Fix threshold progress bar rendering error
## Steps to reproduce

1. Create a Motion. 
2. Fully Stake and Object a motion to initiate voting.
4. See the voting timeout is showing a 0 rendering error.

## Expected behavior

* The threshold progress bar will not display a zero rendering error.
* Help icons need to be aligned accordingly next to the labels.

## Current behavior

Screen Shot 2022-07-04 at 3 04 05 PM

Julianahlk:
## Description

This PR fixes the issue with a 0 rendering error for the threshold progress bar. As well as the alignment of the help icons in the `VoteDetails` component. 

##Story
The threshold value was 0, so when it was being evaluated it kept returning a 0 in the tsx and rendered by React to the screen. In this situation, it is better to use a boolean.

[Figma link](https://www.figma.com/file/CCgGsBzhX3BUENS1IvgP7N/Motions-%26-Disputes?node-id=3756%3A1)

**Before**
Screen Shot 2022-07-04 at 3 27 22 PM

**Changes** 🏗

* Removed unnecessary divs in the `VoteDetails` component that were causing the issue with the misalignment of the help icons.

* Created a new variable that would handle the threshold as a boolean instead of a number.  

**After**
Screen Shot 2022-07-08 at 6 55 36 PM

## TODO
- [x] Fix the 0 rendering error for the threshold progress bar.
- [x] Fix alignment of help icons

Resolves #3572

### DIFF
--- a/src/modules/core/components/ProgressBar/ProgressBar.tsx
+++ b/src/modules/core/components/ProgressBar/ProgressBar.tsx
@@ -42,9 +42,10 @@ const ProgressBar = ({
 }: Props) => {
   const { formatMessage } = useIntl();
   const titleText = formatMessage(MSG.titleProgress, { value, max });
+  const shouldDisplayThreshold = !!(threshold && threshold > 0);
   return (
     <div className={`${styles.wrapper} ${getMainClasses(appearance, styles)}`}>
-      {threshold && (
+      {shouldDisplayThreshold && (
         <div
           style={{
             left: `calc(${threshold}% - 12px)`,
--- a/src/modules/core/components/ProgressBar/ProgressBar.tsx
+++ b/src/modules/core/components/ProgressBar/ProgressBar.tsx
@@ -48,7 +48,7 @@ const ProgressBar = ({
       {shouldDisplayThreshold && (
         <div
           style={{
-            left: `calc(${threshold}% - 12px)`,
+            left: `calc(${threshold}% - 14px)`,
           }}
           className={styles.threshold}
         >
--- a/src/modules/core/components/ProgressBar/ProgressBar.tsx
+++ b/src/modules/core/components/ProgressBar/ProgressBar.tsx
@@ -42,10 +42,10 @@ const ProgressBar = ({
 }: Props) => {
   const { formatMessage } = useIntl();
   const titleText = formatMessage(MSG.titleProgress, { value, max });
-  const shouldDisplayThreshold = !!(threshold && threshold > 0);
+
   return (
     <div className={`${styles.wrapper} ${getMainClasses(appearance, styles)}`}>
-      {shouldDisplayThreshold && (
+      {!!threshold && (
         <div
           style={{
             left: `calc(${threshold}% - 14px)`,
--- a/src/modules/dashboard/components/ActionsPage/VoteWidget/VoteDetails.tsx
+++ b/src/modules/dashboard/components/ActionsPage/VoteWidget/VoteDetails.tsx
@@ -108,17 +108,15 @@ const VoteDetails = ({
     <div>
       <div className={styles.item}>
         <div className={styles.label}>
-          <div>
-            <FormattedMessage {...MSG.votingMethodLabel} />
-            <QuestionMarkTooltip
-              tooltipText={MSG.votingMethodTooltip}
-              className={styles.help}
-              tooltipClassName={styles.tooltip}
-              tooltipPopperOptions={{
-                placement: 'right',
-              }}
-            />
-          </div>
+          <FormattedMessage {...MSG.votingMethodLabel} />
+          <QuestionMarkTooltip
+            tooltipText={MSG.votingMethodTooltip}
+            className={styles.help}
+            tooltipClassName={styles.tooltip}
+            tooltipPopperOptions={{
+              placement: 'right',
+            }}
+          />
         </div>
         <div className={styles.value}>
           {/*
@@ -134,17 +132,15 @@ const VoteDetails = ({
         <>
           <div className={styles.item}>
             <div className={styles.label}>
-              <div>
-                <FormattedMessage {...MSG.reputationTeamLabel} />
-                <QuestionMarkTooltip
-                  tooltipText={MSG.reputationTeamTooltip}
-                  className={styles.help}
-                  tooltipClassName={styles.tooltip}
-                  tooltipPopperOptions={{
-                    placement: 'right',
-                  }}
-                />
-              </div>
+              <FormattedMessage {...MSG.reputationTeamLabel} />
+              <QuestionMarkTooltip
+                tooltipText={MSG.reputationTeamTooltip}
+                className={styles.help}
+                tooltipClassName={styles.tooltip}
+                tooltipPopperOptions={{
+                  placement: 'right',
+                }}
+              />
             </div>
             <div className={styles.value}>
               <div className={styles.reputation}>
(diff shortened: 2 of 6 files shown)


src/modules/core/components/ProgressBar/ProgressBar.tsx
```diff
@@ -42,12 +42,13 @@ const ProgressBar = ({
 }: Props) => {
   const { formatMessage } = useIntl();
   const titleText = formatMessage(MSG.titleProgress, { value, max });
+  const shouldDisplayThreshold = !!(threshold && threshold > 0);
```
willm30, line 45: `!!threshold` would have the same effect here. If `threshold` is falsy (e.g. `undefined` or `0`) it will return `false`, otherwise `true`.  Then you could remove this line and below do `!!threshold && (...)`

willm30 (approved):
Nice alignment of the question marks. I have left a comment re the boolean logic. 

As an aside to this issue, I'm curious to know if anybody else had trouble trying to test the voting mechanism? I found the experience error-prone (either receiving errors or unexpected behaviour). For example, I received the following error on multiple occasions, as well as the voting phase itself going straight to reveal, despite my not having voted.

revert

 I can document more extensively if this is something others have noticed also.

Julianahlk:
> I can document more extensively if this is something others have noticed also.

Yes, please. I am not getting the same error as you. However, sometimes I get something similar if the reputation miner is on but as soon as I toggle it off the staking or objection works for me.

src/modules/core/components/ProgressBar/ProgressBar.tsx
Julianahlk: I agree. I was keeping the threshold inside a variable because generally I like to use variable names to help document the code but in this case its simple enough that it isn't necessary. Thank you for the suggestion. I made the change already.

ArmandoGraterol:
The alignment and `0` left of the progress bar are fixed :+1: 

Something fishy is definitely happening with the threshold and progress percentage thought. For me, they remain `0` even though the operations that calculate them should not return `0`. Don't know if you want to take a look at that in this PR or if you prefer to create a separate issue for it.

@willm30 
That error in the screenshot is expected. The reputation monitor fast-forwards to the future every time it runs in order to calculate the reputation, this is something that happens only in dev's environments to avoid having to wait too much time to generate reputation.

You are staking a motion, which is something that changes the user's balance, while having the monitor ON. The monitor then fast-forwards a few days into the future, making the staking period or even the motion instantly finish because all the timers ran out. This is why you can't stake, or why the voting phase goes straight to revealing (voting phase timer ran out).

Julianahlk:
@ArmandoGraterol Do you know how to test to get a threshold other than 0? I wasn't able to test this in QA either and I figured my dev environment wasn't setup correctly for this type of test. Would it be possible to test in Production? I can create a new issue but wanted to be clear on the correct repro steps for this issue.

src/modules/core/components/ProgressBar/ProgressBar.tsx
```diff
@@ -42,12 +42,13 @@ const ProgressBar = ({
 }: Props) => {
   const { formatMessage } = useIntl();
   const titleText = formatMessage(MSG.titleProgress, { value, max });
+
```
rdig, line 45: So the `value` for the `ProgressBar` does not get loaded/fetched correctly.

Please open a issue to track investigating that

src/modules/core/components/ProgressBar/ProgressBar.tsx
```diff
@@ -42,12 +42,13 @@ const ProgressBar = ({
...
   return (
     <div className={`${styles.wrapper} ${getMainClasses(appearance, styles)}`}>
-      {threshold && (
+      {!!threshold && (
```
rdig, line 48: Threshold now looks good, however the questionmark icon alignment (for that line) looks wonky on my machine.

Screenshot from 2022-07-13 21-25-43

rdig: Same, when you get to the finalize step

Screenshot from 2022-07-13 21-26-33

src/modules/dashboard/components/ActionsPage/VoteWidget/VoteDetails.tsx
```diff
@@ -108,17 +108,15 @@ const VoteDetails = ({
...
-          </div>
+          <FormattedMessage {...MSG.votingMethodLabel} />
+          <QuestionMarkTooltip
+            tooltipText={MSG.votingMethodTooltip}
+            className={styles.help}
+            tooltipClassName={styles.tooltip}
+            tooltipPopperOptions={{
+              placement: 'right',
+            }}
+          />
```
rdig, line 119: Very nice. They render in properly in line now

Screenshot from 2022-07-13 21-25-49

rdig (changes requested):
All good, except there's two more question mark icons that are mis-aligned.

Here's a basic outline to follow in order to test this properly _(and not run into the revert transaction that Will was running into)_

1. Toggle mining on: http://0.0.0.0:3001/reputation/monitor/toggle
2. Create 1st user and a colony
3. mint tokens
4. payout 10 tokens to 1st user
5. log out
6. log in, create 2nd user and join the colony
7. log out
8. log back in with the 1st user
9. pay the 10 tokens to the same user
10. initialize voting reputation
11. enable voting reputation
12. Toggle mining off: http://0.0.0.0:3001/reputation/monitor/toggle
13. refresh app

By this point, you'll have a colony, with voting reputation enabled, and 2 users, each having 50% of the reputation in said colony

If you need to pass time manually, just use this from the command line:
```shell
curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id":1}' localhost:8545
```

ArmandoGraterol:
> @ArmandoGraterol Do you know how to test to get a threshold other than 0? I wasn't able to test this in QA either and I figured my dev environment wasn't setup correctly for this type of test. Would it be possible to test in Production? I can create a new issue but wanted to be clear on the correct repro steps for this issue.

I tried to get a different threshold and it seems like it's not working as expected. By following the steps Raul outlined, and leaving the default extension settings you should be able to test if the threshold and progress bar display works or not, and right now it seems like it's not working properly because when you vote with 1 user the progress remains in 0 as if nobody voted.

src/modules/core/components/ProgressBar/ProgressBar.tsx
Julianahlk: Sure thing, I can do that!

src/modules/core/components/ProgressBar/ProgressBar.tsx
Julianahlk: Oh yes, these icons are annoying but I will get them 💪.

Julianahlk:
> I tried to get a different threshold and it seems like it's not working as expected. By following the steps Raul outlined, and leaving the default extension settings you should be able to test if the threshold and progress bar display works or not, and right now it seems like it's not working properly because when you vote with 1 user the progress remains in 0 as if nobody voted.

Agreed! I did the same test and still I cannot get the progress bar to change. If you hardcode the threshold number it will actually display the threshold percentage in the progress bar but without that nothing works so far. 

As mentioned above, I will open a new issue for the progress bar to investigate further. 

Thank you guys!

willm30:
> ```shell
> curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id":1}' localhost:8545
> curl -X POST --data '{"jsonrpc":"2.0","method":"evm_mine","params":[],"id":1}' localhost:8545
> ```

@rdig I've put some of this info in [Notion](https://www.notion.so/colony/Dev-70f1a0965f8b4070b0c749c5378f5f0c) for my and others' future reference.

src/modules/core/components/ProgressBar/ProgressBar.tsx
Julianahlk: @rdig Changes were made. Can you please review this again? Thanks.

rdig (approved):
They render properly aligned now. Good work. :100: 

Screenshot from 2022-07-15 13-33-37
Screenshot from 2022-07-15 13-34-25

Also, LOL. This are your first and last commits :)

Screenshot from 2022-07-15 13-22-37
Screenshot from 2022-07-15 13-22-42

PR #3592, opened 2022-07-07
Update tooltip positioning for `MembersListItem`

### LINKED ISSUE #3574: Top of Tooltip of Member in Address Book is cut off
## Steps to reproduce

1. Ensure you have a member added to the Address Book
2. Hover over the `check` icon of the member in the address book.

## Expected behaviour

* Top of check's tooltip on a member in the Address Book should not be cut off, and should match the design of the tooltips in the DApp.

## Actual behaviour

* Top of check's tooltip on a member in the Address Book is cut off.

download (43)

chinins:
## Description

This PR fixes the tooltip position that appears when we hover over the greenish checkmark of the verified sign.
Screenshot 2022-07-07 at 15 30 05

resolves #3574

### DIFF
--- a/src/modules/core/components/MembersList/MembersListItem.css
+++ b/src/modules/core/components/MembersList/MembersListItem.css
@@ -15,7 +15,6 @@
 
 .usernameSection {
   composes: section;
-  composes: inlineEllipsis from '~styles/text.css';
   flex-grow: 0.5;
   margin-right: auto;
   max-width: 200px;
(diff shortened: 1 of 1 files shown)


arrenv (commented):
It seems to work, an alternative approach could be to remove the ellipsis on the `usernameSection` and add a width to the tooltio.

I.e. remove this - https://github.com/JoinColony/colonyDapp/blob/77f18cd680de3c458866ad69d3ce5b117d9ba7d9/src/modules/core/components/MembersList/MembersListItem.css#L18

Add `width: 130px` here - https://github.com/JoinColony/colonyDapp/blob/77f18cd680de3c458866ad69d3ce5b117d9ba7d9/src/modules/core/components/MembersList/MembersListItem.css#L78

danbr (approved):
Very good - works for me.

PR #3598, opened 2022-07-08
Add: responsive styles for `/wallet` route

### LINKED ISSUE #3590: Add responsive styles to `/wallet`
## Specification

The `/wallet` route should be mobile responsive. There's no specific design but it should follow the pattern of the other designs in this project.

willm30:
## Description

This PR tracks the work for making the `/wallet` route mobile responsive. Should follow same pattern as rest of project.

**Changes** 🏗

* `ColonyFunding`: moving styles shared with `Wallet` into `TokenCard` and `TokenCardList`, both of which are only used in these two components. 
* `Wallet`: adding mobile styles
* Routes: adding layout on mobile for wallet route

## Screenshots

wallet

Resolves #3590


Julianahlk (approved):
Good work Will! Everything looks good from my end. 

Screen Shot 2022-07-12 at 8 45 53 PM

PR #3602, opened 2022-07-08
Fix screen jumping on members page when changing filters

### LINKED ISSUE #3593: Bug: Members List "jumping" when filtering Watchers
## Steps to reproduce

1. I detected this on branch: feature/3381-add-filters-to-members-page
2. Ensure you have one contributor and one watcher. 
3. Verify the contributor, but not the watcher.
4. Filter watchers, and the list shrinks slightly. 

jumpy-members-list

Following the set-up exactly, e.g. verifying the contributor but not the watcher, is important as I've only been able to reproduce it with this exact set-up.

## Expected behaviour

* List doesn't shrink

## Actual behaviour

* List shrinks

## Suggestions

* Could it be the absence of the blue tick is causing the list to shrink?

chinins:
## Description

This PR fixes the bug with jumping content on the Members page on smaller screens.

To reproduce the initial bug you need to have a relatively small screen, have a contributor verified and then select watchers in the filters section. The contents (title and the watchers list) will jump to the right. This shouldn't happen with my fix.

**Changes** 🏗

- update min-width value in 2 components

resolves #3593


arrenv (approved):
Well no jumping is happening, although, I was still not able to replicate for some reason.

ArmandoGraterol (approved):
Looking good! I had to change the screen size to a smaller one in order to replicate the bug

danbr (approved):
LGTM

willm30 (approved):
Fixed for me

PR #3617, opened 2022-07-12
Fix: Updated logic for funds links

### LINKED ISSUE #3550: Bug: Links are Inactive on Funds Page
## Steps to reproduce

1. Make sure you have a colony created and are connected to a wallet as normal. Make sure the `governance` extension is enabled.
2. Signout and join the same colony with a different wallet.
Bug1: The "Available Funds" heading in the right aside should link to the funds page, but is inactive.
3. Go to funds page (click "manage funds")
Bug2: The three links in the right aside should all be active.

## Expected behaviour

1. Available funds link on home page should be active
2. "Move", "mint" and "manage" links should all be active 

## Actual behaviour

1. Link is inactive
2. Only the "mint" link is active

## Screenshot

**_Bug 1 (highlighting is mine)_**
available-funds-link

**_Bug 2_**
image

Julianahlk:
## Description
This PR fixes the "Available Funds" link that should take the user to the funds page and be enabled regardless of permissions but it is currently inactive.

Additionally, once you go inside the funds page, the Move funds, Mint Tokens, Manage Tokens links should all be active depending on the following:

1. If Governance Extension is enabled:
- Move Funds - visible
- Mint Tokens - visible
- Manage Tokens - visible

2. If Governance Extension is NOT enabled:
- Move Funds - needs the Funding role
- Mint Tokens - needs the Root role
- Manage Tokens - needs the Root role 

**Changes** 🏗
* Removed permissions requirement for available funds link.
Screen Shot 2022-07-12 at 11 42 40 PM

* Made the Move funds, Mint Tokens, Manage Tokens links enabled when the governance extension is active. 
Screen Shot 2022-07-12 at 11 39 26 PM

## TODO
- [x] Activate the "Available Funds" link.
- [x] Make the Move funds, Mint Tokens, Manage Tokens links enabled when the governance extension is active.

Resolves #3550

### DIFF
--- a/src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
+++ b/src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
@@ -5,11 +5,7 @@ import { SpinnerLoader } from '~core/Preloaders';
 import Heading from '~core/Heading';
 import InfoPopover from '~core/InfoPopover';
 import NavLink from '~core/NavLink';
-import {
-  useLoggedInUser,
-  Colony,
-  useTokenBalancesForDomainsQuery,
-} from '~data/index';
+import { Colony, useTokenBalancesForDomainsQuery } from '~data/index';
 
 import TokenItem from './TokenItem';
 
@@ -30,10 +26,6 @@ interface Props {
 const displayName = 'dashboard.ColonyHome.ColonyFunding';
 
 const ColonyFunding = ({ colony, currentDomainId }: Props) => {
-  const { ethereal, username } = useLoggedInUser();
-
-  const canMoveFunds = !!username && !ethereal;
-
   const {
     colonyAddress,
     tokens: colonyTokens,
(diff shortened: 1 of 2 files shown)


src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
```diff
@@ -33,11 +30,9 @@ interface Props {
...
-  const { walletAddress, ethereal, username } = useLoggedInUser();
+  const { ethereal, username } = useLoggedInUser();
 
-  const allUserRoles = useTransformer(getAllUserRoles, [colony, walletAddress]);
-
-  const canMoveFunds = !!username && !ethereal && canFund(allUserRoles);
+  const canMoveFunds = !!username && !ethereal;
 
   const {
     colonyAddress,
```
willm30, line 38: Line 60 below `canMoveFunds ? ...` is the logic that needs to be removed.

willm30 (changes requested):
Part 2 looks good. Part 1, I the 'available funds' link is still disabled if you're not a member of the colony. I have commented where I believe the relevant logic is.

rdig (changes requested):
Agree with Will, the link disabling / enabling works fine, however the like to the Manage Funds page does not.

Here's with a logged out user _(it should work for that case as well)_
Screenshot from 2022-07-13 21-44-39

Note that the incoming funds section works correctly, and even point to the same location. So you can just copy/paste it's logic

While after being logged in, it works correctly

Screenshot from 2022-07-13 21-45-00

Julianahlk:
> Here's with a logged out user (it should work for that case as well)
I didn't know the links should also work for this case when the user is logged out. I will make the changes. Thank you for the help!

src/modules/dashboard/components/ColonyHome/ColonyFunding/ColonyFunding.tsx
Julianahlk: Got it! Thanks Will.

willm30 (approved):
Looking good!

chinins (approved):
Good to go!

PR #3627, opened 2022-07-14
Introduce `ExtendedJsonRpcProvider` proper caching bypass

rdig:
This PR refactors and adds provider caching bypass for more calls, especially the ones that might cause a race condition _(`addr` when creating a new colony)_

Besides the above, I've also refactored the way this is set up, so that we don't rely on hashed signatures anymore, but hash them at compile time, meaning we can use the human readable format.


willm30 (approved):
Works for me too 👍

PR #3630, opened 2022-07-14
Ensure Gas Prices is at least 2 GWEI when on Gnosis Chain

rdig:
Exactly as it says on the tin, when on Gnosis Chain, if the gas estimate from the Oracle will be less than 2, set it to 2, otherwise, use the value from the Oracle.

This is to counteract transactions failing to be sent out, if the Gas paid was lower.

Granted this should't be that much of a problem anymore, with the advent of Metatransactions, but since we've been running this code in production since before the Token Sale, might as well merge it in.

### DIFF
--- a/src/modules/core/sagas/utils/getGasPrices.ts
+++ b/src/modules/core/sagas/utils/getGasPrices.ts
@@ -97,6 +97,23 @@ const fetchGasPrices = async (
       // API prices are in Gwei, so they need to be normalised
       const oneGwei = bigNumberify(10 ** 9);
 
+      let { average } = data;
+      let { fast } = data;
+      /*
+       * @NOTE IF we're on Gnosis Chain, ensure that transactions alway pay at
+       * least 2gwei for gas
+       *
+       * This to counteract some unpleasantness coming from the Blockscout gas oracle
+       *
+       * Locally this is not a problem (and we don't even use the oracle to estimate)
+       */
+      if (
+        DEFAULT_NETWORK === Network.Xdai ||
+        DEFAULT_NETWORK === Network.XdaiFork
+      ) {
+        average = data.average > 2 ? data.average : 2;
+        fast = data.fast > 2 ? data.fast : 2;
+      }
       /*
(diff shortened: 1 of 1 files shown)


src/modules/core/sagas/utils/getGasPrices.ts
```diff
@@ -97,18 +97,33 @@ const fetchGasPrices = async (
       // API prices are in Gwei, so they need to be normalised
       const oneGwei = bigNumberify(10 ** 9);
 
+      let { average } = data;
+      let { fast } = data;
+      /*
```
willm30, line 102: This could be written as
``` js
let { average, fast } = data;
```

src/modules/core/sagas/utils/getGasPrices.ts
```diff
@@ -97,18 +97,33 @@ const fetchGasPrices = async (
...
+       * This to counteract some unpleasantness coming from the Blockscout gas oracle
+       *
+       * Locally this is not a problem (and we don't even use the oracle to estimate)
+       */
+      if (
+        DEFAULT_NETWORK === Network.Xdai ||
+        DEFAULT_NETWORK === Network.XdaiFork
+      ) {
+        average = data.average > 2 ? data.average : 2;
+        fast = data.fast > 2 ? data.fast : 2;
```
willm30, line 115: Could be: 

``` js
average = average > 2 ? average : 2;
fast = fast > 2 ? fast : 2;
```

willm30 (approved):
Looks good to me. Noted a couple of character-saving suggestions.

src/modules/core/sagas/utils/getGasPrices.ts
rdig: I thought about this change, but I wanted to emphasize that the values are coming from the oracle _(`data`)_.

So I'll leave this as it is.

ArmandoGraterol (approved):

chinins (approved):
Beautiful!

PR #3652, opened 2022-07-18
Add: actions menus mobile

### LINKED ISSUE #3651: Style actions menus on mobile
## Specification

When creating actions on mobile, e.g. smite reputation or mint tokens, the menus should be responsive. 

### Design 

No design but should follow same pattern as rest of project. 

- [ ] Every actions menu is mobile responsive

willm30:
## Description

This PR covers the work for making the actions menus (e.g. mint tokens or smite reputation) responsive on mobile. To test, go to home page, click `New Action`, and go through every possible option in the menus and ensure all behave on mobile.

**Changes** 🏗

* Styling all the components connected to the Actions menus for mobile

## Screenshot

**_The menus I'm talking about_**

action-menus

Resolves #3651

### DIFF
--- a/src/modules/core/components/SingleUserPicker/SingleUserPicker.css
+++ b/src/modules/core/components/SingleUserPicker/SingleUserPicker.css
@@ -1,7 +1,7 @@
 @value query700 from '~styles/queries.css';
 
 .main {
-  composes: field from '~styles/fields.css';
+  composes: field from "~styles/fields.css";
   width: 100%;
 }
 
@@ -169,7 +169,10 @@
   }
 
   .arrowIcon {
+    display: flex;
+    padding: 11px 0;
     position: static;
+    border-bottom: 1px solid var(--temp-grey-blue-7);
   }
 
   .widthWide input {
@@ -183,5 +186,15 @@
   .input:placeholder-shown {
     overflow: hidden;
     text-overflow: ellipsis;
+  }
+
+  .arrowIconActive {
+    border-bottom-color: var(--primary);
+  }
+
+  .baseInput,
+  .omniPickerContainer,
+  .recipientName {
+    max-width: none;
   }
 }
--- a/src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.css
+++ b/src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.css
@@ -1,4 +1,5 @@
 @value wideButton: 160px;
+@value query700 from '~styles/queries.css';
 
 .inputContainer {
   margin-bottom: 20px;
@@ -107,4 +108,53 @@
   height: 1px;
   width: 100%;
   background-color: color-mod(var(--temp-grey-blue-7) alpha(15%));
+}
+
+@media screen and query700 {
+  .percentageSign {
+    padding-top: 45px;
+    position: static;
+  }
+
+  .inputText {
+    margin-top: 0;
+    padding-right: 28px;
+    float: none;
+    text-align: right;
+  }
+
+  .inputContainer {
+    margin-bottom: -10px; /* Necessary (because of the min-height) to reduce distance from next section */
+    min-height: 127px;
+    width: auto;
+    column-gap: 6px;
+  }
+
+  .inputContainer > div {
+    display: flex;
+    column-gap: 5px;
+  }
+
+  .inputContainer div > p {
+    justify-content: flex-end;
+    text-align: right;
+  }
+
+  .inputContainer > div > div:first-child {
+    flex: 1;
+  }
+
+  .inputContainer label {
+    margin-top: 0px;
+    position: static;
+  }
+
+  .inputContainer label span {
+    display: inline-block;
+    min-height: 34px;
+  }
+
+  .inputContainer input {
+    padding-right: 0px;
+  }
 }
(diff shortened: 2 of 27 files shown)


src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.css
```diff
@@ -108,3 +109,50 @@
...
+
+  .inputContainer > div:first-child {
+    flex: 1 0 80%;
+    min-height: 95px;
+  }
+
+  .inputContainer > div:nth-child(2) {
+    flex: 1 0 10%;
+    align-self: flex-end;
+    transform: translateY(-26px);
```
Julianahlk, line 156: Correct me if wrong but I think the line above messed up the alignment between the error message and the Max reputation msg?

Screen Shot 2022-07-19 at 10 53 30 AM

Julianahlk (approved):
Tested all the actions. Looking good so far except for the comment I left in the code.

Also, curios about the positioning of the create motion below, will this be also fixed with some other PR?
Screen Shot 2022-07-19 at 10 41 55 AM

I see the network fee message in the above screenshot is also a bit far off. Can you please do this fix here? I think a `bottom: 25px;` will do. ☺️

willm30:
@Julianahlk Thanks, good spots. Have fixed 🙂

Julianahlk:
> @Julianahlk Thanks, good spots. Have fixed 🙂

The Confirm/Create Motion buttons are looking good now in all situations. 👍

I found a new issue with the reputation points in Award Reputation and Smite Reputation points. The points message for both cases look quite far from the input field. You can reproduce this error the very first time you open Award reputation/Smite Reputation, then without clicking anywhere type a ( - ) and after that ( -0 ), then you will see the reputation points message is still far from the input field. 

reputation

Screen Shot 2022-07-19 at 8 18 57 PM
Screen Shot 2022-07-19 at 8 19 24 PM

However, if you click anywhere outside the input field you will see that the gaps are intended/needed for the error messages.
Screen Shot 2022-07-19 at 8 19 50 PM
Screen Shot 2022-07-19 at 8 45 46 PM

I worked on an issue like this before that fixed the Network fee with a similar gap. I think one solution could be not to show the rep points message until the user inputs 0 or something > 0 because I think it will be tricky to fix the gap since there is supposed to be an error message in there.

ArmandoGraterol (changes requested):
Nice work! I found these instances where we could do some changes:

Caret is outside of the input when in the desktop version is inside.
FireShot Capture 668 - Actions - Colony - wonker - localhost

The token symbol in the mint tokens dialog needs to be positioned lower.
FireShot Capture 669 - Actions - Colony - wonker - localhost

willm30:
@ArmandoGraterol Fixed 🙂
@Julianahlk Also fixed 🙂 I've set validation style back to the way it does it desktop. (Changed it initially for styling reasons but no longer necessary and it resolves this problem automatically)

Julianahlk:
> @ArmandoGraterol Fixed 🙂 @Julianahlk Also fixed 🙂 I've set validation style back to the way it does it desktop. (Changed it initially for styling reasons but no longer necessary and it resolves this problem automatically)

I see the changes for the validation. Much better for sure 👍. 

Now that Armando points it out, I see the Award and Smite pts also look different in desktop version than in mobile. They are outside the input field (see below). Is this intended for the mobile version?

Desktop
Screen Shot 2022-07-20 at 4 11 03 PM

Mobile
Screen Shot 2022-07-20 at 4 19 52 PM
Screen Shot 2022-07-20 at 4 20 15 PM

Also, I can you please add a few more pixels (2-3) to the gap between the Award/Smite points message and the input field. It  looks a bit too close. 
Screen Shot 2022-07-20 at 4 05 14 PM

Thank you!

src/modules/core/components/SingleUserPicker/SingleUserPicker.css
```diff
@@ -184,4 +184,14 @@
...
+
+  .container > i {
+    display: flex;
+    padding: 11px 0;
+    border-bottom: 1px solid var(--temp-grey-blue-7);
+  }
+
+  .arrowIconActive {
+    border-bottom-color: var(--primary) !important;
+  }
```
ArmandoGraterol, line 196: This feels hacky to me, could we not do something like this?
Screenshot from 2022-07-20 15-50-00

This way we don't need the borders (and more importantly, that `!important`) and the static positioning

willm30: If you switch your dev tools to responsive mode, and change the width of the viewport, you'll notice that absolute positioning does not respond well (at all) to changes in the viewport width. This means that `left: 299px` works only on one viewport width, whereas currently the positioning responds to all screen sizes up to the breakpoint.

willm30:
@Julianahlk Changes made 👍

src/modules/core/components/SingleUserPicker/SingleUserPicker.css
willm30: However, I agree about the `important!`. There are often times in which it's unnecessary, and this is one of them. The specificity issue had a simple fix, now implemented, of using the correct class name to select the arrow.

ArmandoGraterol: Oh, I see what you are saying. This may indicate that perhaps something else in the styles of the core component needs to change because I feel like this should be more straightforward than how it is right now. 

Incidentally, I'm currently reworking this component so I may have the chance to update the styles so that this may be simpler :)

For now, I feel like this is ok :+1: We could revisit this once the component is updated

ArmandoGraterol (approved):
Great work! :ok_hand:

PR #3655, opened 2022-07-19
Fix Yet More Transactions Message Descriptors

rdig:
This PR fixes three more transactions (and metatransactions) message descriptors which were broken after I've merged #2751

First one, `WrappedToken` and `VestingSimple` was because I've renamed the clients _(to be in line with the expected client naming scheme. Ie: append `Client` an the end)_

Second, `createUser`, was because of batch rename, which didn't account for the work `transaction` happening a second time in the same message descriptor id.


chinins (approved):
Good to go!

danbr (approved):
Makes sense..... LGTM

Issue #3564, opened 2022-06-30
Add "Multiple transaction" preview to Gnosis Control Safe Dialog


ArmandoGraterol:
### Story

As a user, when I'm creating multiple transactions, I want to see the multiple transaction version of the normal preview screen.

### Description

The multiple transaction version of the preview dialog should show each transaction in its respective tab, each tab has a heading with the transaction number and type and it can be opened or closed.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=139%3A9825)

![FireShot Capture 645 - Safe Control – Figma - www figma com](https://user-images.githubusercontent.com/18473896/176565496-ab83d352-aec7-4431-b02f-0b18447e535e.png)

### This issue is complete when...

- [ ] Creating multiple transactions at the same time takes you to the "multiple transaction" version of the preview dialog.
- [ ] Each tab has a heading with the transaction number and type.
- [ ] Each tab can be closed or opened.
- [ ] Each tab shows the info previously added relevant to the transaction.

PR #3656, opened 2022-07-19
Add gnosis transaction preview dialog

### LINKED ISSUE #3543: Add preview dialog for transactions of `GnosisSafeControlDialog`
### Story

As a user creating any transaction from the Gnosis control dialog, I want to see a preview of the transaction before sending it.

### Description

When the user chooses a type, and fills the form of the Gnosis control dialog, after confirming the form and before sending the transaction, they should see a preview of what transaction they are going to send. It should show the relevant information previously filled as a summary and also fields to provide a title for the transaction, plus an optional annotation.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=139%3A20975)
FireShot Capture 638 - Safe Control – Figma - www figma com

- [ ] You click on the submit button of the previous Gnosis Safe Control dialog screen and it takes you to the preview.
- [ ] The information relevant to the transaction type (Transfer funds, transfer NFT, etc) is shown in the preview in a list similar to the way we display info of a action in the `DetailsWidget` of the `ActionsPage`.
- [ ] The preview dialog shows an input where you have to type a title for the transaction/group of transactions
- [ ] The preview dialog shows a textarea where you can, if you want to, include an annotation.
- [ ] The form has relevant validation. If the form is not valid then it should not let the user confirm.

### LINKED ISSUE #3664: Validation in Gnosis Safe control dialog is not working when dialog first opened
When you first open the dialog the create transaction button is enabled when it shouldn't be
Screenshot 2022-07-21 at 17 58 55

chinins:
## Description

This PR adds transaction preview logic for the gnosis safe control dialog. I've already added preview for transfer funds & for raw transaction. NFT & arbitrary transactions are not finished so there is a separate issue to add them when they are.

I've also added a fix for another issue - validate on load.

I've also added `TransactionTypes` enum across the dialog.
Screenshot 2022-07-22 at 13 37 27
Screenshot 2022-07-22 at 13 37 56

Resolves #3543 
Resolves #3664

### DIFF
--- a/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
+++ b/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
@@ -84,6 +84,12 @@ const transactionTypeFieldsMap = {
     },
   ],
   [TransactionTypes.TRANSFER_NFT]: [
+    // To be finished once NFT part of the form is merged
+    {
+      key: 'function',
+      label: MSG.function,
+      value: () => <FormattedMessage {...ConstantsMSG.transferNft} />,
+    },
     {
       key: 'nft',
       label: MSG.nft,
@@ -91,21 +97,7 @@ const transactionTypeFieldsMap = {
     },
   ],
   [TransactionTypes.CONTRACT_INTERACTION]: [
-    {
-      key: 'contract',
-      label: MSG.contract,
-      value: (contract) => contract,
-    },
-    {
-      key: 'abi',
-      label: MSG.abi,
-      value: (abi) => abi,
-    },
-    {
-      key: 'contractFunction',
-      label: MSG.contractFunction,
-      value: (contractFunction) => contractFunction,
-    },
+    // To be finished once NFT part of the form is merged
   ],
   [TransactionTypes.RAW_TRANSACTION]: [
     {
@@ -169,14 +161,6 @@ const SafeTransactionPreview = ({ colony: { tokens }, values }: Props) => {
             <AddressDetailsView
               item={(values.safe as never) as AnyUser}
               isSafeItem
-            />
-          }
-        />
-        <DetailsItem
-          label={MSG.function}
-          value={
-            <FormattedMessage
-              {...ConstantsMSG[values.transactions[0].transactionType]}
             />
           }
         />
--- a/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
+++ b/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
@@ -190,6 +190,7 @@ const SafeTransactionPreview = ({ colony: { tokens }, values }: Props) => {
           label={MSG.transactionsSetTitle}
           name="transactionSetTitle"
           disabled={false}
+          placeholder=""
         />
       </DialogSection>
       <DialogSection>
--- a/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
+++ b/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
@@ -1,0 +1,312 @@
+import React, { useMemo } from 'react';
+import { defineMessages, FormattedMessage } from 'react-intl';
+
+import { DialogSection } from '~core/Dialog';
+import { Annotations, Input } from '~core/Fields';
+import Heading from '~core/Heading';
+import Numeral from '~core/Numeral';
+import TokenIcon from '~dashboard/HookedTokenIcon';
+
+import { Colony, AnyUser } from '~data/index';
+
+import AddressDetailsView from './TransactionPreview/AddressDetailsView';
+
+import { FormValues } from './GnosisControlSafeDialog';
+
+import { TransactionTypes } from './constants';
+import styles from './GnosisControlSafeForm.css';
+
+const MSG = defineMessages({
+  title: {
+    id: 'dashboard.GnosisControlSafeDialog.SafeTransactionPreview.title',
+    defaultMessage: 'Control Safe',
+  },
+  description: {
+    id: 'dashboard.GnosisControlSafeDialog.SafeTransactionPreview.description',
+    defaultMessage: `You can use Control Safe to interact with other third party smart contracts. Be careful. <a>Learn more</a>`,
+  },
+  selectSafe: {
+    id: 'dashboard.GnosisControlSafeDialog.SafeTransactionPreview.selectSafe',
+    defaultMessage: 'Select Safe',
+  },
+  safePickerPlaceholder: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.safePickerPlaceholder`,
+    defaultMessage: 'Select Safe to control',
+  },
+  transactionLabel: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.transactionLabel`,
+    defaultMessage: 'Select transaction type',
+  },
+  transactionPlaceholder: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.transactionPlaceholder`,
+    defaultMessage: 'Select transaction',
+  },
+  buttonTransaction: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.buttonTransaction`,
+    defaultMessage: 'Add another transaction',
+  },
+  buttonCreateTransaction: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.buttonCreateTransaction`,
+    defaultMessage: 'Create transaction',
+  },
+  transactionTitle: {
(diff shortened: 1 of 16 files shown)


ArmandoGraterol (changes requested):
Great work! Just a couple of details/questions and this should be good to go from my side.

1. In the specs now the "Target contract" field is called "Safe".
FireShot Capture 674 - Safe Control – Figma - www figma com

2. Also it seems like the values of the things like `Data` or `Value (wei)` should be bolder. In the specs they are `600`, which looks too different from the `400` we have implemented.

3. "Title of the set of all the arbitrary transactions" should now be simply "Title"

4. "Explain why you’re making arbitrary transaction (optional)" is now "Description" which I'm gonna assume now that it's required because it doesn't have the "(optional)"

5. The field values should be aligned to the right, as per the screenshot above of the specs.

6. If I focus on the title field, then focus out and go back to the previous screen then the "Create transaction" button shows as disabled even though the current form is completed.

7. In the case of the "Transfer funds" transaction form, when I go back the amount field is empty and I have to re-enter the amount.

8. This is more of an idea/suggestion, what do you think of adding `appareance={{ theme: 'sidepadding' }}` to the section with the transaction title? In my opinion it seems like it makes it closer to the specs (in the specs there's not that much space between the title and description).

willm30 (commented):
Correct me if I'm wrong but this now needs to be updated to reflect the added ability to add multiple transactions? I just added 3 but only received a preview for the first one.

willm30:
> Correct me if I'm wrong but this now needs to be updated to reflect the added ability to add multiple transactions? I just added 3 but only received a preview for the first one.

Ah, I see this is a [separate issue ](https://github.com/JoinColony/colonyDapp/issues/3564)👍

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -0,0 +1,205 @@
...
+    id: 'dashboard.GnosisControlSafeDialog.SafeTransactionPreview.previewTitle',
+    defaultMessage: 'Confirm transaction details',
+  },
+  transactionsSetTitle: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.transactionsSetTitle`,
+    defaultMessage: 'Title of the set of all the arbitrary transactions',
+  },
+  explainWhy: {
+    id: 'dashboard.GnosisControlSafeDialog.SafeTransactionPreview.explainWhy',
+    defaultMessage: `Explain why you’re making arbitrary transaction (optional)`,
```
Julianahlk: I think it would sound better if we add the word "the": "Explain why you’re making the arbitrary transaction (optional)"

chinins:
> Great work! Just a couple of details/questions and this should be good to go from my side.
> 
> 1. In the specs now the "Target contract" field is called "Safe".

> 2. Also it seems like the values of the things like `Data` or `Value (wei)` should be bolder. In the specs they are `600`, which looks too different from the `400` we have implemented.
> 3. "Title of the set of all the arbitrary transactions" should now be simply "Title"
> 4. "Explain why you’re making arbitrary transaction (optional)" is now "Description" which I'm gonna assume now that it's required because it doesn't have the "(optional)"
> 5. The field values should be aligned to the right, as per the screenshot above of the specs.
> 6. If I focus on the title field, then focus out and go back to the previous screen then the "Create transaction" button shows as disabled even though the current form is completed.
> 7. In the case of the "Transfer funds" transaction form, when I go back the amount field is empty and I have to re-enter the amount.
> 8. This is more of an idea/suggestion, what do you think of adding `appareance={{ theme: 'sidepadding' }}` to the section with the transaction title? In my opinion it seems like it makes it closer to the specs (in the specs there's not that much space between the title and description).

@ArmandoGraterol, thank you for the feedback Please, see my comments:
1, 3 & 4 - updated (didn't notice the spec changed before). Descriptions should still be optional (saying it in brackets) - confirmed with Arren
2 - in the figma it says font-weight: 400 but I asked Arren & confirmed that we can do 700 - updated.
5 - on my screen it's all aligned to the right like in the specs. Can you please check and maybe offer a fix for your system (can't fix myself as I can't replicate)
Screenshot 2022-07-26 at 11 59 49

7 -  good catch. Fixed now although it take me a while 🙈
8 - talked to Arren and he said use margin above the Title field but not between  title and description - updated.

6 - OMG, this is a nightmare bug! I understand why it's happening but I haven't been able to fix it. Kept trying for almost the whole day. It validates it based on the old validation form although it's already update and it triggers re-validation only after you focus & unfocus a field. I tried all of the Formik validation props & functions but unsuccessfully. I will try again tomorrow... Any suggestions are welcome.

Julianahlk:
I noticed that the avatar does not have a border radius.
Screen Shot 2022-07-26 at 5 25 23 PM

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
Julianahlk: Never mind, I was going through your code and while I was reviewing it you already made the changes.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -0,0 +1,205 @@
...
+import styles from './GnosisControlSafeForm.css';
+
+const MSG = defineMessages({
+  previewTitle: {
+    id: 'dashboard.GnosisControlSafeDialog.SafeTransactionPreview.previewTitle',
+    defaultMessage: 'Confirm transaction details',
+  },
+  transactionsSetTitle: {
+    id: `dashboard.GnosisControlSafeDialog.SafeTransactionPreview.transactionsSetTitle`,
+    defaultMessage: 'Title of the set of all the arbitrary transactions',
```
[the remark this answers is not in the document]
Julianahlk: Never mind, I was going through your code and while I was reviewing it you already made the changes.

ArmandoGraterol:
> 1, 3 & 4 - updated (didn't notice the spec changed before). Descriptions should still be optional (saying it in brackets) - confirmed with Arren

Thank you for confirming that!

> 5 - on my screen it's all aligned to the right like in the specs. Can you please check and maybe offer a fix for your system (can't fix myself as I can't replicate) Screenshot 2022-07-26 at 11 59 49

My bad, I think I didn't explain myself well, I'm talking about the inner alignment of the values:
FireShot Capture 678 - Actions - Colony - wonker - localhost

Data's value is internally aligned to the left when it has multiple lines as you can see from my screenshot, while in the specs even internally they are aligned to the right:

180580112-0ec77df0-8f20-4b99-9d5d-5394501a83a3


> 6 - OMG, this is a nightmare bug! I understand why it's happening but I haven't been able to fix it. Kept trying for almost the whole day. It validates it based on the old validation form although it's already update and it triggers re-validation only after you focus & unfocus a field. I tried all of the Formik validation props & functions but unsuccessfully. I will try again tomorrow... Any suggestions are welcome.

I'll also take a stab at it to see if I can come up with some suggestions :+1:

ArmandoGraterol:
@chinins Suggestion time! This worked for me:

1. Make the initial value of `showPreview` `null`: `const [showPreview, setShowPreview] = useState<boolean | null>(null);`.
2. Add this effect to the `GnosisControlSafeForm` component:
```
  useEffect(() => {
    if (!isNil(showPreview) && !showPreview) {
      validateForm();
    } 
  }, [showPreview, validateForm]);
```

`validateForm` comes from the `FormikProps`.

This way I have been able to re-run the validation when going back to the previous dialog and get rid of the preview error :) Perhaps there's a better way but at least it's a start.

chinins:
> Data's value is internally aligned to the left when it has multiple lines as you can see from my screenshot, while in the specs even internally they are aligned to the right:

I see now - when the data has spaces! Good catch - fixed now.

chinins:
> @chinins Suggestion time! This worked for me:
> 
> 1. Make the initial value of `showPreview` `null`: `const [showPreview, setShowPreview] = useState<boolean | null>(null);`.
> 2. Add this effect to the `GnosisControlSafeForm` component:
> 
> ```
>   useEffect(() => {
>     if (!isNil(showPreview) && !showPreview) {
>       validateForm();
>     } 
>   }, [showPreview, validateForm]);
> ```
> 
> `validateForm` comes from the `FormikProps`.
> 
> This way I have been able to re-run the validation when going back to the previous dialog and get rid of the preview error :) Perhaps there's a better way but at least it's a start.

I think it's a very good solution. Yesterday everything I tried was in the Dialog component (even the validate).
I've added the fix. It seems that the null & isNil check is not even necessary. Working without it.

Thank you for the idea!

willm30 (commented):
Looking good. One question: should there be validation on the title field? If you enter a single space it will let you submit the form without a title.

ArmandoGraterol (approved):
Looks good to me! Great job. Just take into consideration what Will mentioned about the title

chinins:
@willm30 @ArmandoGraterol it's the same validation as we have everywhere else in the dapp - empty strings are accepted.

PR #3659, opened 2022-07-20
Add split equal

### LINKED ISSUE #3635: Expenditures: As a user I can use split funds equal
[Provided by Adrian via Contractors project management platform]

## Description
User has to be able to use split equal option. User can set a max amount of funds, change currency, add/delete and select recipients. After adding recipients user can display a recipients counter and share for each of them.

## Acceptance Criteria

- an option to set max amount of funds
- an option to change currency
- an option to add/delete recipients
- an option to select a recipient
- an option to display recipients counter
- an option to display a share of each recipient

## Design

- [Figma Link - Equal Distribution](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=3807%3A285937)
- [Figma Link - Equal Distribution with selections](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=3802%3A277457)

ace4021ed01dee92b9f7cbcd39385e0b

ace4ce8cce344d39d76085b56539f238

agnieszkajarosikloj:
## Description

This PR adds split equal option. User can set a max amount of funds, change currency, add/delete and select recipients. After adding recipients there is displayed recipients counter and share for each of them.

[Figma Link - Equal Distribution](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=3807%3A285937)
[Figma Link - Equal Distribution with selections](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=3802%3A277457)

**New stuff** ✨

* new part in form after selecting equal split
Screenshot 2022-07-20 at 11 52 50

* yup validation for new fields

Resolves  #3635


willm30 (changes requested):
@agnieszkajarosikloj Made a few notes 🙂
1. There's a duplicate key error when adding recipients.
duplicate-k

1. I don't think you should be able to add the same recipient more than once, and similarly, I don't think you should be able to submit the form if there's only one recipient added, as both are counterintuitive for a split payment function. Is that right?

1. ~~When  you put in a ridiculously large number it overflows the "max" label. Also, if you then click the "max" label it resets the value in the input to 0 and flashes a warning that the value should be greater than 0. (Unsure if this latter functionality is intentional but if so isn't intuitive to me).~~ I see this is a problem with the component not your branch 🙂

willm30 (commented):
@agnieszkajarosikloj I see there's no longer the duplicate key error. 👍

However, the user dropdown picker is off the screen for me when selecting a second user:

off-screen-popover

willm30:
@agnieszkajarosikloj I've made a gif to help illustrate what's happening on my machine. I have four members, and the last one is being left off screen even when scrolled to the bottom of the dropdown and the aside.  

dropdown-offscreen

willm30:
@agnieszkajarosikloj Dropdown menu height is fixed! I am however noticing a problem with the recipients list. Adding unique recipients works fine, but adding the same person twice causes some funny behaviour (and generates a duplicate key error in the console). 

In the gif below, I am clicking the delete icon next to the first "will" user, which is resulting is all non-will users being deleted, additional "wills" being added to the bottom and the recipients count dropping to zero.
  
multiple-recipients

willm30:
@agnieszkajarosikloj Fixed the problem. Just one comment / question:

When you delete the last remaining recipient you get what feels like a duplicate message (0 recipients and no recipients). 
split-view-no

Maybe the latter should be hidden instead?
split1

agnieszkajarosikloj:
@willm30  you're right. I've hidden this section when there are 0 recipients. Thank you.

willm30 (approved):
UI looking nice 👍

Julianahlk:
UI looks good! Just a few things:

1. The the state of the Max input field is not being saved when I toggle from Equal to Unequal. I find this odd because everything else is being saved except for that. 
toogle-equal

2. I think I mentioned this before but I think we shouldn't be able to add users that do not exist:
Screen Shot 2022-08-03 at 5 35 47 PM
Screen Shot 2022-08-03 at 5 35 55 PM

Especially because these recipients are being counted towards the Split Equal count:
recipients

Julianahlk (approved):

src/modules/core/components/UserPickerWithSearch/UserPickerWithSearch.tsx
```diff
@@ -161,6 +162,11 @@ const UserPickerWithSearch = ({
       ? placeholder
       : formatMessage(placeholder);
 
+  const dropdownHeight = useMemo(() => {
+    const height = data.length * 56;
+    return height > 180 ? 180 : height;
```
chinins, line 167: Can you please declare `56` and `180` as variables (you can do it outside of the component) and use variables? Not great to have just numbers.
Please, use capital letters for the variables, e.g. ELEMENT_HEIGHT

src/modules/dashboard/components/ExpenditurePage/Recipient/LockedRecipient.tsx
```diff
@@ -4,15 +4,15 @@ import classNames from 'classnames';
...
+import { Colony } from '~data/index';
 import TokenIcon from '~dashboard/HookedTokenIcon';
 
 import styles from './LockedRecipient.css';
-import { getTokenDecimalsWithFallback } from '~utils/tokens';
-import Numeral from '~core/Numeral';
-import { Colony } from '~data/index';
 import { getRecipientTokens } from '../utils';
 import { MSG } from './Recipient';
+import { Recipient as RecipientType } from '../Payments/types';
```
chinins, line 15: Please, group and separate `../` & `./` imports (as per earlier guidelines)

src/modules/dashboard/components/ExpenditurePage/Recipient/LockedRecipient.tsx
```diff
@@ -4,15 +4,15 @@ import classNames from 'classnames';
...
 
 const displayNameLockedRecipient = 'dashboard.ExpenditurePage.LockedRecipient';
```
chinins, line 17: Should just be `displayName` and you are missing `Recipient` before `LockedRecipient`

src/modules/dashboard/components/ExpenditurePage/Split/SplitEqual/SplitEqual.tsx
```diff
@@ -0,0 +1,221 @@
+import { FieldArray, useField, useFormikContext } from 'formik';
+import React, { useEffect, useMemo } from 'react';
+import { defineMessages, FormattedMessage, useIntl } from 'react-intl';
+
+import { nanoid } from 'nanoid';
```
chinins, line 5: Belongs to the upper group

src/modules/dashboard/components/ExpenditurePage/Split/SplitEqual/SplitEqual.tsx
```diff
@@ -0,0 +1,221 @@
...
+import Numeral from '~core/Numeral';
+import { getTokenDecimalsWithFallback } from '~utils/tokens';
+import UserPickerWithSearch from '~core/UserPickerWithSearch';
+import { filterUserSelection } from '~core/SingleUserPicker';
+import { supRenderAvatar } from '~dashboard/ExpenditurePage/Recipient/Recipient';
+import Icon from '~core/Icon';
+import Button from '~core/Button';
+
+import styles from './SplitEqual.css';
+import { initalRecipient } from '../constants';
```
chinins, line 19: Should be before styles import & separated by an empty line

src/modules/dashboard/components/ExpenditurePage/Split/SplitEqual/SplitEqual.tsx
```diff
@@ -0,0 +1,221 @@
...
+  }, [amount, recipientsCount]);
+
+  useEffect(() => {
+    setFieldValue(
+      'split.recipients',
+      recipients.map((recipient) => {
+        return { ...recipient, amount: calculatedAmount };
+      }),
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
chinins, line 97: Why are `recipients` not included in the dependency array? If there is a specific reason, please, add a comment explaining

src/modules/pages/components/ExpenditurePage/LockedSidebar.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+interface Props {
+  colony: Colony;
+  formValues?: ValuesType;
+}
+
+const LockedSidebar = ({ colony, formValues }: Props) => {
+  const { expenditure, recipients, filteredDomainId } = formValues || {};
+
+  const secondFormSection = useMemo(() => {
```
chinins, line 20: I would prefer to have it directly in the html with just a ternary operator. No need to separate it out with the use memo.

src/modules/dashboard/components/ExpenditurePage/Split/SplitEqual/SplitEqual.tsx
```diff
@@ -0,0 +1,221 @@
...
+                elementOnly
+              />
+            </div>
+          </div>
+        </div>
+      </FormSection>
+      <FormSection appearance={{ border: 'bottom' }}>
+        <div className={styles.counterWrapper}>
+          <span>
+            {recipientsCount}{' '}
```
chinins, line 144: Please, do this string via `FormattedMessage`, not good to have spaces in the html like that

src/modules/dashboard/components/ExpenditurePage/Split/SplitEqual/SplitEqual.tsx
```diff
@@ -0,0 +1,221 @@
...
+            <div className={styles.tokenCounterWrapper}>
+              <TokenIcon
+                className={styles.tokenIcon}
+                token={token}
+                name={token?.name || token?.address}
+              />{' '}
+              <Numeral
+                unit={getTokenDecimalsWithFallback(0)}
+                value={calculatedAmount}
+              />{' '}
```
chinins, line 159: Please, remove spaces and use margin instead in both cases

src/modules/dashboard/components/ExpenditurePage/Split/SplitEqual/SplitEqual.tsx
```diff
@@ -0,0 +1,221 @@
...
+      </FormSection>
+      <FormSection>
+        <FieldArray
+          name="split.recipients"
+          render={({ push, remove }) => (
+            <>
+              {recipients?.length > 0 && (
+                <div className={styles.recipientsWrapper}>
+                  {recipients?.map((recipient, index) => {
+                    return (
```
chinins, line 173: You don't need return word here, you can just do `=> (<div ...`

chinins (changes requested):
Great job! Please, implement the changes requested (mostly small things) and ping me so that I can re-review with priority.

ArmandoGraterol (changes requested):
Looking good! Apart from what Olga already mentioned, I noticed this:
FireShot Capture 691 - Create Expenditure - Colony - wonker - localhost

`.` should count as 0 like the other amount inputs that we have in the dapp, but it's being interpreted as `NaN` and it's showing a user-unfriendly error

ArmandoGraterol (approved):
Nice job! This is good to go from my side :ok_hand:

arrenv (approved):
Nice work @agnieszkajarosikloj, thank you!

PR #3660, opened 2022-07-20
Add Gnosis Safe event message

### LINKED ISSUE #3657: Implement "Add Gnosis Safe" action page
### Story

As a user, I want to be able to add my Gnosis Safe to the Colony and see an action page detailing the action after doing it.

### Description

After adding a safe, the user should be redirected to a corresponding action page showing the correct title, action details, and action events.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=727%3A15054)

FireShot Capture 667 - Safe Control – Figma - www figma com

- [ ] The action page title shows the "Add Gnosis Safe" variant.
- [ ] The action page event list shows a corresponding event for the "Add Gnosis Safe" action.
- [ ] The action details section shows the relevant data for the "Add Gnosis Safe" action.

willm30:
## Description

This PR tracks the work involved with ensuring that the correct messages appear on the Action Event page when a new safe is added, and that the correct `ActionListItem` titles appear in the Colony Actions list. 

Design: https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=727%3A15054

**Changes** 🏗

* DefaultAction & ActionsListItem: Show 'Add Gnosis Safe' title when safe added
* Details Widget: Add AddSafe view when safe added
* Updating `staticMaps`, the i18n files, and various other files to account for the AddSafe action

## Screenshot

add-gnosis-safe

actionsList

## TODO

- [x] Event title says 'Add Gnosis Safe'
- [x] Event text accurately describes the added safe
- [x] Details Widget displays correct data 
- [x] ActionsList on ColonyHome displays correct titles 

Resolves #3657

### DIFF
--- a/src/i18n/en-actions.ts
+++ b/src/i18n/en-actions.ts
@@ -33,6 +33,7 @@ const actionsMessageDescriptors = {
       ${ColonyExtendedActions.AddressBookUpdated} {Address book was updated}
       ${ColonyExtendedActions.TokensUpdated} {Colony tokens were updated}
       ${ColonyExtendedActions.SafeRemoved} {Remove Safe}
+      ${ColonyExtendedActions.SafeAdded} {Add Safe from {chainName}}
       other {Generic action we don't have information about}
     }`,
   [`action.${ColonyActions.SetUserRoles}.assign`]: `Assign the {roles} in {fromDomain} to {recipient}`,
@@ -55,6 +56,7 @@ const actionsMessageDescriptors = {
       ${ColonyActions.Recovery} {Recovery}
       ${ColonyActions.EmitDomainReputationPenalty} {Smite}
       ${ColonyActions.EmitDomainReputationReward} {Award}
+      ${ColonyExtendedActions.SafeAdded} {Add Safe}
       ${ColonyExtendedActions.SafeRemoved} {Remove Safe}
       ${ColonyExtendedActions.TokensUpdated} {Update Tokens}
       ${ColonyExtendedActions.AddressBookUpdated} {Update Address Book}
--- a/src/i18n/en-events.ts
+++ b/src/i18n/en-events.ts
@@ -30,6 +30,7 @@ const eventsMessageDescriptors = {
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.tokens`]: `{initiator} changed this colony's tokens`,
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.verifiedAddresses`]: `{initiator} updated this colony's address book`,
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.safeRemoved`]: `{initiator} removed {removedSafesString} using the Administration and Funding permissions.`,
+  [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.safeAdded`]: `{initiator} added the Safe {addedSafeAddress} from {chainName} using the Administration and Funding permissions`,
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.fallback`]: `{initiator} changed this colony's metadata, but the values are the same`,
   [`event.${ColonyAndExtensionsEvents.DomainMetadata}.all`]: `{initiator} changed teams's name, description, color from {oldName}, {oldDescription}, {oldColor} to {domainName}, {domainPurpose}, {domainColor}`,
   [`event.${ColonyAndExtensionsEvents.DomainMetadata}.nameDescription`]: `{initiator} changed teams's name and description from {oldName}, {oldDescription} to {domainName}, {domainPurpose}`,
--- a/src/modules/core/components/ActionsList/ActionsListItem.tsx
+++ b/src/modules/core/components/ActionsList/ActionsListItem.tsx
@@ -40,7 +40,11 @@ import {
   MotionState,
   MOTION_TAG_MAP,
 } from '~utils/colonyMotions';
-import { useExtendedColonyActionType } from '~modules/dashboard/hooks';
+import {
+  useColonyMetadataChecks,
+  useExtendedColonyActionType,
+} from '~modules/dashboard/hooks';
+import { GNOSIS_SAFE_NAMES_MAP } from '~constants';
 
 import { ipfsDataFetcher } from '../../../core/fetchers';
 
@@ -131,6 +135,13 @@ const ActionsListItem = ({
   const [fetchTokenInfo, { data: tokenData }] = useTokenInfoLazyQuery();
 
   const colonyObject = parseColonyMetadata(metadataJSON);
+  const { addedSafe } = useColonyMetadataChecks(
+    actionType,
+    colony,
+    transactionHash,
+    colonyObject,
+  );
+
   const extendedActionType = useExtendedColonyActionType(
     actionType,
     colony,
@@ -339,6 +350,8 @@ const ActionsListItem = ({
                   reputationChangeNumeral: (
                     <Numeral value={formattedReputationChange} />
                   ),
+                  chainName:
+                    addedSafe && GNOSIS_SAFE_NAMES_MAP[addedSafe.chainId],
                 }}
               />
             </span>
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.tsx
@@ -103,7 +103,7 @@ const DefaultAction = ({
   );
 
   let domainMetadata;
-  const { removedSafes } = useColonyMetadataChecks(
+  const { removedSafes, addedSafe } = useColonyMetadataChecks(
     actionType,
     colony,
     transactionHash,
@@ -231,10 +231,15 @@ const DefaultAction = ({
     isSmiteAction: new Decimal(reputationChange).isNegative(),
     removedSafes,
     removedSafesString,
+    addedSafeAddress: addedSafe && (
+      <MaskedAddress address={addedSafe.contractAddress} />
+    ),
+    chainName: addedSafe && GNOSIS_SAFE_NAMES_MAP[addedSafe.chainId],
+    safeName: addedSafe?.safeName,
   };
 
   const actionAndEventValuesForDocumentTitle = {
-    actionType,
+    actionType: extendedActionType,
     initiator:
       initiator.profile?.displayName ??
       initiator.profile?.username ??
@@ -252,6 +257,7 @@ const DefaultAction = ({
     roles: roleTitle,
     reputationChange: actionAndEventValues.reputationChange,
     reputationChangeNumeral: actionAndEventValues.reputationChangeNumeral,
+    chainName: addedSafe && GNOSIS_SAFE_NAMES_MAP[addedSafe.chainId],
   };
 
   const motionStyles = MOTION_TAG_MAP[MotionState.Forced];
--- a/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
@@ -78,6 +78,18 @@ const MSG = defineMessages({
     id: 'dashboard.ActionsPage.DetailsWidget.safe',
     defaultMessage: 'Safe',
   },
+  chain: {
+    id: 'dashboard.ActionsPage.DetailsWidget.chain',
+    defaultMessage: 'Chain',
+  },
+  safeAddress: {
+    id: 'dashboard.ActionsPage.DetailsWidget.safeAddress',
+    defaultMessage: 'Safe Address',
+  },
+  safeName: {
+    id: 'dashboard.ActionsPage.DetailsWidget.safeName',
+    defaultMessage: 'Safe Name',
+  },
 });
 
 interface Props {
@@ -155,6 +167,36 @@ const DetailsWidget = ({
           </div>
           <div className={styles.value}>
             <DetailsWidgetTeam domain={values.motionDomain} />
+          </div>
+        </div>
+      )}
+      {detailsForAction.Chain && values?.chainName && (
+        <div className={styles.item}>
+          <div className={styles.label}>
+            <FormattedMessage {...MSG.chain} />
+          </div>
+          <div className={styles.value}>
+            <span>{values.chainName}</span>
+          </div>
+        </div>
+      )}
+      {detailsForAction.SafeAddress && values?.addedSafeAddress && (
+        <div className={styles.item}>
+          <div className={styles.label}>
+            <FormattedMessage {...MSG.safeAddress} />
+          </div>
+          <div className={styles.value}>
+            <span>{values.addedSafeAddress}</span>
+          </div>
+        </div>
+      )}
+      {detailsForAction.SafeName && values?.safeName && (
+        <div className={styles.item}>
+          <div className={styles.label}>
+            <FormattedMessage {...MSG.safeName} />
+          </div>
+          <div className={styles.value}>
+            <span>{values.safeName}</span>
           </div>
         </div>
       )}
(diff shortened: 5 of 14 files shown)


ArmandoGraterol (changes requested):
I think something changed in the base PR that broke the add safe action? Seeing it's the same error @danbr pointed out in #3674
Screenshot from 2022-07-27 18-01-40

Left some comments for things I've noticed for now that might need changes

src/modules/dashboard/components/ActionsPage/ActionsComponents/RecoveryAction.tsx
```diff
@@ -271,6 +274,7 @@ const RecoveryAction = ({
                       }}
                       emmitedBy={emmitedBy}
                       colony={colony}
+                      colonyMetadataChecks={metadataChecks}
```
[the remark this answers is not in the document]
willm30: @ArmandoGraterol Yep, have removed from both

src/utils/colonyActions.ts
```diff
@@ -265,6 +274,58 @@ export const sortMetdataHistory = (colonyMetadata) =>
...
+  const currentMetadata = getMetadataHashAtIndexFromCurrent(
+    metadataHistory,
+    transactionHash,
+    0,
+  );
+  if (currentMetadata) {
+    return currentMetadata.metadata;
+  }
+};
+/* eslint-enable consistent-return */
```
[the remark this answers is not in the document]
willm30: Thank you for explaining. Hadn't come across that before and was confused!

willm30:
> I think something changed in the base PR that broke the add safe action? 
> 
> Left some comments for things I've noticed for now that might need changes

@ArmandoGraterol I've fixed this on the base branch and have rebased the others so this should be fixed. Thanks for explaining why the linter wasn't happy with me. I've implemented your other suggestions.

src/modules/core/components/ActionsList/ActionsListItem.tsx
```diff
@@ -287,7 +288,7 @@ const ActionsListItem = ({
               <FormattedMessage
                 id={roleMessageDescriptorId || 'action.title'}
                 values={{
-                  actionType,
+                  actionType: isActionTypeAddSafe ? 'AddSafe' : actionType,
```
ArmandoGraterol, line 291: Are there any reasons why this type cannot be determined in `getActionType` like the types of the other actions? This isn't the first action that uses the same events or it's quite similar to other actions, so I'm wondering why can this be done  in the resolvers when others in a similar context can

src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.tsx
```diff
@@ -178,10 +201,18 @@ const DefaultAction = ({
...
+    ),
+    selectedChain:
+      mostRecentlyAddedSafe &&
+      SUPPORTED_GNOSIS_SAFE_NETWORKS[mostRecentlyAddedSafe.chainId].name,
+    safeName: mostRecentlyAddedSafe?.safeName,
   };
 
   const actionAndEventValuesForDocumentTitle = {
-    actionType,
+    actionType: isActionTypeAddSafe ? 'AddSafe' : actionType,
```
ArmandoGraterol, line 215: Same question here

ArmandoGraterol (commented):
Looking great from what I've been testing! Congrats on your first action page wiring :) The IPFS workaround of passing the data in the router state, if I understand it correctly, seems like a good idea, haven't thought about doing something like that before.

I just have a question regarding determining the action type on the action page.

src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidget.tsx
```diff
@@ -153,6 +165,34 @@ const DetailsWidget = ({
           </div>
         </div>
       )}
+      {values?.isAddSafe && (
```
ArmandoGraterol, line 168: Also, you need to change.  Add the details of this action and the action type itself to the `DETAILS_FOR_ACTION` mapping that we have, that's something that we do for every action. Then each detail should have its own separate conditional like the rest of the details in this component:

```
...
{detailsForAction.Chain && (...)}
{detailsForAction.Safe && (...)}
...
```

This is really useful and important because the component stays adaptable to any other action that might share details with this one.

ArmandoGraterol: Overall, take a look at the `staticMaps` file and make sure that this new action type is correctly integrated just like the other action types that we have

src/modules/core/components/ActionsList/ActionsListItem.tsx
willm30: @ArmandoGraterol I think originally I struggled to see where I'd implement the checks, but after looking again I've seen where they should go and have included them in the transformer (for the Colony Actions list) and the resolver (for the Actions Page).

willm30:
@ArmandoGraterol I've added the checks for whether the action type is "AddSafe" to `getActionType` in the resolver (in the case of the Actions Page) and the `getActionsListData` transformer (for the actions list on the colony home page), as you suggested. I've also updated the StaticMaps so the DetailsWidget follows the same pattern as the other actions.

Julianahlk (changes requested):
Great work so far Will! However, I am experiencing the following:

1. Whenever I add a new safe I see the following data displayed in the DetailsWidget: 
Screen Shot 2022-08-02 at 6 10 14 PM
But I was expecting to see more data about the safe like shown below:
Screen Shot 2022-08-02 at 6 12 13 PM

2. Every time I add a new safe I see two action list items being created but I would expect just one to be created:
Screen Shot 2022-08-02 at 6 12 44 PM

- The first item from the bottom up takes me to the page below and it contains all the expected data details of the newly added safe:
Screen Shot 2022-08-02 at 6 12 44 PM

- The second item takes me to a blank page with some TypeErrors:
Screen Shot 2022-08-02 at 6 38 19 PM

willm30:
> 1. Whenever I add a new safe I see the following data displayed in the DetailsWidget:

This should now be fixed, thanks.

> 2. Every time I add a new safe I see two action list items being created but I would expect just one to be created:

This doesn't happen on my machine. Does this still happen after restarting webpack or after a creating a new colony? And if so, does this happen for all actions or just add safe? Could you let me know anything relevant for me to reproduce, e.g. which browser are you using, which extensions have you got installed? Thanks.

Julianahlk:
> This doesn't happen on my machine. Does this still happen after restarting webpack or after a creating a new colony? And if so, does this happen for all actions or just add safe? Could you let me know anything relevant for me to reproduce, e.g. which browser are you using, which extensions have you got installed? Thanks.

1. Yes, still experiencing issues even after restarting Webpack. 
2. I tried creating a new Colony but for some reason I am unable to do so. 
3. At the moment, every time I create a new action, the actions list is not updating for all actions.
4. 
Screen Shot 2022-08-02 at 8 10 25 PM

Julianahlk:
Hey Will, I re-started my dev env and things are working now. I sent you some videos over Discord of me creating a safe for the first time to make sure everything looks right.

Julianahlk (approved):

chinins (changes requested):
Nice attempt but unfortunately not a correct one.

The approach that you are using is not in line with what we already have in the dapp and also duplicates some of the already existing logic. It's not your fault - you don't know about some things in the dapp. The feature branch is not rebased on latest master where verified recipients is introduced. There, the logic for comparing colony metadata already exists.
I think that you should wait for Armando to rebase on latest master. Then rebase your branch and then update this PR based on the metadata checks that we already have in the dapp and the approach we use.

I've already discussed with Armando that for this issue & for the issue that I have of adding remove safe page we need the rebase and this is on his radar.

Also, for the PR titles - these are not commits, no need to add `Add:` (Feat or anything like this). It seems that this PR is to add safe action page so just `Add sage action page` would be good.

willm30:
@chinins I have updated to reflect your comment and the existing style of checking metadata in the dapp. I'd be grateful if you couuld take another look to confirm this is what you were thinking. 

As an aside, I think we should update all the sagas that interact with colony metadata to use the approach in the verified recipients saga. Do you and @ArmandoGraterol agree with that? When I was working on wiring up the addSafe saga, I changed the editColony saga to get the current metadata direct from IPFS and update it that way, but verified recipients receives all the metadata via its payload. So, I think we should ensure consistency. They also all need to be updated to ensure they're all including the safe data -- I haven't done this is in this branch but would imagine including that in the work to ensure consistency across them all.

chinins (changes requested):
Nice work! I think that you've chosen an approach that is too manual so prone to errors if there is added complexity and it also introduces yet another way of handling actions.

I was able to get into nitty-gritty details because I was working on a similar PR - just removing safes. Please, have a look at my PR #3736 . I tried to generalise and abstract the colony edit actions specifics. The approach for adding and removing safes should be the same so if you and the rest of the team is happy with my approach then please rebase your branch on mine and refactor accordingly. Alternatively I'm open to suggestions.

For the sagas/actions I think it should all be handled in a separate issue/PR. I talked to @ArmandoGraterol  a while ago about refactoring them after a rebase and I think it should be on his radar to add an issue for that. I think it makes sense to combine it all in one PR - to make sure that it's exactly the same approach everywhere. If @ArmandoGraterol  agrees with me I would remove that part and let it be handled separately. But Armando is the feature owner so it's for him to decide.

willm30:
@chinins I have updated to follow your approach. Please review when you get a chance. Thanks

src/utils/colonyActions.ts
```diff
@@ -323,13 +337,14 @@ export const getSpecificActionValuesCheck = (
...
     isWhitelistActivated?: boolean | null;
+    colonySafes?: ColonySafe[];
   },
 ): { [key: string]: boolean } => {
   switch (actionType) {
     case ColonyAndExtensionsEvents.ColonyMetadata: {
       const nameChanged = prevColonyDisplayName !== currentColonyDisplayName;
       const logoChanged = prevColonyAvatarHash !== currentColonyAvatarHash;
-
+      const safeAdded = currentColonySafes.length > prevColonySafes.length;
```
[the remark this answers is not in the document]
willm30: @chinins For this one, I updated the `colonyAction` resolver so that if `colonySafes` comes back from IPFS as `null` it will return an empty array. This is consistent with how we've typed the colonySafes property (i.e. as non-nullable).

ArmandoGraterol (approved):
Looking good to me! Tested by adding several safes and it was working as expected. The code is much simpler than the first I reviewed it :)

About the sagas, yeah it was agreed that after the rebase updating the sagas added in this feature would also be updated with a dedicated issue/PR

src/i18n/en-actions.ts
```diff
@@ -33,6 +33,7 @@ const actionsMessageDescriptors = {
       ${ColonyExtendedActions.AddressBookUpdated} {Address book was updated}
       ${ColonyExtendedActions.TokensUpdated} {Colony tokens were updated}
       ${ColonyExtendedActions.SafeRemoved} {Remove Safe}
+      ${ColonyExtendedActions.SafeAdded} {Add Gnosis Safe}
```
chinins, line 36: We might have a different chain than Gnosis so either you remove Gnosis from the string or you add a variable and pass the relevant chain name.

src/i18n/en-events.ts
```diff
@@ -30,6 +30,7 @@ const eventsMessageDescriptors = {
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.tokens`]: `{initiator} changed this colony's tokens`,
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.verifiedAddresses`]: `{initiator} updated this colony's address book`,
   [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.safeRemoved`]: `{initiator} removed {removedSafesString} using the Administration and Funding permissions.`,
+  [`event.${ColonyAndExtensionsEvents.ColonyMetadata}.safeAdded`]: `{initiator} added the Gnosis Safe {safeAddress} from {selectedChain} using the Administration and Funding permissions`,
```
chinins, line 33: Same here - we might have a safe from a different chain so you need to add the chainName variable and pass it in (same as safeAddress).
@ArmandoGraterol, am I right?

src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.tsx
```diff
@@ -231,10 +231,15 @@ const DefaultAction = ({
     isSmiteAction: new Decimal(reputationChange).isNegative(),
     removedSafes,
     removedSafesString,
+    safeAddress: addedSafe && (
```
chinins, line 234: It's a nitpick but I would call it something like `addedSafeAddress`, so it's clear what operation we are talking about. But up to you...

src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultAction.tsx
```diff
@@ -231,10 +231,15 @@ const DefaultAction = ({
...
+      <MaskedAddress address={addedSafe.contractAddress} />
+    ),
+    selectedChain: addedSafe && GNOSIS_SAFE_NAMES_MAP[addedSafe.chainId],
```
chinins, line 237: A nitpick again but I would go for a more precise name, something like `chainName` (name is an important part here). But up to you...

src/utils/colonyActions.ts
```diff
@@ -364,6 +368,14 @@ export const getSpecificActionValuesCheck = (
         currentColonyTokens?.slice(0).sort() || [],
       );
 
+      const addedSafe =
+        currentColonySafes.length > prevColonySafes.length
+          ? currentColonySafes.filter(
+              (safe) =>
+                !prevColonySafes.some((prevSafe) => isEqual(prevSafe, safe)),
```
chinins, line 375: I think you can just do a simple check of `prevSafe.address === safe.address`.
I think that you can also use `.find` instead of filter, this would give you the element without the need to select `[0]` like you do now. But that's a nitpick.

chinins (approved):
Looks much better and cleaner - good job!

There is only one thing that I also mentioned in my direct comments. The safe name in the title and description should be based on the chain name or removed at all (at least htis is how I understand the feature) but it's hardcoded. So what ends up happening is I add a safe from Aurora but it's called Gnosis. That doesn't seem right. @ArmandoGraterol, @arrenv, can you please confirm?
Unrelated to this issue, but if I'm right I would also suggest to update the title in the advanced dialog from `Gnosis Safe Control` to just `Safe control`.

Screenshot 2022-08-17 at 11 31 55

src/i18n/en-events.ts
ArmandoGraterol: This might be a good question for @arrenv as from my understanding these safes may still be called "Gnosis Safe" even though they may not be in the "Gnosis Chain". 

For example, in `https://gnosis-safe.io` you can create safes from different chains other than the Gnosis chain, my suspicion is that "Gnosis Safe" may just be a "brand name" for the safes.

willm30: I understand it that the product is called "Gnosis Safe", and the use of the term "safe" to refer to a multisig is specific to the Gnosis Safe platform. Therefore something like an "Aurora" safe doesn't exist. It would be a Gnosis Safe deployed on the Aurora network. 

@arrenv Is that right?

arrenv: @willm30 @ArmandoGraterol  Yes, "Gnosis Safe" is a brand. In fact they are rebranding it to just be "Safe" - https://www.cryptotimes.io/gnosis-safe-is-rebranded-to-safe-with-100m-fundraise

Yes, we need to be able to specify what chain the Safe is on.

arrenv: So, this would be: "{initiator} added the Safe {safeAddress} from {selectedChain} using the Administration and Funding permissions`"

src/i18n/en-actions.ts
arrenv: This would be: "Add Safe From Rinkeby Testnet"

arrenv (approved):
Looks and works well! Very nice!

To add to @chinins comments, I also think we should improve this to use "Safe" as the naming and then include the chainName.

PR #3667, opened 2022-07-21
Add "Linked assets" section to home sidebar

### LINKED ISSUE #3624: Implement "Linked Assets" aside in Colony Home
### Story

As a user, I want to be able to see a list of all the linked safes to the colony in the aside of the colony's home.

### Description

This issue is for the UI and connection of the new aside section that should appear when there's at least 1 safe linked to the colony. The list of the safes should be obtained from the colony's metadata with IPFS.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=139%3A9988)

FireShot Capture 664 - Safe Control – Figma - www figma com

- [ ] A new section appears in the colony's home aside when safes are added to the colony showing the name of each safe.

ArmandoGraterol:
FireShot Capture 671 - Actions - Colony - wonker2 - localhost

Resolves #3624

### DIFF
--- a/src/data/graphql/typeDefs.ts
+++ b/src/data/graphql/typeDefs.ts
@@ -418,6 +418,12 @@ export default gql`
     processedBalances(colonyAddress: String!): [ProcessedTokenBalances!]!
   }
 
+  type ColonySafe {
+    safeName: String!
+    contractAddress: String!
+    chainId: String!
+  }
+
   type ProcessedColony {
     id: Int!
     colonyAddress: String!
@@ -428,6 +434,7 @@ export default gql`
     nativeTokenAddress: String!
     tokenAddresses: [String]!
     extensionAddresses: [String!]
+    safes: [ColonySafe!]!
     domains: [ProcessedDomain!]!
     roles: [ProcessedRoles!]!
     tokens: [ProcessedTokens!]!
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -94,6 +94,11 @@ export const getProcessedColony = async (
   let avatarHash: string | null = null;
   let avatarObject: { image: string | null } | null = { image: null };
   let tokenAddresses: Array<Address> = [];
+  let safes: Array<{
+    safeName: string;
+    contractAddress: string;
+    chainId: number;
+  }> = [];
 
   const sortedMetdataHistory = sortMetdataHistory(metadataHistory);
   const currentMetadataIndex = sortedMetdataHistory.findIndex(
--- a/src/data/resolvers/colony.ts
+++ b/src/data/resolvers/colony.ts
@@ -123,10 +123,12 @@ export const getProcessedColony = async (
         colonyDisplayName = null,
         colonyAvatarHash = null,
         colonyTokens = [],
+        colonySafes = [],
       } = JSON.parse(ipfsMetadata);
       displayName = colonyDisplayName;
       avatarHash = colonyAvatarHash;
       tokenAddresses = colonyTokens;
+      safes = colonySafes;
 
       /*
        * Fetch the colony's avatar
@@ -168,6 +170,7 @@ export const getProcessedColony = async (
       ? [...tokenAddresses, token.tokenAddress].map(createAddress)
       : [],
     extensionAddresses: colonyExtensions.map(({ address }) => address),
+    safes,
   };
 };
 
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -1213,6 +1213,12 @@ export type ProcessedTokensProcessedBalancesArgs = {
   colonyAddress: Scalars['String'];
 };
 
+export type ColonySafe = {
+  safeName: Scalars['String'];
+  contractAddress: Scalars['String'];
+  chainId: Scalars['String'];
+};
+
 export type ProcessedColony = {
   id: Scalars['Int'];
   colonyAddress: Scalars['String'];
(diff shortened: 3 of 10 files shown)


src/data/resolvers/colony.ts
```diff
@@ -123,10 +127,12 @@ export const getProcessedColony = async (
         colonyDisplayName = null,
         colonyAvatarHash = null,
         colonyTokens = [],
+        colonySafes = [],
```
danbr, line 130: For V2 metadata in IPFS, will safe data be added to a ColonyMetadata Object (as listed here) or should we create a new MetadataType?

danbr (approved):
Looks good well done. Just a question about metadata & IPFS.....

src/data/resolvers/colony.ts
ArmandoGraterol: As far as I know and from what I've seen in recent implementations of the safe dialogs, it has been very convenient for us to have it with the colony's metadata

src/data/graphql/typeDefs.ts
```diff
@@ -428,6 +434,7 @@ export default gql`
     nativeTokenAddress: String!
     tokenAddresses: [String]!
     extensionAddresses: [String!]
+    safes: [ColonySafe!]!
```
willm30, line 437: Think this should be `safes: [ColonySafe]!` because the array may be empty?

Incidentally, I made the same changes to the query/resolver [here](https://github.com/JoinColony/colonyDapp/pull/3660). Just flagging for your attention.

willm30 (approved):
Looks good. I left one comment. I made some changes to the base branch so a rebase is required.

willm30:
> Looks good. I left one comment. I made some changes to the base branch so a rebase is required.

Also, how come we're calling the component `ColonySafes` but the heading is "LinkedAssets"?

ArmandoGraterol:
> > Looks good. I left one comment. I made some changes to the base branch so a rebase is required.
> 
> Also, how come we're calling the component `ColonySafes` but the heading is "Linked Assets"? If we can only use the section for safes, shouldn't it be titled "Linked safes"?

That's a good question for our Product guy @arrenv, if the section is going to be used to show other assets other than safes then it would make sense for the component's name to be `ColonyAssets` or something like that

src/data/graphql/typeDefs.ts
ArmandoGraterol: `ColonySafe!` is not telling GraphQL that the array cannot be empty, it just telling it that the values inside the array are non-nullable so it shouldn't return something like `[null, {...NORMAL_ITEM_HERE}]`, for example. 

`[ColonySafe!]!` is basically telling GraphQL that we are expecting an array, empty or not, that stores `ColonySafe` objects inside it (no nulls).

More info here in the docs: https://graphql.org/learn/schema/

willm30: I've looked over that before but was confused by "And since Episode! is also non-nullable, you can always expect every item of the array to be an Episode object." But I guess it means _when_ the array contains more than zero items , not _that_ it must return at least one item that is an Episode (or ColonySafe in this case). 

Thanks for clarifying.

ArmandoGraterol: No problem! At first the use of ! for cases like this can be confusing

.prettierignore
```diff
@@ -0,0 +1 @@
+*.css
```
chinins, line 1: Why do we have these files in this PR?

.prettierrc
```diff
@@ -0,0 +1,4 @@
+{
```
chinins, line 1: Why do we have these files in this PR?

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
```diff
@@ -47,6 +47,7 @@ const MSG = defineMessages({
   fetchFailed: {
     id: 'dashboard.AddExistingSafeDialog.AddExistingSafeDialogForm.fetchFailed',
     defaultMessage:
+      // eslint-disable-next-line max-len
       'Could not fetch Safe details. Please check your connection and try again.',
```
chinins, line 51: You can just wrap it in back ticks, then no need for the eslint-disable

chinins (commented):
Should the colonyNetork files be in this PR? Along with the prettier files? It doesn't seem like a PR requiring it.

We are missing the verified recipients work in this branch which impacts the ipfs metadata. When there is a rebase on latest master, there will need to be updates added to `manageVerifiedRecipients` saga and/or dialog. So either we should add an issue about it or rebase everything on master right now and add it straight away.

Generally, I'm a bit concerned that we keep adding things to colony metadata object & keep adding sagas where we touch it. There is a risk that in some of these sagas we might forget about including all the parts for the metadata and things will be lost. But I guess this is more of a big picture and it will be changed when we switch to cdapp.

ArmandoGraterol:
> Should the colonyNetork files be in this PR? Along with the prettier files? It doesn't seem like a PR requiring it.

Oops, those changes were in the branch that this PR is based on as a mistake, and they were recently removed, so that's why they are showing up now in this one. Time for a rebase.

> We are missing the verified recipients work in this branch which impacts the ipfs metadata. When there is a rebase on latest master, there will need to be updates added to `manageVerifiedRecipients` saga and/or dialog. So either we should add an issue about it or rebase everything on master right now and add it straight away.

In that case, I think that once all the current PRs that deal with IPFS that are related to the feature are merged, I could rebase the feature branch and then create an issue to update the parts you are mentioning.

> Generally, I'm a bit concerned that we keep adding things to colony metadata object & keep adding sagas where we touch it. There is a risk that in some of these sagas we might forget about including all the parts for the metadata and things will be lost. But I guess this is more of a big picture and it will be changed when we switch to cdapp.

Considering that these are the last feature we are going to develop for the dapp, yeah it shouldn't be much of a concern once we switch to the cdapp.

chinins:
> In that case, I think that once all the current PRs that deal with IPFS that are related to the feature are merged, I could rebase the feature branch and then create an issue to update the parts you are mentioning.

I would just create and issue and put it as blocked just not to forget - but as you wish :)

chinins (approved):
All good once you rebase. Just make sure that the files not related to this branch are removed.

PR #3668, opened 2022-07-21
Connect `Remove Safe` dialog

### LINKED ISSUE #3623: Connect `Remove Safe` dialog
### Story

As a user, I want to be able to see a list of added safes and remove any of the Gnosis Safes associated with the colony.

### Description

The dialog should display a _connected_ list of the added safes (or none if no safes were added), which is obtained from the colony's metadata using IPFS. Through the dialog, we can remove any of the safes that were added (or even all of them at once) by getting the metadata object, editing it, and uploading it once more like we would if we added a new safe.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=115%3A9509)

FireShot Capture 662 - Safe Control – Figma - www figma com

- [x] A connected list appears in the dialog which displays the current safes in the colony.
- [x] You can delete as many safes as possible and as you want (Just one, more than 1, or all).
- [x] A message appears in the dialog when the list of safes is empty.

ArmandoGraterol:
Connected "Remove Safe" dialog with IPFS. Now we should be able to remove safes that are linked with the colony.

- If there's only 1 safe, it should be preselected when entering the dialog.
- If there are multiple, it should be possible to remove as many as we want.
- Check that the checkbox are displaying the correct info of each safe (name, address, and chain).
- Verify if the safes are deleted by checking the "Linked assets" sections in the colony home.
- Unrelated to the issue but also added permission validation to the add/remove safes UAC menu items.

FireShot Capture 675 - Actions - Colony - wonker - localhost

Resolves #3623


src/modules/dashboard/sagas/actions/removeExistingSafes.ts
```diff
@@ -0,0 +1,221 @@
...
+    const currentMetadata = yield call(
+      ipfsWithFallback.getString,
+      currentMetadataIPFSHash,
+    );
+
+    const colonyMetadata = JSON.parse(currentMetadata);
+
+    const updatedColonySafes = colonyMetadata.colonySafes.filter((safe) =>
+      safeAddresses.find((safeAddress) => safeAddress !== safe.contractAddress),
+    );
```
willm30, line 124: For more than one safe in the `safeAddresses` array, the find predicate will always return true and therefore the filter will return the same array. If there is more than one safe in `safeAddresses`, you will always be able to find a safe address that is not equal to the current item in the filter. 

I suggest something like this, which is saying, if there is a safe in the metadata that is also in `safeAddresses`, don't return it in the updated array. 

``` js
const updatedColonySafes = colonyMetadata.colonySafes.filter(
      (safe) =>
        !safeAddresses.some(
          (safeAddress) => safeAddress === safe.contractAddress,
        ),
    );
```

willm30 (changes requested):
🟩 If there's only 1 safe, it should be preselected when entering the dialog.
🟩 Check that the checkbox are displaying the correct info of each safe (name, address, and chain).

> If there are multiple, it should be possible to remove as many as we want.
> Verify if the safes are deleted by checking the "Linked assets" sections in the colony home.

The actual remove safe functionality isn't working for me. I have added three safes, the two you sent me and the one I had on Goerli. I have made a suggestion in the saga re the filtering logic. 

> Unrelated to the issue but also added permission validation to the add/remove safes UAC menu items.

I think this is a good addition. I can see that the add / remove safe options are disabled when I don't have permissions, but the control safe option is still available. Is that right?

src/modules/dashboard/sagas/actions/removeExistingSafes.ts
ArmandoGraterol: You are completely right! This is the solution. When I tested it I remember that it was working, but it seems like I'm going crazy because, following the logic I had before, it shouldn't hahaha 

Thank you!

ArmandoGraterol:
> I think this is a good addition. I can see that the add / remove safe options are disabled when I don't have permissions, but the control safe option is still available. Is that right?

That option also needs permissions. It's on purpose that I didn't implement the permission logic of the control safe option, didn't want to extend myself too much in a PR that is not related, and well, that option is not as "ready" as add/remove.

I'll probably create a issue for that

willm30 (approved):

src/modules/dashboard/components/Dialogs/RemoveSafeDialog/SafeListItem.tsx
```diff
@@ -30,28 +34,27 @@ const SafeListItem = ({ safe, isChecked }: Props) => {
...
-        title={safe.name || safe.address}
+        seed={safe.contractAddress}
+        title={safe.safeName || safe.contractAddress}
         size="xs"
         className={styles.avatar}
       />
 
       <span className={`${isChecked ? styles.selectedLabel : styles.label}`}>
-        {`${safe.name} (${safe.chain})`}
+        {`${safe.safeName} (${safeNetwork?.name || 'Unknown'})`}
```
chinins, line 49: I think `Unknown` should be done vie a message descriptor (react-intl)

chinins (commented):
A couple of things:
- I think that add safe and remove existing safes are very similar so I would suggest making just one saga & actiontypes and differentiate based on a variable or in another way. But it's your PR and your decision. I am just wondering if you considered that option.
- I found a strange bug that's only somewhat related to this PR. If I login with a different account (with or without relevant permisisons) I don't see linked assets on the side and no safes if I get to the dialog. I think it's something to do with the IPFS data. The screenshots are below.  Can you test on your machine (it might be something with my dev environment but tbh I doubt it).

My initial account
Screenshot 2022-07-30 at 13 12 03
Screenshot 2022-07-30 at 13 11 49

Same colony same time but a different account (logged in both via chrome incognito & via brave):
Screenshot 2022-07-30 at 13 11 21

Screenshot 2022-07-30 at 13 12 16

danbr (approved):
Well done...LGTM. 
The redirection page isnt quite right, but i think this has been already fixed now (outside of this feature)
Screenshot 2022-08-01 at 16 18 06

Julianahlk (changes requested):
Looking good! A few things I found:

1. I am having issues when adding and removing safes. Every time I add a new safe address, it looks like the safe was added successfully but then I go to the remove safe dialog and I only see the "No Safes found to remove." message:

Screen Shot 2022-08-01 at 9 04 18 PM

However, only once I refresh the page the new safe is/are finally there:
Screen Shot 2022-08-01 at 9 06 08 PM

But if I delete the safes they aren't updated inmediatedly and I still see them in the Linked assets section:
Screen Shot 2022-08-01 at 9 26 23 PM
Screen Shot 2022-08-01 at 9 22 17 PM

2. When I add two safes with the same address and the same name I get this warning:
Screen Shot 2022-08-01 at 8 56 18 PM

So, I tried to remove only one of the duplicates but when I tried to click on one of them both copies get selected at the same time so my only choice was to delete both:
Screen Shot 2022-08-01 at 9 12 45 PM

4. The tooltip below is a bit off
Screen Shot 2022-08-01 at 9 15 19 PM

ArmandoGraterol:
@chinins 

1. That sounds like a good idea! I didn't think about that. I'll see if I can implement it in this PR.
2. I can replicate it but only using Incognito. I'm gonna take a look as it is weird that even after reloading they are not appearing.

@danbr The action page wirings are a different issue: [#3669](https://github.com/JoinColony/colonyDapp/issues/3669) and [#3657](https://github.com/JoinColony/colonyDapp/issues/3657) 

@Julianahlk

1. That's unfortunately not happening to me, are you using Incognito or something like that, like Olga? For me, the linked assets sections and the dialog are updating without refreshing.
2. I mean, it's an expected bug if you add 2 identical safes, which you shouldn't be able to do once #3674 is merged :P
3. Thank you! Didn't notice this one :detective:

src/modules/dashboard/sagas/actions/removeExistingSafes.ts
```diff
@@ -0,0 +1,225 @@
...
+      },
+      fetchPolicy: 'network-only',
+    });
+
+    const currentMetadata = yield call(
+      ipfsWithFallback.getString,
+      currentMetadataIPFSHash,
+    );
+
+    const colonyMetadata = JSON.parse(currentMetadata);
```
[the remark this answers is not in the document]
ArmandoGraterol: Yep, @chinins mentioned it to me last week. Once some of the PRs that deal with metadata are merged into the feature branch then I'll rebase and create an issue to deal with updating the actions

Julianahlk:
> 1. That's unfortunately not happening to me, are you using Incognito or something like that, like Olga? For me, the linked assets sections and the dialog are updating without refreshing.

Hey, No I am not using incognito or Brave. I am using Chrome. I will send you a video recording of my screen over Discord so you can see what's happening. 

> 2. I mean, it's an expected bug if you add 2 identical safes, which you shouldn't be able to do once [Add: check if safe already exists in colony #3674](https://github.com/JoinColony/colonyDapp/pull/3674) is merged :P

Oh right! I forgot. Ok 👍

> 3. Thank you! Didn't notice this one 🕵️

☺️

Julianahlk:
Apparently, my local dev was in a weird state, so I re-started my dev env and I no longer encounter the issues I mentioned above. I sent you a video confirmation via Discord. All is good 😅.

Julianahlk (approved):

PR #3643, opened 2022-07-17
Add: Connect add safe dialog

willm30:
## Description

This PR is for the wiring up of the "Add safe" dialog. It:

* adds the functionality to check the safe exists on the desired chain
* adds the saga that uploads the data to the colony's metadata stored in IPFS on form submit
* edits the `editColony` saga to copy over and update existing metadata instead of overwriting it

It does not:
* Update the actions messages on the `ActionsPage`. This is covered here: https://github.com/JoinColony/colonyDapp/pull/3660

**Additions**  ✨

* `addExistingSafe.ts` saga

**Changes** 🏗

* `AddExisitingSafeDialogForm` : wire up form
* `AddExistingSafeDialog`: wire up form
* `constants.ts`: add Goerli test net to chainlist
* `editColony.ts` saga: copy over and update existing metadata instead of overwriting it

## GIF

add-safe

## Notes 

1. For testing you can use: 
* Goerli testnet : `0xD5E1361a921F939b1B01335f1E58ce3f7c743CFf`
* Polygon: 0x7E08C7606aCF5887B7dA9914f7c7cD05c2bec8Ab 
* Ethereum: 0x0DA0C3e52C977Ed3cBc641fF02DD271c3ED55aFe 

1. To verify the safes have been added successfully to the colony metadata, stick a `console.log(ipfsMetadata, 'ipfsMetadata')` on line 119 of `data/resolvers/colony.ts`. Then navigate to the home page with the console open, and after a moment you'll see it show up in the log.

1. Currently, you can add the same safe multiple times. I've opened this PR to fix: https://github.com/JoinColony/colonyDapp/pull/3674)

 ## TODO

- [x] Wire up info message (does safe exist?)
- [x] Wire up submit

Resolves #3622

PR #3674, opened 2022-07-24
Add: check if safe already exists in colony

willm30:
## Description

This PR adds a check to the AddExistingSafeDialog form for if the safe already exists in the colony, and won't permit form completion if so.

To test, add a safe via the Add Existing Safe option in the UAC. 
Then try to add the same safe again. You should see the following error :
safe-already-exists

**Changes** 🏗

* AddExistingSafeDialogForm: Check whether safe already exists in colony and show error if so.

Resolves #3643

### DIFF
--- a/src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
@@ -1,13 +1,15 @@
 import React, { useEffect, useState } from 'react';
 import { FormikProps, useField } from 'formik';
 import { defineMessages, MessageDescriptor, useIntl } from 'react-intl';
+import { isAddress } from 'web3-utils';
 
 import Button from '~core/Button';
 import DialogSection from '~core/Dialog/DialogSection';
 import { Select, Input, Annotations, SelectOption } from '~core/Fields';
 import Heading from '~core/Heading';
-import { GNOSIS_SAFE_NETWORKS } from '~constants';
+import { GNOSIS_SAFE_NETWORKS, SAFE_ALREADY_EXISTS } from '~constants';
 import { SimpleMessageValues } from '~types/index';
+import { ColonySafe } from '~data/index';
 
 import { FormValues } from './AddExistingSafeDialog';
 
@@ -38,6 +40,11 @@ const MSG = defineMessages({
     id: 'dashboard.AddExistingSafeDialog.AddExistingSafeDialogForm.safeLoading',
     defaultMessage: 'Loading Safe details...',
   },
+  alreadyExists: {
+    id:
+      'dashboard.AddExistingSafeDialog.AddExistingSafeDialogForm.alreadyExists',
+    defaultMessage: 'Safe already exists in this colony.',
+  },
   safeCheck: {
     id: 'dashboard.AddExistingSafeDialog.AddExistingSafeDialogForm.safeCheck',
     defaultMessage: `Safe {safeData, select,
@@ -54,6 +61,7 @@ const MSG = defineMessages({
 interface Props {
   back: () => void;
   networkOptions: SelectOption[];
+  colonySafes: ColonySafe[];
 }
 
 interface SafeContract {
@@ -70,7 +78,7 @@ interface StatusText {
   statusValues?: SimpleMessageValues;
 }
 
-type SafeData = SafeContract | undefined | null | {};
+type SafeData = SafeContract | undefined | null | {} | 'alreadyExists';
 
 const AddExistingSafeDialogForm = ({
   back,
@@ -79,6 +87,7 @@ const AddExistingSafeDialogForm = ({
   isSubmitting,
   isValid,
   dirty,
+  colonySafes,
 }: Props & FormikProps<FormValues>) => {
   const { formatMessage } = useIntl();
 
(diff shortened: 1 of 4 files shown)


src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
```diff
@@ -120,11 +130,34 @@ const AddExistingSafeDialogForm = ({
...
+      /*
+       * Check if safe already exists in colony
+       */
+      if (
+        colonySafes.filter((safe) => {
+          return (
+            safe?.chainId === selectedChain.value &&
+            safe?.contractAddress === address
+          );
+        }).length > 0
```
ArmandoGraterol, line 142: I think something like this would be more readable:
```
const isSafeAlreadyAdded = !colonySafes.find(safe => 
  safe?.chainId === selectedChain.value && safe?.contractAddress === address
);

if (isSafeAlreadyAdded) { ...
```

What do you think?

ArmandoGraterol (changes requested):
Mostly good! I noticed something:

I wrote an address, deleted it, and focus out of the field:
Screenshot from 2022-07-25 18-42-36

We should check that the address used for the URL is actually an address before making the request, otherwise, we will be making unnecessary requests

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
willm30: Yes like it (but without the negation as that would be if safe is not already added).

willm30:
> We should check that the address used for the URL is actually an address before making the request, otherwise, we will be making unnecessary requests

Fixed!

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
```diff
@@ -119,12 +129,34 @@ const AddExistingSafeDialogForm = ({
   // When selected chain or address changes
   useEffect(() => {
     // If there's a valid address in the address input
-    if (addressTouched && !addressError) {
-      // Get safe data
-      setIsLoadingSafe(true);
-      getSafeData(`${baseURL}api/v1/safes/${address}/`);
+    if (addressTouched && !addressError && address.length !== 0) {
```
ArmandoGraterol, line 132: Crashing error here:
Screenshot from 2022-07-26 15-27-20

This happens when I focus in and out of the address field

A better alternative to this could be `isAddress(address)`, that way you are covering `undefined`, `null`, `''`, `' '` and checking that the string is actually a valid address. The current implementation is only covering the `''` case.

ArmandoGraterol (changes requested):

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
willm30: @ArmandoGraterol Good suggestion 👍

ArmandoGraterol (approved):
Looks good to me :+1: Nice job

danbr (commented):
@willm30 Started looking at this PR but its failing for me.... I added a console.log to help

Screenshot 2022-07-27 at 22 29 45

danbr:
> @danbr Sorry about that, I updated the createColony saga to include `colonySafes` but I guess colonies that were created before these updates go in won't have it. I've put in a check so should you should be able to test it now.

@willm30 No worries.... Thank you. I'll take a look when i have some spare time today.

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
```diff
@@ -179,13 +195,28 @@ const AddExistingSafeDialogForm = ({
...
+        /*
+         * If not, run api check
+         */
+        setSafeNotFoundError(undefined);
+        setIsLoadingSafe(true);
+        getSafeData(`${baseURL}api/v1/safes/${address}/`);
+      }
     }
-  }, [baseURL, address, addressTouched, addressError]);
+  }, [baseURL, address, addressTouched, addressError, colonySafes]);
```
danbr, line 219: There is a dependancy missing here

Screenshot 2022-07-28 at 21 20 25

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
```diff
@@ -197,7 +228,12 @@ const AddExistingSafeDialogForm = ({
   };
 
   useEffect(() => {
-    if (!isLoadingSafe && (safeData === undefined || safeData === null)) {
+    if (
+      !isLoadingSafe &&
+      (safeData === undefined ||
+        safeData === null ||
+        safeData === 'alreadyExists')
```
danbr, line 235: 'alreadyExists' is used many places in the file, maybe a constant could be an alternative?

danbr (approved):
Its now functioning well.  I'm approving but there are a couple of points in the code to consider. Well done.

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
willm30: @danbr Good suggestion, I've implemented.

PR #3677, opened 2022-07-25
Fix: confirm button flashing on dialog open

### LINKED ISSUE #3632: Bug: "ban / unban address" button flashing
## Steps to reproduce

1. Go to `{colony_home}/members/`  and click either ban address or unban address. 
2. As the modal opens, the confirm button flashes the colour corresponding to whether it's ban / unban. 

## Expected behaviour

No colour flashing should be observable

## Actual behaviour

Button is flashing 

flashing

willm30:
## Description

Small pr to fix a small bug with the confirm button of the Ban User form that was flashing upon first open. Now, when opening the dialog via ban / unban address, the confirm button does not flash.

**Changes** 🏗

* Form is now disabled if the input is an empty string.

## Screenshot
_Undesired flashing_

flashing

Resolves #3632

### DIFF
--- a/src/modules/core/components/Comment/Dialogs/BanUser/BanUser.tsx
+++ b/src/modules/core/components/Comment/Dialogs/BanUser/BanUser.tsx
@@ -175,7 +175,12 @@ const BanUser = ({
       onSubmit={handleSubmit}
       enableReinitialize
     >
-      {({ isValid, isSubmitting, submitForm }: FormikProps<FormValues>) => (
+      {({
+        isValid,
+        isSubmitting,
+        values,
+        submitForm,
+      }: FormikProps<FormValues>) => (
         <Dialog cancel={cancel}>
           <DialogSection>
             <Heading
@@ -219,7 +224,7 @@ const BanUser = ({
                   size: 'large',
                 }}
                 text={isBanning ? MSG.banish : MSG.deactivateBan}
-                disabled={!isValid || isSubmitting}
+                disabled={!isValid || isSubmitting || values.userAddress === ''}
                 loading={loadingBanAction || isSubmitting}
                 onClick={submitForm}
               />


ArmandoGraterol (changes requested):
Unfortunately, it can't be this simple:
FireShot Capture 676 - Members - Colony - wonker - localhost

Now when I want to ban a user through the "Ban user" option in the 3 dots tooltip of the member list I can't because the user address is prefilled and therefore the form is not dirty

willm30:
> Now when I want to ban a user through the "Ban user" option in the 3 dots tooltip of the member list I can't because the user address is prefilled and therefore the form is not dirty

Nice catch. Should be fixed now (still fairly simple 🙂).

ArmandoGraterol (approved):
Nice one! It's working properly now. I like Dan's suggestion, it's even more simple than what we currently have :)

arrenv (approved):
Looks good, I also prefer Dan's suggestion.

src/modules/core/components/Comment/Dialogs/BanUser/BanUser.tsx
```diff
@@ -175,58 +175,66 @@ const BanUser = ({
...
-                disabled={
-                  loadingBannedUsers || loadingBanAction || isSubmitting
-                }
+      {({
+        isValid,
+        isSubmitting,
+        values,
+        submitForm,
+      }: FormikProps<FormValues>) => {
+        const isEmptyString = values.userAddress === '';
```
chinins, line 184: This is not good practice to have a stateless variable like this.

chinins (approved):
Please, make sure that you update it based on Dan's suggesiton.

willm30:
@danbr When you go to ban an address that is prepopulated, there are a couple of renders where the initial values are actually `{userAddress: undefined}`, which means that with `!values.userAddress` you still get a flash. 

ban-user-flash
 

Is that ok or should I do something like `disabled={!isValid || isSubmitting || values.userAddress === ''}`?

chinins:
> @danbr When you go to ban an address that is prepopulated, there are a couple of renders where the initial values are actually `{userAddress: undefined}`, which means that with `!values.userAddress` you still get a flash.
> 
> 
> Is that ok or should I do something like `disabled={!isValid || isSubmitting || values.userAddress === ''}`?

That's fine as well

PR #3680, opened 2022-07-26
Restructure metadata parser - replacement

danbr:
## Description

PR replacement for colonyDapp#3535. The merged with the feature branch was screwed up!

Contributes to #3413


chinins (approved):
All good - same as before.

PR #3681, opened 2022-07-26
Restructure & extract IPFS metadata

### LINKED ISSUE #3413: Restructure and Extract IPFS Metadata Parser functionality

danbr:
## Description

This feature PR introduces Colony Metadata Event updates to restructure and extract IPFS Metadata Parser functionality

Resolves #3413


chinins (approved):
All good!

willm30 (approved):
Looks great. Tested by creating some metadata & annotations on master and switching to this branch, and didn't notice any issues.

PR #3683, opened 2022-07-26
Fix: syntax errors with spread operator

### LINKED ISSUE #3662: Fix syntax errors for FormattedMessage that uses types: string | MessageDescriptor
## Steps to reproduce

1. Go to any file listed here:

Screenshot 2022-07-21 at 09 55 36

2.  See error
Screenshot 2022-07-21 at 09 55 26

The error is displayed as the `FormattedMessage` prop could be either string | MessageDescriptor. The spread operator should only be used on MessageDescriptor. Guard against this.

Search & destroy other instances of this error.

willm30:
## Description

This PR fixes a few syntax errors. If a prop type is `string | MessageDescriptor`, you get an error because you can't spread a string.

**Changes** 🏗

* Implementing \<FormatMessage\> conditionaly if it's an object

Resolves #3662

### DIFF
--- a/src/modules/core/components/PermissionsLabel/PermissionsLabel.tsx
+++ b/src/modules/core/components/PermissionsLabel/PermissionsLabel.tsx
@@ -68,10 +68,10 @@ const PermissionsLabel = ({
     <Tooltip
       content={
         <div className={styles.tooltip}>
-          {infoMessage && infoMessage === 'string' ? (
-            infoMessage
-          ) : (
+          {infoMessage && typeof infoMessage === 'object' ? (
             <FormattedMessage {...infoMessage} values={infoMessageValues} />
+          ) : (
+            infoMessage
           )}
         </div>
       }
--- a/src/modules/dashboard/components/ActionsPageFeed/ActionsPageSystemInfo/ActionsPageSystemInfo.tsx
+++ b/src/modules/dashboard/components/ActionsPageFeed/ActionsPageSystemInfo/ActionsPageSystemInfo.tsx
@@ -14,16 +14,16 @@ export interface Appearance {
 
 interface Props {
   appearance?: Appearance;
-  tip?: MessageDescriptor | string;
+  tip: MessageDescriptor | string;
   tipValues?: UniversalMessageValues;
 }
 
 const ActionsPageSystemInfo = ({ tip, tipValues, appearance }: Props) => (
   <div className={getMainClasses(appearance, styles)}>
-    {tip && tip === 'string' ? (
-      tip
-    ) : (
+    {typeof tip === 'object' ? (
       <FormattedMessage {...tip} values={tipValues} />
+    ) : (
+      tip
     )}
   </div>
 );
--- a/src/modules/core/components/Fields/Radio/CustomRadio.tsx
+++ b/src/modules/core/components/Fields/Radio/CustomRadio.tsx
@@ -97,21 +97,19 @@ const CustomRadio = ({
         </div>
       )}
       <div className={styles.content}>
-        {label && (
-          <span className={styles.label}>
-            {label === 'string' ? (
-              label
-            ) : (
-              <FormattedMessage {...label} values={labelValues} />
-            )}
-          </span>
-        )}
+        <span className={styles.label}>
+          {typeof label === 'object' ? (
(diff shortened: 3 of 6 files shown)


arrenv (changes requested):
Nice @willm30, it looks to be working well.

I noticed a couple more though, as of the screenshot.

Screenshot 2022-07-26 181332

chinins (changes requested):
@willm30, I'm sorry for being a bit harsh but this is very poor quality. I have an impression that you have not tested it. It also shows that you don't understand how FormattedMessage works. This is not a solution. It crashes the app if you pass the string:
permisisons-demo

The solution is the following (Example from `PermissionsLabel`:
```
 {infoMessage && typeof infoMessage === 'object' ? (
            <FormattedMessage {...infoMessage} values={infoMessageValues} />
          ) : (
            infoMessage
          )}
          ```
Which produces the desired outcome
correct-string

Please, don't rush solutions but  first  understand the problem(s) that you are trying to solve and then provide quality result.

**Comment by chinins** [comment]
In the gifs you can't see my mouse but it's hovering on the permissions label and the app crashes because your change is in the tooltip (on hover).

**Comment by willm30** [comment]
@chinins Apologies, I should've been more thorough. Will keep in mind going forward. I have implemented your suggestions and followed the same pattern for the files @arrenv mentioned also needed changing.

**Review by ArmandoGraterol — CHANGES_REQUESTED** [review]
Nice work! While testing I didn't see any weird errors. I think there's some changes that could help tidy up the old and new code introduced in this PR :)

**Comment by willm30** [comment]
> Nice work! While testing I didn't see any weird errors. I think there's some changes that could help tidy up the old and new code introduced in this PR :)

Thanks. I have updated to reflect your suggestions -- not rendering the container if the data could be undefined, else not including the check at all.

**Review by ArmandoGraterol — APPROVED** [review]
Looks good to me! Thank you for implementing the changes :ok_hand:

File: src/modules/core/components/PermissionsLabel/PermissionsLabel.tsx [hunk 1]
```diff
@@ -68,10 +68,10 @@ const PermissionsLabel = ({
     <Tooltip
       content={
         <div className={styles.tooltip}>
-          {infoMessage && infoMessage === 'string' ? (
-            infoMessage
-          ) : (
+          {infoMessage && typeof infoMessage === 'object' ? (
```
danbr, line 71: why did you keep the truthy check on infoMessage? you removed this check on all other 
```
typeof ...  === 'object'
```

danbr (approved):
All looks very reasonable........ and i don't see spread operator errors any longer. Well done!!

src/modules/core/components/PermissionsLabel/PermissionsLabel.tsx
willm30: The rationale for this one is that infomessage could be undefined according to its typing, so I check for it. Whereas on the ones where I removed it, they are not optional according to their typings, therefore I can safely assume they won't be undefined.

src/modules/dashboard/components/ActionsPageFeed/ActionsPageSystemInfo/ActionsPageSystemInfo.tsx
```diff
@@ -14,16 +14,16 @@ export interface Appearance {
 
 interface Props {
   appearance?: Appearance;
-  tip?: MessageDescriptor | string;
+  tip: MessageDescriptor | string;
```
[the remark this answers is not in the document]
willm30: Yep that's what I was thinking

src/modules/core/components/PermissionsLabel/PermissionsLabel.tsx
chinins: If you do `typeof undefined` it will be 'undefined' and definitely won't equal `string` or `object` so it's really not necessary

chinins: I take my words back. It's actually correct. I've misread the code initially

chinins (approved):
All good!

Julianahlk (approved):
LGTM 👍

arrenv (approved):
Looks good to me Will :)

PR #3684, opened 2022-07-26
Sync single user picker with Address Book in award/smite dialogs

### LINKED ISSUE #3641: With Address Book enabled, all SingleUserPicker fields should filter members
## Steps to reproduce

1. Add a Member to the Address Book
2. Go to the Smite or Award modals
3. Try to select a member using the `SingleUserPicker`
4. The picker will show all members and not the filtered members as of the Address Book

This should apply to all `SingleUserPicker` fields, so this filter may work better at a higher level.

## Expected behaviour

* All `SingleUserPicker` fields should filter members when the Address Book is enabled

## Actual behaviour

* The `SingleUserPicker` is not filtering addresses in the Address Book

user-picker-filter

willm30:
## Description

This PR  ensures that when the Address Book is enabled, the single user picker in the Award/Smite dialogs shows verified members instead of all members. When the Address Book is inactive, or no members have been added, the single user picker in both components should show all colony members. 

Note: this was already happening in the Create Payment dialog. I have extracted this logic for reuse in the Award/Smite components.
 
To test: 

1. With no members added, verify that the `SingleUserPicker` in the Award/Smite show all users in the colony. 
2. Add a member or two to the address book. Now verify the `SingleUserPicker` only shows those members. 
3. Turn off the address book. Now verify that the `SingleUserPicker` is showing all users again.

**Changes** 🏗

`CreatePaymentDialog.tsx`: Extracted the `verifiedUsers` logic for reuse in award/smite.
`ManageReputationContainer.tsx`: Added the `ColonyMembers` prop so that when the component loads, `ColonyMembers` is defined. This ensures that the selected user doesn't rerender after the component has been opened and therefore avoids a "flashing" in the ui. Also added the `verifiedUsers` logic. 
`ColonyHomeActions.tsx`: I moved the ColonyMembers query here because this contains the loading component nearest to the ManageReputation Container.
`useSelectedUser.ts`: updated to take an array of `AnyUser` (making it compatible with an array of verified or subscribed users)

**Additions**

`verifiedRecipients.ts`: Extracted logic for filtering subscribed (i.e. all) users by whether they're verified or not.

Resolves #3641

**Note:** The issue says that this should apply to all `SingleUserPicker` instances however after checking this with Arren we agreed that it should only apply to the Payments and Award/Smite dialogs for now (see issue comments).

### DIFF
--- a/src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
+++ b/src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, useState } from 'react';
+import React, { useCallback, useMemo, useState } from 'react';
 import { FormikProps } from 'formik';
 import * as yup from 'yup';
 import { ROOT_DOMAIN_ID } from '@colony/colony-js';
@@ -14,6 +14,7 @@ import { ActionTypes } from '~redux/index';
 import { useMembersSubscription } from '~data/index';
 import { pipe, withMeta, mapPayload } from '~utils/actions';
 import { useSelectedUser } from '~utils/hooks/useSelectedUser';
+import { getVerifiedUsers } from '~utils/verifiedRecipients';
 
 import DialogForm from '../ManageReputationDialogForm';
 import {
@@ -51,6 +52,12 @@ const ManageReputationContainer = ({
   const { data: colonyMembers } = useMembersSubscription({
     variables: { colonyAddress },
   });
+
+  const subscribedUsers = colonyMembers?.subscribedUsers || [];
+
+  const verifiedUsers = useMemo(() => {
+    return getVerifiedUsers(colony.whitelistedAddresses, subscribedUsers) || [];
+  }, [subscribedUsers, colony]);
 
   const updateReputationCallback = (
     userRepPercentage: number,
@@ -127,8 +134,10 @@ const ManageReputationContainer = ({
     [totalReputationData, isSmiteAction],
   );
 
-  const selectedUser = useSelectedUser(colonyMembers);
-
+  const { isWhitelistActivated } = colony;
+  const selectedUser = useSelectedUser(
+    isWhitelistActivated ? verifiedUsers : subscribedUsers,
+  );
   return (
     <ActionForm
       initialValues={{
@@ -163,6 +172,9 @@ const ManageReputationContainer = ({
               isVotingExtensionEnabled={isVotingExtensionEnabled}
               back={() => callStep(prevStep)}
               ethDomainId={ethDomainId}
+              verifiedUsers={
+                isWhitelistActivated ? verifiedUsers : subscribedUsers
+              }
               updateReputation={
                 isSmiteAction ? updateReputationCallback : undefined
               }
--- a/src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationDialogForm/ManageReputationDialogForm.tsx
@@ -29,7 +29,6 @@ import {
   OneDomain,
   useUserReputationQuery,
   useLoggedInUser,
-  useMembersSubscription,
 } from '~data/index';
 import { useDialogActionPermissions } from '~utils/hooks/useDialogActionPermissions';
 import { useTransformer } from '~utils/hooks';
@@ -108,6 +107,7 @@ const MSG = defineMessages({
 interface Props extends ActionDialogProps {
   isVotingExtensionEnabled: boolean;
   nativeTokenDecimals: number;
+  verifiedUsers: AnyUser[];
   ethDomainId?: number;
   updateReputation?: (
     userPercentageReputation: number,
@@ -135,6 +135,7 @@ const ManageReputationDialogForm = ({
   ethDomainId: preselectedDomainId,
   isVotingExtensionEnabled,
   nativeTokenDecimals,
+  verifiedUsers,
   isSmiteAction = false,
 }: Props & FormikProps<ManageReputationDialogFormValues>) => {
   const { walletAddress, username, ethereal } = useLoggedInUser();
@@ -171,10 +172,6 @@ const ManageReputationDialogForm = ({
   );
 
   const inputDisabled = !userHasPermission || onlyForceAction;
-
-  const { data: colonyMembers } = useMembersSubscription({
-    variables: { colonyAddress },
-  });
 
   const { data: userReputationData } = useUserReputationQuery({
     variables: {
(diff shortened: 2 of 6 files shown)


danbr:
@willm30 Maybe my expectation is wrong?
The updates seem to be working except when I deactivate the address book. The list of potential awardees are still restricted.

Empty address book, with 1 contributor & 1 watcher
Screenshot 2022-08-15 at 15 11 22
With User2 added to address book & deactivated
Screenshot 2022-08-15 at 15 10 22
User options, expecting Dan & User2 to be displayed
Screenshot 2022-08-15 at 15 18 25

willm30:
> @willm30 Maybe my expectation is wrong? The updates seem to be working except when I deactivate the address book. The list of potential awardees are still restricted.

@danbr Did you refresh the browser after deactivating the address book? For me, I need to do this first, then the updates propagate and the form shows all users. (Incidentally, I would like to open an issue to update the fetch-policies of the relevant queries so a browser refresh isn't needed after adding a member to the address book to see the updated address book).

danbr:
> > @willm30 Maybe my expectation is wrong? The updates seem to be working except when I deactivate the address book. The list of potential awardees are still restricted.
> 
> @danbr Did you refresh the browser after deactivating the address book? For me, I need to do this first, then the updates propagate and the form shows all users. (Incidentally, I would like to open an issue to update the fetch-policies of the relevant queries so a browser refresh isn't needed after adding a member to the address book to see the updated address book).

@willm30 Thanks for the suggestion. I re-setup, and did several page refreshes and the behaviour remained the same with only 1 address shown. When she is available, i will ask @chinins to test, perhaps it will work for her.

src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -163,6 +169,9 @@ const ManageReputationContainer = ({
               isVotingExtensionEnabled={isVotingExtensionEnabled}
               back={() => callStep(prevStep)}
               ethDomainId={ethDomainId}
+              verifiedUsers={
+                (isWhitelistActivated && verifiedUsers) || subscribedUsers
+              }
```
willm30, line 174: @danbr Hmm, interesting. I am unable to reproduce. With a contributor and a watcher, I add the watcher to the address book, then deactivate the address book via the toggle switch, and after a browser refresh on the ColonyHome page I can see both users in the Award/Smite and CreatePayment dialogs. 

This is where I implement the check for whether the whitelist is activated. It should pass the form (and therefore the Single User Picker) all users (i.e `subscribedUsers`) if `isWhitelistActivated` is false (even if there is a member added to the address book).  Maybe you can spot something I'm overlooking?

danbr: @willm30 Finally, I deleted my local branch and now its behaving as expected, thankfully!! I'll review the code next....

danbr (approved):
Looks good. Working well..... Nice job.

src/modules/dashboard/components/ColonyHomeActions/ColonyHomeActions.tsx
```diff
@@ -79,6 +80,13 @@ const ColonyHomeActions = ({ colony, ethDomainId }: Props) => {
     variables: { address: colony.colonyAddress },
   });
 
+  const {
+    data: colonyMembers,
+    loading: isLoadingColonyMembers,
+  } = useMembersSubscription({
+    variables: { colonyAddress: colony.colonyAddress },
+  });
```
chinins, line 88: First of all, the approach should be similar in all cases - both for Smite/Award dialogs and for Create payment dialog which is not the case right now. There is no reason for the approaches to be different.

Secondly, I think that the query shouldn't be moved here. Will we now move all the queries from all of the dialogs here? It doesn't make sense. I think that it should removed from here and the same approach should be takes as for the Create payment dialog.

src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -48,9 +49,12 @@ const ManageReputationContainer = ({
...
   const history = useHistory();
 
-  const { data: colonyMembers } = useMembersSubscription({
-    variables: { colonyAddress },
-  });
+  const { subscribedUsers } = colonyMembers;
+
+  const verifiedUsers = useMemo(() => {
+    const verifiedAddresses = colony.whitelistedAddresses;
+    return getVerifiedUsers(verifiedAddresses, subscribedUsers);
```
chinins, line 56: You can just pass `colony.whitelistedAddresses` to the function, there is no need to create a variable.

src/modules/dashboard/components/Dialogs/AwardAndSmiteDialogs/ManageReputationContainer/ManageReputationContainer.tsx
```diff
@@ -163,6 +169,9 @@ const ManageReputationContainer = ({
               isVotingExtensionEnabled={isVotingExtensionEnabled}
               back={() => callStep(prevStep)}
               ethDomainId={ethDomainId}
+              verifiedUsers={
+                (isWhitelistActivated && verifiedUsers) || subscribedUsers
```
chinins, line 173: You should use ternary in cases like this (as per my other comments)

chinins (changes requested):
As per my direct comments.

src/modules/dashboard/components/ColonyHomeActions/ColonyHomeActions.tsx
willm30: @chinins The intention here is to ensure that `colonyMembers` is defined when the component mounts. If it isn't, as is the case when you run the query in the component, you can get a slight flashing in the UI, as below (see recipient field): 

flashingRep

If the colonyMembers data is undefined, then the selectedUser will also be undefined. Once the query returns, the component rerenders and the selectedUser gets set properly -- hence the flash. 

The reason this wouldn't be necessary in the CreatePayment dialog is because no user is preselected, so the effect of colonyMembers being undefined initially is invisible -- by the time a user has opened the SingleUserPicker, the data is there. 

What do you think? I'm easy either way, but in general prefer to avoid UI flashes, even small ones.

chinins: To be honest, I don't mind this flash. I think we shouldn't be add queries in improper places just to avoid it. You can have a loading spinner instead for the whole dialog if you want. And then when we move to the cdapp with new react features there either won't be any flashing at all, just a tiny delay in showing the form, or you will have a spinner for a couple of seconds.

willm30:
> As per my direct comments.

@chinins Have updated and restored the Memebrs Query to the component. Please review when you get a moment.

Julianahlk (approved):
It works well for me!

ArmandoGraterol (approved):
Working as expected! Nice job

chinins (approved):
Very nice - much cleaner, I prefer it this way.

PR #3609, opened 2022-07-11
Add edits in the locked state

agnieszkajarosikloj:
## Description

This PR adds ability to edit a locked expenditure, after clicking edit icon. The state of the expenditure is revert to an editable. Also adds a modal where user can confirm the changes and displays newly created linked motion. UI only.
[Figma Link - Showing edit icon](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2416%3A313415)
[Figma Link - Showing edit state](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2416%3A314175)
[Figma Link - Showing confirm edit modal](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2416%3A314175)
[Figma Link - Showing linked edit motions](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=2420%3A1074809)

**New stuff** ✨

* Edit button that reverts to an editable state
Screenshot 2022-07-14 at 14 48 48
* Modal where user can confirm changes
Screenshot 2022-07-14 at 15 20 52

**Changes** 🏗

* The Stages component has been changed to separate the Stages inside the form from the one with locked values. It has been done to avoid calling the 'useFormikContext' hook outside of formik context.
* StakeExpenditureDialog has been changed - used only Toggle with correct props
Screenshot 2022-07-27 at 18 41 33

Resolves #3492

PR #3693, opened 2022-07-28
Add split locked

### LINKED ISSUE #3671: Expenditures: As a user I can display a locked state
[Provided by Adrian via Contractors project management platform]

## Description
User has to be able to display state where split funds are locked.

## Acceptance Criteria

- an option to display and click an edit icon
- an option to display total amount of funds and currency
- an option to display recipient and his share with currency
- an option to display updated log event
- an option to display updated stage status
- equal/unequal toggle should not be visible

## Design

- [Figma Link - Split locked state](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=3746%3A315142)

download (61)

agnieszkajarosikloj:
## Description

This PR adds locked state of split payment.

Screenshot 2022-08-24 at 12 06 10

[Figma Link - Split locked state](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=3746%3A315142)

## TODO

- add edition in locked state after adding this [PR](https://github.com/JoinColony/colonyDapp/pull/3609)

Resolves #3671


ArmandoGraterol (changes requested):
Solid work! I just noticed some small things:
FireShot Capture 699 - Create Expenditure - Colony - wonker - localhost
FireShot Capture 700 - Create Expenditure - Colony - wonker - localhost

1. The recipient's containers look like they have different padding for the top and bottom, so it doesn't look centered.
2. We could use more space between the token icon and the amount, currently it's `4px`, if we use `8px` it could look closer to the specs.
3. The amount label doesn't look aligned to me, but only by like `1px` - `2px`. Might be screen/OS but it's noticeable
4. With really long usernames it goes out of the container. We should establish a width in which we should cut off the name with an ellipsis, like how we do it in other places in the dapp.

arrenv (approved):
Looks good to me @agnieszkajarosikloj

chinins (approved):
Well done!

src/modules/dashboard/components/ExpenditurePage/Split/LockedSplit.tsx
```diff
@@ -0,0 +1,123 @@
...
+                        token={token}
+                        name={token.name || token.address}
+                      />
+                      <Numeral
+                        unit={getTokenDecimalsWithFallback(0)}
+                        value={recipient.amount || 0}
+                      />
+                      {token.symbol}
+                    </div>
+                  </>
```
ArmandoGraterol, line 110: The unnecessary empty bracket are still here

ArmandoGraterol (approved):
Excellent work! Just noticed that the empty brackets are still there, but other than that detail, this is good to go

PR #3700, opened 2022-07-31
Add support for multiple transaction preview

Julianahlk:
## Description

This PR adds the multiple transaction version of the preview dialog to Gnosis Control Safe Dialog. 

- Each transaction should be in its respective tab
- Each tab has a heading with the transaction number and type 
- Each tab can be opened or closed

**Design**

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=139%3A9825)

Screen Shot 2022-08-04 at 4 45 58 PM

**New stuff** ✨
Screen Shot 2022-08-01 at 12 33 08 AM

Screen Shot 2022-08-02 at 9 36 07 PM

## TODO
- [x] Creating multiple transactions at the same time takes you to the "multiple transaction" version of the preview dialog
- [x] Each tab has a heading with the transaction number and type
- [x] Each tab can be closed or opened
- [x] Each tab shows the info previously added relevant to the transaction

(Resolves | Contributes to) #3564


willm30:
@Julianahlk This is working really well from my testing. 

Just have some notes re the design. Comparing the screenshot below to the figma, the title looks to have a bit too much space beneath it, and the colour of the opened heading is the same grey-blue in the design as the closed header. Also the button text is "Confirm" in the design, instead of "Create Transaction". 

multipl-preview

Design: 

design-multiple

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,173 @@ interface Props {
...
+
+  const handleTabToggle = useCallback(
+    (newIndex: number) => {
+      const newTransactionTabs = transactionTabStatus.map((tab, index) =>
+        index === newIndex ? !tab : tab,
+      );
+      setTransactionTabStatus([...newTransactionTabs]);
+    },
+    [transactionTabStatus, setTransactionTabStatus],
+  );
```
willm30, line 163: What's the value in memoising this function? As far as I can tell, it gets re-run every render anyway because inside it you call `setTransactionTabStatus` and have `transactionTabStatus` as one of its dependencies.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,173 @@ interface Props {
...
+    (transactionTypeValue: string) => {
+      const transactionType = transactionOptions.find(
+        (transaction) => transaction.value === transactionTypeValue,
+      );
+      return transactionType ? (
+        <FormattedMessage {...transactionType.label} />
+      ) : null;
+    },
+    [],
+  );
```
willm30, line 174: Same question here re memoisation. Putting a console.log in the function shows it gets re-run every render, so not sure why we'd want to memoise it. The `Heading` that consumes this function also gets recreated every render because the `values.transactions.map` on line 195 is re-run every render, creating new props for that component.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
Julianahlk: For handleTabToggle the idea was to only rerun this function when the user clicks to open/close a tab but I see your point. I was originally following the pattern I found in GnosisControlSafeForm.tsx as that component works very similar and contains the same function. Do you agree it makes sense to not have the useCallback there as well? Line 209 GnosisControlSafeForm.tsx

Julianahlk:
> @Julianahlk This is working really well from my testing.
> Just have some notes re the design. Comparing the screenshot below to the figma, the title looks to have a bit too much space beneath it, and the colour of the opened heading is the same grey-blue in the design as the closed header. Also the button text is "Confirm" in the design, instead of "Create Transaction".

Thanks Will! I did notice some inconsistencies regarding the tab colors designs in Figma which is why I left the tabs without the grey-blue color when opened. For example:

I see the grey-blue heading in some of the screens is not always there, mostly when the tab is opened:
Screen Shot 2022-08-03 at 2 03 02 PM
Screen Shot 2022-08-03 at 2 03 59 PM

But for consistency, I will make the changes so the tabs remain grey-blue even when closed but what do you think about the cases above?

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
willm30: I got it wrong in my first comment, it's not being re-run _every_ render. I meant to say that it seems to be run the same number of times with and without the usecallback.

And yes, would have the same question re the rationale for memoisation in other similar cases.

willm30:
> But for consistency, I will make the changes so the tabs remain grey-blue even when opened but what do you think about the cases above?

I think that is one for @arrenv -- what colour should the heading be when the preview transaction tab is open?

ArmandoGraterol (changes requested):
Overall, looking good! 
FireShot Capture 684 - Actions - Colony - wonker - localhost

I think we need to reduce space in some places, like between the transaction title and transaction values, between the values of one transaction and the title of the one below, and between the dialog's title and the first transaction title. Also, we should remove the border in the last item value when it's not the last transaction, so as per the screenshot, there shouldn't be a border-bottom between the `Amount` field and the transaction title.

arrenv:
> > But for consistency, I will make the changes so the tabs remain grey-blue even when opened but what do you think about the cases above?
> 
> I think that is one for @arrenv -- what colour should the heading be when the preview transaction tab is open?

@Julianahlk @willm30 There are inconsistencies in the design, all screens should follow the styling as of this screen - https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=898%3A20492

i.e. same colour font for text, blue background for each transaction in closed or open state, the number format being `1.`, `2.` etc.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,173 @@ interface Props {
...
+    },
+    [transactionTabStatus, setTransactionTabStatus],
+  );
+  const getTransactionTypeLabel = useCallback(
+    (transactionTypeValue: string) => {
+      const transactionType = transactionOptions.find(
+        (transaction) => transaction.value === transactionTypeValue,
+      );
+      return transactionType ? (
+        <FormattedMessage {...transactionType.label} />
```
danbr, line 170: If you use `formatMessage` then you will not need to cast the return value of getTransactionTypeLabel

```
return transactionType ? formatMessage(transactionType.label) : null;
```

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,173 @@ interface Props {
...
+                  theme: 'dark',
+                }}
+                text={MSG.transactionTitle}
+                textValues={{
+                  transactionNumber: index + 1,
+                  transactionType: getTransactionTypeLabel(
+                    transaction.transactionType,
+                    /* Need casting as `formatValues` func used in Heading
+                            doesn't have correct types (coming from react-intl) */
+                  ) as PrimitiveType,
```
danbr, line 216: See comment above...

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,173 @@ interface Props {
...
+              [styles.tabContentClosed]:
+                values.transactions.length > 1 && !transactionTabStatus[index],
+            })}
+          >
+            <DialogSection>
+              <div className={styles.transactionDetailsSection}>
+                {values.transactions[index].transactionType !== '' &&
+                  values.transactions.length === 1 && (
+                    <div className={styles.transactionTitle}>
+                      1.{' '}
```
danbr, line 245: I see that @arren has already answered a question about style & format, so this will be the standard numbering approach.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,173 @@ interface Props {
...
-              label={label}
-              value={value(values.transactions[0][key], token)}
-            />
-          ))}
+                {values.transactions[index].transactionType !==
+                  TransactionTypes.CONTRACT_INTERACTION && (
+                  <DetailsItem
+                    label={MSG.to}
+                    value={
+                      <AddressDetailsView
```
danbr, line 267: This component needs to be refactored (not in this PR). Like SingleUserPicker, <AddressDetailsView> should be generic. 
You're having to make lots of casts be able to pretend to be a user. But the component should handle, User, Safe, NFT.... etc.

Also this component never displays an avatar for me. Perhaps this is due to my dev env. I check and saw that the avatars are shown in other peoples screen captures.
Screenshot 2022-08-04 at 15 54 31

danbr (changes requested):
Looking good....couple of minor things, I left several comments.

Julianahlk:
> I think we need to reduce space in some places, like between the transaction title and transaction values, between the values of one transaction and the title of the one below, and between the dialog's title and the first transaction title. Also, we should remove the border in the last item value when it's not the last transaction, so as per the screenshot, there shouldn't be a border-bottom between the `Amount` field and the transaction title.

Thanks @ArmandoGraterol ! I will reduce the spaces as you requested. Just letting you know Arren provided the actual design we need to follow [here](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=898%3A20492) since there has been some inconsistencies in the designs.

Julianahlk:
> @Julianahlk @willm30 There are inconsistencies in the design, all screens should follow the styling as of this screen - https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=898%3A20492
> i.e. same colour font for text, blue background for each transaction in closed or open state, the number format being `1.`, `2.` etc.

I was very confused with the designs and the number format but now I have a clear path. Thanks @arrenv!

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/DetailsItem/DetailsItem.tsx
```diff
@@ -1,21 +1,30 @@
...
-  <div className={styles.detailsItem}>
-    <div className={styles.detailsItemLabel}>
-      <FormattedMessage {...label} />
+  isLastItem?: boolean;
+}) => {
+  return (
+    <div
+      className={classnames(styles.detailsItem, {
+        [styles.separator]: !isLastItem,
+      })}
```
ArmandoGraterol, line 20: Couldn't you have done a `.detailsItem:last-child` to select the element that doesn't need the border? That way you avoid the extra prop and logic

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -119,84 +138,174 @@ interface Props {
...
-        <Annotations label={MSG.description} name="annotation" />
-      </DialogSection>
+      ))}
+      <div className={styles.footer}>
+        <DialogSection appearance={{ theme: 'sidePadding' }}>
+          <Input
+            appearance={{ colorSchema: 'grey', theme: 'fat' }}
+            label={MSG.transactionsTitle}
+            name="transactionsTitle"
+            placeholder=""
```
ArmandoGraterol, line 302: I know that you didn't introduce this but can you remove this empty placeholder? :) I don't think it's needed

ArmandoGraterol (commented):
The redesign is looking good! However, I'm seeing something strange:
FireShot Capture 685 - Actions - Colony - wonker - localhost

For me the carets of the tabs are now transparent.

Also, there's still too much space between the transaction title and its value, and the values of one transaction and the transaction title below it

Julianahlk:
> For me the carets of the tabs are now transparent.
@ArmandoGraterol I investigated this issue and it looks like the name of the caret's SVG file was changed from "caret-up-small" to "caret-up"during the merge to Master, so I fixed the name to match the change and it should be working now :) 

> Also, there's still too much space between the transaction title and its value, and the values of one transaction and the transaction title below it
I made the changes for this as well. Let me know what you think. Thanks.

ArmandoGraterol (approved):
Looking really good! Nice work Juliana
FireShot Capture 686 - Actions - Colony - wonker - localhost

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
Julianahlk: @willm30 I made the changes and removed the useCallback function. I also, confirmed with Olga that in some cases like in this one, the useCallback is not necessary.

Julianahlk:
> Looking really good! Nice work Juliana !

Thanks Armando 👍

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
Julianahlk: Hey Dan, I made the changes you requested. Please take a look whenever you get a chance. Thanks.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/SafeTransactionPreview.tsx
Julianahlk: > This component needs to be refactored (not in this PR). Like SingleUserPicker, should be generic. You're having to make lots of casts be able to pretend to be a user. But the component should handle, User, Safe, NFT.... etc.
I created a new task for the refactoring [here](https://github.com/JoinColony/colonyDapp/issues/3730).

> Also this component never displays an avatar for me. Perhaps this is due to my dev env. I check and saw that the avatars are shown in other peoples screen captures. Screenshot 2022-08-04 at 15 54 31
Can you please check again? It seems to be working for everyone else.

danbr (approved):
Looking good.... well done.
I'm now getting my avatar displayed!!!

PR #3704, opened 2022-08-02
Simple Decisions

danbr:
## Description

This is the feature branch for "Simple Decisions".

Figma designs: https://www.figma.com/file/oRaTwsD5BUH3FJNzoZdSCv/Discussions?node-id=1018%3A54646

Notion docs with useful info: https://www.notion.so/colony/Simple-Decisions-2d869923870640a5889f9f24b183c65b

Screenshot 2022-08-02 at 12 02 30
Screenshot 2022-08-02 at 12 16 47
Screenshot 2022-08-02 at 12 16 55

Screenshot 2022-08-02 at 12 17 02
Screenshot 2022-08-02 at 12 17 12

Screenshot 2022-08-02 at 12 17 23


willm30 (commented):
Great feature!

Noticed some small things:

1. The font size of the h3 in the preview is 14px, whereas it's 16px in the create decision form. I think the user would expect the preview to reflect exactly the text they enter into the form.

font14



2. There is a space missing before "Create a new Decision"
space-missing

3. Create a draft with one user. Switch user. Click "create new decision". The form is not clean, which I would expect it to be. 

wron

danbr:
> 2\. There is a space missing before "Create a new Decision"

@willm30 Good catch.... Some non-obvious bugs.

1. Is fixed with #3985
2. looks find on my system, its a button so some margin is added rather than a space.
Screenshot 2022-10-13 at 15 01 28

3. is being resolved in https://github.com/JoinColony/colonyDapp/issues/3978

danbr:
> > 2. There is a space missing before "Create a new Decision"
> 
> @willm30 Good catch.... Some non-obvious bugs.
> 
> 1. Is fixed with [Decisions: Handle Decisions heading formatting in preview & staking pages #3985](https://github.com/JoinColony/colonyDapp/pull/3985)
> 2. looks find on my system, its a button so some margin is added rather than a space.
> 
> Screenshot 2022-10-13 at 15 01 28
> 
> 3. is being resolved in [Decisions: local storage is not correctly associating with a user when storing a Decision #3978](https://github.com/JoinColony/colonyDapp/issues/3978)

I actually simplified 2. - margin is now removed. The fix is included in #3985
Screenshot 2022-10-13 at 21 54 40

willm30 (commented):
I've tested again and can confirm all previous comments have been addressed. In my clicking around, I've gathered some more, quite minor notes. Not all may need actioning but I mention just in case.

### Drafts

1. Two users can't both have draft decisions simultaneously. If I open a draft in Account 1 and switch to Account 2 and open a draft, then switch back to Acccount 1, the draft in Account 1 disappears. I would expect drafts to be preserved across accounts.

### The editor

1. When you have a list in the editor, if you set a colour for a list item and then hit "Enter" to move to the next list item, it resets the colour. You can then highlight the whole list if you want it all one colour, so this doesn't prevent a user from have an entire list the same colour, it was just something I wasn't expecting. 

2. When I'm in the title input, if I hit "tab", it doesn't focus the textarea, but the toolbar. It would be a nice convenience to jump straight into the textarea when hitting tab from the title input. 

### The motion

1. There's an unfinished tooltip [TO BE ADDED WHEN AVAILABLE]. 
tooltip2

2. I get a warning in the console when I click the IPFS refresh link after first publishing a decision. You might have seen this already but just in case:

warning

danbr:
> I've tested again and can confirm all previous comments have been addressed. In my clicking around, I've gathered some more, quite minor notes. Not all may need actioning but I mention just in case.
> 
> ### Drafts
> 1. Two users can't both have draft decisions simultaneously. If I open a draft in Account 1 and switch to Account 2 and open a draft, then switch back to Acccount 1, the draft in Account 1 disappears. I would expect drafts to be preserved across accounts.
> 
> ### The editor
> 1. When you have a list in the editor, if you set a colour for a list item and then hit "Enter" to move to the next list item, it resets the colour. You can then highlight the whole list if you want it all one colour, so this doesn't prevent a user from have an entire list the same colour, it was just something I wasn't expecting.
> 2. When I'm in the title input, if I hit "tab", it doesn't focus the textarea, but the toolbar. It would be a nice convenience to jump straight into the textarea when hitting tab from the title input.
> 
> ### The motion
> 1. There's an unfinished tooltip [TO BE ADDED WHEN AVAILABLE].
>    tooltip2
> 2. I get a warning in the console when I click the IPFS refresh link after first publishing a decision. You might have seen this already but just in case:
> 
> warning

@willm30 Thanks for your comments. 
In general we are close to delivering this feature, so unless we find something significant i would be inclined to deliver and put it into the hands of our users. I'm sure we will find minor things to tweak and improve on, which can be delivered later.

To reply:

1. I spoke with @chinins - and this was the agreed implementation. This feature was not originally included in the product spec and since proposed its grown in complexity. We want to keep this feature simple so consequently, I dont think we care gonna cover all scenarios,, at least in the first version. Perhaps this extra functionality will be requested by users and then implemented.

2 & 3. They are quite minor. I can create issues to fix in the future, but i dont think it should hold up delivery of this feature.

4 & 5. These issue currently exist in Master.

PR #3534, opened 2022-06-24
Refactor and add filters to members page

chinins:
## Description

This PR adds filters to the members page. It also refactors the code a bit to split and simplify it. I've also added fixed width to the MembersListItem so that watchers item list doesn't shrink when there are no contributors.

Please, note that the hover state of the `reset` button for the filters is not as in the specs. It was agreed with Arren that it will stay the same as all other buttons above.

Update figma specs
https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=7705%3A91695

Screenshot 2022-07-01 at 16 08 23

**New stuff** ✨

- add filters checkboxes and logic

**Changes** 🏗

- extract MemberControls to a separate component
- fix the MembersListItem width

resolves #3381

PR #3738, opened 2022-08-11
Update useCallback dependency array in ManageWhiteListDialog

### LINKED ISSUE #3588: Bug: Can only have one verified address at once
## Steps to reproduce

1. Switch to base feature branch from master, ensure you have no members verified to start with
2. Add a member to the address book and refresh so you can see the tick confirming they're verified
3. Add another member to the address book and refresh
4. The first member has lost their tick. 

## Expected behaviour

* Should be able to verify multiple members

## Actual behaviour

* Can only verify one member at any given time

## GIF

losing-verification

willm30:
## Description

A while ago, a noticed a bug when adding a member to the address book. It would override any previous members I had added, such that I could only have one verified member at a time. While working on another issue I stumbled across what I believe is the fix for this. 

**Changes** 🏗

* Ensure useCallback dependency array is complete in transform function.

## Before
**Adding a new member replaces existing**
one-mem-only

## After
**Adding a new member does not replace existing**
multiple-members

Resolves #3588

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/ManageWhitelistDialog/ManageWhitelistDialog.tsx
@@ -157,7 +157,7 @@ const ManageWhitelistDialog = ({
       ),
       withMeta({ history }),
     ),
-    [tabIndex],
+    [tabIndex, storedVerifiedRecipients],
   );
 
   return (


Julianahlk (commented):
This is working as expected for me 👍. 

Questions:
- Are you getting this warning message "Cannot update a component (`dashboard.ColonyMembers`) while rendering a different component (`Formik`)"?
I only see it every time I interact with members. 
- Also, is it normal that I have to refresh the page again to be able to see the tick?
addressbook1
addressbook1

willm30:
@Julianahlk 

> * Are you getting this warning message "Cannot update a component (`dashboard.ColonyMembers`) while rendering a different component (`Formik`)"? I only see it every time I interact with members.

Yes, see: https://github.com/JoinColony/colonyDapp/pull/3534#issuecomment-1173681104

> * Also, is it normal that I have to refresh the page again to be able to see the tick?

I would like to address this in a separate issue as I have noticed it as well. I believe it can be solved by updating the fetch policy of the relevant query in the Members component.

Julianahlk:
> I would like to address this in a separate issue as I have noticed it as well. I believe it can be solved by updating the fetch policy of the relevant query in the Members component.

Sounds good @willm30! I think if we fix this it will be hugeee. Thank you!

Julianahlk (approved):

ArmandoGraterol (approved):
Working as expected! Good job

PR #3745, opened 2022-08-15
Remove references to showWhiteListWarning

### LINKED ISSUE #3744: Remove `showWhitelistWarning` element from `CreatePaymentDialogForm`
## Description

Currently, when using the `CreatePaymentDialog`, if the address book is active, the `SingleUserPicker` only shows members that have been added to the address book.

However, in the `CreatePaymentDialog`, there is logic to determine whether, when the address book is activated, the selected user is in the address book. Since the `SingleUserPicker` is now filtered according to whether members are in the address book, this check is redunant and can be removed.  

To confirm this, try and generate the warning. To do this, you would need the address book to be active, and to select a user that is not on it. However, you can't, because the `SingleUserPicker` will only show you members in the address book. 

## Screenshot

`showWhitelistWarning` and all references to it should be removed.

showwhitelistwarning

showwhitelistwarning

This issue is related to, and was discovered when working on, #3641

willm30:
## Description

When working on #3684, I noticed a small piece of logic that is now reduntant. In the `CreatePaymentDialog`, when the address book is activated, we don't need to check if the selected address is not in the address book, because the SingleUserPicker will only display users in the address book. 

I have removed the related logic and styles.

If you want to verify this, on `master`, try and generate the warning. To do this, you would need the address book to be active, and to select a user that is not on it. However, you can't, because the SingleUserPicker will only show you members in the address book.

**Deletions** ⚰️

* `showWhiteListWarning` and related logic / styles

Resolves #3744

### DIFF
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialog.tsx
@@ -121,15 +121,6 @@ const CreatePaymentDialog = ({
     return getVerifiedUsers(colony.whitelistedAddresses, subscribedUsers) || [];
   }, [subscribedUsers, colony]);
 
-  const showWarningForAddress = (walletAddress) => {
-    if (!walletAddress) return false;
-    return isWhitelistActivated
-      ? !colonyData?.processedColony?.whitelistedAddresses.some(
-          (el) => el.toLowerCase() === walletAddress.toLowerCase(),
-        )
-      : false;
-  };
-
   const transform = useCallback(
     pipe(
       mapPayload((payload) => {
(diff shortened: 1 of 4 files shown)


ArmandoGraterol (commented):
What if the user copy-pastes the address of one of the members of the colony like I do here?
FireShot Capture 703 - Actions - Colony - wonker - localhost

Yeah, for the address I pasted I'm shown a warning but that's just the confusable library being silly, and probably something we should consider fixing... And if that's fixed then no warning will be shown regarding that the wallet address is not on the whitelist if we take into account your changes.

willm30:
> What if the user copy-pastes the address of one of the members of the colony like I do here? 

@ArmandoGraterol Hmm I didn't of that. Let's leave as is.

PR #3746, opened 2022-08-15
Fix 99% staking bug

### LINKED ISSUE #3739: Update Motions Slider with Max and Min buttons
### Description

Motions have had a 99% issue on the staking slider for almost 1 year with Shapeshift repeatedly reporting it and Colony recently experiencing it ourselves on the MetaColony. So, we need to prioritise a fix for it.

An alternate way of approaching this that we would like to implement is to add a `min` and a `max` button to the slider.

- The `min` button will be used to set the staking to the 10% threshold that allows a Motion to appear in the actions list.
- If the `min` 10% threshold of the Motion has already been met, then the `min` button will be disabled with a tooltip stating "Minimum threshold has already been met for this Motion.".

- The `max` button will be used to set the full value of the required stake. This will use the actual value and not the value set by using the slider, however, the slider should still fill up. (We are hoping this will fix the issue)
- If the user does not have enough to cover the required stake, then the `max` button will use all users available activated tokens.
- If the Motion already has some stake, then it will use the required stake minus the value of what is already staked.

You can see a reference to the issue history here for context - https://github.com/JoinColony/colonyDapp/issues/2733

### Design

[Figma link](https://www.figma.com/file/CCgGsBzhX3BUENS1IvgP7N/Motions-%26-Disputes?node-id=6692%3A186298)

View already staked sees

chinins:
## Description

This PR fixes the max staking in slider that in some cases produced only 99% staked.

It has been decided not to implement min and max buttons as it wouldn't help the bug and then the UI would get confusing and the slider logic would be even more complicated.

The bug is caused by rounding calculations in `Decimal`. I've added a direct check if the slider is 100% staked.

I've also added `log.verbose` with all the relevant staking values in case the bug persists and we need to investigate further in production.

This bug is hard to replicate, but what worked for me is either really large mounts or really small amounts and a combination of both (for reputation, so basically paying a user). I paid 1 billion and then 2 more tokens. You can try these amounts to replicate.

I'll try to replicate on QA, will add a link to the colony/motion if I manage to.

Resolves #3739

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.tsx
@@ -135,13 +135,21 @@ const StakingWidget = ({
     pipe(
       mapPayload(({ amount }) => {
         if (data?.motionStakes) {
-          const { minUserStake } = data.motionStakes;
+          const { minUserStake, maxUserStake } = data.motionStakes;
+
+          let finalStake;
           const stake = getDecimalStake(amount);
-          const stakeWithMin = new Decimal(minUserStake).gte(stake)
-            ? new Decimal(minUserStake)
-            : stake;
+
+          if (amount === 100) {
+            finalStake = maxUserStake;
+          } else if (amount === 0 || new Decimal(minUserStake).gte(stake)) {
+            finalStake = minUserStake;
+          } else {
+            finalStake = stake.toString();
+          }
+
           return {
-            amount: stakeWithMin.round().toString(),
+            amount: finalStake,
             userAddress: walletAddress,
             colonyAddress,
             motionId: bigNumberify(motionId),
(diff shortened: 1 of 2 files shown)


arrenv (commented):
It seems to make sense. I have not been able to reliably replicate. But after talking to Alex, we should be able to get it to way we think is the cause after 2 reputation mining cycles, which should happen after 2 payments. I will test again when I run the dev environment again later today.

ArmandoGraterol (approved):
Managed to reproduce the bug on master, but unfortunately, I had the monitor active (I had it on so that I would have more "random" values, it was a last effort of mine to replicate this bug), and I couldn't stay on the exact rep amount where I replicated it.

The logic looks good and from what I've been testing the slider works as expected in this branch, hopefully, the problems with the infamous 99% are solved by this.

arrenv (approved):
Again, a difficult one to replicate, but when I do, it seemingly seems resolved.

danbr (approved):
Looks reasonable..... I tried, but i couldn't provoke the issue.

arrenv:
Just noting here that we have been able to replicate this consistently on QA with 50,000 Minted and one payout of 40,000.

PR #3755, opened 2022-08-18
Implement contract functions parser

### LINKED ISSUE #3692: Add function parser to "Contract Interaction" transaction type
### Story

As a user, when I select a function to use for the interaction, if the function has parameters then a series of fields for each param should appear.

### Description

Each param of the function should be parsed in order to determine how many fields are going to be added to the form, the name, and the type of the param.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A24680)

FireShot Capture 680 - Safe Control – Figma - www figma com

- [ ] You select a function with params and fields appear for each param.
- [ ] Each field's label, placeholder, and validation corresponds to the data previously parsed.

### Suggestions

* Clearly getting functions to develop or test it is not possible right now, so a good alternative for placeholder data could be using the functions we have in the ABIs of Network, for example.

ArmandoGraterol:
Added parser for the "Contract interaction" transaction type which fetches the contract ABI and parses it in order to get the functions and their params.

FireShot Capture 694 - Actions - Colony - wonker - localhost
FireShot Capture 695 - Actions - Colony - wonker - localhost

- We can use the contracts here in order to test the parser: https://etherscan.io/contractsVerified
- In order to get the contract ABI, we need to use the etherscan API, which requires a API KEY. You can get 1 free following the steps here: https://docs.etherscan.io/getting-started/creating-an-account . After getting it, plug it in the `.env` file as `ETHERSCAN_API_KEY`
- Each generated field should have its own validation.

Resolves #3692

### DIFF
--- a/src/modules/dashboard/hooks/useContractABIParser.ts
+++ b/src/modules/dashboard/hooks/useContractABIParser.ts
@@ -1,5 +1,13 @@
+import { isEmpty } from 'lodash';
 import { useEffect, useState } from 'react';
-import { AbiItem, keccak256 } from 'web3-utils';
+import { AbiItem, isAddress, keccak256 } from 'web3-utils';
+
+import {
+  BINANCE_NETWORK,
+  GNOSIS_NETWORK,
+  GNOSIS_SAFE_NETWORKS,
+  POLYGON_NETWORK,
+} from '~constants';
 
 export interface AbiItemExtended extends AbiItem {
   name: string;
@@ -9,14 +17,35 @@ export interface AbiItemExtended extends AbiItem {
   signatureHash: string;
 }
 
-export const fetchContractABI = async (contractAddress: string) => {
+const fetchContractABI = async (
+  contractAddress: string,
+  safeChainId: number,
+) => {
   if (!contractAddress) {
     return [];
   }
 
   try {
-    const apiUri = `https://api.etherscan.io/api?module=contract&action=getAbi&address=${contractAddress}&apiKey=${process.env.ETHERSCAN_API_KEY}`;
-    const response = await fetch(apiUri);
+    const currentNetworkData =
+      GNOSIS_SAFE_NETWORKS.find((network) => network.chainId === safeChainId) ||
+      GNOSIS_NETWORK;
+    const getApiKey = () => {
+      if (currentNetworkData.chainId === POLYGON_NETWORK.chainId) {
+        return process.env.POLYGONSCAN_API_KEY;
+      }
+      if (currentNetworkData.chainId === BINANCE_NETWORK.chainId) {
+        return process.env.BSCSCAN_API_KEY;
+      }
+      if (currentNetworkData.chainId === GNOSIS_NETWORK.chainId) {
+        return process.env.GNOSIS_API_KEY;
+      }
+      return process.env.ETHERSCAN_API_KEY;
+    };
+    const getApiUri = () => {
+      const apiUri = `${currentNetworkData.apiUri}?module=contract&action=getabi&address=${contractAddress}`;
+      return `${apiUri}&apiKey=${getApiKey()}`;
+    };
+    const response = await fetch(getApiUri());
 
     if (!response.ok) {
       return [];
@@ -24,14 +53,14 @@ export const fetchContractABI = async (contractAddress: string) => {
 
     const responseJSON = await response.json();
 
-    return responseJSON.result;
+    return responseJSON.result || responseJSON.message;
   } catch (error) {
     console.error('Failed to retrieve ABI', error);
     return '';
   }
 };
 
-export const extractUsefulMethods = (abi: AbiItem[]): AbiItemExtended[] => {
+const extractUsefulMethods = (abi: AbiItem[]): AbiItemExtended[] => {
   const extendedAbiItems = abi.filter(
     ({ name, type }) => type === 'function' && !!name,
   ) as AbiItemExtended[];
@@ -61,33 +90,60 @@ export const extractUsefulMethods = (abi: AbiItem[]): AbiItemExtended[] => {
     });
 };
 
-export const useContractABIParser = (contractAddress?: string) => {
+export const useContractABIParser = (
+  contractAddress: string | undefined,
+  contractABI: string | undefined,
+  safeChainId: number,
+  handleContractABIChange: (abi: string) => void,
+) => {
   const [
     currentParsedContractAddress,
     setCurrentParsedContractAddress,
   ] = useState<string>('');
-  const [contractABI, setContractABI] = useState<string>('');
   const [usefulMethods, setUsefulMethods] = useState<AbiItemExtended[]>([]);
 
   useEffect(() => {
-    if (contractAddress && currentParsedContractAddress !== contractAddress) {
-      const contractPromise = fetchContractABI(contractAddress);
+    if (isEmpty(contractAddress) && currentParsedContractAddress) {
+      setCurrentParsedContractAddress('');
+      handleContractABIChange('');
+      setUsefulMethods([]);
+    }
+  }, [contractAddress, currentParsedContractAddress, handleContractABIChange]);
+
+  useEffect(() => {
+    if (
+      contractAddress &&
+      isAddress(contractAddress) &&
+      currentParsedContractAddress !== contractAddress
+    ) {
+      const contractPromise = fetchContractABI(contractAddress, safeChainId);
       contractPromise.then((data) => {
-        setContractABI(data);
+        handleContractABIChange(data);
       });
 
       setCurrentParsedContractAddress(contractAddress);
     }
-  }, [currentParsedContractAddress, contractAddress]);
+  }, [
+    currentParsedContractAddress,
+    contractAddress,
+    handleContractABIChange,
+    safeChainId,
+  ]);
 
   useEffect(() => {
     if (contractABI) {
-      setUsefulMethods(extractUsefulMethods(JSON.parse(contractABI)));
+      let parsedContractABI: AbiItem[];
+      try {
+        parsedContractABI = JSON.parse(contractABI);
+      } catch (error) {
+        console.error(error);
+        parsedContractABI = [];
+      }
+      setUsefulMethods(extractUsefulMethods(parsedContractABI));
     }
   }, [contractABI]);
 
   return {
-    contractABI,
     usefulMethods,
   };
 };
--- a/src/modules/dashboard/hooks/useContractABIParser.ts
+++ b/src/modules/dashboard/hooks/useContractABIParser.ts
@@ -1,0 +1,93 @@
+import { useEffect, useState } from 'react';
+import { AbiItem, keccak256 } from 'web3-utils';
+
+export interface AbiItemExtended extends AbiItem {
+  name: string;
+  type: 'function';
+  action: string;
+  methodSignature: string;
+  signatureHash: string;
+}
+
+export const fetchContractABI = async (contractAddress: string) => {
+  if (!contractAddress) {
+    return [];
+  }
+
+  try {
+    const apiUri = `https://api.etherscan.io/api?module=contract&action=getAbi&address=${contractAddress}&apiKey=${process.env.ETHERSCAN_API_KEY}`;
+    const response = await fetch(apiUri);
+
+    if (!response.ok) {
+      return [];
+    }
+
+    const responseJSON = await response.json();
+
+    return responseJSON.result;
+  } catch (error) {
+    console.error('Failed to retrieve ABI', error);
+    return '';
+  }
+};
+
+export const extractUsefulMethods = (abi: AbiItem[]): AbiItemExtended[] => {
+  const extendedAbiItems = abi.filter(
+    ({ name, type }) => type === 'function' && !!name,
+  ) as AbiItemExtended[];
+  const getMethodSignatureAndSignatureHash = (
+    method: AbiItem,
+  ): { methodSignature: string; signatureHash: string } => {
+    const params = method.inputs?.map((x) => x.type).join(',');
+    const methodSignature = `${method.name}(${params})`;
+    const signatureHash = keccak256(methodSignature).toString();
+    return { methodSignature, signatureHash };
+  };
+  const getMethodAction = ({ stateMutability }: AbiItem): 'read' | 'write' => {
+    if (!stateMutability) {
+      return 'write';
+    }
+    return ['view', 'pure'].includes(stateMutability) ? 'read' : 'write';
+  };
+
+  return extendedAbiItems
+    .map((method) => ({
+      ...getMethodSignatureAndSignatureHash(method),
+      ...method,
+      action: getMethodAction(method),
+    }))
+    .sort(({ name: a }, { name: b }) => {
+      return a.toLowerCase() > b.toLowerCase() ? 1 : -1;
+    });
+};
+
+export const useContractABIParser = (contractAddress?: string) => {
+  const [
+    currentParsedContractAddress,
+    setCurrentParsedContractAddress,
+  ] = useState<string>('');
+  const [contractABI, setContractABI] = useState<string>('');
+  const [usefulMethods, setUsefulMethods] = useState<AbiItemExtended[]>([]);
+
+  useEffect(() => {
+    if (contractAddress && currentParsedContractAddress !== contractAddress) {
+      const contractPromise = fetchContractABI(contractAddress);
+      contractPromise.then((data) => {
+        setContractABI(data);
+      });
+
+      setCurrentParsedContractAddress(contractAddress);
+    }
+  }, [currentParsedContractAddress, contractAddress]);
+
+  useEffect(() => {
+    if (contractABI) {
+      setUsefulMethods(extractUsefulMethods(JSON.parse(contractABI)));
+    }
+  }, [contractABI]);
+
+  return {
+    contractABI,
+    usefulMethods,
+  };
+};
--- a/.env.example
+++ b/.env.example
@@ -56,3 +56,6 @@ SUBGRAPH_WS_ENDPOINT=ws://127.0.0.1:8001/subgraphs/name/joinColony/subgraph
 METATRANSACTIONS=true
 # METATRANSACTIONS BROADCASTER ADDRESS
 BROADCASTER_ENDPOINT=http://127.0.0.1:3004
+
+# Add an etherscan key in order to allow contract interaction parsing
+ETHERSCAN_API_KEY=
--- a/.env.example
+++ b/.env.example
@@ -57,5 +57,11 @@ METATRANSACTIONS=true
 # METATRANSACTIONS BROADCASTER ADDRESS
 BROADCASTER_ENDPOINT=http://127.0.0.1:3004
 
-# Add an etherscan key in order to allow contract interaction parsing
+# Add API keys in order to allow contract interaction parsing
 ETHERSCAN_API_KEY=
+
+POLYGONSCAN_API_KEY=
+
+BSCSCAN_API_KEY=
+
+GNOSIS_API_KEY=
--- a/src/modules/constants.ts
+++ b/src/modules/constants.ts
@@ -12,6 +12,7 @@ export type NetworkInfo = {
   name: string;
   chainId: number;
   shortName: string;
+  apiUri: string;
   description?: string;
   displayENSDomain?: string;
   /**
@@ -85,6 +86,7 @@ export const GNOSIS_NETWORK: NetworkInfo = {
   contractAddressLink: 'https://blockscout.com/poa/xdai/address',
   rpcUrl: 'https://rpc.gnosischain.com',
   safeTxService: 'https://safe-transaction.xdai.gnosis.io/',
+  apiUri: 'https://blockscout.com/xdai/mainnet/api',
 };
 
 export const ETHEREUM_NETWORK: NetworkInfo = {
@@ -98,6 +100,7 @@ export const ETHEREUM_NETWORK: NetworkInfo = {
   contractAddressLink: 'https://etherscan.io/address',
   rpcUrl: 'https://mainnet.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161',
   safeTxService: 'https://safe-transaction.mainnet.gnosis.io/',
+  apiUri: 'https://api.etherscan.io/api',
 };
 
 export const GOERLI_NETWORK: NetworkInfo = {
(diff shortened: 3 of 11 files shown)


src/modules/dashboard/hooks/useContractABIParser.ts
```diff
@@ -0,0 +1,93 @@
...
+    const responseJSON = await response.json();
+
+    return responseJSON.result;
+  } catch (error) {
+    console.error('Failed to retrieve ABI', error);
+    return '';
+  }
+};
+
+export const extractUsefulMethods = (abi: AbiItem[]): AbiItemExtended[] => {
```
willm30, line 34: No need for export here

src/modules/dashboard/hooks/useContractABIParser.ts
```diff
@@ -0,0 +1,93 @@
...
+
+export interface AbiItemExtended extends AbiItem {
+  name: string;
+  type: 'function';
+  action: string;
+  methodSignature: string;
+  signatureHash: string;
+}
+
+export const fetchContractABI = async (contractAddress: string) => {
```
willm30, line 12: No need for export here

src/modules/dashboard/hooks/useContractABIParser.ts
```diff
@@ -0,0 +1,93 @@
...
+  if (!contractAddress) {
+    return [];
+  }
+
+  try {
+    const apiUri = `https://api.etherscan.io/api?module=contract&action=getAbi&address=${contractAddress}&apiKey=${process.env.ETHERSCAN_API_KEY}`;
```
willm30, line 18: This will only work for contracts on mainnet, presumably? If I can add a safe on Aurora, I should be able to interact with contracts on Aurora as well? Should there be compatibility for using other apis like https://aurorascan.dev/apis?

willm30 (changes requested):
Nice work! Very cool to see it coming together. 

I noticed a few things for your attention:

1. If you insert an invalid contract address it crashes the app
2. I don't think the validation messages are very user friendly at the moment, especially when it's expecting a number and you give it a string (which goes off the page)
3. Address array and boolean input types don't validate currently. See below (using: 0x3462DBcC45e049634A5eA6DB288da8aceD2192aF)
form-val-gnosis
4. The form is disabled even when every input is error-free
5. If I remove the entries in the ABI and contract address inputs, I can still interact with the contract functions. My sense would be that if you change the text in the ABI field, the functions dropdown should recalculate/update itself.
6. And final thought, I like the "read / write" label that is in the Gnosis Safe app. Makes scrolling through 15+ functions easier (as would a search bar but that could be a separate issue if it was thought to be useful). Maybe we could implement something similar?
rw

src/modules/dashboard/hooks/useContractABIParser.ts
arrenv: You could look at how Gnosis Safe do it.

The TextField - https://github.com/safe-global/safe-react/blob/c413702a0ad5887b356f6bc6f30aabcfcba1882d/src/routes/safe/components/Balances/SendModal/screens/ContractInteraction/ContractABI/index.tsx###

As well as processing the ABI to get the methods - here https://github.com/safe-global/safe-react/blob/c413702a0ad5887b356f6bc6f30aabcfcba1882d/src/logic/contractInteraction/sources/ABIService/index.ts

ArmandoGraterol: @arrenv @willm30 In conclusion, we are going to need 4 API keys in order to cover all the current networks:
- Polygonscan
- BSCSCAN (Binance)
- Gnosis BlockScout
- Etherscan (the rest of networks)

ArmandoGraterol:
@willm30 Thank you realli much for the feedback! I believe I've addressed everything you pointed out. The logic in this PR has a lot of moving parts so please let me know if you find something weird again

About point 6, that sounds like a good idea to me! It could be useful for a next iteration if Product also thinks it's a good idea.

src/modules/dashboard/hooks/useContractABIParser.ts
arrenv: @ArmandoGraterol That seems like a lot to deal with, is that what Gnosis Safe are doing?

ArmandoGraterol: @arrenv Gnosis Safe is not clear on what it does with most of the networks, at least it's not clear looking at the repo we are taking as an example as only the Rinkeby, Polygon, and Ether configs are there. Those 3 we know can share the etherscan API.

I recently tested making requests with the Gnosis, BSC, and Polygon WITHOUT API keys and:
- It seems like BlockScout allows not having an API key. But, according to the docs: 
`If My Account is enabled, users can add up to 3 API keys to ensure 10 request/second limits. When API Keys are not used, the global default is 50 requests/second.`

- BSC and Polygon also allow requests without API key but they come with this little message:
`K-Missing/Invalid API Key, rate limit of 1/5sec applied`

Re-tested all of them, just to be sure, with the etherscan API key: 
- Polygon and BSC outright tell you that the keys are invalid.
- BlockScout doesn't tell you if the API keys are wrong or right tbh, you can write anything in the API key param and it still gonna deliver the ABI, so I'm assuming they are only going to complain if we reach the limit I mentioned.

What do you think?

arrenv: @ArmandoGraterol I have not looked into it fully either. Just seems a little odd that you would need an API to be able to read the ABIs.

willm30: @ArmandoGraterol The following would suggest that Gnosis Safe just use the Etherscan api key: https://github.com/safe-global/safe-react/blob/c413702a0ad5887b356f6bc6f30aabcfcba1882d/src/config/index.ts#L137

willm30 (commented):
@ArmandoGraterol From what I can tell, the abi parser is working nicely. Just noticed a few things with the form:

1. I don't think the function input validation is working. I tried a few different data types and none were validating. There's also no validation on the target contract address field. 

2. When you click into the target contract address field, the address disappears and you need to go re-copy / paste the address if you didn't intend for that to happen. Maybe add the address to the `SingleUserPicker` dropdown so you can easily select recently inputted addresses?

3. When you select a function from the abi, the Create Transaction button is active, even if the function input fields are empty. 

4. Maybe show a loading spinner while the abi is being fetched?  I've noticed the fetching isn't instant and so wasn't sure for a second if I'd entered everything correctly. Also, if there's no network connection, there's currently no error in the UI so you might be confused about what's happening. 

5. This might be intentional but just mentioning in case: whenever you manually edit the text in the ABI field, it automatically returns as it was before (e..g if you delete it, it will come straight back).

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -47,7 +74,104 @@ const renderAvatar = (address: Address, item: AnyUser) => (
...
+    },
+    [
+      transactionFormIndex,
+      setFieldValue,
+      transactionValues.abi,
+      selectedContractMethods,
+      handleSelectedContractMethods,
+    ],
+  );
+  const selectedSafe = safes.find((safe) => safe.address === values.safe);
```
willm30, line 113: The "safe" value is an object, not a string, so this would be `values.safe.id`

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -47,7 +74,104 @@ const renderAvatar = (address: Address, item: AnyUser) => (
...
+
+  const usefulMethods: AbiItemExtended[] = useMemo(
+    () =>
+      getContractUsefulMethods(
+        contractAddress,
+        transactionValues.abi,
+        Number(selectedSafe?.chain) || GNOSIS_NETWORK.chainId,
```
willm30, line 120: In the mock data, the chain is the name, not chainId, so this will also default to Gnosis.

src/modules/dashboard/hooks/useContractABIParser.ts
ArmandoGraterol: Yeah, it only uses the etherscan API key. So with the comment I left above I was more or less asking if you guys thought that the limit is not a problem for us, because if that's the case then yeah we can survive with just the etherscan key

ArmandoGraterol:
@willm30 

> 1. I don't think the function input validation is working. I tried a few different data types and none were validating. There's also no validation on the target contract address field.

The validation for `address` should now be fixed. I'm assuming that's the only one you found? The others work for me.

The contract address field does have validation. Remember that Picker fields don't show error messages like the other inputs but they do have validation, at least this one has from what I've checked.

> 2. When you click into the target contract address field, the address disappears and you need to go re-copy / paste the address if you didn't intend for that to happen. Maybe add the address to the `SingleUserPicker` dropdown so you can easily select recently inputted addresses?

I would prefer for this to be its own issue as it's an improvement upon a core component and it's not on the scope. This issue is already packed as it is.

> 3. When you select a function from the abi, the Create Transaction button is active, even if the function input fields are empty.

Unfortunately, I tried a few solutions but I haven't been able to fix this. It seems like the validation doesn't react instantly to the new function input fields and we may need to do some hacky stuff to make it work properly. 

I will create a separate issue to track and tackle this at another time.

> 4. Maybe show a loading spinner while the abi is being fetched?  I've noticed the fetching isn't instant and so wasn't sure for a second if I'd entered everything correctly. Also, if there's no network connection, there's currently no error in the UI so you might be confused about what's happening.

Good idea! Right now implementing it is not as straightforward as I would like it to be so I will create an issue for it. (We may need to get a few more hooks going and rework the function that returns the useful methods).

> 5. This might be intentional but just mentioning in case: whenever you manually edit the text in the ABI field, it automatically returns as it was before (e..g if you delete it, it will come straight back).

Oh yeah, if you delete all of it then it should come back if you still have a contract address. However, you can, for example, remove a function and it should stay that way and also update the Select field accordingly.

ArmandoGraterol:
@willm30 Thank you for clarifying, the validation for bytes and uint/int should now be more fleshed out. Also increased by a bit the space between the error messages and the other fields

Julianahlk (changes requested):
Looking good Armando! Although, I found a few things:

1. When I click on the Target contract address input field without me actually writing anything I get the “it looks like I couldn’t find what you are looking for” message. My expectation is that there wouldn’t be any messages until the user had a change to write or paste something into the input.
Screen Shot 2022-09-01 at 12 11 47 AM

2. If you paste the target contract address but you don’t actually select it from the drop down and instead move to the ABI/JSON field and start typing or try to paste something, it doesn’t let you do anything. I would expect that you don’t need to select the address from the drop down. And even if you do need to select the address from the drop down option there is no visible error for the Target contract address input field. 
Screen Shot 2022-09-01 at 12 33 17 AM

3. When you do actually select the target contract address from the drop down, the address appears in read, which makes me think this is signaling that there is an error but there is no actual error message for the target contract address field. 

4. If you try to delete the text “contact source code not verified” from the ABI/JSON field, it adds the text back before you have the chance to type or paste anything my expectation would be that this text “contact source code not verified” would appear as an error message below the input box.|
Screen Shot 2022-09-01 at 12 37 38 AM

5. If you try to type in the ABI/JSON field while the target contract address field is empty, it briefly flashes an error message “Please enter a value” but it doesn’t let you type in the field. My expectation would be that an error would appear why you can’t type in the ABI/JSON field below the Target contract address field (assuming that’s the problem).

6.  After you fill out all the fields in the form, if you accidentally click on the target contact address field again it clears all the fields (Target contract address, ABI/JSON and Select function to interact with), which is a bit frustrating after spending the time to copy and pasting the address and the ABI/JSON.
Sep-01-2022 00-48-41

7. After selecting a function the create transaction button is active by default but as soon as I select the new option fields that show up below the function drop down the create transaction becomes disabled. The expectation is that the create transaction button would be already be disabled. 


8. If you delete the “contract source code not verified” text and immediately paste the ABI/JSON the text you just pasted will be replaced by the text “contract source code not verified”. In my opinion, this could be solved if that text was displayed as an error message and not inside the input box.

willm30:
> @willm30 Thank you for clarifying, the validation for bytes and uint/int should now be more fleshed out. Also increased by a bit the space between the error messages and the other fields

@ArmandoGraterol Nice one. I think it might be worth opening a separate issue to deal with the rest of the validation permutations because I'm realising that there are a few more e.g. int/uint8,16,24... in step 8 up to 256. Also I've seen for the first time today the [tuple](0xA45D5ddf6E1B98d55A5FE6FE54ffFB29149A3901) type, and I'm sure there are other less common ones. ([the official list](https://docs.soliditylang.org/en/v0.8.11/types.html))

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
```diff
@@ -63,25 +91,146 @@ type Props = DialogProps &
...
+        return yup.number().when('contractFunction', {
+          is: (contractFunction) =>
+            contractFunction === contractName || isArraySchema,
+          then: yup
+            .number()
+            .test(
+              'is-integer',
+              () => MSG.notIntegerError,
+              (value) => new Decimal(value || 0).isInteger(),
+            )
```
willm30, line 137: Think you could use `number().integer()` here?

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
```diff
@@ -63,25 +91,146 @@ type Props = DialogProps &
...
+            .test(
+              'is-integer-safe',
+              () => MSG.notSafeIntegerError,
+              (value) =>
+                new Decimal(value || 0).lte(new Decimal(MaxUint256.toString())),
+            )
```
willm30, line 143: When I enter `115792089237316195423570985008687907853269984665640564039457584007913129639935`, which is the max safe integer, I get an error that it's not safe, so maybe the Decimal library doesn't handle these large numbers so well?

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -74,16 +202,50 @@ const ContractInteractionSection = ({
...
+        >
+          <div className={styles.inputParamContainer}>
+            <Input
+              label={`${input.name} (${input.type})`}
+              name={`transactions.${transactionFormIndex}.${input.name}`}
+              appearance={{ colorSchema: 'grey', theme: 'fat' }}
+              disabled={disabledInput}
+              placeholder={`${input.name} (${input.type})`}
+              formattingOptions={
+                input.type === 'uint256' || input.type === 'int256'
```
willm30, line 237: If you changed this so it matched a type containing `uint/int` it would cover all the int/uint permutations. Then there'd only be the max safe value left to test in a separate issue.

willm30 (commented):
Just some small notes and then I think it'll be good. If you want a hand with creating some of the other issues I can assist 🙂

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
ArmandoGraterol: I didn't use it because `.integer()` uses `Number.isInteger()` in the background, which may not be as precise as using Decimal.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
ArmandoGraterol: We use Decimal for more complex calculations than this, so this ain't it. 

The value that arrives at this line of the code it's already not the same amount you inputted, from what I've tested it's coming rounded up, for some reason. So something is happening to the number earlier in the process.

ArmandoGraterol:
@Julianahlk Thank you for the feedback!
> 1. When I click on the Target contract address input field without me actually writing anything I get the “it looks like I couldn’t find what you are looking for” message. My expectation is that there wouldn’t be any messages until the user had a change to write or paste something into the input.
> 2. ... And even if you do need to select the address from the drop down option there is no visible error for the Target contract address input field.
> 3. When you do actually select the target contract address from the drop down, the address appears in read, which makes me think this is signaling that there is an error but there is no actual error message for the target contract address field.

These points are more about how the `SingleUserPicker` core component works than this issue/PR/feature. You could create some issues in `Maintenance`, perhaps check Figma to see if Product has updated the component, or if not then ask in the Product channel what they think of updating the component, I'm sure they will find this feedback useful if they haven't addressed this yet :)

> 2. If you paste the target contract address but you don’t actually select it from the drop down and instead move to the ABI/JSON field and start typing or try to paste something, it doesn’t let you do anything ...

Fixed!

> 4. If you try to delete the text “contact source code not verified” from the ABI/JSON field, it adds the text back before you have the chance to type or paste anything my expectation would be that this text “contact source code not verified” would appear as an error message below the input box.|

Fixed! That message should now appear as an `inputStatus`, which is used by Formik to inform the user of API's error and things like that.

> 5. If you try to type in the ABI/JSON field while the target contract address field is empty, it briefly flashes an error message “Please enter a value” but it doesn’t let you type in the field. My expectation would be that an error would appear why you can’t type in the ABI/JSON field below the Target contract address field (assuming that’s the problem).

This is the same thing as point 2, fixed! For future reference, there's no difference between pasting an address and not selecting something, and not interacting with the picker at all. In both cases, you are not setting a value until you actually click on an item in the dropdown, that's why this point is no different than point 2.

> 6. After you fill out all the fields in the form, if you accidentally click on the target contact address field again it clears all the fields (Target contract address, ABI/JSON and Select function to interact with), which is a bit frustrating after spending the time to copy and pasting the address and the ABI/JSON.

Well, ideally you should not need to copy and paste the ABI as that's fetched by selecting the contract's address, the only copy-pasted thing in the form should be the contract address. Regardless, I've made it so that it doesn't reset the form when the contract address goes from having a value to empty.

> 7. After selecting a function the create transaction button is active by default but as soon as I select the new option fields that show up below the function drop down the create transaction becomes disabled. The expectation is that the create transaction button would be already be disabled.
 
This is already addressed in one of my comments above as Will also noticed it:
>> Unfortunately, I tried a few solutions but I haven't been able to fix this. It seems like the validation doesn't react instantly to the new function input fields and we may need to do some hacky stuff to make it work properly.

The issue to fix this is [#3803](https://github.com/JoinColony/colonyDapp/issues/3803)

> 8. If you delete the “contract source code not verified” text and immediately paste the ABI/JSON the text you just pasted will be replaced by the text “contract source code not verified”. In my opinion, this could be solved if that text was displayed as an error message and not inside the input box.

Same bug as point 4, fixed.

willm30 (changes requested):
@ArmandoGraterol 

I'm seeing this "NOTOK" status when entering an address for the first time:

notok

Also, when the error "please enter a value" is showing beneath the ABI field, the parser seems to stop working. To reproduce, 

1. Select Pipe's Dream (Ethereum)
2. Enter an ethereum contract address
3. Delete the text in the ABI field, prompting the "please enter a value"
4. Select All Saints (Gnosis Chain)
5. Either re-enter an ethereum address or enter a gnosis chain address
6. In either case, nothing happens. 

Also, just bringing to your attention that the following is no longer happening, i.e. when you delete the abi it doesn't come back anymore. 
 
> > 5. This might be intentional but just mentioning in case: whenever you manually edit the text in the ABI field, it automatically returns as it was before (e..g if you delete it, it will come straight back).

> Oh yeah, if you delete all of it then it should come back if you still have a contract address. However, you can, for example, remove a function and it should stay that way and also update the Select field accordingly.

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
willm30: @ArmandoGraterol Will this be addressed in a separate issue?

ArmandoGraterol:
@willm30 
> I'm seeing this "NOTOK" status when entering an address for the first time:

Fixed! Weird that I didn't see this when I was testing.

> Also, when the error "please enter a value" is showing beneath the ABI field, the parser seems to stop working. 

It's not that it stopped working, it's just that the validation error had more priority than the status message (The inputs can only show one or the other). I've just changed it so that the status of the input is always shown even when there's a validation error as I feel like it makes more sense that way.

> Also, just bringing to your attention that the following is no longer happening, i.e. when you delete the abi it doesn't come back anymore.

Yes, I changed this as per Juliana's feedback. Now the ABI textfield should only be automatically filled when a contract address is selected, or when you at first select a contract that is not in the safe's chain and then change the safe to one in a different chain where the contract is actually found, for example.

willm30 (approved):
@ArmandoGraterol Nice job. Looks good to me. I have opened [an issue](https://github.com/JoinColony/colonyDapp/issues/3827) to track the various other validation permutations.

Julianahlk:
> These points are more about how the `SingleUserPicker` core component works than this issue/PR/feature. You could create some issues in `Maintenance`, perhaps check Figma to see if Product has updated the component, or if not then ask in the Product channel what they think of updating the component, I'm sure they will find this feedback useful if they haven't addressed this yet :)

Thanks @ArmandoGraterol for explaining and addressing the issues I encountered before, I will follow up with product regarding the`SingleUserPicker`.

While I was  re-testing this PR, I noticed something different regarding the Selected Safe. Last week, I was testing contract addresses in the All Saints (Gnosis Safe) and today I was trying to do the same but it kept getting the message "Contract source code not verified" but when I switched to the Pipe's Dream (Etherium), the contract address was found and tested this with different contracts and kept getting the same result.

safes

Did something change? or am I missing something?

ArmandoGraterol:
@Julianahlk 

> Did something change? or am I missing something?

Have you checked that the contracts you were trying to get in Gnosis exist and are verified? In Ethereum they seem to exist and also are verified but that may not be the case for Gnosis.

This address here should exist and be verified in the Gnosis chain: `0xefb2a96b21e686164f347983be5664a65ee14042`. Otherwise, you can check that the address you are using exists and is verified through Blockscout

PR #3758, opened 2022-08-18
Add initial setup for safe core SDK transactions

### LINKED ISSUE #3754: Implement initial Safe Core SDK setup
Completing this issue is needed for the transactions implementations. 

1. In order to make transactions with safes, we need first to install the required libraries which are:
```
@gnosis.pm/safe-core-sdk-types
@gnosis.pm/safe-core-sdk
@gnosis.pm/safe-service-client
@gnosis.pm/safe-web3-lib
```
The documentation proposes using either `@gnosis.pm/safe-ethers-lib` or `@gnosis.pm/safe-web3-lib`, we will use the latter because `safe-ethers-lib` requires ethersv5, which we currently don't support in the dapp.

2. Then, after installing the libraries the idea that we have is that we could create a utility function that could be reused to get the safe SDK required to create transactions. This function would be used in future sagas of the transaction types.

The documentation available about initializing and getting a safe SDK is right [here](https://github.com/safe-global/safe-core-sdk/blob/main/packages/guides/integrating-the-safe-core-sdk.md).

ArmandoGraterol:
### Usage

`const safeSDK = await getSafeCoreSDK(walletAddress, safeAddress)`
 
And if we `console.log` that const then it should look like this:
Screenshot from 2022-08-18 20-02-43

I've been using this safe deployed in Rinkeby for testing: `0x05b990D01A2f86852feD1E7d609dd3E52aA9EA27` (https://gnosis-safe.io/app/rin:0x05b990D01A2f86852feD1E7d609dd3E52aA9EA27). If you want to become owners for future testing please let me know and I'll add you :)

Otherwise, you can use any Gnosis Safe for testing, just make sure that the RPC URL corresponds with the chain where the safe is deployed.

Rinkeby is being used as the default testing network because that's what Gnosis Safe uses for development and testing.

Resolves #3754

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,6 +16,10 @@
         "@formatjs/intl-pluralrules": "^1.5.7",
         "@formatjs/intl-relativetimeformat": "^4.5.14",
         "@formatjs/intl-unified-numberformat": "^3.3.5",
+        "@gnosis.pm/safe-core-sdk": "^2.3.2",
+        "@gnosis.pm/safe-core-sdk-types": "^1.4.0",
+        "@gnosis.pm/safe-service-client": "^1.2.0",
+        "@gnosis.pm/safe-web3-lib": "^1.4.0",
         "@popperjs/core": "^2.3.3",
         "@purser/core": "^3.0.0",
         "@purser/metamask": "^3.2.0",
@@ -79,6 +83,7 @@
         "subscriptions-transport-ws": "^0.9.18",
         "userflow.js": "^2.2.0",
         "utility-types": "^3.7.0",
+        "web3": "^1.7.5",
         "web3-utils": "^1.3.5",
         "yup": "^0.29.0"
       },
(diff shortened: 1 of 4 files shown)


arrenv (approved):
Looks to be working to me!

31609029-3e4c0e4df162a4f1172313f69dbd6f06

Julianahlk (approved):
Good work Armando! 👍
Screen Shot 2022-08-22 at 5 45 17 PM

I tested a few other safes but I kept getting this RPC error:
Screen Shot 2022-08-22 at 5 58 18 PM

I realized the RPCs we had in `constants.ts` were different for GOERLI_NETWORK and ETHEREUM_NETWORK from the ones posted in [this](https://rpc.info/) website you sent me. Can these be updated? 

GOERLI_NETWORK
rpcUrl: 'https://goerli.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161',

ETHEREUM_NETWORK
rpcUrl: 'https://mainnet.infura.io/v3/9aa3d95b3bc440fa88ea12eaa4456161',

PR #3772, opened 2022-08-25
Add NFT preview

### LINKED ISSUE #3729: Add "Transfer NFT" transaction type to preview dialog
### Story

As a user creating a "Transfer NFT" transaction from the Gnosis control dialog, I want to see a preview of the transaction before sending it.

### Description

When the user chooses the "Transfer NFT", and fills the form of the Gnosis control dialog, after confirming the form and before sending the transaction, they should see a preview of what transactions they are going to send, along with the preview of other transactions (in the case of the user doing a multi transaction).

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5418%3A22229)
FireShot Capture 687 - Safe Control – Figma - www figma com

- [ ] You click on the submit button of the previous Gnosis Safe Control dialog screen with a "Transfer NFT" transaction added and it takes you to the preview with a summary of that transaction.
- [ ] The summary includes the transfer request and approval of the request. So, a "Transfer NFT" is a multi transaction by default.

willm30:
## Description

This PR adds the NFT preview screen to Gnosis Control Safe form. 

[Design](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=115%3A7477)

To test:

* Go to Gnosis Safe Control
* Click Transfer NFT and create transaction
* Preview should look like the design. 
* Test with multiple transactions.

**Changes** 🏗

* Adding NFT Transfer preview

## Screenshots

gs-nft-prev

gs-nft-prev-multiple

Resolves #3729

## Note

I have a created a separate issue for the "Approve NFT" transaction type & preview.  https://github.com/JoinColony/colonyDapp/issues/3844

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/SafeTransactionPreview.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/SafeTransactionPreview.tsx
@@ -10,7 +10,7 @@ import Numeral from '~core/Numeral';
 import TokenIcon from '~dashboard/HookedTokenIcon';
 import Button from '~core/Button';
 import Icon from '~core/Icon';
-
+import Avatar from '~core/Avatar';
 import { Colony, AnyUser } from '~data/index';
 
 import AddressDetailsView from './TransactionPreview/AddressDetailsView';
@@ -23,6 +23,8 @@ import {
 import DetailsItem from './DetailsItem';
 
 import styles from './SafeTransactionPreview.css';
+import { testNFTData } from './ControlSafeForm';
+import { getFilteredNFTData } from './utils';
 
 const MSG = defineMessages({
   previewTitle: {
@@ -65,6 +67,22 @@ const MSG = defineMessages({
     id: 'dashboard.ControlSafeDialog.SafeTransactionPreview.nft',
     defaultMessage: 'NFT',
   },
+  transactionType: {
+    id: `dashboard.ControlSafeDialog.SafeTransactionPreview.transactionType`,
+    defaultMessage: 'Transaction type',
+  },
+  nftHeldByTheSafe: {
+    id: `dashboard.ControlSafeDialog.SafeTransactionPreview.nftHeldByTheSafe`,
+    defaultMessage: 'NFT held by the safe',
+  },
+  targetContract: {
+    id: `dashboard.ControlSafeDialog.SafeTransactionPreview.targetContract`,
+    defaultMessage: 'Target contract',
+  },
+  nftId: {
+    id: 'dashboard.ControlSafeDialog.SafeTransactionPreview.nftId',
+    defaultMessage: 'Id',
+  },
   data: {
     id: 'dashboard.ControlSafeDialog.SafeTransactionPreview.data',
     defaultMessage: 'Data',
@@ -102,11 +120,56 @@ const transactionTypeFieldsMap = {
     },
   ],
   [TransactionTypes.TRANSFER_NFT]: [
-    // To be finished once NFT part of the form is merged
     {
-      key: 'function',
-      label: MSG.function,
+      key: 'transactionType',
+      label: MSG.transactionType,
       value: () => <FormattedMessage {...ConstantsMSG.transferNft} />,
+    },
+    {
+      key: 'nft',
+      label: MSG.nftHeldByTheSafe,
+      value: (nft) => (
+        <div className={styles.nftContainer}>
+          <Avatar
+            avatarURL={undefined}
+            placeholderIcon="circle-close"
+            seed={nft.id.toLocaleLowerCase()}
+            title=""
+            size="xs"
+          />
+          <div>{nft.profile.displayName}</div>
+        </div>
+      ),
+    },
+    {
+      key: 'nft',
+      label: MSG.targetContract,
+      value: (nft) => (
+        <div className={styles.nftContainer}>
+          <Avatar
+            avatarURL={undefined}
+            placeholderIcon="circle-close"
+            seed={nft.id.toLocaleLowerCase()}
+            title=""
+            size="xs"
+          />
+          <div>{getFilteredNFTData(testNFTData, nft.id)?.name}</div>
+        </div>
+      ),
+    },
+    {
+      key: 'nft',
+      label: MSG.nftId,
+      value: (nft) => (
+        <div>{getFilteredNFTData(testNFTData, nft.id)?.tokenID}</div>
+      ),
+    },
+    {
+      key: 'recipient',
+      label: MSG.to,
+      value: (recipient) => (
+        <AddressDetailsView item={recipient} isSafeItem={false} />
+      ),
     },
   ],
   [TransactionTypes.CONTRACT_INTERACTION]: [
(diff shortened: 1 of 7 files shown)


willm30:
@ArmandoGraterol I'm not sure what to do about [this screen](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5418%3A22229). 

image

It differs from the [single transaction preview](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A22136) in a number of respects (the `1. Transfer NFT` section has different row headings and then there's the presence of the section `2. Approve NFT`), and I'm not sure how the user flow would be differ to get here compared with the other screen.

What do you recommend?

arrenv:
> Yeah I understand why it's confusing and I believe the Transfer NFT section in the single transaction preview specs is the correct one, and the approve NFT part might just be a contract interaction. We could confirm if this is the case with @arrenv when he comes back as I believe he's taking some time off and the NFT transactions are not as high-priority as the others.

The main purpose of this it to transfer ownership of the NFT from one person to another. So, however many transactions are required to do that is the goal.

If referring to the [EIP-721 NFT standard](https://eips.ethereum.org/EIPS/eip-721), there are two methods that are required `Transfer` and `Approval`. Which, is why both are included here.

willm30:
@arrenv I've been looking into this and this is my understanding so far:

 `Approve` [can only be called by the token owner](https://docs.openzeppelin.com/contracts/2.x/api/token/erc721#ERC721-approve-address-uint256-). Assuming the Safe owns the NFT, then it would not be possible for a Colony to call the `approve` method on an ERC721 contract (NFT). 

A user would first need to interact with the NFT from within the Safe UI & approve the Colony before we could call `safeTransferFrom` from the Colony UI. 

In which case, there'd be no need for the "approve" block in the screen above. 

Do I understand this correctly?

arrenv:
@willm30 The AMB would be able to trigger the Safe to do both of these transactions though, right?

willm30:
> @willm30 The AMB would be able to trigger the Safe to do both of these transactions though, right?

@arrenv Ah good point. I was assuming the Colony contract would be calling the function in the ERC721 contract, but of course it actually gets called more like this:

Colony --> AMB 1 --> Bridge Monitor --> AMB2 --> Zodiac module --> Safe --> ERC721. So the safe would still be calling the functions on the ERC721 contract. 

I will assume we can make the calls from the colony UI and test using our mock deployment.

src/modules/dashboard/components/Dialogs/ControlSafeDialog/SafeTransactionPreview.tsx
```diff
@@ -242,36 +310,39 @@ const SafeTransactionPreview = ({ colony, values }: Props) => {
...
+                      }
+                    />
+                  )}
                 {values.transactions[index].transactionType !== '' &&
                   transactionTypeFieldsMap[
                     values.transactions[index].transactionType
                   ].map(({ key, label, value }) => (
                     <DetailsItem
-                      key={key}
+                      key={`${key}${label.defaultMessage}`}
```
ArmandoGraterol, line 345: I believe we could remove the key prop from `transactionTypeFieldsMap` and make this key better if we use autogenerated ids. Check out how we `nanoid` to generate ids in cases like this where there's no clear way to get a unique id.

ArmandoGraterol (changes requested):
Looking great to me :+1: Just noticed some small things and I also have a question:

What's the reason for separating the "Approve" section from this issue? I don't have anything against implementing that in a separate issue, just wanna know.

willm30:
> What's the reason for separating the "Approve" section from this issue? I don't have anything against implementing that in a separate issue, just wanna know.

@ArmandoGraterol I'm assuming it'll be it's own transaction type, so thought to separate it from this one which is just the `transfer NFT` flow.

Julianahlk (commented):
As I was testing this I realized the NFT's images are not connected to the safes yet (which you are doing in your other PR) but also, whenever I change Safes, the SinglePickerUser for the NFTs continues giving me the same options no matter what safe I pick. 
NFTallthesame

But the preview is looking good:
Screen Shot 2022-09-14 at 8 55 17 PM

ArmandoGraterol:
> @ArmandoGraterol I'm assuming it'll be its own transaction type, so thought to separate it from this one which is just the `transfer NFT` flow.

As far as I know and according to the specs that we have, the `Approve` transaction is not its own type like the way the 4 transaction types are. 

This transaction only appears in the preview screen and its only appears following an NFT transfer, so when creating an NFT transfer transaction it's by default a multi-transaction.

ArmandoGraterol:
> ...whenever I change Safes, the SinglePickerUser for the NFTs continues giving me the same options no matter what safe I pick. 

That's expected, the PR you mentioned is about connecting the NFT picker: #3838

This issue/PR is only for adding the "Transfer NFT" transaction type to the Preview dialog, so only UI

willm30:
> As far as I know and according to the specs that we have, the `Approve` transaction is not its own type like the way the 4 transaction types are.
> 
> This transaction only appears in the preview screen and its only appears following an NFT transfer, so when creating an NFT transfer transaction it's by default a multi-transaction.

@arrenv Are you able to confirm whether `approve NFT` should be its own transaction type? If it is not, then that would imply that every `transfer NFT` comes with an `approve NFT` call, which would seem unnecessary if the Colony had already been approved. And I'm not sure why [this screen](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A22136) would exist if every `transfer NFT` call were also to include an `approve NFT` call. Can you clarify?

willm30:
> > ...whenever I change Safes, the SinglePickerUser for the NFTs continues giving me the same options no matter what safe I pick.
> 
> That's expected, the PR you mentioned is about connecting the NFT picker: #3838
> 
> This issue/PR is only for adding the "Transfer NFT" transaction type to the Preview dialog, so only UI

@Julianahlk

arrenv:
> @arrenv Are you able to confirm whether `approve NFT` should be its own transaction type? If it is not, then that would imply that every `transfer NFT` comes with an `approve NFT` call, which would seem unnecessary if the Colony had already been approved. And I'm not sure why [this screen](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A22136) would exist if every `transfer NFT` call were also to include an `approve NFT` call. Can you clarify?

From reading the documentation on Approve, it implies that you are giving permission to the Safe to transfer the NFT. This would only apply for a single NFT, so, it would make sense that every NFT Transfer includes both transactions in most cases.

If the exact same NFT was transferred back to the Safe and that was transferred out again, then I would assume that it would already have permission, so then it would not need the approve. In this case, I believe that the Transfer should be the only transaction.

If you discover something else though, please do update me.

Julianahlk (approved):
LGTM 👍

I found the warning below after creating only one transaction.  Steps to repro:

1. Create only one transaction
2. Create transaction
3. Go back to add a second one (it only happens the first time after refreshing the page)

Screen Shot 2022-09-15 at 4 24 02 PM

willm30:
> I found the warning below after creating only one transaction. Steps to repro:

@Julianahlk Thanks. Have created an issue for it as it's a bug on the main branch.

ArmandoGraterol (approved):
Nice work Will! This is good to go :ok_hand:

Issue #3687, opened 2022-07-27
Force action toggles should have a tooltip


arrenv:
## Steps to reproduce

1. Ensure you have the Governance extension installed
2. Go to any action dialog that has the "Force" action toggle.

## Expected behaviour

* The force toggles should have an info tooltip.

This is already built into the `Toggle` component, and will only require a prop. However, you may want to update the usage of the `QuestionMarkTooltip` component to the `IconTooltip` component.

You can use this copy for the tooltip: `Toggle "Force" to bypass governance and perform this action immediately. This is only possible with the right permissions.`

Once merged, please also create an issue in the `Expenditures` project to ensure that it is updated in that branch to match the DApp's implementation.

- [x] Added Tooltip to Force Toggle
- [x] Created issue in `Expenditures` project

Figma Link - https://www.figma.com/file/CCgGsBzhX3BUENS1IvgP7N/Motions-%26-Disputes?node-id=3511%3A0

![Screenshot 2022-07-27 113705](https://user-images.githubusercontent.com/33682027/181215410-77e2efcb-b8bf-4c07-ac2d-06cf05c41625.png)

## Actual behaviour

* The force toggles throughout the DApp don't have an info tooltip.

![Screenshot 2022-07-27 114318](https://user-images.githubusercontent.com/33682027/181216703-7ab56559-7d4b-4a74-afc0-07138d09cd36.png)

PR #3775, opened 2022-08-25
Add tooltip to force action toggles

Julianahlk:
## Description

This PR adds a tooltip to the force action toggles. To test this make sure you have the Governance extension installed and then go to any action dialog that has the "Force" action toggle.

**The text for the tooltip should say the following:**
- "Toggle "Force" to bypass governance and perform this action immediately. This is only possible with the right permissions."

**New stuff** ✨

Screen Shot 2022-08-26 at 12 27 29 AM

(Resolves | Contributes to) #3687

### DIFF
--- a/src/modules/core/components/Fields/ForceToggle/ForceToggle.tsx
+++ b/src/modules/core/components/Fields/ForceToggle/ForceToggle.tsx
@@ -1,0 +1,73 @@
+import React from 'react';
+import { PopperOptions } from 'react-popper-tooltip';
+import { MessageDescriptor } from 'react-intl';
+
+import { Toggle } from '~core/Fields';
+
+import { SimpleMessageValues } from '~types/index';
+
+import styles from './ForceToggle.css';
+
+const displayName = 'ForceToggle';
+
+interface Appearance {
+  theme?: 'primary' | 'danger';
+}
+
+interface Props {
+  appearance?: Appearance;
+  name?: string;
+  label?: string | MessageDescriptor;
+  labelValues?: SimpleMessageValues;
+  disabled?: boolean;
+  elementOnly?: boolean;
+  tooltipTextValues?: SimpleMessageValues;
+  tooltipText?: string | MessageDescriptor;
+  tooltipClassName?: string;
+  tooltipPopperOptions?: PopperOptions;
+  onChange?: (value: boolean) => any;
+}
+
+const ForceToggle = ({
+  appearance,
+  name = 'forceAction',
+  label = { id: 'label.force' },
+  labelValues,
+  disabled = false,
+  elementOnly,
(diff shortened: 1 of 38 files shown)


chinins (commented):
Many of the dialogs don't have the icon with the tooltip. One example of having it (Mint tokens), all other print screens are without:
Screenshot 2022-08-26 at 15 34 07
Screenshot 2022-08-26 at 15 34 15
Screenshot 2022-08-26 at 15 34 22
Screenshot 2022-08-26 at 15 34 35
Screenshot 2022-08-26 at 15 34 40
Screenshot 2022-08-26 at 15 34 48
Screenshot 2022-08-26 at 15 35 05
Screenshot 2022-08-26 at 15 33 27
Screenshot 2022-08-26 at 15 33 36
Screenshot 2022-08-26 at 15 33 56

Julianahlk:
> Many of the dialogs don't have the icon with the tooltip. One example of having it (Transfer funds), all other print screens are without

Hey, that's so strange. I added the tooltip to all the dialogs see my last commit, it has all the changes. I will force push my changes again. See my last commit, it contains 34 file changes.

chinins (commented):
It's all good now. Probably your last commit was not reflected. 

It's working well. I do see a lot of repetitive code both in .tsx & .css and I would suggest to extract it to avoid this manual repetition. What do you think?

ArmandoGraterol:
> It's all good now. Probably your last commit was not reflected.
> 
> It's working well. I do see a lot of repetitive code both in .tsx & .css and I would suggest to extract it to avoid this manual repetition. What do you think?

I agree, perhaps it's time for a `ForceToggle` component

Julianahlk:
I agree with both of you @chinins and @ArmandoGraterol as well, so I created the new `ForceToggle` component. Let me know what you think.

src/modules/core/components/Fields/ForceToggle/ForceToggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+import styles from './ForceToggle.css';
+
+const displayName = 'ForceToggle';
+
+interface Appearance {
+  theme?: 'primary' | 'danger';
+}
+
+interface Props {
+  appearance?: Appearance;
```
chinins, line 18: You are not passing appearance from anywhere, so I would just remove it. I'm on the side of making it minimal and then adding stuff as needed. Same with all other props really except tooltipPopperOptions and obviously disabled....

But your

chinins (approved):
All good. Much cleaner and better now. As I mentioned directly in my comment, I wouldn't add additional props as they are not used right now, but it's up to you.

ArmandoGraterol (approved):
Looks good to me :ok_hand:

PR #3776, opened 2022-08-25
Add activate button when part of tokens is activated

### LINKED ISSUE #3771: Not able to Stake full amount of available funds
## Steps to reproduce

1. Ensure you have only enough activated funds to stake part of a Motion.
2. Try to Stake your full amount of available funds using the slider.
3. You should hit the max limit, which is the same as your balance but you are not able to stake it.

This was originally found on production, however, I was able to replicate it locally with the following:
 - 2 members in a Colony
 - 1,000,000 minted
 - 457 paid to first member
 - 20 paid to second member
 - Try to create a Motion and stake with the second member

## Expected behaviour

* You should still be able to to stake if you have maxed out the slider.
* The error message should still appear, as long as the button to activate more.

## Actual behaviour

* You are not able to stake the full amount of available funds.

staking-limit

chinins:
## Description

This PR:

- adds another condition for activate button
- fixes the rounding error for the stake button
- moves back button to the left (UI is confirmed with Arren)

To replicate the bug, please, follow Arren's instructions from the issue.

resolves #3771

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.css
+++ b/src/modules/dashboard/components/ActionsPage/StakingWidget/StakingWidget.css
@@ -84,8 +84,3 @@
 .loading {
   padding-top: 35px;
 }
-
-.backButtonWrapper button {
-  margin-left: 14px;
-  width: fit-content;
-}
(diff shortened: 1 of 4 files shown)


danbr (commented):
@chinins I'm testing (with Arren's method) and wanna double check the behaviour with you, when you're free.

ArmandoGraterol (approved):
Looks good to me :+1: I wasn't able to replicate the rounding error on master but other than that the slider is working as expected

PR #3784, opened 2022-08-30
Add Raw Transaction validation

### LINKED ISSUE #3777: Improve Gnosis Control Form validation
Currenly, it's possible to submit the form when the following are true:

1. Recipient is not a valid address
1. The value for wei is a decimal
1. Byte data is invalid
1. The signer is not an owner of the chosen safe

These should all be addressed such that the form will not submit if any one of them is true. 

With respect to checking if the signer owns the safe, I have already written the relevent function (`isSafeOwnedByUser`) in `safeHelpers.ts` in [`feature/3721-connect-raw-transaction`](https://github.com/JoinColony/colonyDapp/pull/3765). It just needs to be added to the form submission.

willm30:
## Description

When connecting up the raw transaction form to the saga, I noticed a number of problems with validation that meant that the interaction with the Gnosis Safe SDK would fail. 

These were:

-- That you could enter an invalid address for the recipient
-- That you could enter a decimal for the value
-- That you could enter a non-hex string for the byte data

This PR addresses these. 

**New stuff** ✨

* Add form validation as detailed above

Resolves #3777

### DIFF
--- a/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
@@ -3,6 +3,7 @@ import { FormikProps } from 'formik';
 import * as yup from 'yup';
 import toFinite from 'lodash/toFinite';
 import { defineMessages } from 'react-intl';
+import { ethers } from 'ethers';
 
 import { AnyUser } from '~data/index';
 import Dialog, { DialogProps, ActionDialogProps } from '~core/Dialog';
@@ -20,9 +21,17 @@ const MSG = defineMessages({
     id: 'dashboard.GnosisControlSafeDialog.requiredFieldError',
     defaultMessage: 'Please enter a value',
   },
-  amountZero: {
+  gtZeroError: {
     id: 'dashboard.GnosisControlSafeDialog.amountZero',
     defaultMessage: 'Amount must be greater than zero',
+  },
+  notIntegerError: {
+    id: 'dashboard.GnosisControlSafeDialog.integer',
+    defaultMessage: 'Amount must be an integer',
+  },
+  notHexError: {
+    id: 'dashboard.GnosisControlSafeDialog.notHexError',
+    defaultMessage: 'Value must be a valid hex string',
   },
 });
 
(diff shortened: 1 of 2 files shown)


Julianahlk (approved):
Makes sense! LGTM 👍

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
```diff
@@ -114,7 +130,15 @@ const validationSchema = (isPreview) =>
         data: yup.string().when('transactionType', {
           is: (transactionType) =>
             transactionType === TransactionTypes.RAW_TRANSACTION,
-          then: yup.string().required(() => MSG.requiredFieldError),
+          then: yup
+            .string()
+            .required(() => MSG.requiredFieldError)
+            .matches(/^0x/, () => MSG.startWith0x)
```
ArmandoGraterol, line 136: How come you need to specifically check that the string starts with 0x? Isn't `.isHexString` enough?

ArmandoGraterol (commented):
Looking good! I just have a question regarding data's validation

src/modules/dashboard/components/Dialogs/GnosisControlSafeDialog/GnosisControlSafeDialog.tsx
willm30: @ArmandoGraterol Yes, can do. Was trying to be as explicit as possible but I'll follow your lead in the abi parser branch.

ArmandoGraterol (approved):
Looking good to me! Just wanted to inform you that for the validation this is a valid integer:
FireShot Capture 711 - Actions - Colony - wonker - localhost

It's quite the edge case. If you want to address this, in this PR or in another one, I would recommend to limit the decimals to 0 as I did in the function parser PR, it's a quick and easy solution and we don't want decimals anyways

chinins (approved):
Other than what Armando mentioned all good - nice work!

willm30:
> Looking good to me! Just wanted to inform you that for the validation this is a valid integer: FireShot Capture 711 - Actions - Colony - wonker - localhost
> 
> It's quite the edge case. If you want to address this, in this PR or in another one, I would recommend to limit the decimals to 0 as I did in the function parser PR, it's a quick and easy solution and we don't want decimals anyways

@ArmandoGraterol Javascript 😂 I have removed ability to enter decimals.

PR #3789, opened 2022-08-31
Validate selected safe 

### LINKED ISSUE #3788: Add chosen safe validation in gnosis safe control
When using the Gnosis Safe Control Form, we should validate that the signer is the owner of the chosen safe and that the safe is on the correct network.

Currently, if either of these isn't true, the interaction with the SDK will fail and there will be no indication in the UI as to why.

willm30:
# This PR is back in draft since we're not using the SDK. It may become a wontfix, tbc...

## Description

When connecting up the raw transaction form to the saga, I noticed that you could select a safe on the wrong network / that wasn't owned by the logged-in user. This would cause the interaction with the Gnosis Safe SDK to fail. 

This PR implements validation that checks whether the safe selected is on the correct network, and whether it is owned by the signer of the transaction (i.e. the logged in user in production, and the owner of the provided private key in dev). Make sure you have set your SIGNER_PRIVATE_KEY `.env` variable with an account that owns the safe you'll be testing.

It also corrects a small error in the `SingleUserPicker` that was causing the invalid state to display incorrectly. 
And, it removes the references to "testnet" for Rinkeby and Goerli in `constants.ts`. I added these in an earlier pr thinking it'd be useful to distinguish between them, but it needlessly lengthens the safe display name in the `SingleSafePicker` so have removed. 

To Test: 

* Ensure you have a safe that you own in the `safes` array in `GnosisControlSafeDialog` and that your SIGNER_PRIVATE_KEY .env variable is set
* Go to gnosis control form, and select a safe. 
* Selecting All Saints does not show an error. This is because the address is invalid and we already verify for this when adding a safe. 
* Selecting the safe on Mainnet will show an error because you don't own it and it's not on Rinkeby (assuming you leave rinkeby as the network you're testing on)
* Selecting the safe you have added should not show an error (provided it's on rinkeby and the key you provided is for an account that owns it) and you should be able to submit the form. 

**New stuff** ✨

* Display error message in `SingleSafePicker` if safe is not owned by logged in user or is on wrong network
* Add test to schema validation 

**Changes** 🏗

* Fix invalid state in `SingleUserPicker` not composing from base input
* Remove references to "testnet" in constants. I did this thinking it would be useful to distinguish between a testnet and a non-testnet but I no longer think so. 

Resolves #3788


willm30:
No longer relevant due to the use of zodiac bridge module.

PR #3794, opened 2022-09-01
Add empty string as ipfs hash fallback where missing

### LINKED ISSUE #3787: Add empty string as a fallback for missing IPFS hash everywhere in the dapp
- all annotations
- all edit colony actions
- anywhere else if found

chinins:
## Description

This PR adds string as ipfs upload fallback to avoid app crashing (like we had in production recently).

I've renamed the util function to show that it's not only annotation that we will upload. I've also added try-catch block and logs there.

resolves #3787

### DIFF
--- a/src/modules/dashboard/sagas/actions/createDomain.ts
+++ b/src/modules/dashboard/sagas/actions/createDomain.ts
@@ -16,7 +16,6 @@ import {
   createTransactionChannels,
   getTxChannel,
 } from '../../../core/sagas';
-import { ipfsUpload } from '../../../core/sagas/ipfs';
 import {
   transactionReady,
   transactionPending,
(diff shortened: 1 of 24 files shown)


willm30 (commented):
@chinins On this branch I tried to edit the Colony display name via the actions menu (governance extn uninstalled, so normal action), but the transaction doesn't load on the ActionsPage, I only see this: 

naked-rat

Is that happening for you also? I checked my env was OK by doing the same on `master` and it went through as normal.

chinins:
> @chinins On this branch I tried to edit the Colony display name via the actions menu (governance extn uninstalled, so normal action), but the transaction doesn't load on the ActionsPage, I only see this:

> Is that happening for you also? I checked my env was OK by doing the same on `master` and it went through as normal.

oh, good catch! It was missing my last push. It's all pushed now.

Julianahlk (commented):
Hey, I found a few things while testing your PR:

1. After creating a motion for a Payment action, I created an Objection but as soon as I refreshed the page I received the following errors in the console that seem to be related to IPFS:
Screen Shot 2022-09-01 at 10 40 55 PM
Screen Shot 2022-09-01 at 10 49 20 PM
objectiontest

Creating a **Forced Payment** seems to be working fine:
FORCED ACTION

2. When I try to Edit the Colony Details the app just stays in the "loading transaction" state and then I get no response:

**Force Action** 
COLONYDETAILS

**Motion**
TEST22

chinins:
@Julianahlk, can you please, try again. I keep missing the push and not catching it 🙈 Remember to pull again.

Screenshot 2022-09-02 at 15 57 49

ArmandoGraterol (approved):
Working great for me :+1: Tested several actions and motions and I didn't see anything weird

willm30 (approved):
Working for me now.

PR #3797, opened 2022-09-01
Fix SingleSafePicker width

### LINKED ISSUE #3796: Select safe dropdown breaking layout
## Steps to reproduce

Go to select a safe when a transaction type has been selected and it breaks the layout. (dropdown should be on same line)

layout-br

willm30:
## Description

This PR implements a small fix to the `SingleSafePicker` that stops it from breaking onto a new line when the scroll bar is present. See issue for screenshot of problem.

**Changes** 🏗

* Changing input and button width / max-width to the same as in the `TransactionTypesSection.css`

## Screenshot

**Doesn't break onto a new line**
safepicker

Resolves #3796


ArmandoGraterol (approved):
Looks good to me :+1: Nice catch
FireShot Capture 712 - Actions - Colony - wonker - localhost

chinins (approved):
Good catch!

PR #3801, opened 2022-09-01
Set AddItemButton disabled cursor

### LINKED ISSUE #3800: Change cursor on `Add Another Transaction`
## Steps to reproduce

When "Add Another Transaction" button is disabled, cursor is a pointer whereas it should be the default like the "Create transaction" button.

## Expected behaviour

Cursors should be the same.

## Actual behaviour

safe-control-cursor

willm30:
## Description

Small tweak to the cursor when the `AddItemButton` is disabled.

**Changes** 🏗

* Setting the cursor on `AddItemButton` when disabled to default

## Before 

safe-control-cursor

## After

aatcursor

Resolves #3800

### DIFF
--- a/src/modules/core/components/Button/AddItemButton.css
+++ b/src/modules/core/components/Button/AddItemButton.css
@@ -17,6 +17,7 @@
 .main[disabled] {
   background-color: color-mod(var(--text-disabled) alpha(12%));
   color: var(--text-disabled);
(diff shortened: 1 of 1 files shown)


chinins (approved):
Short and sweet

Julianahlk (approved):
👍

PR #3806, opened 2022-09-03
Fix history in edit decision (adding missing commits from older branch)

chinins:
## Description

I guess my connection broke and last 2 commits were not pushed and I didn't notice. This provides the fix for the routing history.


danbr (approved):
Looks familiar.... Happy to welcome the updates!

PR #3822, opened 2022-09-06
Add clickable heading styling in right aside

### LINKED ISSUE #3814: Add clickable heading styling in right aside
### Story

When users are using Colony most are unaware that the headings in the right aside are clickable.

### Description

This requires adding improved styling and visual indication to the headings that are clickable in the right aside. This should help indicate to users that these headings are clickable and help them explore more features of Colony.

### Design

[Figma link](https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=7653%3A90396)

Screenshot 2022-09-05 144912

- [ ] Add the default blue colour to headings that are clickable
- [ ] Add the right chevron to the clickable headings
- [ ] Add the hover state

jakubcolony:
## Description

Implementing https://www.figma.com/file/qQoBbyyDXkuHwoqHIOGjEu/Colony-Simplified?node-id=7599%3A89178

**New stuff** ✨

* `ClickableHeading` component to put the styling for both funding and members headings in one place

**Changes** 🏗

* Added `smallish` size to `Heading` to avoid overriding the size in CSS files

Resolves #3814


arrenv (commented):
I'm assuming the introduction of the new font size for the heading is due to matching the font size in the Figma. There have been rendering differences with the old font currently used in Figma, so it should be okay to leave the font size as it was to be more consistent with the rest of the DApp.

Note: I could not test visually, as I am stuck in a nasty rebase. So, feel free to correct me if visually it is correct and consistent.

Also, congrats on your first PR :)

src/modules/core/components/Heading/Heading.css
```diff
@@ -78,6 +78,10 @@
   font-size: var(--size-small);
 }
 
+.sizeSmallish {
```
chinins, line 81: I agree with Arren, I would not add another size to the heading to be consistent with the rest of the dapp.

src/modules/dashboard/components/ColonyHome/SidebarClickableHeading/SidebarClickableHeading.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { ReactChild } from 'react';
```
chinins, line 1: I am not totally sure but I think that it would be better to move it to the core components...

chinins (commented):
Very nice work! I agree with Arren about not adding extra size to the Heading, we already have enough so probably would be better to use the existing one.

 I'm a bit surprised that we are removing invite link - @arrenv, could you please confirm?

arrenv:
> Very nice work! I agree with Arren about not adding extra size to the Heading, we already have enough so probably would be better to use the existing one.
> 
> I'm a bit surprised that we are removing invite link - @arrenv, could you please confirm?

Good pickup @chinins, no, we should not be removing the Invite link, that is an oversight on the product side. Please leave the invite link in there.

chinins:
As talked on discord, it's ok to leave the size smallish - better option than doing manually.

src/modules/dashboard/components/ColonyHome/ColonyMembers/MembersSubsection.tsx
```diff
@@ -120,27 +120,28 @@ const MembersSubsection = ({
...
+            <InviteLinkButton
+              colonyName={colonyName}
+              buttonAppearance={{ theme: 'blueWithBackground' }}
+            />
+          </div>
         )}
       </div>
     ),
-    [members, membersPageRoute, isContributorsSubsection, colonyName],
+    [isContributorsSubsection, membersPageRoute, members],
```
chinins, line 144: github shows me that you have a missing dependency `colonyName`

chinins (approved):
Looks good - nice work! Just please check about the missing dependency warning.

src/modules/dashboard/components/ColonyHome/ColonyMembers/MembersSubsection.tsx
jakubcolony: Good shout, thanks!

ArmandoGraterol (commented):
Looking good! I've noticed that the icon in the second link is not aligned for me:
FireShot Capture 715 - Actions - Colony - wonker - localhost

How does it look in your machine?

src/modules/core/components/ClickableHeading/ClickableHeading.tsx
```diff
@@ -0,0 +1,30 @@
...
+}
+
+const ClickableHeading = ({ children, linkTo }: Props) => {
+  return (
+    <div className={styles.heading}>
+      <Heading appearance={{ size: 'smallish', weight: 'bold' }}>
+        <NavLink to={linkTo}>
+          <span className={styles.contents}>
+            {children}
+
```
Julianahlk, line 21: I think you have an extra space here 😉

jakubcolony: Hmm, it's aligned for me. I wouldn't think a whitespace can make a difference if the parent is using flex. 
image
Which browser are you using @ArmandoGraterol?

Julianahlk (approved):
Good work! It looks good for me!
Screen Shot 2022-09-13 at 12 05 12 PM
 
Before you merge make sure you follow up with Armando's misalignment issue. 👍

src/modules/core/components/ClickableHeading/ClickableHeading.tsx
```diff
@@ -0,0 +1,32 @@
+import React, { ReactChild } from 'react';
+
+import Heading from '~core/Heading';
+import NavLink from '~core/NavLink';
+import Icon from '~core/Icon';
+
+import styles from './ClickableHeading.css';
+
+interface Props {
+  children?: ReactChild;
```
ArmandoGraterol, line 10: Optional props should go last

ArmandoGraterol (approved):
Looking good! Nice job with the alignment, they seem to be aligned properly in my machine now

PR #3823, opened 2022-09-06
Decisions: Connect Decisions list to query

### LINKED ISSUE #3790: Decisions: Wire-up Decisions List on Decisions page
### Story

As a user I want to see the list of Decisions, on the Decision page.

### Description 

Using #3717 connect UI and retrieve any existing Decisions, and display.
Remove dummy data.

Figma:  https://www.figma.com/file/oRaTwsD5BUH3FJNzoZdSCv/Discussions?node-id=1018%3A56590

image

danbr:
## Description

This PR:

- Adds extra prop to Subscription SubgraphMotion to exclude Decision motions.
- Creates Subscription SubgraphDecision that performs a query to get the annotation hash which is storing the Decision details.
- Removes Decisions from ColonyActions listItems
- Display spinner as Decisions are being loaded
- Clicking on Decision listItem navigates to transaction (Decision)

Screenshot 2022-09-09 at 19 17 59 (2)

Screenshot 2022-09-09 at 18 50 47

*I noticed a small bug where, if you create a Decision and then nav back to Decisions homepage the decisions list is not refreshed.  The subscription seems to be called but returns zero results. I'm not sure why, but i will raise an issue.

Resolves #3790

### DIFF
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -2650,7 +2650,6 @@ export type SubgraphMotionsSubscriptionVariables = Exact<{
   first?: Maybe<Scalars['Int']>;
   colonyAddress: Scalars['String'];
   extensionAddress: Scalars['String'];
-  motionAction?: Maybe<Scalars['String']>;
   motionActionNot?: Maybe<Scalars['String']>;
 }>;
 
@@ -7271,8 +7270,8 @@ export function useSubgraphEventsThatAreActionsSubscription(baseOptions?: Apollo
 export type SubgraphEventsThatAreActionsSubscriptionHookResult = ReturnType<typeof useSubgraphEventsThatAreActionsSubscription>;
 export type SubgraphEventsThatAreActionsSubscriptionResult = Apollo.SubscriptionResult<SubgraphEventsThatAreActionsSubscription>;
 export const SubgraphMotionsDocument = gql`
-    subscription SubgraphMotions($skip: Int = 0, $first: Int = 1000, $colonyAddress: String!, $extensionAddress: String!, $motionAction: String, $motionActionNot: String) {
-  motions(skip: $skip, first: $first, where: {associatedColony: $colonyAddress, extensionAddress: $extensionAddress, action: $motionAction, action_not: $motionActionNot}) {
+    subscription SubgraphMotions($skip: Int = 0, $first: Int = 1000, $colonyAddress: String!, $extensionAddress: String!, $motionActionNot: String) {
+  motions(skip: $skip, first: $first, where: {associatedColony: $colonyAddress, extensionAddress: $extensionAddress, action_not: $motionActionNot}) {
     id
     fundamentalChainId
     associatedColony {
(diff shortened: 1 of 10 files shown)


ArmandoGraterol (commented):
Looking good to me :ok_hand: 

A small suggestion, instead of going back to the action list after exiting a decision action page you could go back to the decision list. I feel like this could improve UX, what do you think?

ArmandoGraterol (approved):

willm30 (changes requested):
1. When I successfully stake a motion, clicking claim is producing the following error: 

claimerr

This doesn't happen after successfully objecting to a motion.
2. Agree that the "back" button should go to `/decisions`
3. If, after publishing a decision, you refresh the browser without staking or objecting, the motion fails. However, it doesn't show up as a failed decision in the decisions list. Should it?
4. If, after publishing a decision, I click the back button and navigate to `/decisions`, I only see "No decisions exist" (even if they do). I need to manually refresh to see my decisions list. 


And relatedly, if not directly relevant to this issue, I would think the decisions list default sort should be most recent first (reverse chronological), so the decision I just made is at the top. Also the label "Ending soonest/latest" is a bit confusing for decisions that have already ended (i.e. it's not obvious which sorting direction it refers to). This could be a separate issue if you agree.

Julianahlk (commented):
Perhaps not part of this issue but thought I should mention my findings below: 

* If I refresh the Preview page before hitting the publish button, for a few secs you can see the Connect Wallet to Log in popup: 
decisionrefresh

* In the **Should "Discussion" be approved?** if you vote **no** I am getting the following message:
Screen Shot 2022-09-13 at 12 37 20 PM

* RichTextEditor messages are not being posted to the ActionPage feed.

chinins (commented):
It's working properly except for the decision is not showing up in the list before it's staked 10%. I thought that we don't have the 10% threshold fr decisions. Could you, please, confirm?

Also, the loading should show for the whole time until the subscription is loaded so that we don't have a message about installing voting extension. I assume, you need to move around the loading states.

chinins:
@willm30, the back button is not related to this issue and is being fixed some place else.

@Julianahlk, the connect wallet is there because the app needs some time to reconnect the user and it's a wallet protected route. The warning that you have in the console is the same as in the motions.

Julianahlk:
> @willm30, the back button is not related to this issue and is being fixed some place else.
> 
> @Julianahlk, the connect wallet is there because the app needs some time to reconnect the user and it's a wallet protected route. The warning that you have in the console is the same as in the motions.

Nice! Good to know 👍 Thanks.

danbr:
> Looking good to me 👌
> 
> A small suggestion, instead of going back to the action list after exiting a decision action page you could go back to the decision list. I feel like this could improve UX, what do you think?

@ArmandoGraterol Thank you. Yes thank you. That is the plan, its been implemented in another PR.

chinins (approved):
Other than the bug you mention and that is treated separately, everything is good - well done!

PR #3833, opened 2022-09-09
Feature/3809-streaming-empty-form

joannasia:
## Description

- User has to be able to see an empty form for streaming payment. Starts date should be pre-filled with current date.
- User has to be able to add recipient for streaming payment.
- User has to be able to choose how streaming payment should end. User has 3 options: "fixed time", "when canceled", "limit is reached".

**New stuff** ✨

* Streaming form with different fields.

**Changes** 🏗

* Changed ExpenditureSettings.tsx, ExpenditurePage/types.ts files

**TODO**

* Add datepicker.

Resolves #3809 #3812

PR #3851, opened 2022-09-13
Date Picker Core Component

### LINKED ISSUE #3834: Date Picker Core Component
### Story

As a User creating a streaming payment, I would like to be able to set a start date and time in the future as well as potentially set an end date and time for a streaming payment.

### Description

We are adding in a date picker component in the Dapp that first will be used within the Expenditures project. We will need to support both date and time entry.

We are proposing to use this library - https://reactdatepicker.com/ - as it has also recently been added to Coin Machine here - https://github.com/JoinColony/coinMachine/pull/56 and additional update here - https://github.com/JoinColony/coinMachine/pull/196

Rough proof of concept for Datepicker with time as a child - https://codesandbox.io/s/react-datepicker-playground-forked-x94yc0

### Design

[Fimga Date Picker Component](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=6364%3A334590)

Screenshot 2022-09-09 115502

Screenshot 2022-09-09 115714

Screenshot 2022-09-09 115720

jakubcolony:
## Description

This PR adds a new component, `DatePicker` that will allow users to pick date and, optionally, time. It's based on [React Datepicker](https://reactdatepicker.com/) library.

"Plain" DatePicker:
image

DatePicker with time select:
image

DatePicker with options:
image

You will notice some odd-looking date manipulation. `date-fns`, the library used for date formatting doesn't allow you to switch from local timezone to UTC so everytime I wanted to format the date in UTC, I had to create a "fake" Date object that represents the UTC date as a local one. Simplified example:
10:00 AM BST is 9:00 AM UTC
I would create another Date object that is set to 9:00 AM BST so that it imitates the UTC time for formatting purpose. Hope this is clear, happy to answer any questions! 

**New stuff** ✨

 * `DatePicker` component
 * `date-fns` package for formatting Date objects
 * `@floating-ui` package - the designs required a time dropdown to be nested inside the date dropdown and the Popper library used by both `react-datepicker` and our internal components (`Popover`) cannot handle these very well
 * `patch-package` - the library above uses TypeScript 4+ features while the dapp runs on Typescript 3.x. This package allowed me to apply a custom "patch" to make it compatible with dapp's TS version. I spoke to @chinins about this and she confirmed the cdapp will use a newer version of TS so this is only a temporary solution.

**Changes** 🏗

**Deletions** ⚰️

## TODO

- [x] Basic DatePicker component
- [x] DatePicker with time selection
- [x] DatePicker within select (https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=1195%3A241884)

Resolves #3834


willm30 (commented):
@jakubcolony Good work Jakub! I will take a closer look at the code, but for now my first thought is that this should be based off https://github.com/JoinColony/colonyDapp/pull/3833 (streaming payment), as it does not belong in the "advanced" expenditure type.  Then you can connect it to "Starts" (and "Ends" too or leave that for a follow up pr if it adds too much additional size to this one).  

Also a few notes on the styles:
1. in [the design](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=5755%3A331599) the border is not rounded and the border color is   `var(--grey-blue-2)`;
1. the scrollbar on time picker drodown should be rounded and styled per design
1. time dropdown border is var(--temp-grey-blue-8) [in figma](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=6364%3A334590)

But to reiterate, looks like really solid work. Good job!

PR #3858, opened 2022-09-15
Refactor Add Existing Safe flow

### LINKED ISSUE #3849: Refactor AddExistingSafe dialog
### Story

As a user, when adding a safe to the colony, I need to go through a multistep process inside the dialog.

### Description

Due to the requirements of creating a cross-chain transaction, the `AddExistingSafeDialog` should now include 3 steps to link a safe to the colony. 

- In the 1st step, the user should input the chain and address of the safe, and this stage should check that the safe exists in the chain, as we are currently doing. 

- In the 2nd step, the user should see a warning, a description explaining this step, and 3 **read-only** inputs with data that we are going to provide:
  1. The AMB field should display a corresponding AMB address for the chain of the safe. _(It is my understanding that this should be the HOME bridge address)_
  2. The controller address field is just the colony's address.
  3. The chain id field should be `100` as we are deployed on the Gnosis chain after all.

- The final step should include a summary of the details in the first step and 3 more inputs: 
  1. The Zodiac module address input, which should also check that the module is added to the safe. 
  2. The safe's name input.
  3. The annotation text field.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=7108%3A22501)

FireShot Capture 716 - Safe Control – Figma - www figma com

FireShot Capture 717 - Safe Control – Figma - www figma com

FireShot Capture 718 - Safe Control – Figma - www figma com

- [ ] The steps are implemented as per the specs and have all the required inputs with corresponding validation.
- [ ] Submitting the form adds the safe as we did before, but now includes the Zodiac Module Address (The rest of the new fields like AMB address, controller address, and chain id are things that we can deduce and should know already, and are only there for the user to input when installing the zodiac module).
- [ ] The action page shows in the details widget the zodiac bridge address along with the other details that we have.

Link for the 2nd step's description: https://colony.gitbook.io/colony/advanced-features/safe-control-gnosis-safe
Link for the 3rd step's "Where to find this": https://colony.gitbook.io/colony/advanced-features/safe-control-gnosis-safe#finding-module-contract-address

willm30:
## Description

This PR refactors the add existing safe form. 

There's one outstanding todo which is what to do with chains that don't have bridges. I will add the change to this PR or to a follow-up PR when I have more info but don't envision it being a significant change so want to get the review process rolling in the meantime. 

## To test
1. You need a safe deployed.
2. You need the zodiac bridge module connected to that safe. ([Follow this tutorial](https://gnosis.github.io/zodiac/docs/tutorial-module-bridge/get-started))

The following info can be used if you don't have your own (Rinkeby): 
Safe 1: `0x93ebb9b7b0a3d42d3AF76930EF569F196e3821C1`, Module 1: `0x25603925a75D39AFFb22e810aC4a2A92BCfEe6B4`
Safe 2: `0xF8F46F65b7365b97691C6794d0ff235B05BFAf78`, Module 2: `0xcae11900da49EB16a14D20f31237E04542F8Da35`

You should be able to change the inputs and it should always display the correct status / error, and you should be able to move forwards and backwards through the form without state changing or any other funny business. Submitting the form should take you to the actions page, where you should see all the info including the `moduleAddress`.

**New stuff** ✨

* `CheckSafe, ConnectSafe, ConfirmSafe`
* `DetailsWidgetAddSafe` (extracting logic to own component)
* `useClipboardCopy` (extracting logic to own hook)

**Changes** 🏗

* ActionsPage now includes `moduleAddress` in Details Widget
* Add onblur to core `Input` component (did not work previously. Used to set loading state in `CheckSafe` when field is not touched (i.e. first time user enters an address))
* Moved address validations into the yup schema in `AddExistingSafeDialog` (where they belong) instead of having them in effect hooks in the form. 

## Screenshots

step3
deets

Resolves #3849


src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialog.tsx
```diff
@@ -1,45 +1,268 @@
...
   close,
 }: Props) => {
+  const { formatMessage } = useIntl();
   const history = useHistory();
+  const loadingSafeState = useState<boolean>(false);
+  const loadingModuleState = useState<boolean>(false);
+  const [abortController, setAbortController] = useState<
+    AbortController | undefined
+  >(undefined);
+  const stepState = useState<number>(1);
```
jakubcolony, line 87: Not saying this is wrong or anything but why do you not destructure the array returned by `useState`? You refer to the state value later by using `stepState[0]` which could be less readable than something like `stepIndex`. And you end up destructuring this array in the `AddExistingSafeDialogForm` component anyway

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/CheckSafe.tsx
```diff
@@ -0,0 +1,235 @@
...
+  contractAbi: any;
+  trustedForDelegateCall: boolean;
+}
+
+export interface StatusText {
+  status: MessageDescriptor;
+  statusValues?: SimpleMessageValues;
+}
+
+type SafeData = SafeContract | undefined | null | {} | 'alreadyExists';
```
jakubcolony, line 74: You have the same type defined here: https://github.com/JoinColony/colonyDapp/pull/3858/files#diff-17516f245a322de3ed786172e8d5da66e2ee65d4a4b2d163242ffea703a56d10R22. Maybe you could share it between these two files?

jakubcolony (commented):
Can't say too much about the code (besides 2 minor comments) apart from the fact that it runs (I don't have a safe deployed yet). 
When I was adding some icons recently Arren told me we tend to use 30 30 SVG viewbox for consistency so you might want to have a look at that (yours is 32 32).

ArmandoGraterol (changes requested):
Excellent work with all of this Will! From what I can test with a safe on Rinkeby it is looking good! I've noticed some things in the code and I also have some questions but overall I quite like how it is looking

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/CheckSafe.tsx
```diff
@@ -0,0 +1,235 @@
...
+      setIsLoadingSafe(true);
+    }
+  };
+
+  const handleNextStep = () => {
+    setStep((step) => step + 1);
+    if (moduleContractAddress) {
+      setIsLoadingModule(true);
+      // Don't run validation until step has increased.
+      setTimeout(() => validateField('moduleContractAddress'), 0);
```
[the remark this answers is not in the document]
willm30: @ArmandoGraterol This validation needs to wait until the step has been incremented. It won't run if it thinks we're still on the first page (so as not to unnecessarily slow down the validation that page). State updates are async so without the setTimeout, `validateField` gets called before `setState`. 

The only other way I can think of is using [flushSync](https://reactjs.org/docs/react-dom.html#flushsync) to make state update syncronously but it's poorly documented and seems to impact performance. Seems worse than just making `validateField` wait until state has been updated. 

Open to suggestions.

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/CheckSafe.tsx
willm30: Good idea

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialog.tsx
willm30: Another good idea

willm30:
@jakubcolony As in you haven't been able to test? You can use the following safe addresses on Rinkeby to test. Both have modules installed.

Safe 1: `0x93ebb9b7b0a3d42d3AF76930EF569F196e3821C1`, Module 1: `0x25603925a75D39AFFb22e810aC4a2A92BCfEe6B4`
Safe 2: `0xF8F46F65b7365b97691C6794d0ff235B05BFAf78`, Module 2: `0xcae11900da49EB16a14D20f31237E04542F8Da35`

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/ConnectSafe.tsx
```diff
@@ -0,0 +1,148 @@
...
+              ),
+            }}
+          />
+        </div>
+      </DialogSection>
+      <DialogSection appearance={{ theme: 'sidePadding' }}>
+        <div className={styles.info}>
+          <CopyableData
+            label={MSG.amb}
+            text={GNOSIS_AMB_BRIDGES[Number(chainId)]?.foreignAMB} // TODO: something for safes without a bridge...
```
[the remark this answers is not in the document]
willm30: @ArmandoGraterol This one was more so I don't forget. But Arren has said we're only allowing selection of chains for which we have a bridge so question answered.

willm30:
@ArmandoGraterol Thank you sir. Yet another pr thwarted by the imports. I'll get there eventually.

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialogForm.tsx
```diff
@@ -1,249 +1,97 @@
...
-    id: 'dashboard.AddExistingSafeDialog.AddExistingSafeDialogForm.safeCheck',
-    defaultMessage: `Safe {safeData, select,
-      true {found}
-      other {not found}
-    } on {selectedChain}`,
-  },
-  fetchFailed: {
-    id: 'dashboard.AddExistingSafeDialog.AddExistingSafeDialogForm.fetchFailed',
-    defaultMessage: `Could not fetch Safe details. Please check your connection and try again.`,
+    id: 'dashboard.AddExistingSafeDialog.CheckSafe.title',
```
ArmandoGraterol, line 26: Id that needs to be updated

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/CheckSafe.tsx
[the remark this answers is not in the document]
ArmandoGraterol: I was mostly curious as to why it was needed, but if this is the best solution right now then sure

ArmandoGraterol (approved):
Nice changes! From my side this is looking good :ok_hand:

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialog.tsx
```diff
@@ -1,53 +1,278 @@
...
+      return new Promise((res) => res(result));
+    }
+    // If fetching was aborted, don't show error.
+    return new Promise((res) => res(true));
+  };
   const validationSchema = yup.object().shape({
     chainId: yup.string().required(),
-    contractAddress: yup.string().address().required(),
-    safeName: yup.string().required().max(20),
+    contractAddress: yup
```
chinins, line 123: What do you think of taking validationSchema out to split the code a bit? The component seems quite loaded...

chinins (approved):
Very nice work!

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/AddExistingSafeDialog.tsx
willm30: @chinins I think this is a very good idea and have implemented.

PR #3866, opened 2022-09-19
Decision: Display objection section

### LINKED ISSUE #3779: Decisions: Implement Decision showing Objection UI
### Story

As a user, I would like to object to a Decision.

### Description

After making an objection, implement the necessary changes to show the details.

Figma: https://www.figma.com/file/oRaTwsD5BUH3FJNzoZdSCv/Discussions?node-id=1018%3A54646

- [ ] Display Decisions details
- [ ] Display objection details
- [ ] Update existing routines to support functionality

danbr:
## Description

This PR implements the page that:

- [x]  Display Decisions details
- [x]  Display objection details
- [x]  Update existing routines to support functionality

Screenshot 2022-09-19 at 13 52 09

Please note: I will raise another issue to implement the timer/clocks located next to the user.

Resolves #3779


willm30 (commented):
Just to confirm, this is the screen I should be comparing with? https://www.figma.com/file/oRaTwsD5BUH3FJNzoZdSCv/Discussions?node-id=1386%3A57367

I'm noticing a few differences:
1. Action type says "Discussion" not "Proposal"
2. Beneath "Congratultions, your side won", I see the heading "Should "Proposal" be approved?", even though it's not the action type.
3. Space between claim button and the rule line above is 30px in the design, whereas about 7px here
4. The tag says "Failed" instead of "Objection" and is grey with red text instead of red with white text
5. In the objection container, there's no "Objection" tag
6. In the objection container, there is no username in the design

obj

You will know better than me which are intentional and which are oversights!

danbr:
> Just to confirm, this is the screen I should be comparing with? https://www.figma.com/file/oRaTwsD5BUH3FJNzoZdSCv/Discussions?node-id=1386%3A57367
> 
> I'm noticing a few differences:
> 
> 1. Action type says "Discussion" not "Proposal"
> 2. Beneath "Congratultions, your side won", I see the heading "Should "Proposal" be approved?", even though it's not the action type.
> 3. Space between claim button and the rule line above is 30px in the design, whereas about 7px here
> 4. The tag says "Failed" instead of "Objection" and is grey with red text instead of red with white text
> 5. In the objection container, there's no "Objection" tag
> 6. In the objection container, there is no username in the design
> 
> obj
> 
> You will know better than me which are intentional and which are oversights!

Thanks for the comments @willm30.
Indeed it's a balance of staying consistent with the existing system and incorporating the new designs. Generally, I have tried to wiring in Decisions to the existing motion pages. Therefore, the UI & behaviour should be identical to motions.
@arrenv What are your thoughts? (I have already created an issue to add the times & tag next to the username #.5)

arrenv:
> @arrenv What are your thoughts? (I have already created an issue to add the times & tag next to the username #.5)

For this at the moment, we should be leveraging Motions as they are without any changes.

willm30 (approved):
In which case, LGTM 👍

chinins (approved):
Looking good!

PR #3878, opened 2022-09-21
Fix the display of decisions list after creating decision motion

### LINKED ISSUE #3870: Decisions: DecisionsList incorrectly showing empty after navigation
Screenshot 2022-09-19 at 17 28 57

## Steps to reproduce

1. Assume 1 or 2 Decisions already exist. Can be in any state.
2. Create a Decision
3. Navigate back to Decisions homepage

## Expected behaviour

* The Decisions list shows the current Decisions from the subgraph subscription.

## Actual behaviour

* The subscription is performed, but zero Decisions are returned.

A page refresh will retrieve the appropriate Decision from subgraph and populate the DecisionsList

chinins:
## Description

The ipfs hash was not available right after decision creation (without refresh), so the whole decisions list was undefined. I added the try catch block to guard from the error. Now we just have a placeholder for the decision title until we refresh (nothing we can do about that part).

resolves #3870

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -58,6 +58,7 @@ import {
   UserReputationQueryVariables,
   UserReputationDocument,
 } from '~data/index';
+import { log } from '~utils/debug';
 
 import {
   ActionsPageFeedType,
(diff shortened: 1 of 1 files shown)


danbr (approved):
Very good Sherlock!! Well done.

Julianahlk (approved):
LGTM ✅

PR #3882, opened 2022-09-21
Date Picker Core Component

### LINKED ISSUE #3834: Date Picker Core Component

jakubcolony:
I closed my previous DatePicker [pull request](https://github.com/JoinColony/colonyDapp/pull/3851) as I was made aware it should have originated from a different branch. I also connected the DatePicker to the streaming form.

## Description

This PR adds a new component, `DatePicker` that will allow users to pick date and, optionally, time. It's based on [React Datepicker](https://reactdatepicker.com/) library.

"Plain" DatePicker:
image

DatePicker with time select:
image

DatePicker with options:
image

**New stuff** ✨

 * `DatePicker` component
 * `date-fns` package for formatting Date objects
 * `@floating-ui` package - the designs required a time dropdown to be nested inside the date dropdown and the Popper library used by both `react-datepicker` and our internal components (`Popover`) cannot handle these very well
 * `patch-package` - the library above uses TypeScript 4+ features while the dapp runs on Typescript 3.x. This package allowed me to apply a custom "patch" to make it compatible with dapp's TS version. I spoke to @chinins about this and she confirmed the cdapp will use a newer version of TS so this is only a temporary solution.

**Changes** 🏗

**Deletions** ⚰️
* Deleted the old `DatePicker` component which wasn't used anywhere in the dapp.

## TODO

- [x] Basic DatePicker component
- [x] DatePicker with time selection
- [x] DatePicker within select (https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=1195%3A241884)

Resolves #3834


jakubcolony:
@willm30's comment on my previous PR:

>@jakubcolony Good work Jakub! I will take a closer look at the code, but for now my first thought is that this should be based off https://github.com/JoinColony/colonyDapp/pull/3833 (streaming payment), as it does not belong in the "advanced" expenditure type. Then you can connect it to "Starts" (and "Ends" too or leave that for a follow up pr if it adds too much additional size to this one).

Thank you for spotting this!

>Also a few notes on the styles:

> * in [the design](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=5755%3A331599) the border is not rounded and the border color is var(--grey-blue-2);
> * the scrollbar on time picker drodown should be rounded and styled per design
> * time dropdown border is var(--temp-grey-blue-8) [in figma](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=6364%3A334590)

All should be fixed now. I checked with Arren and it's okay to leave the system's default scrollbar

>But to reiterate, looks like really solid work. Good job!

Thanks! 👍

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+});
+
+interface DatePickerFieldValue {
+  date: Date;
+  option?: any;
+}
+
+export interface DatePickerOption {
+  label: string | MessageDescriptor;
+  value: any;
```
willm30, line 33: How come value is `any` here instead of `string`?

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+      : currentOption?.label;
+
+  return (
+    <button
+      type="button"
+      className={styles.dateButton}
+      onClick={onClick}
+      ref={ref}
+    >
+      {currentOption && currentOption.hideDatePicker ? labelText : value}
```
willm30, line 71: Could read `currentOption?.hideDatePicker ?...`

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureStreamingForm.tsx
```diff
@@ -39,14 +40,16 @@ export const MSG = defineMessages({
...
     value: ExpenditureEndDateTypes.WhenCancelled,
+    hideDatePicker: true,
   },
   {
     label: MSG.limitIsReached,
     value: ExpenditureEndDateTypes.LimitIsReached,
+    hideDatePicker: true,
   },
   {
     label: MSG.fixedTime,
```
willm30, line 55: Add `hideDatePicker: false` to this one for consistency? And then remove as optional property from `DatePickerOption`? Purely a matter of personal preference.

willm30 (changes requested):
My first pass impression is that this is a solid component! Left a few small notes on the code but overall looks fabulous. 

Main question is with this UTC / local business. 

Time is being displayed as UTC, but stored in the form state as local. Why this difference? Should we not just store it as UTC if we're also displaying it as UTC?

dates

I also notice in Figma the date string displayed explicitly includes "UTC" at the end. Not sure if that's been omitted intentionally. 

Separately, the component goes off screen for longer months in the End Date section:

offscreen

And currently the default state looks like the selected state. Default in figma is just the color highlight, without the background.

Figma:
def

Branch:
branch

But good job!

willm30:
Oh, and btw. I don't think you invited `colony/frontend` to review this pr? I think I got here via the old one. Also just noting the branch naming convetion generally includes the issue number. And finally, do you know if this component is intended to supersede the existing date picker we have (`components/DatePicker`)? Or we'll hang on to both?

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,252 @@
...
+      /** Convert the "fake UTC date" to a correct date */
+      const updatedDate = new Date(UTCDate);
+      updatedDate.setUTCFullYear(UTCDate.getFullYear());
+      updatedDate.setUTCMonth(UTCDate.getMonth());
+      updatedDate.setUTCDate(UTCDate.getDate());
+      updatedDate.setUTCHours(UTCDate.getHours());
+      updatedDate.setUTCMinutes(UTCDate.getMinutes());
+
+      helpers.setValue({
+        ...field.value,
```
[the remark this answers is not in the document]
jakubcolony: I'm overriding date but want to persist `option` property

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
jakubcolony: That's an interesting one as I didn't want to restrict the type of the option's value. Many HTML elements (button, input) accept a `value` attribute which is of type `string | number | string[] | undefined` so I suppose I'll change my typing accordingly

jakubcolony:
>Main question is with this UTC / local business.

>Time is being displayed as UTC, but stored in the form state as local. Why this difference? Should we not just store it as UTC if we're also displaying it as UTC?

Actually, all Date objects in JavaScript store a single moment in time in UTC but "the basic methods to fetch the date and time or its components all work in the local (i.e. host system) time zone and offset" ([mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date)). Adding that to the limitations of `date-fns` library, I need to convert the correct UTC date into an offset date which doesn't represent the correct moment in time, but makes the "local component" of Date equal to the correct UTC date (eg. 10 AM UTC becomes 10 AM BST/CEST/whichever timezone you're in). 
I hope this explanation is clear but ping me on Discord and I'll be happy to go over it in more detail

>And currently the default state looks like the selected state. Default in figma is just the color highlight, without the background.

I was under impression this isn't marking the default but today's date:
image
I think "selected" and "default" are interchangeable in this case as there must be some "selected" date by default.
@arrenv are you able to confirm this?

>And finally, do you know if this component is intended to supersede the existing date picker we have (components/DatePicker)? Or we'll hang on to both?

I don't know but the existing one doesn't seem to be used anywhere. @arrenv should I get rid of the old one?

src/modules/pages/components/ExpenditurePage/constants.ts
```diff
@@ -54,6 +54,13 @@ export const initialValues = {
       { ...initalRecipient, key: nanoid() },
     ],
   },
+  streaming: {
+    startDate: { date: new Date() },
```
willm30, line 58: With respect to setting the UTC time, could you not do it here? Then you wouldn't need to do those conversions onChange. 

Eg:

``` js

const nowUTC = () => {
  const date = new Date();
  const utc = new Date(
    date.getUTCFullYear(),
    date.getUTCMonth(),
    date.getUTCDate(),
    date.getUTCHours(),
    date.getUTCMinutes(),
  );
  return utc;
};
{
  startDate: { date: nowUTC() }
...
}

```

And remove lines 98-102 and 122-132 in `DatePicker`, and just have the component always reading from the form state (`value.date`).

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,261 @@
...
+  const dateFormatOrDefault =
+    dateFormat ?? showTimeSelect
+      ? `${DEFAULT_DATE_FORMAT}, ${DEFAULT_TIME_FORMAT}`
+      : DEFAULT_DATE_FORMAT;
+
+  return (
+    <div>
+      <ReactDatePicker
+        selected={UTCDate}
+        onChange={handleDateChange}
```
willm30, line 232: Should do  `setTouched(true)` on change also?

src/modules/core/components/Fields/DatePicker/TimePicker.tsx
```diff
@@ -0,0 +1,119 @@
...
+    selectedOptionRef.current.scrollIntoView();
+  }, [isOpen]);
+
+  const formattedTime = useMemo(
+    () => format(selectedDate, DEFAULT_TIME_FORMAT),
+    [selectedDate],
+  );
+
+  const timeOptions = useMemo(() => {
+    return [...new Array((24 * 60) / 30)].map((_, idx) => {
```
willm30, line 57: Small one. `24 * 2` makes more sense to me intuitively. "24 hours in a day, times 2 to get half hours".

willm30 (commented):
Added a few comments. Also, when using the "Ends" date picker, the month of October is still going off screen for me.

src/modules/core/components/Fields/DatePicker/TimePicker.tsx
jakubcolony: I've only just realised there's a bug in this line. The value `30` should be `timeInterval` instead

src/modules/pages/components/ExpenditurePage/constants.ts
jakubcolony: Since the "local UTC time" represents an incorrect moment in time, I think it makes more sense to do all the hacky conversion in the DatePicker component and only ever expose the correct Date object. I checked with Arren and we're not planning support for timezones other than UTC so this solution seemed like the most sensible one

jakubcolony:
@arrenv as discussed I changed the picker to work in local timezone instead of UTC and also added a tooltip that shows you the selected date in UTC. Regarding the timezone abbreviations, it seems pretty difficult to reliably get the abbreviation of the user's timezone so I opted for the default "GMT-x" notation:
image

src/modules/core/components/Fields/DatePicker/TimePicker.tsx
```diff
@@ -0,0 +1,120 @@
+import React, { useEffect, useMemo, useRef, useState } from 'react';
+import { format } from 'date-fns';
+import { defineMessages } from 'react-intl';
+import { flip, useFloating } from '@floating-ui/react-dom';
+import classnames from 'classnames';
+
+import Icon from '~core/Icon';
+import styles from './DatePicker.css';
+import { DEFAULT_TIME_FORMAT } from './constants';
```
willm30, line 9: I'm always the one who forgets to check this, but styles come last in the imports list ⬆️🆙

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -0,0 +1,275 @@
...
+import Icon from '~core/Icon';
+
+import TimePicker from './TimePicker';
+import {
+  DEFAULT_DATE_FORMAT,
+  DEFAULT_TIME_FORMAT,
+  MONTH_NAMES,
+} from './constants';
+import styles from './DatePicker.css';
+import { Tooltip } from '~core/Popover';
```
willm30, line 19: Same as below (styles last)

willm30 (approved):
Nice work Jakub! Looks good to me. 🙌🙌

src/modules/core/components/Fields/DatePicker/constants.ts
```diff
@@ -0,0 +1,18 @@
+export const MONTH_NAMES = [
```
arrenv, line 1: These should probably use our language system.

arrenv (changes requested):
@jakubcolony Looks great, nice work.

A couple of comments and requests.

1. I think we should get rid of the `GMT +X`. It should simplify the appearance, and with the very nice tooltip, it provides enough explanation for the timezones.

2. The alignment of the month and the arrows is off.

32786407-d9f808a75974763b7169e304794c5fed

3. However, with that, there is an oversite from the product side. I think we really need to include the year in the date picker.

This has been updated here - https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=6364%3A334590

4. We have six caret arrows and two down ones. Do we need to introduce another?

Julianahlk (approved):
Great work Jakub! The date picker is working smoothly 🙌
Screen Shot 2022-09-27 at 12 12 40 AM
Screen Shot 2022-09-27 at 12 18 47 AM
---
I do have a question regarding old dates, should the user be able to set a start date in the past? And would it make more sense to grey out any date in the past to make it clearer that is also an old date?

arrenv:
> I do have a question regarding old dates, should a user be able to set a start date in the past? And would it make more sense to grey out any date in the past to make it clearer that the selected date is also an old date?

Good pickup @Julianahlk, for inspiration, @thomassdk solved this here fairly recently - https://github.com/JoinColony/coinMachine/pull/196

jakubcolony:
>The alignment of the month and the arrows is off.

Can you check now? They're aligned on my machine:
image

I added an optional `minDate` prop (similarly to Thomas's solution) which disables the days and times before the specified min date. 
@arrenv which colour should I use for past dates?

arrenv:
> > The alignment of the month and the arrows is off.
> 
> Can you check now? They're aligned on my machine: image
> 
> I added an optional `minDate` prop (similarly to Thomas's solution) which disables the days and times before the specified min date. @arrenv which colour should I use for past dates?

Hey @jakubcolony, you can use this color `--temp-grey-6`.

jakubcolony:
@arrenv that has now been updated:
image

src/modules/dashboard/components/ExpenditurePage/ExpenditureSettings/ExpenditureStreamingForm.tsx
```diff
@@ -39,18 +40,21 @@ export const MSG = defineMessages({
...
+const endDateOptions: DatePickerOption[] = [
   {
     label: MSG.whenCancelled,
     value: ExpenditureEndDateTypes.WhenCancelled,
+    hideDatePicker: true,
   },
   {
     label: MSG.limitIsReached,
     value: ExpenditureEndDateTypes.LimitIsReached,
+    hideDatePicker: true,
```
arrenv, line 52: Wouldn't it be better to have this as toggle on, i.e. showDatePicker: true, then default, should be not having to include it.

PR #3891, opened 2022-09-22
Refactor contract parser validation

### LINKED ISSUE #3827: Improve contract parser validation
In the contract interaction section of the Safe control form, there are a few more validation checks that could be added. 

For example: 
1. int/uint begin at size 8 bits and go up in step 8 to 256. We can validate they don't overflow.
2. Currently the max safe int `115792089237316195423570985008687907853269984665640564039457584007913129639935` triggers an error, which it shouldn't.
3. Functions can accept tuples as inputs. 

I'm sure there are more, less common types we could validate. ([official list](https://docs.soliditylang.org/en/v0.8.11/types.html))

willm30:
## Description

This PR refactors the contract parser validation to accomodate arrays, and all of the various uint/int and byte sizes. It also moves the validation schema logic into its own file as it is getting rather large.

Please see the end of the issue for some relevant observations about performance.

**New stuff** ✨

* Validation helpers
* Add `formatjs/intl` package to format messages outside of React components (facilitates extraction of validation schema)

**Changes** 🏗

* Extracted validation schema into its own file 
* Refactor contract function parser validation

To test:
1. Go to Contract Interaction in Safe Control
2. Use the abi below if you want or use your own
3. Test out the validation. Should permit arrays and have the correct error messages. 

BytesX needs to be length x * 2 + 2 and start with 0x
intX needs to be within  +-((2 ** (X - 1)) - 1)
uintX needs to be 0 or above and less than 2 ** X  - 1
Arrays need to start with '[', end with ']', and each item needs to validate correctly
bool needs to be 'true' or 'false'
address needs to begin '0x' and be a valid address

 
## Test ABI

The following ABI covers a good cross-section of the added validation.

```
[
	{
		"inputs": [
			{
				"internalType": "bytes4[]",
				"name": "a",
				"type": "bytes4[]"
			},
			{
				"internalType": "bytes30[]",
				"name": "b",
				"type": "bytes30[]"
			},
			{
				"internalType": "bytes16",
				"name": "c",
				"type": "bytes16"
			},
			{
				"internalType": "address[]",
				"name": "d",
				"type": "address[]"
			},
			{
				"internalType": "address",
				"name": "e",
				"type": "address"
			},
			{
				"internalType": "uint56",
				"name": "f",
				"type": "uint56"
			}
		],
		"name": "one",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	},
	{
		"inputs": [
			{
				"internalType": "uint200[]",
				"name": "g",
				"type": "uint200[]"
			},
			{
				"internalType": "uint256[]",
				"name": "h",
				"type": "uint256[]"
			},
			{
				"internalType": "int256[]",
				"name": "i",
				"type": "int256[]"
			},
			{
				"internalType": "int8",
				"name": "j",
				"type": "int8"
			},
			{
				"internalType": "bool[]",
				"name": "k",
				"type": "bool[]"
			},
			{
				"internalType": "bool",
				"name": "l",
				"type": "bool"
			},
			{
				"internalType": "string[]",
				"name": "m",
				"type": "string[]"
			},
			{
				"internalType": "string",
				"name": "n",
				"type": "string"
			}
		],
		"name": "two",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	}
]
```

## Discussion

The `Safe Control` form can be of arbitrary size given the user's ability to add transactions. Currently, the entire form validates when any of the inputs change. Validation takes time to run, and with the increasing complexity added here, I'm starting to notice a bit of input lag. For example, hold down a single key in an input field and it gets a bit choppy. 

Performance will degrade with form size, since a larger form requires more validation to be run, and could degrade noticeably if, say, we had a large form full of contract interaction transactions, all of which had many inputs with many large arrays. 

I tested this out as a proof of concept. In the gif below, I have 5 Contract Interaction Tabs open, all with the abi in this issue, but only the first one contains any user input. You can see that there's a lag as I hold down the "1" key, and that's only doing so for a second or so (the longer you hold it down for, the worse it gets).
 

I think this indicates a problem with how we're validating this form. I haven't thought too much about the solution since I wanted to first confirm my observations. However, my first thought is that we validate inputs at the field level onchange (or even onblur), and only run entire form validation once when the form is submitted . The easiest way to do this may to be alter the `Input` component to accept a `validateOnChange/OnBlur` prop and call `validateField()` inside the input's `onChange/Blur` function if it is true.

Thoughts very welcome.

Resolves #3827


ArmandoGraterol (approved):
From what I could test, this is working as expected :+1: 

I've noticed the same thing with the performance, I believe we should create a separate issue to investigate and explore what are our possibilities.

I feel like the approach of doing field-level validation instead of form-wide validation could be a worthwhile approach to explore.

src/modules/dashboard/components/Dialogs/ControlSafeDialog/validation.ts
```diff
@@ -0,0 +1,255 @@
...
+import { defineMessages, MessageDescriptor } from 'react-intl';
+import * as yup from 'yup';
+
+import { intl } from '~utils/intl';
+import { validateType } from '~utils/safes';
+import { TransactionTypes } from './constants';
+
+const MSG = defineMessages({
+  requiredFieldError: {
+    id: 'dashboard.ControlSafeDialog.requiredFieldError',
```
Julianahlk, line 12: You are missing the validation part in all the ids, i.e: 
id: 'dashboard.ControlSafeDialog.validation.requiredFieldError'

willm30: @Julianahlk Thanks. Please re-review when you can.

src/utils/safes/contractParserValidation.ts
```diff
@@ -0,0 +1,163 @@
...
+const getArrayFromString = (array: string) => {
+  if (array === '[]') {
+    return [];
+  }
+  const vals = array.substring(1, array.length - 1);
+  return vals.split(',');
+};
+
+const isAddressValid = (value: string) => {
+  return isAddress(value);
```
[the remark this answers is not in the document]
willm30: The validationFn is passed a second input that isn't the same type as the expected second input in `isAddress`

willm30:
> I feel like the approach of doing field-level validation instead of form-wide validation could be a worthwhile approach to explore. What do you think?

@ArmandoGraterol Agree. Have opened this issue: https://github.com/JoinColony/colonyDapp/issues/3917

Julianahlk (approved):
Good work with the new validation Will 👍

PR #3892, opened 2022-09-22
Refactor: Move DefaultMotion content into separate component

### LINKED ISSUE #3828: Decisions: Refactor `pageFeedContent` into separate component
## Specification

DefaultMotion is growing too large and now is a suitable time to refactor.

### Description 

Refactor the page feed content to support Decisions and existing. 
Reconsider this refactoring with new information about supporting existing comments & system messages.

*This should be branched from Feature:Simple Decisions, after the resolver PR has been merged.

jakubcolony:
## Description

This PR refactors `DefaultMotion` component by extracting the `pageFeedContent` and the queries it depends on into a separate component, `ActionPageMotionContent`. 

Resolves #3828

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionPageMotionContent/ActionPageMotionContent.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionPageMotionContent/ActionPageMotionContent.tsx
@@ -1,0 +1,161 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import { CommentInput } from '~core/Comment';
+import ActionsPageFeed, {
+  ActionsPageFeedItemWithIPFS,
+  SystemMessage,
+} from '~dashboard/ActionsPageFeed';
+
+import {
+  AnyUser,
+  Colony,
+  ColonyActionQuery,
+  useEventsForMotionQuery,
+  useMotionObjectionAnnotationQuery,
+  useMotionsSystemMessagesQuery,
+  useUser,
+} from '~data/index';
+
+import ActionPageDecisionWithIPFS from '../ActionPageDecisionWithIPFS';
+
+import styles from './ActionPageMotionContent.css';
+
+interface Props {
+  colony: Colony;
+  colonyAction: ColonyActionQuery['colonyAction'];
+  isDecision: boolean;
+  initiator: AnyUser;
+  walletAddress: string;
+  actionAndEventValues: any;
+  roleTitle: string;
+  transactionHash: string;
+  userHasProfile: boolean;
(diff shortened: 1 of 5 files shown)


danbr:
@jakubcolony Hey - Could you please rebase this with the feature branch? Its hard to see your changes with so many updated files. Thanks

src/modules/dashboard/components/ActionsPage/ActionPageMotionContent/ActionPageMotionContent.tsx
```diff
@@ -0,0 +1,161 @@
...
+                colony={colony}
+                user={objectionAnnotationUser}
+                annotation
+                hash={objectionAnnotation.motionObjectionAnnotation.metadata}
+                appearance={{ theme: 'danger' }}
+              />
+            </div>
+          )}
+
+      <ActionsPageFeed
```
chinins, line 133: I would prefer that ActionsPageFeed & CommentInput should stay in the DefaultMotion. There is no reason to take them out. It would be more consistent with DefaultAction structure as well... What do you think?

chinins (commented):
Looking good - just not sure about the ActionsPageFeed & CommentInput as per my comment.

src/modules/dashboard/components/ActionsPage/ActionPageMotionContent/ActionPageMotionContent.tsx
danbr: @jakubcolony @chinins Sorry if this adds a little confusion, but the way I see it, is that all of the decision/motion content is handled in the new component. It's encapsulated. It will be easier to maintain. Which is how its been refactored.

Otherwise whats the responsibility of the new component? only handling the heading/title & objections? I think thats a bit messy. 
But If so, then the components name should be changed. 

I'll go along with the option that you decide........

danbr (changes requested):
Overall looks good. 
But unfortunately there is a bug thats making the dapp crash...

It can be recreated by creating an objection to the Decision, supplying a description, and staking.....

Before:
Screenshot 2022-09-28 at 22 38 34

After:
Screenshot 2022-09-28 at 22 36 53

You can msg me if you need help recreating.....

src/modules/dashboard/components/ActionsPage/ActionPageMotionContent/ActionPageMotionContent.tsx
jakubcolony: I'm leaning towards Dan's view that this new component handles all the content rather than just the heading but I'll wait to see if I can get another reviewer's opinion

PR #3904, opened 2022-09-23
Refactor: SingleUserPicker in Control Safe dialog

### LINKED ISSUE #3865: Refactor SingleUserPicker to provide a better user experience for Safe Control
### Story

1. When clicking on the target contract address input field without someone writing anything, the message “it looks like I couldn’t find what you are looking for” pops up. The expectation is that there wouldn’t be any dropdown message until the user had a chance to write or paste something into the input.
2. If you paste the target contract address into the input field but you don’t actually select the address from the given drop down option but instead move to the ABI/JSON field and start typing or try to paste something, you can't actually do anything. The expectation would be the user won't need to select the address only from the drop down options, the user should be able to paste any contract address.
3. When you do actually select the target contract address from the drop down, the address appears in red, which could confuse the user into thinking there is an error. 

singlepicker

## Implementation/Suggestions

1. Use dirty to detect whether or not the field has been modified. Note, this does not seem to be an issue in the Payment modal for example.
2. This should be able to work if you enter in a contract without a dropdown selection. The validation could happen onBlur.
3. Please change the color of the contract address (only for the Safe Control version). You can re-use the same style used in the (input: select safe). 

image

jakubcolony:
## Description

This PR improves the user experience of SingleUserPicker used in Target contract address field in the Control Safe Dialog by addressing the following issues raised in #3865:

>When clicking on the target contract address input field without someone writing anything, the message “it looks like I couldn’t find what you are looking for” pops up. The expectation is that there wouldn’t be any dropdown message until the user had a chance to write or paste something into the input.

>If you paste the target contract address into the input field but you don’t actually select the address from the given drop down option but instead move to the ABI/JSON field and start typing or try to paste something, you will notice immediately after you can't do any other actions. The expectation would be the user won't need to select the address only from the drop down options, the user should be able to paste any contract address.

There's a new prop in `SingleUserPicker` controlling whether the currently entered value should be set on blur (by default it's only set after clicking an item)

>When you do actually select the target contract address from the drop down, the address appears in red, which could confuse the user into thinking there is an error.

https://user-images.githubusercontent.com/112586815/192036354-1c4cb1f7-440c-40ed-80be-d76b18102990.mov

**Changes** 🏗

* New props in `SingleUserPicker`: `renderEmpty` and `setValueOnBlur`
* Tweaks in `ContractInteractionSection`

Resolves #3865


Julianahlk (approved):
Well done Jakub! The SingleUserPicker is now working much better 👍

addressfound
noaddressfound

Julianahlk:
As discussed on this [PR](https://github.com/JoinColony/colonyDapp/pull/3905#issuecomment-1262838545). The work done here can be used in the future when the refactoring of the [SingleUserPicker](3587) resumes. 

>Apply the changes in this PR to the core component itself so that every instance of the picker is not only consistent but has improved UX.

chinins (approved):
Looks reasonable and working as expected.

Issue #3587, opened 2022-07-06
Refactor SingleUserPicker to be generic & create specialised picking components (User, NFT, Safe)


danbr:
## Specification

SingleUserPicker provides great functionality that is now `hacked` or used in several places. Its popularity has made it a good candidate to refactor.

Please refactor SingleUserPicker to be generic (SinglePicker), that can be specialised  into components such as:

- SingleUserPicker
- SingleNFTPicker
- SingleSafePicker

There could be other uses that i am not aware of.

PR #3905, opened 2022-09-26
Fix "Address format is wrong" displayed next to the SafePicker

### LINKED ISSUE #3899: "Address format is wrong!" displayed next to input in SafePicker
## Steps to reproduce

Go to control safe. 
Click into the SafePicker input. 
Type something random. Select what you've typed. 
The error message is conjoined with your input.

## Expected behaviour

The error message should either not appear and the input text should be red, as we do with the user picker, or the error message should appear beneath the input. 

## Actual behaviour

addwrong

jakubcolony:
## Description

This PR fixes the bug described in #3899:
image

Upon investigation, I realised this happens not only in the safe picker mentioned above, but in any instance where the `SingleUserPicker` is being passed `showMaskedAddress` prop, so also in the Recipient field of the Transfer NFT section: 
image

I added an optional prop to the `MaskedAddress` component that hides the "Address format is wrong" message and set it in `SingleUserPicker` whenever the `showMaskedAddress` option is enabled to achieve consistent behaviour. @willm30 mentioned in the original issue we could potentially change the colour of the safe picker to red when it's invalid, I noticed however that the `SingleUserPicker` text is always red, not only when the field is invalid. I believe no changes are needed for error messages beneath the field since there's support for Formik/yup validation already.

The end result is that the "Address format is wrong" message doesn't get appended to both field's value anymore:
image
image

Resolves #3899

### DIFF
--- a/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
+++ b/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
@@ -4,6 +4,7 @@ import compose from 'recompose/compose';
 import classnames from 'classnames';
 
 import { useField } from 'formik';
+import { isAddress } from '@colony/colony-js/lib/utils';
 import { AnyUser } from '~data/index';
 import { Address, SimpleMessageValues } from '~types/index';
 import { getMainClasses } from '~utils/css';
@@ -233,7 +234,8 @@ const SingleUserPicker = ({
                   {value.profile.displayName ||
                     value.profile.username ||
                     value.profile.walletAddress}
-                  {showMaskedAddress && (
+                  {/* eslint-disable-next-line max-len */}
+                  {showMaskedAddress && isAddress(value.profile.walletAddress) && (
                     <span className={styles.maskedAddress}>
                       <MaskedAddress address={value.profile.walletAddress} />
                     </span>
--- a/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
+++ b/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
@@ -194,9 +194,7 @@ const SingleUserPicker = ({
       ? placeholder
       : formatMessage(placeholder);
 
-  const isValidWalletAddress = value
-    ? isAddress(value?.profile.walletAddress)
-    : false;
+  const isValidWalletAddress = isAddress(value?.profile.walletAddress || '');
 
   return (
     <div className={styles.omniContainer}>
--- a/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
+++ b/src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
@@ -194,6 +194,10 @@ const SingleUserPicker = ({
       ? placeholder
       : formatMessage(placeholder);
 
+  const isValidWalletAddress = value
+    ? isAddress(value?.profile.walletAddress)
+    : false;
+
   return (
     <div className={styles.omniContainer}>
       <OmniPickerWrapper className={getMainClasses(appearance, styles)}>
@@ -234,8 +238,7 @@ const SingleUserPicker = ({
                   {value.profile.displayName ||
                     value.profile.username ||
                     value.profile.walletAddress}
-                  {/* eslint-disable-next-line max-len */}
-                  {showMaskedAddress && isAddress(value.profile.walletAddress) && (
+                  {showMaskedAddress && isValidWalletAddress && (
                     <span className={styles.maskedAddress}>
                       <MaskedAddress address={value.profile.walletAddress} />
                     </span>


ArmandoGraterol (changes requested):
I don't think this is the correct approach. 

MaskedAddress, which is a core component that is used in many other components other than the user picker, shouldn't need an extra prop to handle a bug that actually originates from the SingleUserPicker and no other place.

A better solution would be that we should not only check if `showMaskedAddress` is `true` but also if the address that is going to be masked is a valid address, so something like this:
`showMaskedAddress && isAddress(value.profile.walletAddress)`

Also, I agree with what Will said in the issue, the value should be pink red regardless of whether the value is valid or not or the error state of the picker because that's how all our pickers in the rest of the dapp work (You can check the Payment dialog for an example on how an address picker should look, minus the masked address).

Julianahlk:
> Also, I agree with what Will said in the issue, the value should be pink red regardless of whether the value is valid or not or the error state of the picker because that's how all our pickers in the rest of the dapp work (You can check the Payment dialog for an example on how an address picker should look, minus the masked address).

@ArmandoGraterol @willm30 I talked to product and our designer a few weeks ago about changing the red color of the SingleUserPicker for Safe control only . They both agreed having the color red all the time could cause confusion about the validity of the contract address which is why I created [this](https://github.com/JoinColony/colonyDapp/pull/3904) issue. 

Sorry for the confusion, I think this should have been communicated to the feature owner (Armando) before creating an issue for this change. Making sure the user clearly knows when a contract address is valid, or not, it is very important. I also agree that consistency throughout the dapp is equally important. I propose our decision making is based on the former, creating a great experience, and then that experience should be consistent throughout the dapp. Keeping consistent with a confusing design, specifically in this use case, could be troublesome in the future.

ArmandoGraterol:
> ...which is why I created [this](https://github.com/JoinColony/colonyDapp/pull/3904) issue.

I don't agree with this issue/PR being related to Safe Control. Why should we create a Safe Control specific solution to a problem in a core component that is used in many other places in the dapp? This should have been something for `Maintenance`, hell, there's an issue already to rework the SingleUserPicker: #3587 which I was working on before Safe Control came along, the changes you are suggesting in #3904 could have been integrated into that issue.

If we want those changes in this feature branch, we have to be aware that a few of the changes #3904 is introducing are Safe Control specific, I'm almost 100% sure that these will be discarded anyways when the time comes of reworking the whole `SingleUserPicker` if the idea is that eventually, the other pickers are consistent with the ones we have in the control safe form in terms of UX and styles.

>  I propose our decision making is based on the former, creating a great experience, and then that experience should be consistent throughout the dapp. Keeping consistent with a confusing design, specifically in this use case, could be troublesome in the future.

I agree with you on this, I just don't agree with the approach, as this way I feel like we are wasting our time making specific solutions for the control safe form when the problem is not specific to this feature, and so any solution that we implement here may most likely be made obsolete anyways due to the future rework of the core component.

src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
```diff
@@ -233,7 +234,8 @@ const SingleUserPicker = ({
                   {value.profile.displayName ||
                     value.profile.username ||
                     value.profile.walletAddress}
-                  {showMaskedAddress && (
+                  {/* eslint-disable-next-line max-len */}
```
jakubcolony, line 237: I had to add this line as there seemed to be a weird bug with stylelint and eslint. I can't add a line break because it gets "fixed" back to one line, presumably by stylelint, and then eslint flags this as an issue. Have you come across this before?

Julianahlk:
@ArmandoGraterol, thanks for your input. I was not aware you were still working on the SingleUserPicker refactoring. Basically, I saw something that could be improved, I asked product and design and based on their recommendations I decided to create an issue for it thinking it could make the feature better. I agree that this could have been addressed outside of Safe Control and communicated ahead of time. Definitely, a great situation to learn from. 
Should I go ahead and close [this](https://github.com/JoinColony/colonyDapp/pull/3904) issue then?

jakubcolony:
Thanks for your suggestion @ArmandoGraterol, I initially thought about keeping the address validation encapsulated in `MaskedAddress` but this solution is much neater.

ArmandoGraterol:
@Julianahlk 

> @ArmandoGraterol, thanks for your input. I was not aware you were still working on the SingleUserPicker refactoring.

Unofficially speaking, no hahaha I haven't worked on that because I've been working on Safe Control but yeah my hope is to eventually pick up the refactor once I'm out of Safe Control arm's length (If there's nothing else more important) or someone else picks it up (If any of you guys want to take it I have no problems).

> Basically, I saw something that could be improved, I asked product and design and based on their recommendations I decided to create an issue for it thinking it could make the feature better.

Agreed :+1: I think our only mistake was just that we should have communicated more beforehand to avoid all of this confusion as you've already mentioned

> Should I go ahead and close [this](https://github.com/JoinColony/colonyDapp/pull/3904) issue then?

I don't think that's necessary, we could apply the changes that you discussed with Product but on `master` to the core component itself so that every instance of the picker is not only consistent but has improved UX, and that way the changes will not go to waste when I, or someone else, reworks the picker.

src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
ArmandoGraterol: I've come across this but with strings, but in this case, I don't think there's a solution other than storing the wallet address in a const so that you can do `... isAddress(userAddress)` or something like that

src/modules/core/components/SingleUserPicker/SingleUserPicker.tsx
```diff
@@ -193,6 +194,10 @@ const SingleUserPicker = ({
       ? placeholder
       : formatMessage(placeholder);
 
+  const isValidWalletAddress = value
+    ? isAddress(value?.profile.walletAddress)
```
ArmandoGraterol, line 198: If `value` is true then that means the `?` operator is not needed here

ArmandoGraterol (approved):
LGTM :ok_hand:

chinins (approved):
Looking good!

Julianahlk:
> I don't think that's necessary, we could apply the changes that you discussed with Product but on `master` (Actually it can also stay in this feature branch if you want but the changes would have to be made specifically to the core components, not the sections that we use in Safe Control for example) to the core component itself so that every instance of the picker is not only consistent but has improved UX, and that way the changes will not go to waste when I, or someone else, reworks the picker.

Thanks Armando. I agree with you. I will write a note on issue #3904 so Jakub's work can be used in the future, when the work on the refactoring resumes.

jakubcolony:
I will merge this after #3935 has been merged

jakubcolony:
Waiting for #3936 to be merged before I can merge this PR

willm30:
@jakubcolony Both merged :)

PR #3907, opened 2022-09-26
Add subdomain decision creation and reputation check for decisions dialog

### LINKED ISSUE #3880: Team for decision creation is always root
If you create a decision in a team other than root, it will still show that it was created in root. Despite the parameter that's passed to saga being correct.

chinins:
## Description

This PR adds 2 things:

- reputation check in the decision dialog. The form is disabled if there is no reputation and a warning message is shown
- update decision saga to have the motion id passed into the transaction (not just ROOT id). colonyNetwork had to be updated, which lead to colonyjs and subgraph update. SO you need to run `npm i` & `npm run provision` & restart the dev environment to run the dapp correclty

resolves #3880


ArmandoGraterol (approved):
Looks good to me :+1: 

FireShot Capture 721 - Create Decision - Motion - Colony - wonker - localhost

ArmandoGraterol (commented):
Just noticed this but when I try to claim the motion created in a team I get these errors:

Screenshot from 2022-10-03 22-50-43

Is this expected? I did make sure to install and run provision before running the dev env

jakubcolony (approved):
The code looks good (as far as I can tell)! 

I messaged you on Slack regarding reputation in specific teams but apart from that all works for me. One thing I noticed is the textarea, while having the disabled styling, is not really disabled and you can still type in it:
Screenshot 2022-10-04 at 08 39 51

danbr:
> One thing I noticed is the textarea, while having the disabled styling, is not really disabled and you can still type in it:

@chinins 
Good catch @jakubcolony . I will address this in a separate issue.

danbr (approved):
Looks good. I've created a separate issue for the bug @jakubcolony found.

PR #3910, opened 2022-09-27
Fix: Contract parser crash when no safe is selected

### LINKED ISSUE #3894: Contract parser crashes when safe not selected
This is separate from but related to https://github.com/JoinColony/colonyDapp/issues/3901
## Steps to reproduce

Go to safe control.
Don't select a safe. Instead select "contract interaction". Enter an address in "target contract address". 
It crashes the dapp.

## Expected behaviour

Relevant error message displayed beneath text area.

## Actual behaviour

crash

jakubcolony:
## Description

This PR fixes the bug described in #3894. The crash seems to happen in this line in `InputStatus` component:
```
const statusText =
    typeof status === 'object' ? formatMessage(status, statusValues) : status;
```
Because `typeof (null)` is `object` (🤦‍♂️), when `status` is null and not undefined it still gets passed to `formatMessage` which tries to read some properties of it, therefore crashing. 

I solved this by stopping the `onContractChange` callback execution if there is no safe selected. There might be alternative solutions, e.g.:
 * Improving UX of the entire Control Safe by hiding/disabling the target contract address field when no safe is selected,
 * Refactoring `InputStatus` to handle `null` statuses.

**Changes** 🏗

* Added condition checking if safe is selected to `ContractInteractionSection`'s `onContractChange` callback

Resolves #3894

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
@@ -151,6 +151,10 @@ const ContractInteractionSection = ({
 
   const onContractChange = useCallback(
     (contract: AnyUser) => {
+      if (!selectedSafe) {
+        return;
+      }
+
       const contractPromise = fetchContractABI(
         contract.profile.walletAddress,
         Number(selectedSafe?.chainId),


willm30 (changes requested):
Nice work Jakub. I think it would be good to add to this so that it displays an appropriate error message beneath the text area, so the user knows to select a safe.

willm30:
@jakubcolony Would you mind pinging me / clicking the re-request review button when you want me to take another look? I'm getting a lot of github notifications atm and can't always be sure.

ArmandoGraterol (approved):
I think we should expand the message to "You must select a safe first before fetching the contract's ABI" or any other better message you guys might think of. With this status message, I feel like it's important that the user knows that using the safe's chain is how we know the chain where the contract is and so that's why it's important that they select the safe first. 

Other than this, looks good to me :+1:

jakubcolony:
@ArmandoGraterol I changed the error message in line with your suggestion:
image

willm30 (changes requested):
This should be an error, as opposed to a status, in my opinion. 

This will most likely require you to add a test in the form validation for the abi. You can access the form context in yup using `this` inside a normal `function`, (`this.from` in this case as the safe is not in the parent context), so you can check whether the safe is null and if so, return the error message by using `(formatMessage(MSG...)`

Something like:

``` js
abi: yup.string().when('transactionType', {
          is: (transactionType) =>
            transactionType === TransactionTypes.CONTRACT_INTERACTION,
          then: yup
            .string()
            .required(() => MSG.requiredFieldError)
            .test(
              'is-safe-selected',
              formatMessage(MSG.safeNotSelected), // will be used as error when test returns false
              function isSafeSelected() {
                if(!this.from[1].safe) {
                    return false 
                }
              },
            ),
          otherwise: false,
        }),

```

safefirst

willm30:
@jakubcolony It may be that we could actually simplify the UI error state. With the addition of the loading spinner in https://github.com/JoinColony/colonyDapp/pull/3902, I propose we display errors like we do in the Transfer NFT section, essentially hiding the TextArea if there's a fetching error. (See here: https://github.com/JoinColony/colonyDapp/pull/3926)

In which case, you wouldn't need to add any validation. Just use setError with the message formatted as a string. See #3926. 

@ArmandoGraterol Do you think this makes sense, to display the error not beneath the textarea but instead of it?

ArmandoGraterol:
@willm30 
> @ArmandoGraterol Do you think this makes sense, to display the error not beneath the textarea but instead of it? Reasoning being the textarea and function selector aren't useful if there's an error fetching the abi.
> 
> notve

I quite like this! That sounds like a good idea to me :+1:

jakubcolony:
Thank you for your comments @willm30 and @ArmandoGraterol, I also like the idea of hiding the ABI parser. I'm adding a Blocked label to this issue and will come back to it once #3926 has been merged.

willm30:
@jakubcolony As we discussed, I've incorporated your changes here into https://github.com/JoinColony/colonyDapp/pull/3926. So should be good to close this one.

PR #3925, opened 2022-09-29
Prevent jumping of UI when displaying the transaction fee

### LINKED ISSUE #3774: Prevent jumping of UI when displaying the transaction fee
## Steps to reproduce

1. Go to create a payment in the UI
2. Notice the UI jump to show the fee when adding in a value.

## Expected behaviour

* We should be able to avoid the jumping by setting the height of the `.networkFee` class to 0.

## Actual behaviour

* UI jumps when the network fee value appears.

jumping-for-fee-text

danbr:
## Description

Simple PR that prevents jumping when displaying the transaction fee.

Can be tested by:

1. Go to create a payment dialog
2. Add payment amount
3. There shouldn't be any jumping.

Resolves #3774

### DIFF
--- a/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
+++ b/src/modules/dashboard/components/Dialogs/CreatePaymentDialog/CreatePaymentDialogForm.css
@@ -26,6 +26,7 @@
 }
 
 .domainPotBalance, .networkFee {
+  height: 0px;
   position: relative;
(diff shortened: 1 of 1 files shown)


jakubcolony (approved):
Looks good! No jumping anymore:

https://user-images.githubusercontent.com/112586815/193560847-b6354b7f-6167-44e1-a5e5-60f450c4ba38.mov

chinins (approved):
Good to go!

Issue #3917, opened 2022-09-28
Improve Safe Control Form performance


willm30:
## Description

Investigate and implement a solution to improve the performance of the Safe Control form. 

## Context
From #3891 

The `Safe Control` form can be of arbitrary size given the user's ability to add transactions. Currently, the entire form validates when any of the inputs change. Validation takes time to run, and with the increasing complexity added here, I'm starting to notice a bit of input lag. For example, hold down a single key in an input field and it gets a bit choppy. 

Performance will degrade with form size, since a larger form requires more validation to be run, and could degrade noticeably if, say, we had a large form full of contract interaction transactions, all of which had many inputs with many large arrays. 

I tested this out as a proof of concept. In the gif below, I have 5 Contract Interaction Tabs open, all with the abi in this issue, but only the first one contains any user input. You can see that there's a lag as I hold down the "1" key, and that's only doing so for a second or so (the longer you hold it down for, the worse it gets).
 
![111](https://user-images.githubusercontent.com/64402732/192004652-c876f15b-2ef2-48e4-9ea8-02d13f705dbe.gif)

I think this indicates a problem with how we're validating this form. I haven't thought too much about the solution since I wanted to first confirm my observations. However, my first thought is that we validate inputs at the field level onchange (or even onblur), and only run entire form validation once when the form is submitted . The easiest way to do this may to be alter the `Input` component to accept a `validateOnChange/OnBlur` prop and call `validateField()` inside the input's `onChange/Blur` function if it is true.

willm30:
MERGED #3927

PR #3927, opened 2022-09-29
Disable submit btn in Contract Interaction

### LINKED ISSUE #3803: Submit button enabled on contract function selection
## Steps to reproduce

1. Go to the control safe dialog
2. Select the "Contract interaction" type and fill the contract address field so that we can extract useful methods from the ABI
3. Select a function with parameters

## Expected behaviour

* The submit button of the form should be disabled, as the fields of the parameters need to be filled.

## Actual behaviour

* The submit button is enabled as if the form was completed, even though the param inputs just appeared. If you interact with the inputs, then the validation updates and disables the button, but that should be the case from the start.

**NOTE:** This should be based on #3755, if that branch hasn't been merged yet in the feature branch.

FireShot Capture 710 - Actions - Colony - wonker - localhost

willm30:
## Description

This PR fixes two problems. It made sense to fit them both into the same pr since they both needed form-wide onChange validation to be disabled.

#3803 : the problem where the form was valid after selecting a function to interact with in the contract interaction section, even if the inputs were empty. 

#3917: the issue where the contract function inputs would be laggy when inputting data because validation was running immediately on each change event. 

## Contract Interaction section validation

To solve #3803, I had to turn off form level validation on change. This was to avoid the form validating before the state had updated, which is why you could click through to the form preview. I then manually call `handleValidate` in the relevant component's change handler, which ensures that the form doesn't validate until the state has changed.

You might be wondering why not just turn off the validation at the component level. Unfortunately, not so simple. Formik by default will validate when the form state changes, and again when a value is set imperatively using `setValue`. At the component level, you can only turn off the validation on `setValue`. You cannot turn off onChange validation when a specific form field changes. Formik validates the form onChange before the state has updated (see https://github.com/jaredpalmer/formik/issues/529 for a related discussion), so if I then call `handleValidate` to ensure the errors object is correct, the result is a flashy ui where the form state quickly shifts from valid to invalid after selecting a contract function. So form-wide onchange needs to be disabled.

The down side of this is that I've had to manually add some form of onChange validation to all of the input / select components in the form. I've either done this in an effect when the input changes, or by modifying the component to accept a `validateOnChange` prop, depending on the need. I don't think there is a better way around it, but invite all such investigation!

## Form input performance

To solve #3917, I have debounced the onChange validation in the contract function inputs, so it doesn't block the component from rerendering when there's new input. This is also made possible by disabling the form-wide onChange validation. The form will only validate once a user has paused entering their input, and therefore doesn't feel that the validation is blocking them from entering input.

## Testing

Most important thing with this PR is to test that all of the inputs correctly validate on change, across the whole form. This is a pain. I'm sorry about that. I don't believe there's a better way since with formik you can't prevent just one field from validating onChange -- you have to do it at the form-wide level. Also, don't forget to test the Contract Interaction section in particular, since this was the inspiration for the PR to begin with.

Secondly, it's worth testing that you can input in a big form (e.g. lots of contract interactions) without the ui lagging.

**Changes** 🏗

* Disable form-wide onChange validation
* Add custom onChange validation individually to all components in the form to compensate. 
* Debounce the onChange validation in the form's `Inputs`, so the ui doesn't lag in large forms

Resolves #3803 and #3917


src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
```diff
@@ -86,6 +91,14 @@ const ControlSafeDialog = ({
     expandedValidationSchema,
   );
 
+  const handleSelectedContractMethods = (
+    updatedMethods: UpdatedMethods,
+    validateForm: FormikHelpers<FormValues>['validateForm'],
+  ) => {
+    setSelectedContractMethods(updatedMethods);
+    setTimeout(validateForm, 0);
+  };
```
ArmandoGraterol, line 100: The things we do to disable a button

ArmandoGraterol (commented):
Hmmm I'm noticing a flickering when selecting a different function and this flickering might be longer when there are more transactions or if a slower machine does it.

Are you noticing the same?

src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
jakubcolony: @willm30 Can I just ask why you are using `setTimeout` here? Not saying it's wrong, I'm just curious

willm30: @jakubcolony It's to avoid a problem in formik where the form validates against state that is incorrect/not uptodate. If you take out the setTimeout, the errors object will be incorrect. For example, I've noticed that even if a safe is selected, form validation will produce an error saying that the safe should not be null (when it isn't). It's a strange one for sure, and may have something to do with https://github.com/jaredpalmer/formik/issues/529.

willm30:
@ArmandoGraterol I have fixed the flickering problem. It was being caused because the form was validating immediately when the function was selected (before the form state had updated), and then I was subsequently validating the form against the correct state, resulting in the flicker from valid to invalid.

But, I have had to turn of form-wide level validation. We did mention doing this anyway to solve the performance issue in large forms, but I'm flagging as it means we need to test every form input in this PR now instead of just the contract interaction section.

And since it was related, I've thrown in the fix for the laggy ui when inputting to large forms into this pr aswell. I've just debounced the call to validateForm instead of it running immediately on change (formik's default behaviour, which i've disabled). This ensures the validation doesn't block the user's input.

jakubcolony (approved):
Nice work @willm30, the submit button is correctly disabled upon selecting a function and gets enabled once the form is valid

I didn't notice any flickering earlier or now so can't confirm if this has been fixed but the updated validation works well overall. There's a little delay due to debouncing but I think it's a good compromise in order to achieve better performance.

ArmandoGraterol (changes requested):
Looking good!

- The submit button is disabled as expected when changing functions
- Regarding performance, I think the form is a bit more resilient. You can still get stuck if you have like 4 transactions and you hold a key on an input, however, I don't know just how much of a problem that would be considering that users are going to most likely just copy-paste values.
- With this review, I'm just requesting that we revert back the imports of `lodash` as getting any kind of performance gain is important when it comes to the dapp as you can already tell :)

src/utils/lodash/index.ts
```diff
@@ -0,0 +1,7 @@
+import debounce from 'lodash/debounce';
```
willm30, line 1: @ArmandoGraterol Do you think this is a reasonable compromise? Only importing the functions we use but still getting them into the relevant file on one line.

chinins: I think it's a great idea

chinins (approved):
Good job!

willm30:
@ArmandoGraterol Can you please re-review when you get a second?

PR #3929, opened 2022-09-29
Fix: Avatar uploader gets stuck

### LINKED ISSUE #3786: User stuck if invalid avatar uploaded
## Steps to reproduce

1. Log in to the Dapp
2. Go to the `edit-profile` page 
3. Click on `choose` and enter an image that is too large (over 1mb)

## Expected behaviour

* The user is presented with an error message that explains what went wrong and can then try again.

## Actual behaviour

* No error message is given apart from a terse and wrong `filetypeError` on tooltip on hover (wrong as the error was due to attempting to upload a file over 1mb)
* The user is unable to try again with the `Remove` button being disabled and the `Choose` button being broken due to the `Input` component it is trying to trigger being removed from the DOM to show the error image. The `Choose` button now gives the following error in the console when clicked:
```
Uncaught TypeError: b.current.open is not a function
    at main.0056045e2d11cabb3bc6.js:2065:618247
    at Object.s (vendors.c5931cd20a20152fd19e.js:2:4056030)
    at p (vendors.c5931cd20a20152fd19e.js:2:4056173)
    at vendors.c5931cd20a20152fd19e.js:2:4056319
    at b (vendors.c5931cd20a20152fd19e.js:2:4056405)
    at ot (vendors.c5931cd20a20152fd19e.js:2:4071933)
    at it (vendors.c5931cd20a20152fd19e.js:2:4071743)
    at st (vendors.c5931cd20a20152fd19e.js:2:4072099)
    at pt (vendors.c5931cd20a20152fd19e.js:2:4073306)
    at B (vendors.c5931cd20a20152fd19e.js:2:4170091)
```  

screenshot

## Suggestions (optional)

* Change the choose button to call the `reset()` function for the formik form if an error is detected

jakubcolony:
## Description

This PR fixes the bug described in #3786. 

The investigation took way longer than expected and I'm still not 100% sure of the solution. The issue affects both avatar uploaders that use `AvatarUploader` which is `UserAvatarUploader` and the colony avatar uploader.
It looks like there is some error handling logic in `FileUpload` which is used internally by `AvatarUploader` but it'd be quite hard to get right considering other components depend on it. Instead, I opted for a fix that's specific to `AvatarUploader` and followed the error handling convention set out in the colony avatar uploader (handling error in its own local state).

Before:
- Clicking on Choose does nothing:
image

- The browse button disappears:
image

After:
Screenshot 2022-09-29 at 19 40 04

Screenshot 2022-09-29 at 20 35 03

**Changes** 🏗

* Reset file uploader form upon error so that another upload attempt can be made
* Display appropriate error message if invalid file is uploaded

Resolves #3786

### DIFF
--- a/src/modules/core/components/AvatarUploader/AvatarUploadItem.tsx
+++ b/src/modules/core/components/AvatarUploader/AvatarUploadItem.tsx
@@ -64,11 +64,14 @@ const AvatarUploadItem = ({
     if (file && !error && !uploaded) {
       uploadFile();
     }
-    if (error && handleError) {
-      handleError({ ...value, file });
+    if (error) {
+      handleError?.({ ...value, file });
+
+      // reset the form to allow for another upload attempt
+      reset();
     }
     // Only on first render
-  }, [handleError, file, error, uploadFile, uploaded, value]);
+  }, [handleError, file, error, uploadFile, uploaded, value, reset]);
 
   return (
     <div className={styles.main}>
--- a/src/modules/users/components/UserProfileEdit/UserAvatarUploader.css
+++ b/src/modules/users/components/UserProfileEdit/UserAvatarUploader.css
@@ -1,0 +1,3 @@
+.inputStatus {
+  margin-top: 10px;
+}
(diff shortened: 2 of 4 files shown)


chinins (approved):
Working as expected - nice work!

Julianahlk (approved):
Perfecto! 👌
Screen Shot 2022-10-04 at 7 28 42 PM
Screen Shot 2022-10-04 at 7 29 39 PM

PR #3935, opened 2022-10-03
Update colony safe type in helpers

willm30:
Small fix to fix type checking issue in the `safeHelpers.ts`, after the safe data passed into the saga was updated here https://github.com/JoinColony/colonyDapp/pull/3765


jakubcolony (approved):
The error is gone!

Julianahlk (approved):
Tested and error is not showing anymore. LGTM.

PR #3936, opened 2022-10-03
Fix webpack build error

willm30:
## Description

This PR fixes an error in github actions when building the safe control branch that was being caused by the `require(~lib/colonyNetwork/build/contracts...)`  in the `safeHelpers` file not being properly guarded for development only.

The only way I got this to work was by using a custom webpack plugin. Not sure why, but the approach of calling it only if `process.env.NODE_ENV === 'development'` didn't work. 

**New stuff** ✨

* custom webpack variable `isProduction`

### DIFF
--- a/src/modules/dashboard/sagas/utils/safeHelpers.ts
+++ b/src/modules/dashboard/sagas/utils/safeHelpers.ts
@@ -74,9 +74,19 @@ export const getHomeBridge = (safe: Omit<ColonySafe, 'safeName'>) => {
 };
 
 /* Currently only used to get the ForeignBridgeMock contract from Colony Network, for testing locally. */
-const getBuildFromColonyNetwork = (contractName: string) =>
-  // eslint-disable-next-line @typescript-eslint/no-var-requires, max-len, global-require, import/no-dynamic-require
-  require(`~lib/colonyNetwork/build/contracts/${contractName}.json`);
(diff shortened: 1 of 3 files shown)


jakubcolony (approved):
LGTM! 👍

jakubcolony:
I came across this line in Armando's PR: https://github.com/JoinColony/colonyDapp/pull/3937/files#diff-f4ddde2fbee5af57a71ed6a75340dfa1549b457f222aeea0cd3bc674d1ba0b22R347
This might be irrelevant but weren't you trying to solve the same issue yesterday (detecting if you're in dev/prod)?

willm30:
> I came across this line in Armando's PR: https://github.com/JoinColony/colonyDapp/pull/3937/files#diff-f4ddde2fbee5af57a71ed6a75340dfa1549b457f222aeea0cd3bc674d1ba0b22R347 This might be irrelevant but weren't you trying to solve the same issue yesterday (detecting if you're in dev/prod)?

In [this build](https://github.com/JoinColony/colonyDapp/actions/runs/3188729062/jobs/5201723961) I tried guarding the offending line with that variable, to no avail. So seems the webpack config approach is the solution.

danbr (approved):
LGTM

chinins (approved):
Good creative solution :)

ArmandoGraterol (approved):
:ok_hand:

PR #3937, opened 2022-10-03
Connect "Contract interaction" transaction

### LINKED ISSUE #3726: Connect "Contract interaction" transaction type action
### Description

This issue corresponds to the `Contract interaction` transaction type inside the "Control safe" dialog. In order to make the transaction, we are going to use a bridge to execute the transactions across chains.

The transaction is expected to work for cross-chain safes/contracts.

**NOTE**: This can be worked on so long as it's based on #3765, if it hasn't been merged yet.

## Implementation

* Steps to get the dev env ready for testing transactions:
```
0. Have your normal dev env already running.
1. cd src/lib/colonyNetwork 
2. git fetch
3. git checkout maint/cross-chain-safe 
4. git submodule update --init
5. yarn
6. yarn provision:safe:contracts
7. cd lib/safe-contracts then yarn
8. cd ../.. (back into colonyNetwork)
9. bash scripts/start-bridging-environment.sh
10. node scripts/start-bridging-environment.js
```
You'll see this in the console if the steps were run successfully:
```
Bridge Monitor running
Home bridge address: 0xE0EeC1f1d5Cc344206710Ad526E180e5CeEE5B1c
Foreign bridge address: 0x312787b78BA0A2C4e9092a7871E49c421c24e7Fa
Gnosis Safe address: 0xC33533885f666baB63243b9E0B8432602d8dDb52
Zodiac Bridge module address: 0x0a72DCB0584A80cada11cC95aB2386f23225C63d
```

- [ ] A help message shows below the contract address input indicating that the contract exists
- [ ] Submitting the form creates and executes a correct transaction/many transactions (in the case of a multi transaction).

### Suggestions

* These links could be helpful for testing/development:
How to find the emitted events if needed: https://docs.ethers.io/v4/api-contract.html?highlight=events#filtering-events
Example of how to make the transactions (In the context of Network, however): https://github.com/JoinColony/colonyNetwork/commit/e4484246889a4d24b19e58e9a496330e00a77f0f

ArmandoGraterol:
**To test:**

- Run `npm run dev:heavy -- --skip-webpack`
- Follow the instructions in https://github.com/JoinColony/colonyDapp/pull/3765 to get the foreign bridge running. Make sure you have the latest changes from `maint/cross-chain-safe`.
- Add the env variable values to your .env and then start webpack (`npm start webpack` in `/colonyDapp`).
- Add a safe. For example: `0x4cb09344de5bCCD45F045c5Defa0E0452869FF0f` (Rinkeby)
- Go to safe control, select "contract interaction", and fill out the form. Use any verified contract that is part of the chain of the selected safe.
- Confirm the transaction.
- In the terminal window running the foreign bridge, you should see a confirmation that funds have been transferred out of the safe to a 0 address. In the console in the browser, you should see proof that the bridging transaction was successful.

NOTE: 
- In the local env, it's not expected that the safe you added is used to transfer the funds, or that the function you select in the form is the one used for the transaction. 
- A local safe with 100 test tokens in it is going to be used.
- I've used `transferFrom` out of convenience, the process of testing this transaction is quite similar to testing the transfer funds transaction with the exception that the transfer funds version uses `transfer`  

Resolves #3726

### DIFF
--- a/src/modules/dashboard/sagas/utils/safeHelpers.ts
+++ b/src/modules/dashboard/sagas/utils/safeHelpers.ts
@@ -362,7 +362,13 @@ export const getContractInteractionData = async (
     ];
 
     const transactionValues = onLocalDevEnvironment
-      ? { ...transaction, src: safeAddress, dst: AddressZero, wad: 1 } // src, dst and wad are the param names of the transferFrom function
+      ? {
+          ...transaction,
+          // src, dst and wad are the param names of the transferFrom function
+          [`src-transferFrom`]: safeAddress,
+          [`dst-transferFrom`]: AddressZero,
+          [`wad-transferFrom`]: 1,
+        }
       : transaction;
     const args = inputs?.map(extractMethodArgs(name, transactionValues)) || [];
 
--- a/src/modules/dashboard/sagas/utils/safeHelpers.ts
+++ b/src/modules/dashboard/sagas/utils/safeHelpers.ts
@@ -352,6 +352,10 @@ export const getContractInteractionData = async (
   let abi: string;
   let contractFunction: string;
 
+  if (!safeAddress) {
+    throw new Error('LOCAL_SAFE_ADDRESS not set in .env.');
+  }
+
   if (!contractAddress) {
     throw new Error('LOCAL_SAFE_TOKEN_ADDRESS not set in .env.');
   }
--- a/src/modules/dashboard/sagas/utils/safeHelpers.ts
+++ b/src/modules/dashboard/sagas/utils/safeHelpers.ts
@@ -4,6 +4,7 @@ import { Contract, ethers } from 'ethers';
 import abis from '@colony/colony-js/lib-esm/abis';
 import moveDecimal from 'move-decimal-point';
 import { AddressZero } from 'ethers/constants';
+import { BigNumber } from 'ethers/utils';
 
 import { erc721 } from './abis'; // Temporary
 import { SafeTransaction } from '~dashboard/Dialogs/ControlSafeDialog/ControlSafeDialog';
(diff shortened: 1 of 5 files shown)


jakubcolony (approved):
LGTM! 👍

After a few issues with setting up I managed to get it running:
image
image

And was redirected to:
image

src/modules/dashboard/sagas/utils/safeHelpers.ts
```diff
@@ -280,6 +281,92 @@ export const getTransferFundsData = async (
...
+      }
+      return value;
+    });
+  } catch (err) {
+    return null;
+  }
+};
+
+const isArrayParameter = (parameter: string): boolean =>
+  /(\[\d*])+$/.test(parameter);
```
willm30, line 301: How come you're only checking for arrays containing numeric digits? If I understand correctly, this is intended to test whether the start of the string is '[' and the end ']' (i.e. an array)? In which case, the following might be simpler: 
`parameter[0] + parameter[parameter.length - 1] === '[]'`

src/modules/dashboard/sagas/utils/safeHelpers.ts
```diff
@@ -280,6 +281,92 @@ export const getTransferFundsData = async (
   ]);
 };
 
+const getParsedJSONOrArrayFromString = (
+  parameter: string,
+): (string | number)[] | null => {
+  try {
+    const arrayResult = JSON.parse(parameter);
+    return arrayResult.map((value) => {
```
willm30, line 289: I believe JSON.parse also introduces rounding errors for large numbers. E.g. 2 ** 255 - 1 (the largest safe int) will be turned into 
`5.78960446186581e+76`, which does not `===` 2 ** 255 - 1. 

Whereas `getArrayFromString` will give you an array of the numbers as strings, so you don't need to convert to a big number and then back to a string.

src/modules/dashboard/sagas/utils/safeHelpers.ts
```diff
@@ -280,6 +281,92 @@ export const getTransferFundsData = async (
...
+};
+
+const isArrayParameter = (parameter: string): boolean =>
+  /(\[\d*])+$/.test(parameter);
+
+const extractMethodArgs = (
+  signature: string,
+  transaction: Record<string, any>,
+) => ({ name = '' }) => {
+  if (isArrayParameter(transaction[name || signature])) {
```
willm30, line 307: Do we store the signature as a key in the SafeTransaction object? I've not come across this, I've only noticed that we store the function name as the key and the input data as the value.

src/modules/dashboard/sagas/utils/safeHelpers.ts
```diff
@@ -280,6 +281,92 @@ export const getTransferFundsData = async (
...
+
+  const getData = () => {
+    const foreignProvider = getForeignProvider(safe);
+    const contract = new ethers.Contract(contractAddress, abi, foreignProvider);
+    const { inputs, name = '', signature } = contract.interface.functions[
+      contractFunction
+    ];
+
+    const transactionValues = onLocalDevEnvironment
+      ? { ...transaction, src: safeAddress, dst: AddressZero, wad: 1 }
```
willm30, line 352: Perhaps worth a comment here that `src`, `dst` and `wad` are the inputs to the `transferFrom` function

willm30 (commented):
Working for me! 

Left a question about the way we're turning a string array into a javascript array.

Julianahlk (commented):
When having interactions with functions that do not contain array data, the transactions go through smoothly:
simpleinteraction
Screen Shot 2022-10-06 at 9 39 11 PM

But I am having some issues when interacting with the following array data:  
arraydata
arrays2

If there is just one single array[] there is not issue:
onearray

src/modules/dashboard/sagas/utils/safeHelpers.ts
ArmandoGraterol: You are right, I don't remember why did I include the `\d*`

src/modules/dashboard/sagas/utils/safeHelpers.ts
```diff
@@ -280,6 +281,92 @@ export const getTransferFundsData = async (
   ]);
 };
 
+const getParsedJSONOrArrayFromString = (
+  parameter: string,
+): (string | number)[] | null => {
+  try {
+    const arrayResult = JSON.parse(parameter);
```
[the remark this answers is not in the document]
ArmandoGraterol: Sure, I'll update and use that one then

src/modules/dashboard/sagas/utils/safeHelpers.ts
ArmandoGraterol: True, I think this is a leftover from when I was testing some things

ArmandoGraterol: Gonna have to update this anyways with the merge of #3952

chinins (approved):
Looking good!

ArmandoGraterol:
@Julianahlk Hey! Can you test again? I think this is some bug in the form itself rather than the code I added to connect the transaction, at least it looks like it from the gif you added, but I've rebased and fixed some weird stuff in the validation. Also, can you provide me the contract address you used to get the param of the array inside an array?

Julianahlk:
> @Julianahlk Hey! Can you test again? I think this is some bug in the form itself rather than the code I added to connect the transaction, at least it looks like it from the gif you added, but I've rebased and fixed some weird stuff in the validation. Also, can you provide me the contract address you used to get the param of the array inside an array?

I just tested again and I am getting the same `Error: invalid BigNumber`. It does look like this issue is more related to the validation and not your code. Here is the address I am using for the multidimensional array: 
`0x44762835EeB462fCe0150935593Fd119A935B16a`
Function name: unstake

And this one has single arrays:
`0xE65841eb30548E7be3FFB9e496DB1B540C5eD8C6`
Function name: addLeaderboardEntry

willm30:
@Julianahlk I've opened an issue for nested array validation: https://github.com/JoinColony/colonyDapp/issues/3986 . Good catch.

willm30 (commented):
@ArmandoGraterol Receiving the following error. The  `const = paramName` in `extractMethodArgs` doesn't account for the param names used in the token contract in testing:

arry
extractMethodArgs

ArmandoGraterol:
> @ArmandoGraterol Receiving the following error. The `const = paramName` in `extractMethodArgs` doesn't account for the param names used in the token contract in testing:
> 
> arry extractMethodArgs

Damn, sorry about that. I was tunnel visioned into the other bugs :face_with_head_bandage:

willm30 (approved):
Nice one, working for me👍

PR #3949, opened 2022-10-05
Disable `Reveal` button in voting widget when the user didn't vote

### LINKED ISSUE #3930: "Reveal" button is active for a user who didn't vote
## Steps to reproduce

1. Ensure you have 3 members in a Colony with reputration.
2. Create a Motion and fully stake with one member and fully object with another.
3. Vote as one member for, vote as one member against.
4. Log in as the third member, the "Reveal" button will be active, but won't trigger signing, and won't give an error.

## Expected behaviour

* The "Reveal" button should be disabled if there is no vote to reveal for the user.

## Actual behaviour

* The "Reveal" button is active and clickable, but does nothing.

reveal-votes-non-user

chinins:
## Description

This PR fixes the condition in motion resolver whether the user voted or not.

resolves #3930

### DIFF
--- a/src/data/resolvers/motions.ts
+++ b/src/data/resolvers/motions.ts
@@ -568,9 +568,10 @@ export const motionsResolvers = ({
           fetchPolicy: 'network-only',
         });
 
-        const hasCurrentUserVoted = !!data?.motionVoteSubmittedEvents.filter(
-          (event) => event.args.includes(userAddress),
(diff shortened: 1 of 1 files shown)


danbr (approved):
LGTM

ArmandoGraterol (approved):
LGTM :ok_hand:

PR #3950, opened 2022-10-05
Display contract name in control safe dialog

### LINKED ISSUE #3932: Refactor Contract Interaction Target Address Field
## Description

Per [this screen](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A24680), the target contract address field should display the contract's name. This can be found at this endpoint:

https://api.etherscan.io/api
   ?module=contract
   &action=getsourcecode
   &address=YourContractAddress
   &apikey=YourApiKeyToken

Or here for BSC:

https://api.bscscan.com/api
   ?module=contract
   &action=getsourcecode
   &address=YourContractAddress
   &apikey=YourApiKeyToken 

The preview screen implemented in #3868 should also be updated per [this screen](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A22878)

jakubcolony:
## Description

This PR updates the target contract address field in control safe dialog to match [the designs](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=5416%3A24680). After selecting contract address, if we're able to fetch the name, it will be displayed in the picker:
Screenshot 2022-10-05 at 18 52 54

The contract name is also shown in the transaction preview dialog:
Screenshot 2022-10-13 at 16 00 09

**Changes** 🏗

* Added `fetchContractName` function and call it on contract address change
* Updated `SafeTransactionPreview` to display contract name instead of an address

## TODO

- [x] Display contract name in the target contract address field 
- [x] Display contract name in contract interaction preview screen

Resolves #3932

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
@@ -183,13 +183,10 @@ const ContractInteractionSection = ({
           Number(selectedSafe?.chainId),
         );
 
-        setFieldValue(`transactions.${transactionFormIndex}.contract`, {
-          ...contract,
-          profile: {
-            ...contract.profile,
-            displayName: contractName || 'Unknown contract',
-          },
-        });
+        setFieldValue(
+          `transactions.${transactionFormIndex}.contract.profile.displayName`,
+          contractName || 'Unknown contract',
+        );
         setIsLoadingABI(false);
       } else {
         if (!isAddress(contract.profile.walletAddress)) {
@@ -197,13 +194,10 @@ const ContractInteractionSection = ({
         } else {
           setFetchABIError(formatMessage(MSG.noSafeSelectedError));
         }
-        setFieldValue(`transactions.${transactionFormIndex}.contract`, {
-          ...contract,
-          profile: {
-            ...contract.profile,
-            displayName: 'Unknown contract',
-          },
-        });
+        setFieldValue(
+          `transactions.${transactionFormIndex}.contract.profile.displayName`,
+          'Unknown contract',
+        );
         setIsLoadingABI(false);
       }
     },
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
@@ -190,7 +190,7 @@ const ContractInteractionSection = ({
             ...contract,
             profile: {
               ...contract.profile,
-              displayName: name,
+              displayName: name || 'Unknown contract',
             },
           });
         });
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
@@ -201,6 +201,13 @@ const ContractInteractionSection = ({
           setFetchABIError(formatMessage(MSG.noSafeSelectedError));
         }
         setIsLoadingABI(false);
+        setFieldValue(`transactions.${transactionFormIndex}.contract`, {
+          ...contract,
+          profile: {
+            ...contract.profile,
+            displayName: 'Unknown contract',
+          },
+        });
       }
     },
     [
@@ -218,9 +225,10 @@ const ContractInteractionSection = ({
   );
 
   useEffect(() => {
-    if (
+    if (!selectedSafe) {
+      setPrevSafeChainId('');
+    } else if (
       transactionValues.contract &&
-      selectedSafe &&
       // only run effect if safe chain changes
       prevSafeChainId !== selectedSafe.chainId
     ) {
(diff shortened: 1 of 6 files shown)


willm30:
@jakubcolony Should be unblocked now?

ArmandoGraterol (changes requested):
Generally, this looks good! Nice job man. There are still some cases that we have to consider:
FireShot Capture 724 - Actions - Colony - wonker - localhost

This happens when the chain is known because there's no selected safe. I suggest that instead of displaying 2 times the contract we use the "Unknown Contract" name we already use when the contract is not found.

If I select a safe, the name stays the same and the form even allows you to confirm the transaction:
FireShot Capture 726 - Actions - Colony - wonker - localhost

The form not being disabled might not be related to this issue but it's worth a look.

On my computer, I've noticed that when you select the contract, there's a brief period in which the address is shown as the display name (Like the screenshots above) before the actual contract name is fetched and the state is updated, so in the end, it looks like the input flickers. This brief period could be even longer on QA/Production, we should display the loading state when we are still fetching and updating the state of the contract's name.

willm30 (commented):
> If I select a safe, the name stays the same and the form even allows you to confirm the transaction: FireShot Capture 726 - Actions - Colony - wonker - localhost
> 
> The form not being disabled might not be related to this issue but it's worth a look.

I have not been able to reproduce the form being enabled. This is what I get: 

contractInt

However, I do agree that the message "You must select a safe" shouldn't be there when the contract is reseelcted. I have been fixed this [here](https://github.com/JoinColony/colonyDapp/pull/3998/commits/eeaf807151871457c815aec2157b5542d6b8283e)

ArmandoGraterol:
> I have not been able to reproduce the form being enabled. 

Yeah, I almost couldn't reproduce it today as I wasn't remembering what I did yesterday to do it, but here it is:

[6fed7eb7-9a80-49de-ae7a-e2e83642ac2e.webm](https://user-images.githubusercontent.com/18473896/195919854-c03da05c-068e-4fdf-a363-d6dba73d7c6b.webm)

jakubcolony:
@ArmandoGraterol I think I have fixed the issue you recorded. 
I didn't include any loading indicator as I'm not sure how that should look like. Ideally, the OmniPicker which all of our Pickers depend on would handle asynchronous data so that we can display the contract name before the user selects it rather than fetching it after selection. I didn't refactor it as part of this PR as I think we might be re-working this component during the cdapp migration soon.
What do you think? Should I include some loading indicator now, refactor the OmniPicker or is it okay to leave it for the time being?

willm30:
> Yeah, I almost couldn't reproduce it today as I wasn't remembering what I did yesterday to do it, but here it is:

Nice catch. I believe this is fixed by [the fix ](https://github.com/JoinColony/colonyDapp/pull/3998/commits/eeaf807151871457c815aec2157b5542d6b8283e)I mentioned above for the message being there when it shouldn't be.

ArmandoGraterol:
@jakubcolony 

> ... I didn't include any loading indicator as I'm not sure how that should look like. Ideally, the OmniPicker which all of our Pickers depend on would handle asynchronous data so that we can display the contract name before the user selects it rather than fetching it after selection. I didn't refactor it as part of this PR as I think we might be re-working this component during the cdapp migration soon. What do you think? Should I include some loading indicator now, refactor the OmniPicker or is it okay to leave it for the time being?

You are right in that the ideal case would be for that logic to be handled inside the picker, hell, I feel like this field shouldn't have been a picker and just an input. I think Product did it this way because they were following Gnosis Safe app example but in their case, it makes sense because they have an "Address book" that stores addresses for quick access later when making transactions.

I don't think we need to introduce changes right now to the picker with how we implemented the loading. Take a look at how we change the loading state in `onContractChange`, we are exiting the loading state when the contract's ABI is fetched but now that we are also fetching the name, we should not set loading to `false` after just fetching the ABI but after setting up the contract's name.

jakubcolony:
@ArmandoGraterol I've done it the way you suggested:

https://user-images.githubusercontent.com/112586815/196370024-04f5e28e-d27a-4db2-a4f1-b0d0220ff608.mov

I was hesitant to use the existing loading indicator at first as I didn't want contract name fetching to affect ABI loading but the impact seems minimal. Thanks for the suggestion!

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -161,32 +162,58 @@ const ContractInteractionSection = ({
...
+        } else {
+          setFetchABIError(formatMessage(MSG.fetchFailedError));
+        }
+
+        const contractName = await fetchContractName(
+          contract.profile.walletAddress,
+          Number(selectedSafe?.chainId),
+        );
+
+        setFieldValue(`transactions.${transactionFormIndex}.contract`, {
```
willm30, line 186: I believe this can be simplified to:
``` js
setFieldValue(
   `transactions.${transactionFormIndex}.contract.profile.displayName`,
   contractName || 'Unknown contract',
);
```

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -161,32 +162,58 @@ const ContractInteractionSection = ({
...
+          },
         });
+        setIsLoadingABI(false);
       } else {
         if (!isAddress(contract.profile.walletAddress)) {
           setFetchABIError(formatMessage(MSG.invalidAddressError));
         } else {
           setFetchABIError(formatMessage(MSG.noSafeSelectedError));
         }
+        setFieldValue(`transactions.${transactionFormIndex}.contract`, {
```
willm30, line 200: Same

willm30 (approved):
LGTM

jakubcolony:
Thanks for your comment Will, I made the suggested changes. You'll notice the build has failed due to style issues, I will fix that later as I have another branch checked out

ArmandoGraterol (approved):
Thank you for addressing the feedback Jakub! Looks good to me :+1:

PR #3951, opened 2022-10-05
Scroll to the currently checked option in Select component

### LINKED ISSUE #3943: Scroll to selected option in Select component
### Story

As a user ~~interacting with a contract~~, I would like ~~the functions~~ any dropdown in the ~~control safe form~~ dapp to display the currently selected ~~function~~ option, if there is one, instead of always beginning at the top. 

## Current Behaviour

List always begins from top

scroll function selector

- [ ] Opening the function selector displays the currently selected function

jakubcolony:
## Description

This PR solves the problem @willm30 described in #3943 by adding scrolling to the currently checked option. Unless there are some reasons against, I think it makes sense for every `Select` to have this behaviour so I applied the changes to all `Select` instances rather than just function select in contract interaction. Let me know if you disagree and I can change it.

https://user-images.githubusercontent.com/112586815/194142586-7f741004-c631-4a43-86eb-8c813f5a7a84.mov

**Changes** 🏗

* Moved existing ref "callback" to useEffect in `SelectOption`
* Added useEffect that scrolls to the checked option

Resolves #3943

### DIFF
--- a/src/modules/core/components/Fields/SelectOption/SelectOption.tsx
+++ b/src/modules/core/components/Fields/SelectOption/SelectOption.tsx
@@ -1,4 +1,10 @@
-import React, { KeyboardEvent, SyntheticEvent, useCallback } from 'react';
+import React, {
+  KeyboardEvent,
+  SyntheticEvent,
+  useCallback,
+  useEffect,
+  useRef,
+} from 'react';
 import { defineMessages, useIntl } from 'react-intl';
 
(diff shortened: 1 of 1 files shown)


willm30 (approved):
I've tested this in a few places the `Select` component is used. Can't see any problems! Think it's a good idea.

chinins (approved):
Very good!

PR #3952, opened 2022-10-05
Improve validation for same param names

### LINKED ISSUE #3896: Validation is overwritten when there are two params of the same name in different transactions
## Steps to reproduce

Go to safe control.
Go to Contract Interaction.
Enter a target contract, select a function, validate the transaction and add another Contract Interaction transaction.
Now you have two transactions open, enter this ABI in transaction one:

``` json
[
	{
		"inputs": [
			{
				"internalType": "uint8",
				"name": "a",
				"type": "uint8"
			}
		],
		"name": "one",
		"outputs": [
			{
				"internalType": "uint8",
				"name": "",
				"type": "uint8"
			}
		],
		"stateMutability": "pure",
		"type": "function"
	}
]
```

And this ABI in transaction 2:

``` json
[
	{
		"inputs": [
			{
				"internalType": "bool",
				"name": "a",
				"type": "bool"
			}
		],
		"name": "b",
		"outputs": [
			{
				"internalType": "bool",
				"name": "",
				"type": "bool"
			}
		],
		"stateMutability": "pure",
		"type": "function"
	}
]

```

Both functions `one` and `two` share a parameter called `a`. 

Both should maintain their respective type validation. However, now you will notice that the first `a`  is validated as a bool instead of a unit8.

## Expected behaviour

Function params in previous transactions don't have their validation schema overwritten by params of the same name in subsequent transactions.

## Actual behaviour

overwrite

## Suggestions (optional)

* We could rename duplicate function params to `a(1), a(2), a(3)...` so they don't get overwritten? But hide this from the user, so it just ensures that our validation works correctly.

ArmandoGraterol:
Fixed validation of contract interaction transactions overriding validation when there is more than 1 param with the same name but different types. For an example, take a look at the issue's description.

Resolves #3896

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
@@ -71,9 +71,12 @@ const ControlSafeDialog = ({
 
       Object.values(selectedContractMethods).forEach((method) => {
         method?.inputs?.forEach((input) => {
-          updatedExpandedValidationSchema[
-            `${input.name}(${input.type})`
-          ] = getMethodInputValidation(input.type, method.name);
+          const inputName = `${input.name}(${input.type})-${method.name}`;
+          if (!updatedExpandedValidationSchema[inputName]) {
+            updatedExpandedValidationSchema[
+              inputName
+            ] = getMethodInputValidation(input.type, method.name);
+          }
         });
       });
 
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
@@ -72,7 +72,7 @@ const ControlSafeDialog = ({
       Object.values(selectedContractMethods).forEach((method) => {
         method?.inputs?.forEach((input) => {
           updatedExpandedValidationSchema[
-            input.name
+            `${input.name}(${input.type})`
           ] = getMethodInputValidation(input.type, method.name);
         });
       });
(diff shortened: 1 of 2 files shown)


jakubcolony:
Hey @ArmandoGraterol, your fix works great for parameters with the same name and different types but I'm experiencing issues with parameters that share both name and type. I edited Will's example from #3896 but made both parameters of type "bool" (ABIs below). The validation only seems to be working for the latter field:

image

Here are the ABIs I used:
```
[
	{
		"inputs": [
			{
				"internalType": "bool",
				"name": "a",
				"type": "bool"
			}
		],
		"name": "one",
		"outputs": [
			{
				"internalType": "uint8",
				"name": "",
				"type": "uint8"
			}
		],
		"stateMutability": "pure",
		"type": "function"
	}
]
```
```
[
	{
		"inputs": [
			{
				"internalType": "bool",
				"name": "a",
				"type": "bool"
			}
		],
		"name": "b",
		"outputs": [
			{
				"internalType": "bool",
				"name": "",
				"type": "bool"
			}
		],
		"stateMutability": "pure",
		"type": "function"
	}
]
```

ArmandoGraterol:
@jakubcolony Thank you for catching that! That was weird, Formik is behaving really weird when it comes to fields with the same name that is dynamically added. Couldn't really find the root of the problem but I did find a solution/workaround

willm30:
@ArmandoGraterol If it's any consolation, I've also found formik to be behaving strangely in this form (hence, e.g. the need to use setTimeout to get correct state to validate against).

willm30 (approved):
Working for me 👍 Nice fix.

jakubcolony (approved):
Works great now, cool stuff! 👍

chinins (approved):
Good to go!

PR #3953, opened 2022-10-06
Disable finalize button during escalation

### LINKED ISSUE #3934: Finalize button is incorrectly active while still in an escalation phase
## Steps to reproduce

1. Have at least three members in a Colony with reputation. No single person should have over the voting threshold in reputation (default is 70%)
2. Disable the Reputation monitor so that it does not fast forward you through the Motion on each transaction - http://localhost:3001/reputation/monitor/toggle
3. Create a Motion in a sub-domain.
4. Fully stake and fully object to the Motion to get to a Vote.
5. Vote so that the Motion has more Reputation waited for "No".
6. Reveal all votes.
7. The escalation phase should start.
8. The finalize button will appear clickable, but won't work, because the Motion is in the escalation phase.

## Expected behaviour

* The "Finalize" button should be disabled while in a phase that won't allow you to finalize, i.e. the escalation phase.

## Actual behaviour

* You are able to click "Finalize" while in the escalation phase, however, the transaction fails.

can-finalize-in-escalation

Screenshot 2022-09-30 132221

Screenshot 2022-09-30 132149

danbr:
## Description

This PR, changes the behaviour of the dapp so that the finalise button is disabled when a motion is in the escalation phase.

Resolves #3934

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/FinalizeMotionAndClaimWidget/FinalizeMotionAndClaimWidget.tsx
@@ -340,7 +340,10 @@ const FinalizeMotionAndClaimWidget = ({
                     appearance={{ theme: 'primary', size: 'medium' }}
                     text={MSG.finalizeButton}
                     disabled={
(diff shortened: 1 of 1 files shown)


chinins (approved):
Goog to go!

ArmandoGraterol (approved):
Nice one!

PR #3954, opened 2022-10-06
Upgrade voting & reveal widgets

### LINKED ISSUE #3931: Vote is not being detected, allowing me to vote multiple times
## Steps to reproduce

1. Have at least two members in a Colony with reputation. No one should have over the voting threshold in reputation(default is 70%),
2. Disable the Reputation monitor so that it does not fast forward you through the Motion on each transaction - http://localhost:3001/reputation/monitor/toggle
3. Create a Motion
4. Fully stake with one member and fully object with the other member on the Motion to get to a Vote.
5. Vote on the Motion.

This may take a couple of tries to get, as it was not happening every time. It was first reported by DAO Club on mainnet and I have replicated it locally following this process.

Note: I have noticed that this seems to fix it self once all people have voted. i.e. the vote is being detected, however, the reveal option is visible one all people have voted.

The same issue seems to also happen with the "Reveal" button, i.e. all people need to reveal, before it is update in the UI for each person that they clicked the reveal button.

## Expected behaviour

* The vote should be detected, let the user know their vote, but also allow them to change their vote.
* Clicking the reveal button should notify the user that they have successfully revealed their vote.

## Actual behaviour

* Voting was not detected, and user can vote multiple times.
* Reveal button does not notify the user that they successfully clicked the reveal button.

vote-not-saving

chinins:
## Description

This PR adds `useHasVoted` condition to the VotingWidget & also updates state management in `RevealWidget`

resolves #3931

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/RevealWidget/RevealWidget.tsx
+++ b/src/modules/dashboard/components/ActionsPage/RevealWidget/RevealWidget.tsx
@@ -1,4 +1,4 @@
-import React, { useCallback, RefObject } from 'react';
+import React, { useCallback, RefObject, useState, useEffect } from 'react';
 import { FormikProps } from 'formik';
 import { defineMessages, FormattedMessage } from 'react-intl';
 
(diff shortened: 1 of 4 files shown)


ArmandoGraterol (approved):
Nice changes! It's working as expected :+1:

danbr (approved):
Very fiddly but working well for me.... Behaving as expected. Well done - good job.

arrenv (approved):
Working great @chinins, very nice!!

PR #3963, opened 2022-10-10
Add validation schema in streaming payment

agnieszkajarosikloj:
## Description

This PR adds validation schema in streaming payment type.

## TODO

- [ ] Add TODOs


jakubcolony (changes requested):
Hey @agnieszkajarosikloj, nice work with the validation!

A couple of comments:

In Figma, there's an error message under this field if it's invalid:
image
image

There seems to be no validation for limit (if you select "When limit is reached option"):
image
I can submit the form if it's empty too

It seems that the end date colours got broken (not sure if it happened in this PR or not). They should look like this:
image
And now they look like this (also DatePicker colours were affected):
image

agnieszkajarosikloj:
@jakubcolony I have added changes. Can you re-review please?

jakubcolony (changes requested):
Thanks for making the changes @agnieszkajarosikloj!

I spotted a few things with Limit field validation:

If you enter a character that is allowed in a number input but is not a valid number (e.g. `,`, `.`, `-`) you get an ugly-looking error:
image
It works nicely in the Rate field however:
image

If you enter something invalid in the limit field (particularly a non-numeric characters such as above), then change the "Ends" option to anything other than "Limit is reached", you can't see any errors yet the form submission is blocked:

https://user-images.githubusercontent.com/112586815/196156489-3613e8bb-3ee1-410c-9385-ca8502a770a6.mov

agnieszkajarosikloj:
@jakubcolony I have added fix. It's ready for re-review.

jakubcolony (approved):
@agnieszkajarosikloj great job, everything works nicely now:
image
I can also confirm the validation works if you left something over in the limit field and subsequently switched to a different "Ends" option

chinins (approved):
Looking good - great job!

src/modules/dashboard/components/ExpenditurePage/Streaming/FieldError/FieldError.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { useIntl } from 'react-intl';
+import classNames from 'classnames';
+
+import styles from './FieldError.css';
+
+interface Props {
+  name: string;
+}
+
+const FieldError = ({ name }: Props) => {
```
jakubcolony, line 12: I just had a thought (and I might be 100% wrong), is this component needed or could you maybe use the existing `InputStatus`?

PR #3965, opened 2022-10-10
Fix displayed decision & objetion initiator

### LINKED ISSUE #3959: Decisions: Current user is being shown next to Decision details
## Description

The logged in user is showing at the top, even if they didn't create the Decision.

The author should be shown next to the Decision details. Please also check that the author of the objection should be used next to the objection details.

Screenshot 2022-10-10 at 16 25 27

chinins:
## Description

This PR fixes the display of the initiator of a decision & an objection. Previously, it was the logged in user that was displayed incorrectly

Screenshot 2022-10-10 at 12 46 54

resolves #3959


willm30 (approved):
Can confirm that the correct users are being displayed for initiator of the decision and objector 👍

danbr (approved):
Well done..... working as expected. Good job!

PR #3975, opened 2022-10-11
Fix: GasStationPopover causing horizontal overflow

### LINKED ISSUE #3781: Horizontal overflow when clicking from user profile to address dropdown on first clicks
## Steps to reproduce

1. Load the colony Dapp
2. Click on the user profile to view the dropdown
3. Click on the address dropdown 

## Expected behaviour

* There is no horizontal overflow

## Actual behaviour

* Horizontal overflow occurs when clicking between these 2 dropdowns, the profile dropdown has to be clicked first otherwise this doesnt happen, this happens regardless of the dropdown being closed or not.

https://user-images.githubusercontent.com/32598350/187181342-126aad81-f15c-46bb-905f-db1cd781790a.mp4

jakubcolony:
## Description

This PR fixes the issue described in #3781. Unfortunately after spending good few hours on it I was unable to pinpoint what the issue is exactly (my bet is either Popper positioning not working properly or a Chrome-specific issue, since it doesn't occur in other browsers). After a call with @chinins I decided to amend the placement of the gas station popover so that it "faces away" from the right edge of the window and doesn't cause overflow:
image

The issue was quite tricky to reproduce in first place so here are the steps I followed:
 - Connect a wallet
 - Go to the Profile page
 - Click on the address button
You should see a horizontal overflow equal to roughly the width of the popover

Resolves #3781

### DIFF
--- a/src/modules/users/components/GasStation/GasStationPopover.css
+++ b/src/modules/users/components/GasStation/GasStationPopover.css
@@ -1,3 +1,1 @@
-@value refWidth: 106px;
-@value horizontalOffset: 20px;
 @value verticalOffset: 24px;
--- a/src/modules/users/components/GasStation/GasStationPopover.css.d.ts
+++ b/src/modules/users/components/GasStation/GasStationPopover.css.d.ts
@@ -1,3 +1,1 @@
-export const refWidth: string;
-export const horizontalOffset: string;
 export const verticalOffset: string;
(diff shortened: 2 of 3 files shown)


chinins (approved):
I think that this is a good solution.

Julianahlk (commented):
If this is the intended behavior then it looks good to me:
overflow
@jakubcolony can you confirm?

PR #3976, opened 2022-10-12
Fix colonies list positioning, height & horizontal scroll

### LINKED ISSUE #3658: Horizontal scroll moves page over Colony list
## Steps to reproduce

1. (mac)F12 to bring up dev console or reduce window size to make horizontal scroll necessary
2. Scroll horizontal

The page will move over the Colony list on the left side.

Jul-20-2022 08-06-20

chinins:
## Description

This PR updates css in the `Default` layout so that colonies list looks correct on all screens (I've tested on different pages, but please, test as well, including landing page, home page when you have vertical there etc).

resolves #3658


jakubcolony (approved):
Nice fix! Works for me 👍

willm30 (commented):
Horizontal fix is working for me. 

I'm noticing, however, an overlap of the colonies list and the add colony button when many colonies are present. Could that be addressed here or should I open a separate issue? 

many-cols

chinins:
> Horizontal fix is working for me.
> 
> I'm noticing, however, an overlap of the colonies list and the add colony button when many colonies are present. Could that be addressed here or should I open a separate issue?
> 

Open a new issue, please. I think that product should confirm the UI that we should have in this case.

willm30 (approved):

Julianahlk (approved):
Working well 👍

PR #3977, opened 2022-10-12
Preselect token in Transfer Funds

### LINKED ISSUE #3970: Opening second TransferFund section, token isn't shown
## Steps to reproduce

Go to safe control
Open a Transfer Funds transaction and validate
Open another one

## Expected behaviour

The token should be preselected

## Actual behaviour

No token preselected

no-sele

willm30:
## Description

This fix ensures that the Token is preselected when opening a new transfer funds tab in Safe Control.

## To test
1. Open two Transfer Funds transaction tabs
2. The second should have the token preselected. 
3. If you use this safe on rinkeby, you'll be able to select an ERC20 token in the safe:
`Safe address: 0x358E6D7C8Fa5e55fc6A0D50E23b875c6c49fF09A`
`Zodiac module address: 0xa838Bd5B1e66fE90fbCb5264D3fd88f2E4e245bB`
4. Select the ERC20. Then switch safes to a safe without that token. The token selector should still be preselected with that safe's native token. 

Basically token should be preselected no matter what you try. If you add / remove a tab, token state should be preserved.

**Changes** 🏗

* Use address in token data as token address and set default to the first token in the safeBalances array
* Remove token address from form state

Resolves #3970

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/AmountBalances/AmountBalances.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/AmountBalances/AmountBalances.tsx
@@ -56,10 +56,6 @@ const AmountBalances = ({
   handleChange,
   handleValidation,
 }: Props) => {
-  const [, { value: tokenAddress }] = useField(
-    `transactions.${transactionFormIndex}.tokenAddress`,
-  );
-
   const [
     ,
     { value: selectedTokenData },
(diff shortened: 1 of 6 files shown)


ArmandoGraterol (approved):
Excellent work :ok_hand: Working as expected

jakubcolony (approved):
All working well, good job! 👍

https://user-images.githubusercontent.com/112586815/196546303-a385c5ae-3180-4f43-8c78-dbc0311fc131.mov

PR #3982, opened 2022-10-12
Move Transfer Funds validation to validation schema

willm30:
## Description

- [] TODO write description

**New stuff** ✨

* New stuff goes here

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


src/modules/dashboard/components/Dialogs/ControlSafeDialog/validation.ts
```diff
@@ -52,14 +57,23 @@ const MSG = defineMessages({
     id: 'dashboard.ControlSafeDialog.validation.invalidAtIndexError',
     defaultMessage: 'Item {idx} is not a valid {type}',
   },
+  insuffienctFundsError: {
+    id: `dashboard.ControlSafeDialog.validation.insuffienctFundsError`,
```
ArmandoGraterol, line 61: Typos here

ArmandoGraterol (commented):
It's working as expected! Something that we should also do is validate when we click the `max` button as right now when I click it after having a wrong amount the error persists:
FireShot Capture 727 - Actions - Colony - wonker - localhost

willm30:
@ArmandoGraterol Good spot, have implemented.

ArmandoGraterol (approved):
Nice work mate! Good to go

jakubcolony (approved):
All working now, nice fix!

I'm just wondering if there should be any additional validation in this case:
image
Both transactions transfer 100% of the balance so clearly one of them must fail. Shouldn't we check that the total from all transactions doesn't exceed the balance?

willm30:
@jakubcolony I think it's a good idea but would double check with @ArmandoGraterol as it might be an example of scope creep that we might wish to avoid.

ArmandoGraterol:
> @jakubcolony I think it's a good idea but would double check with @ArmandoGraterol as it might be an example of scope creep that we might wish to avoid. Maybe create an issue for it in the meantime?

Yeah, this looks like an improvement to me, and when it comes to priority a "nice to have" at most.

PR #3983, opened 2022-10-12
Fix `ProgressBar` threshold in `DefaultMotion`

### LINKED ISSUE #3633: Fix threshold value for `ProgressBar`
## Description 

When voting has been initiated, see the voting progress bar is not displaying any threshold value.

## Expected behavior
Screen Shot 2022-07-14 at 11 55 55 PM

## Actual behavior
Screen Shot 2022-07-14 at 11 54 57 PM

## Steps to reproduce

1. Toggle mining on: http://0.0.0.0:3001/reputation/monitor/toggle
2. Create 1st user and a colony
3. Mint tokens
4. Payout 10 tokens to 1st user
5. Log out
6. Log in, create 2nd user and join the colony
7. Log out
8. Log back in with the 1st user
9. Pay the 10 tokens to the same user
10. Initialize voting reputation
11. Enable voting reputation
12. Toggle mining off: http://0.0.0.0:3001/reputation/monitor/toggle
13. Refresh app
14. By this point, you'll have a colony, with voting reputation enabled, and 2 users, each having 50% of the reputation in said colony

If you need to pass time manually, just use this from the command line:

```
curl -X POST --data '{"jsonrpc":"2.0","method":"evm_increaseTime","params":[3600],"id":1}' localhost:8545

```

This relates to issue #3572

chinins:
## Description

This PR updates threshold calculations for the progress bar threshold

Screenshot 2022-10-12 at 19 13 14

resolves #3633

### DIFF
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.css
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.css
@@ -24,6 +24,10 @@
   width: 116px;
 }
 
+.progressBarContainer > div > div {
+  top: -21.5px;
+}
+
 .voteResultsWrapper {
   margin: 14px 0;
   padding: 17px 23px;
--- a/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
+++ b/src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
@@ -222,14 +222,16 @@ const DefaultMotion = ({
   const skillRepValue = bigNumberify(
     votingStateData?.votingState?.skillRep || 0,
   ).div(bigNumberify(10).pow(18));
+
   const currentReputationPercent = !totalVotedReputationValue.isZero()
     ? Math.round(
         totalVotedReputationValue.div(skillRepValue).mul(100).toNumber(),
       )
     : 0;
   const thresholdPercent = !skillRepValue.isZero()
-    ? Math.round(threshold.div(skillRepValue).mul(100).toNumber())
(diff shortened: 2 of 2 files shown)


Julianahlk (commented):
When I test this it seems the thresholdPercent is always 70% is this intended? skillRepValue is always 115 and threshold is always 81 as well. Is there a way to test so that the threshold percent does not end up as 70%?

chinins:
> When I test this it seems the thresholdPercent is always 70% is this intended? skillRepValue is always 115 and threshold is always 81 as well.What's the best way to test so that the threshold percent does not end up as 70%?

It should be always the same once you installed and enabled the voting extension - it comes from the setting and then the threshold is calculated on the contracts side based on your skillrep value. If you change reputation in the colony, then the threshold value will change to fit 70%. You can change the percentage in the settings when enabling voting extension (the default value is 70).

Screenshot 2022-10-13 at 08 39 01

I've changed my setting to 60 and now it's 60 although the reputation didn't change and I didn't do anything else in the colony
Screenshot 2022-10-13 at 08 40 04

Hope this clarifies it.

Julianahlk (approved):
> It should be always the same once you installed and enabled the voting extension - it comes from the setting and then the threshold is calculated on the contracts side based on your skillrep value. If you change reputation in the colony, then the threshold value will change to fit 70%. You can change the percentage in the settings when enabling voting extension (the default value is 70).

> I've changed my setting to 60 and now it's 60 although the reputation didn't change and I didn't do anything else in the colony 
Screenshot 2022-10-13 at 08 40 04

> Hope this clarifies it.

Oh I didn't know that! Thank you for explaining. It makes much more sense now. 👍

danbr (approved):
LGTM....

PR #3984, opened 2022-10-12
Remove unused Gnosis Safe libraries and env variables

ArmandoGraterol:
As the title says, we are removing libraries that were added when using the safe core SDK to interact with safes and env variables that correspond to networks that don't have AMBs available.

If you guys can think of any other thing that could be removed that is not used related to the feature, this is a opportunity to point them out :)


chinins (approved):
Very good to clean things up!

willm30 (approved):
LGTM

PR #3995, opened 2022-10-13
Set scroll position to bottom when going back from preview

### LINKED ISSUE #3942: Scroll position to bottom when going back from preview
### Story

As a user, when I click "back" from the transactions preview screen in the SafeControl form, I want to add a new transaction or change something in the transaction without having to scroll down the form again. Therefore, it'd be more convenient if the form is scrolled to the bottom when I click back, rather than starting at the top.

(I think it's more likely a user would want to "add new transaction" after clicking back than changing the safe. Also, given that the user would have been scrolled at the bottom when they navigated to the preview screen, we're effectively remembering their scroll position). 

## Current Behaviour
Scrollbar is always at the top when clicking back

scrolled-top

### Description 

After clicking "back" from the contract preview screen, ensure scroll is set to bottom (which is where the user would have left it when clicking through to the preview). 

- [x] The form is scrolled to the bottom automatically when clicking back from the SafeControl transaction preview screen

Julianahlk:
## Description

This PR adds a small fix so when you click the "back" button from the transaction preview screen and return to the ControlSafe form the scroll position is set to the bottom instead of the top.

scrollbottom

**To Test:**

* Create a few transactions 
* Press Create transaction 
* Once you are in the preview hit  the "back" button 
* Your view should be fixed at the bottom of the screen

Resolves #3942

### DIFF
--- a/src/modules/core/components/Dialog/DialogSection.tsx
+++ b/src/modules/core/components/Dialog/DialogSection.tsx
@@ -1,4 +1,4 @@
-import React, { ReactNode } from 'react';
+import React, { ReactNode, RefObject, forwardRef } from 'react';
 
 import { getMainClasses } from '~utils/css';
 
@@ -18,8 +18,12 @@ interface Props {
   children?: ReactNode;
 }
 
-const DialogSection = ({ appearance, children }: Props) => (
-  <section className={getMainClasses(appearance, styles)}>{children}</section>
+const DialogSection = forwardRef(
+  ({ appearance, children }: Props, ref: RefObject<any>) => (
+    <section ref={ref} className={getMainClasses(appearance, styles)}>
+      {children}
+    </section>
+  ),
 );
 
 export default DialogSection;
--- a/src/modules/core/components/Dialog/DialogSection.tsx
+++ b/src/modules/core/components/Dialog/DialogSection.tsx
@@ -18,12 +18,13 @@ interface Props {
   children?: ReactNode;
 }
 
-const DialogSection = forwardRef(
-  ({ appearance, children }: Props, ref: RefObject<any>) => (
-    <section ref={ref} className={getMainClasses(appearance, styles)}>
-      {children}
-    </section>
-  ),
+const DialogSection = (
+  { appearance, children }: Props,
+  ref: RefObject<any>,
+) => (
+  <section ref={ref} className={getMainClasses(appearance, styles)}>
+    {children}
+  </section>
 );
 
-export default DialogSection;
+export default forwardRef(DialogSection);
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
@@ -157,7 +157,7 @@ const ControlSafeForm = ({
   setFieldTouched,
   handleSelectedContractMethods,
 }: FormProps & FormikProps<FormValues>) => {
-  const ref = useRef<HTMLSpanElement>(null);
+  const ref = useRef<HTMLElement>(null);
   const [transactionTabStatus, setTransactionTabStatus] = useState([true]);
   const [prevSafeAddress, setPrevSafeAddress] = useState<string>('');
   const [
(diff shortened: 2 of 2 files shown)


willm30 (changes requested):
Looks to be working in most cases! However, not for Transfer NFT. I'm still seeing the behaviour of the gif I posted in the issue. When it's single tx or when Transfer NFT is the last of multiple.

Julianahlk:
> Looks to be working in most cases! However, not for Transfer NFT. I'm still seeing the behaviour of the gif I posted in the issue. When it's single tx or when Transfer NFT is the last of multiple.

Hey @willm30, After spending a few hours yesterday and today trying to find a solution for this issue and based on your suggestions, I found out that whenever a `SpinnerLoader` is rendered, the form is scrolled to display the loader which overrides calling `scrollIntoView`.  I am not sure how we could prevent the spinner loaders from doing this and I don’t know if we would want to prevent the users from seeing the loaders?

One possible solution, if there are multiple transactions, is to collapse the multiple transactions so the spinners are not displayed and therefore, making the scrollIntoView work as expected. However, this won’t work for the situation of a single transaction as it is not collapsable. 

I don’t think the setTimeout(ref.current.scrollIntoView.bind(ref.current), 0) is a viable solution because the setTimeOut has to be set to last longer than the loader in the form. 

Let me know your thoughts.

willm30:
@Julianahlk When moving back to the NFT preview, there should be no spinner loader displayed. This has been fixed https://github.com/JoinColony/colonyDapp/pull/3998. Would you mind rebasing on this branch and seeing if that solves your problem? 

Great work by the way figuring that out. I'll have to quiz you in detail about your debugging strategy!

Julianahlk:
> @Julianahlk When moving back to the NFT preview, there should be no spinner loader displayed. This has been fixed #3998. Would you mind rebasing on this branch and seeing if that solves your problem?

Yes! I rebased off your branch in #3998 and the issue seems to haven been fixed now. Can you please test it again when you have a chance? Thanks again for your help and suggestions.

previewbottom

willm30:
@Julianahlk I will look at not showing the loader when clicking back in the contract interaction section. It should be fine but it's a bit tricky since we're using a bunch of effects and I need to be careful I don't break anything else.

willm30:
@Julianahlk Was easier than I expected. Can you rebase off this pr https://github.com/JoinColony/colonyDapp/pull/4030 and confirm it now works for contract interaction?

willm30:
Note that the rinkeby safe transaction endpoint has been deprecated, which means the transfer funds and transfer nft sections will not work with a rinkeby safe. I think it's fine to not worry about these for now, given they were working previously. Should only need to confirm the contract interaction and raw tx sections

willm30:
Alternatively, you could comment out the code that prevents you from moving to-from the preview just for testing, e.g. error components and the button's disabled logic.

Julianahlk:
> @Julianahlk Was easier than I expected. Can you rebase off this pr #4030 and confirm it now works for contract interaction?

After I rebased I can confirm all transactions are working now, specifically the contract interaction and raw transactions. 
contractinteractionandraw
cnotractandraw2

willm30 (approved):
Working for me! Good job.

chinins (approved):
Looking good. Have you tried adding ref to DialogSection instead of span? It seems that it might work without extra span element...

Julianahlk:
> Looking good. Have you tried adding ref to DialogSection instead of span? It seems that it might work without extra span element...

@chinins I added the ref to the DialogSection. Can you please check if this is what you were thinking of? Thanks 😉.

Julianahlk:
> > Looking good. Have you tried adding ref to DialogSection instead of span? It seems that it might work without extra span element...
> 
> @chinins I added the ref to the DialogSection. Can you please check before I merge if this is what you were thinking of ? Thanks 😉.

chinins:
> @chinins I added the ref to the DialogSection. Can you please check if this is what you were thinking of? Thanks 😉.

I think that this is a better solution but I'm also open to what other people think.

Julianahlk:
> > @chinins I added the ref to the DialogSection. Can you please check if this is what you were thinking of? Thanks 😉.
> 
> I think that this is a better solution but I'm also open to what other people think.

Any thoughts @ArmandoGraterol, @willm30? :)

ArmandoGraterol (approved):
Looks good to me :) 

The only thing I would change is to wrap the export of `DialogSection` with the `forwardRef` instead of the component assignment so that it looks less convoluted but that's just my opinion

ArmandoGraterol:
> > wrap the export of `DialogSection` with the `forwardRef`
> 
> @ArmandoGraterol , I made the change. Let me know if that's what you had in mind. wink Thanks.

That's it! Thank you Juliana

PR #3979, opened 2022-10-12
Refactor transfer funds loading 

willm30:
**Note, only the first commit belongs uniquely to this PR. The others are from a merge of another PR**

## Description

This PR refactors the loading pattern in the transfer funds section. 

First, it ensures there's an appropriate error message if the safe isn't selected. 
Second, it implements a cache, so that we only fetch for token data once per safe.  
Third, it abstracts the `loading` and `error` message elements so that they can be used in other sections ([#3998](https://github.com/JoinColony/colonyDapp/pull/3993/commits/950bf41b69bce90802f3502f23bcd4c6f0c37423)).

**Changes** 🏗

* Safe not selected message
* Remember previously fetched tokens
* Abstracted loading and error messages

Resolves #3969

PR #3998, opened 2022-10-14
Improve Transfer NFT loading pattern and tidy other sections

willm30:
**Note: only the first commit belongs uniquely to this PR. Others from merge of other pr.**
## Description

This PR refactors the Transfer NFT's loading flow  (inspired by doing the same at #3979). It also implements a few other tweaks to make the transaction sections more consistent, such as updating message names and reusing common components.
 
**Changes** 🏗

* Extracting common code into `shared.ts` file (1) and refactoring Transfer NFT loading flow (1)

## To test

1. Just double check that in the case of Transfer NFT, the loading spinner only shows up the first time we fetch the NFTs for the safe and that there's no ui flashing when, e.g. clicking "back" from the Transaction Preview.


jakubcolony (approved):
Nice improvements @willm30!

Screenshot 2022-10-19 at 18 10 18
Screenshot 2022-10-19 at 18 40 56

>Just double check that in the case of Transfer NFT, the loading spinner only shows up the first time we fetch the NFTs for the safe and that there's no ui flashing when, e.g. clicking "back" from the Transaction Preview.

That's how it works for me

One small thing I noticed is that the avatar in the recipient fields in both Transfer NFT and Raw transaction got smaller (22px vs 26px previously) and got an outline on hover:
image

willm30:
> One small thing I noticed is that the avatar in the recipient fields in both Transfer NFT and Raw transaction got smaller (22px vs 26px previously) and got an outline on hover: image

@jakubcolony Previously it wasn't using the same avatar as the other form sections, now they're all consistent. Although, it is a good spot since we usually reserve that style for user icons with a popover... so maybe, actually, the others were incorrect. @ArmandoGraterol Do you think it's fine or should we change it to the avatar that doesn't have an outline on hover?

ArmandoGraterol:
@willm30 
> @ArmandoGraterol Do you think it's fine or should we change it to the avatar that doesn't have an outline on hover?

I agree that the outline should only apply to the user icons when they are interactable like in the member sections, otherwise it seems like you can click on the avatar but nothing happens actually

willm30:
@ArmandoGraterol Have swapped them over

jakubcolony (approved):
The user picker avatar looks good now @willm30! 👍

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/RawTransactionSection.tsx
```diff
@@ -2,40 +2,36 @@ import React, { useEffect } from 'react';
...
     defaultMessage: 'Select or paste a wallet address',
   },
 });
 
-const displayName = `dashboard.ControlSafeDialog.ControlSafeForm.RawTransactionSection`;
-
-const renderAvatar = (address: Address, item: AnyUser) => (
-  <UserAvatar address={address} user={item} size="xs" notSet={false} />
-);
+const displayName = `dashboard.ControlSafeDialog.TransactionTypesSection.RawTransactionSection`;
```
ArmandoGraterol, line 34: Same with these ids, they are correct without the `TransactionTypesSection`

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferFundsSection.tsx
```diff
@@ -16,82 +13,48 @@ import {
...
       'Unable to fetch Safe balances. Please check your connection',
   },
-  noSafeSelectedError: {
-    id: `dashboard.ControlSafeDialog.ControlSafeForm.TransferFundsSection.noSafeSelectedError`,
-    defaultMessage: 'You must first select a safe',
-  },
 });
 
-const displayName = `dashboard.ControlSafeDialog.ControlSafeForm.TransferFundsSection`;
+const displayName = `dashboard.ControlSafeDialog.TransactionTypesSection.TransferFundsSection`;
```
ArmandoGraterol, line 48: Same here

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferNFTSection.tsx
```diff
@@ -16,57 +16,58 @@ import {
...
   nftError: {
-    id: `dashboard.ControlSafeDialog.TransferNFTSection.nftError`,
-    defaultMessage: 'Unable to fetch NFTs. Please check your connection.',
+    id: `dashboard.ControlSafeDialog.TransactionTypesSection.TransferNFTSection.nftError`,
+    defaultMessage: 'Unable to fetch NFTs. Please check your connection',
   },
 });
 
-const displayName = 'dashboard.ControlSafeDialog.TransferNFTSection';
+const displayName = `dashboard.ControlSafeDialog.TransactionTypesSection.TransferNFTSection`;
```
ArmandoGraterol, line 70: Same here

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/shared.tsx
```diff
@@ -0,0 +1,48 @@
...
+import Avatar from '~core/Avatar';
+
+import { DialogSection } from '~core/Dialog';
+import { SpinnerLoader } from '~core/Preloaders';
+import { AnyUser } from '~data/index';
+import { Address, Message } from '~types/index';
+
+import styles from './TransactionTypesSection.css';
+
+export const AvatarXs = (address: Address, item: AnyUser) => (
```
ArmandoGraterol, line 12: Just my opinion and a nitpick but I feel like the name of this component could be understood better if both letters were capitalized `XS`. For me, when talking about sizes, whether in this context or in the context of clothing, for example, it's more common for both letters to be lower or upper case.

ArmandoGraterol: Tbh I think this reusable component is not really necessary, but I don't feel strongly about not including it. My only concern is that too many components like this one can make the dev experience a bit messy at times

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/shared.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
+import { FormattedMessage } from 'react-intl';
+import Avatar from '~core/Avatar';
```
ArmandoGraterol, line 3: This import should be grouped with other similar imports

ArmandoGraterol (commented):
@willm30 Do you have any safe that I can use to test the section? One that has NFT and the module enabled

willm30:
> @willm30 Do you have any safe that I can use to test the section? One that has NFT and the module enabled

@ArmandoGraterol Now that the rinkeby tx service has been deprecated, no. I've messaged you re resolving this.

ArmandoGraterol (approved):
Looks good to me :ok_hand: NIce work

Julianahlk (approved):
LGTM. Good work Will!

PR #4003, opened 2022-10-17
Update invalidSafeError in Safe Control

### LINKED ISSUE #4002: Add couple of additional error  messages in Safe Control
Currently, we don't have a specific message for when a user enters a safe manually into the safe picker instead of selecting a safe from the dropdown menu. 

We also don't have a message for when an ABI fetched in the contract interaction section returns no useful methods. The f unction selector just looks like it isn't working.

willm30:
## Description

This PR adds a couple of extra error messages to Safe Control and fixes a bug in the Contract Interaction section. 

**Changes** 🏗

1. Replace "noSafeSelected" error with "invalidSafeError", which covers the previous case in addition to the case of a user entering a safe address directly into the safe picker. 
2. Fixes a bug in the Contract Interaction section where it wasn't refetching if there was an error displayed
3. Adds a specific message if there are no useful methods on the abi in the contract interaction section

## To test

Open a form section with either no safe selected or something manually entered into the safe picker (valid address or not). Should receive the corresponding error. Make sure it then duly disappears when selecting a safe correctly.

If you fetch a contract such as `0x05b990D01A2f86852feD1E7d609dd3E52aA9EA27` (rinkeby), which has no useful methods, you should see the corresponding message instead of a function selector with no options.

## Screenshots

Invalid Address:

invalid

No external methods:

noexternals

Resolves #4002

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
@@ -67,8 +67,12 @@ const MSG = defineMessages({
     defaultMessage: `Unable to fetch contract. Please check your connection`,
   },
   noUsefulMethodsError: {
-    id: `dashboard.ControlSafeDialog.ControlSafeForm.TransactionTypesSection.ContractInteractionSection.noUsefulMethodsError`,
+    id: `dashboard.ControlSafeDialog.TransactionTypesSection.ContractInteractionSection.noUsefulMethodsError`,
     defaultMessage: `No external methods were found in this ABI`,
+  },
+  unknownContract: {
+    id: `dashboard.ControlSafeDialog.TransactionTypesSection.ContractInteractionSection.unknownContract`,
+    defaultMessage: `Unknown contract`,
   },
 });
 
@@ -182,7 +186,7 @@ const ContractInteractionSection = ({
 
         setFieldValue(
           `transactions.${transactionFormIndex}.contract.profile.displayName`,
-          contractName || 'Unknown contract',
+          contractName || formatMessage(MSG.unknownContract),
         );
         setIsLoadingABI(false);
       } else {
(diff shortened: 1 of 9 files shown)


src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/ContractInteractionSection.tsx
```diff
@@ -157,57 +157,69 @@ const ContractInteractionSection = ({
...
+
   useEffect(() => {
     if (
       transactionValues.contract &&
-      selectedSafe &&
-      // only run effect if safe chain changes
-      prevSafeChainId !== selectedSafe.chainId
+      safe &&
+      // only run effect if safe chain changes or there was previously an error
+      (prevSafeChainId !== selectedVerifiedSafe?.chainId || ABIError)
```
willm30, line 203: This fixes the bug where it wouldn't refetch if you had an invalid safe error and then updated the safe input

src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/TransferFundsSection.tsx
```diff
@@ -87,23 +87,28 @@ const TransferFundsSection = ({
     setBalanceError('');
     if (safe) {
       setIsLoadingBalances(true);
+      let baseUrl: string;
       try {
         const chainName = getChainNameFromSafe(safe.profile.displayName);
```
willm30, line 92: Move `getChainNameFromSafe` to own try catch in case the user inputs manually into the safe picker, which will throw an error we can catch and set the error to `invalidSafeError`

jakubcolony (approved):
Nice work @willm30

The only thing I think you could improve is the alignment of the noUsefulMethodsError so that it sits right in the middle between the ABI textarea and Add another transaction button:
Screenshot 2022-10-20 at 19 09 40

Otherwise, everything looks good!
Screenshot 2022-10-20 at 18 57 46

Screenshot 2022-10-20 at 19 06 04

Julianahlk (approved):
The new error messages are looking great Will! I do agree with Jakub the below error message can be adjusted to match the alignment of other error messages. 
Screen Shot 2022-10-20 at 4 51 47 PM

PR #4009, opened 2022-10-18
Fix const variable name

joannasia:
## Description

- fixed name of const to FIX_TRIGGER_EVENT_NAME

(Resolves | Contributes to) #


joannasia:
@jakubcolony @chinins i have missed one comment from previous PR #3945 - so here is the change for that comment

jakubcolony (approved):
LGTM 👍

chinins (approved):
Very good, thank you for making sure it's updated!

Issue #3850, opened 2022-09-13
Expenditures: Update or add a tooltip to all force action toggles


Julianahlk:
## Specification

Ensure the implementation of adding a tooltip to all force action toggles matches the way it was done in this **[PR](https://github.com/JoinColony/colonyDapp/pull/3775)**  to maintain consistency across the DApp.

Figma Link - https://www.figma.com/file/CCgGsBzhX3BUENS1IvgP7N/Motions-%26-Disputes?node-id=3511%3A0

Resolves #3687

jakubcolony:
This issue can only be worked on once Juliana's changes have been merged/rebased with expenditures branch, moving back to "To Do"

PR #4012, opened 2022-10-18
Integrate motion components with Control Safe dialog

### LINKED ISSUE #4005: Add motion domain selector and force toggle to Control Safe dialog
### Story

As a user creating safe transactions, I want to be able to change the domain of the motion and, if I have permission, to force or not the action.

### Description

When the motion & disputes extension is enabled, the control safe dialog should display the motion domain selector and force toggle that we show in any other action dialog, they should be displayed in the preview dialog. This issue is just to integrate those components and make the values available for the motion saga (AKA: Values are available on submit).

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=724%3A15842)

FireShot Capture 731 - Safe Control – Figma - www figma com

- [x] The force toggle is integrated into the preview transactions dialog.
- [x] The motion domain selector is integrated into the preview transactions dialog.

ArmandoGraterol:
Added force action toggle and motion domain selector.

The motion domain selector is disabled because you are not supposed to create the motion in any other domain that is not the main one.

FireShot Capture 732 - Actions - Colony - wonker - localhost
FireShot Capture 733 - Actions - Colony - wonker - localhost

Resolves #4005

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
@@ -70,6 +70,10 @@ const MSG = defineMessages({
     id: `dashboard.ControlSafeDialog.ControlSafeForm.buttonCreateTransaction`,
     defaultMessage: 'Create transaction',
   },
+  buttonCreateMotion: {
+    id: `dashboard.ControlSafeDialog.ControlSafeForm.buttonCreateMotion`,
+    defaultMessage: 'Create motion',
+  },
   buttonConfirm: {
     id: `dashboard.ControlSafeDialog.ControlSafeForm.buttonConfirm`,
     defaultMessage: 'Confirm',
@@ -285,6 +289,17 @@ const ControlSafeForm = ({
   );
 
   const savedNFTState = useState({});
+
+  const submitButtonText = (() => {
+    if (!showPreview) {
+      if (isVotingExtensionEnabled && !values.forceAction) {
+        return MSG.buttonCreateMotion;
+      }
+      return MSG.buttonCreateTransaction;
+    }
+    return MSG.buttonConfirm;
+  })();
+
   return (
     <>
       {!showPreview ? (
@@ -492,7 +507,7 @@ const ControlSafeForm = ({
           onClick={() =>
             showPreview ? handleSubmit() : handleShowPreview(!showPreview)
           }
-          text={showPreview ? MSG.buttonConfirm : MSG.buttonCreateTransaction}
+          text={submitButtonText}
           loading={isSubmitting}
           disabled={
             !isValid ||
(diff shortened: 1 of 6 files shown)


willm30 (changes requested):
~~I'm noticing the form's valid even if force is toggled off.~~ Sorry, I got confused. The colony had reputation but the user had none so I was expecting the form to be disabled, but looking at other forms this isn't an issue. 

However, should we change the button text to "Create Motion" when it's not being forced? 


Also, do we want to add issues for adding the motion components to add / remove safe forms?

src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
```diff
@@ -498,14 +518,15 @@ const ControlSafeForm = ({
           onClick={() =>
             showPreview ? handleSubmit() : handleShowPreview(!showPreview)
           }
-          text={showPreview ? MSG.buttonConfirm : MSG.buttonCreateTransaction}
```
willm30, line 501: We could use the existing pattern `{id: button.confirm}` and `{id: button.createMotion}` here?

willm30 (commented):
It's showing up for me in the step before instead of on the preview screen:

createMotion

ArmandoGraterol:
@willm30 
> It's showing up for me in the step before instead of on the preview screen:

 Yeah, it's the step before the one that has the "Create transaction" text (which would correspond to a change to "Create motion" when the extension is installed), and the step after its the one with "Confirm". Does that make sense to you?

willm30:
@ArmandoGraterol Shouldn't it be "Create motion" instead of "confirm"? And then when you toggle force it goes back to "confirm"? There's no toggle on the non-preview page so "Create motion" looks a little out of place, at least comparing with other forms, in my opinion.

ArmandoGraterol:
@willm30 
> @ArmandoGraterol Shouldn't it be "Create motion" instead of "confirm"? And then when you toggle force it goes back to "confirm"?

So the way I see it is like this:

The "Confirm" button text still makes sense in the context of having the "Motion & Disputes" extension enabled, you are confirming that you want a motion created for the transactions, therefore it doesn't really need to be replaced. Now if we compare that with "Create transaction", in the context of having the extension enabled, you are actually creating a motion that is going to execute the transactions you added.

> There's no toggle on the non-preview page so "Create motion" looks a little out of place, at least comparing with other forms, in my opinion.

But then again, no other form of us has a preview dialog :P 

I feel like we could argue that the transition from "Create transaction" > "Create motion" also looks awkward in comparison with "Create transaction" > "Confirm" or "Create motion" > "Confirm", or that the place of the toggle and selector is weird because they are not in the dialog with the actual transaction inputs, like our other actions but again no other action has a preview dialog.

Tbh, if it's still confusing I'm good with leaving the texts as they were before as they still make sense ("Create transaction" > "Confirm") and if there's any issue with it surely anyone with Product will point out a better solution (Move the motion components to the first dialog and then change it to "Create motion" perhaps?)

willm30 (approved):
@ArmandoGraterol Thanks for  clarifying. I agree with you that one set is not obviously more / less correct than another. I would have no issues with your suggestion of reverting to how it was before I raised the initial query 😄

jakubcolony (approved):
LGTM!

Screenshot 2022-10-20 at 22 07 18

Screenshot 2022-10-20 at 22 07 06

Should I create an issue similar to #3850 for safe control?

PR #4016, opened 2022-10-19
Fix: Input max button buggy behaviour when clicked twice

### LINKED ISSUE #4011: The Max button in Transfer Funds has an odd behaviour when clicked twice
## Steps to reproduce

It doesn't seem to happen for all safes and tokens but I've already seen it happen a few times.

1. Add a safe, e.g.:
```
Safe address: 0x358E6D7C8Fa5e55fc6A0D50E23b875c6c49fF09A
Zodiac module address: 0xa838Bd5B1e66fE90fbCb5264D3fd88f2E4e245bB
```
2. Select Transfer Funds and choose a token (for the safe above, select `WILL`).
3. Click on Max twice

## Expected behaviour

The amount is set to and stays at the maximum allowable

## Actual behaviour

The amount is set to the maximum after the first click, then it changes to some other value:

https://user-images.githubusercontent.com/112586815/196544353-44d0a360-e91c-4eb7-8978-8012b0bd3c62.mov

https://user-images.githubusercontent.com/112586815/196544362-5b3f96fb-fdbf-4603-9017-77f2a9331eda.mov

jakubcolony:
## Description

This PR fixes the issue described in #4011. By the time I got to the bottom of it, it became clear the bug has been present in the dapp for at least a year and its impact is likely to be minimal, since there are two conditions for it to happen:
- The max allowable amount must be less than 0.000001
- The Max button must be clicked at least twice

Anyway, the fix is just one line.

To test, add any safe and create a transaction to transfer a token of which there's less than 0.000001 in the safe (you can use the safe from the issue and select `WILL` token). You should be able to click Max as many times as you wish without the value changing to something odd.

I decided to branch this off `feature/gnosis-safe-control` (although it potentially affects all other inputs with max button) due to the bug's impact being minimal and easiest to notice inside control safe dialog. If there are any objections I'll branch it off master instead. 

**Changes** 🏗

* Removed formatting of tiny numbers in `InputComponent`

Resolves #4011

### DIFF
--- a/src/modules/core/components/Fields/Input/InputComponent.tsx
+++ b/src/modules/core/components/Fields/Input/InputComponent.tsx
@@ -150,9 +150,7 @@ const InputComponent = ({
             maxButtonParams?.customOnClickFn?.();
             const decimalValue = new Decimal(maxButtonParams.maxAmount);
             if (decimalValue.lt(0.00001) && decimalValue.gt(0)) {
-              cleave?.setRawValue(
-                decimalValue.toSD(5, Decimal.ROUND_DOWN).toNumber(),
-              );
+              cleave?.setRawValue(maxButtonParams.maxAmount);
             } else {
               cleave?.setRawValue(
                 new Decimal(maxButtonParams.maxAmount)


Julianahlk (approved):
Nice fix! Working well 👍
Oct-20-2022 16-16-54

ArmandoGraterol (approved):
LGTM my friend :+1:

PR #4017, opened 2022-10-19
Fix: Many colonies overlap Add New button

### LINKED ISSUE #4007: Many colonies overlap Add New button
## Originally posted by @willm30 in https://github.com/JoinColony/colonyDapp/pull/3976#pullrequestreview-1143928767 

I'm noticing, however, an overlap of the colonies list and the add colony button when many colonies are present. Could that be addressed here or should I open a separate issue? 

many-cols

>@willm30 , open a new issue, please. I think that product should confirm the UI that we should have in this case.

jakubcolony:
## Description

I was unable to reproduce the bug myself (it might as well be a Chrome/Windows specific issue) but looking at the code I found some CSS that I believe could possibly be the offender here. 

In Chrome 106 on Mac OS, the Add New button wasn't overlapped before or after:
Screenshot 2022-10-19 at 23 05 17

@willm30 Please let me know if this fix works for you. I'll request review from the rest of the team once you confirm that.

**Deletions** ⚰️

* Deleted CSS setting the Add New button position to sticky

Resolves #4007

### DIFF
--- a/src/modules/dashboard/components/SubscribedColoniesList/SubscribedColoniesList.css
+++ b/src/modules/dashboard/components/SubscribedColoniesList/SubscribedColoniesList.css
@@ -50,11 +50,6 @@
   fill: var(--temp-grey-blue-7);
 }
 
-.newColonyItem {
-  position: sticky;
-  bottom: 0;
-}
-
 .loadingColonies {
   padding: 8px 0 2px;
 }
@@ -62,15 +57,6 @@
 @media screen and query700 {
   .main {
     align-items: center;
-  }
-
-  .newColonyItem {
-    transform: scale(0.76) translateX(-8px);
-  }
-
-  .newColonyItem a {
-    height: auto;
-    width: auto;
   }
 
   .scrollableContainer {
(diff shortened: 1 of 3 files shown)


willm30 (approved):
Fixed! Good work. Arren is a fellow window's user in case you want a windows-specific second opinion. 
colonylist

src/modules/dashboard/components/SubscribedColoniesList/SubscribedColoniesList.css
```diff
@@ -122,19 +117,19 @@
     box-shadow: 0px 1px 6px var(--drop-shadow);
   }
 
-  .scrollableContainer > div[role="tooltip"] {
+  .scrollableContainer > div[role='tooltip'] {
```
chinins, line 120: Is this the auto fix of your linter? I think we should keep it consistent with how it is in the dapp now and it seems it's double quotes.

chinins (approved):
I couldn't reproduce it neither but this css works as well and it looks reasonable. Just please make sure to revert to double quotes in css files as per my specific comment.

PR #4018, opened 2022-10-19
Implement safe transaction motion page

### LINKED ISSUE #4004: Implement motion page for Safe transactions
### Story

As a user, after making a safe transaction, within a colony with the motions & disputes extension installed and not forcing the action, I have to be redirected to the corresponding motion page.

### Description

After making a transaction, or several transactions, related to safes, the user should be redirected to the motion page, where the title, description, and data previously inputted by the user will be displayed. 

**NOTE**: This issue should address ALL the transaction types and whether the user did single or multiple transactions. There's going to be a lot of shared code between the transactions, better to get them all in 1 issue to avoid code repetition and "stepping on each other foot" situations.

### Design

[Figma link](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=727%3A14616)

FireShot Capture 730 - Safe Control – Figma - www figma com

## Implementation

When using `createMotion`, the method only allows colony actions but we should be able to use the arbitrary transaction action to pass along the safe transactions that need to be executed once the motion is finalized.
 
- [ ] The user is redirected to the motion page after making any number of safe transactions
- [ ] The motion page shows the title as the title inputted by the user in the dialog
- [ ] The motion page shows the annotation as the description inputted by the user in the dialog
- [ ] The details widget shows the relevant info of the safe transaction. If multiple transactions are made, it also shows them all.

ArmandoGraterol:
## Description

This PR wires in the `DefaultMotion` page when creating a motion with one or multiple safe transactions. 

## To test
1. Set up your bridging environment (instructions copied below). 
2. Add a safe
3. Generate reputation before installing the motion extension
4. Create as many different transactions as you like, from 1 to multiple.
5. The detail widget should look like the one we already use on the action page, and the title of the transactions should be displayed on the finalize and vote widgets.
6. On the colony home, the event list should display the motions and their state (Mind you, they should only appear when they are staked up to 10% of the stake).
7. Bonus if you test the other non-safe motions, just to make sure the changes added are not affecting the other motions

## Bridging env instructions

0. have your normal dev env running. 
1. `cd src/lib/colonyNetwork` 
8. `git fetch` 
9. `git checkout maint/cross-chain-safe` 
10. `git submodule update --init` 
11. `yarn` 
12. `yarn provision:safe:contracts` 
13. `cd lib/safe-contracts` then `yarn` (ignore errors)
14. `cd ../..` (back into colonyNetwork) 
15. `bash scripts/start-bridging-environment.sh` 
16. `node scripts/start-bridging-environment.js` 
17. Copy the data in console into your .env and restart webpack (home bridge address, local safe address, local token address etc.)

```
# For testing Safe Control locally. Ensure you have a foreign chain running locally.

LOCAL_HOME_BRIDGE_ADDRESS=

LOCAL_FOREIGN_BRIDGE_ADDRESS=

LOCAL_SAFE_ADDRESS=

ZODIAC_BRIDGE_MODULE_ADDRESS=

LOCAL_ERC721_ADDRESS=

LOCAL_SAFE_TOKEN_ADDRESS=
```

FireShot Capture 736 - NFT moment - Motion - Colony - wonker - localhost
FireShot Capture 737 - Actions - Colony - wonker - localhost

Resolves #4004


src/utils/colonyActions.ts
```diff
@@ -231,8 +231,12 @@ export const getSafeTransactionActionType = (
     if ((safeTransactions || []).length >= 2) {
       return TransactionTypes.MULTIPLE_TRANSACTIONS;
     }
-    const type = safeTransactions[0].transactionType;
-    return type;
+
+    if (safeTransactions.length === 1) {
```
willm30, line 235: To guard against `safeTransactions[0]` being undefined (I can't off the top of my head think of why that would be the case; wouldn't there always be at least one tx?), could do `const type = safeTransactions[0]?.transactionType` instead?

willm30 (changes requested):
Looks like we're nearly there! 

Would you be able to rebase so I can review your changes more easily? 

We need to update the version of colonyJs from `^4.2.1-beta.3` to `^4.2.1`, else I can't access the governance extension. Makes sense to do it straight in this branch.

When doing that and creating a motion, I'm currently seeing this:

moti

src/utils/colonyActions.ts
ArmandoGraterol: I was having problems with the subgraph because the annotations were not being indexed fast enough (That's why you saw that the title in the screenshot is not formatted), but I think I've solved it so this can be reverted to how it was before

ArmandoGraterol: For the record, I tried with `?` and it still wasn't working due to the `[0]` if I remember correctly

ArmandoGraterol:
> Would you be able to rebase so I can review your changes more easily?

Done!
> We need to update the version of colonyJs from `^4.2.1-beta.3` to `^4.2.1`, else I can't access the governance extension. Makes sense to do it straight in this branch.

Done!x2

> When doing that and creating a motion, I'm currently seeing this:
> 
> moti

Fixed! This was because the subgraph query was pulling the data before it was indexed, and then on refetch instead of pulling the newly indexed data it would return the old non-indexed version

willm30 (changes requested):
I'm getting the following error: 

crash

er

Also, should we update the gas station to follow the pattern in actions ? (I.e. to use the transaction title instead of "Create Motion" and the subtitle of "Safe Transaction initiated")
gstn

src/utils/colonyActions.ts
```diff
@@ -219,12 +224,14 @@ export const getSafeTransactionActionType = (
...
-    actionType === ColonyExtendedActions.SafeTransactionInitiated &&
+    (actionType === ColonyExtendedActions.SafeTransactionInitiated ||
+      actionType === AddedMotions.SafeTransactionInitiatedMotion) &&
     safeTransactions
   ) {
     if ((safeTransactions || []).length >= 2) {
       return TransactionTypes.MULTIPLE_TRANSACTIONS;
     }
+
     const type = safeTransactions[0].transactionType;
```
willm30, line 235: Just need the `?` here to capture the undefined case. Maybe add a comment two about why it's coming through as undefined and update the typing?

willm30: Hmm, actually I take that back. I just get what appears to be a fallback and have to refresh to get the actual motion screen: 

fallback`

willm30 (commented):
I'm noticing the safe control motions are taking a little while longer than other motions / actions to appear in the actions list. 

In the gif below, I refresh the page, and you can see a payment motion at the top of the list, along with safe control actions I passed previously, all show up immediately. Then, the motions I've been putting through show up last. 

list-slow

That said, I've been playing with the motions themselves and apart from the issue I mention above, all seems in splendid order.

ArmandoGraterol:
@willm30 
> I'm getting the following error

Is your local branch up-to-date? This is not happening to me. My latest changes should prevent this from happening as this is related to what I mentioned about the subgraph not indexing fast enough and returning old data.

> Also, should we update the gas station to follow the pattern in actions ? (I.e. to use the transaction title instead of "Create Motion" and the subtitle of "Safe Transaction initiated")

I don't think so. Mostly because it's the safe transactions title, and if you are not creating and executing those transactions right away like in the action's version then it could be confusing.

> I'm noticing the safe control motions are taking a little while longer than other motions/actions to appear in the actions list.
> 
> In the gif below, I refresh the page, and you can see a payment motion at the top of the list, along with safe control actions I passed previously, all show up immediately. Then, the motions I've been putting through show up last.

Yes, motions are expected to take longer than the other actions as they are last in the queries and perhaps heavier? You can take a look at  `ColonyActions`, we first fetch payments, then the other actions, and finally motions.

That payment you see at the top when it's loading the motions is not really a payment motion. It's the payment that is made after you finalized the real motion (The payment motion that appears at the top when it loads all the motions). I think it's appearing because as I've mentioned payments are retrieved first, then when the motions are fetched the dapp filters out the initial payment action because it notices that the action was actually the result of a finalized motion and not an "independent" action.

src/utils/colonyActions.ts
ArmandoGraterol: Can you check that your branch is up-to-date? In your screenshot it seems like the subgraph didn't return the IPFS data of the transaction and that's something that it should be fixed by my latest changes

willm30:
@ArmandoGraterol  

> Is your local branch up-to-date? This is not happening to me. My latest changes should prevent this from happening as this is related to what I mentioned about the subgraph not indexing fast enough and returning old data.

Yes, I run `git pull --rebase` followed by `npm i --legacy-peer-deps` and I experience the error shown above. Just retried.

ArmandoGraterol:
@willm30 

> Yes, I run `git pull --rebase` followed by `npm i --legacy-peer-deps` and I experience the error shown above. Just retried.

I've added the optional operator so that it doesn't crash at least, but I can't still replicate it :( If it's still showing the fallbacks when you create the motion (without reloading), can you help me and check from your side what's happening? It should be something related to `getAnnotationFromSubgraph`, the function we use to get all the safe transaction data of the motion, otherwise, I can't really look into it as it's not happening to me no matter how many motions I create. (I will continue to try tho)

willm30:
@ArmandoGraterol I've taken a lil' look into it. The subgraph queries don't return the latest data when called the first time inside the resolver. Evidently we've got a race going on, and we call the queries miliseconds before the latest data becomes available. A refresh solves the issue because by that point, the data is available. 

I think there are a few ways to tackle the problem from our side. 

1) We pass the data in over the location object, given that we have it inside the saga. This would guarantee we have the latest data, independent of the race with the subgraph.

2) It seems Simple Decisions faced the same problem. They prompt the user to refresh the page if the data isn't available.

racing

3) We implement some arbitrary delay before initiating the query to the subgraph. The following has worked for me in testing, right before you call `getAnnotationFromSubgraph`: ` await new Promise((resolve) => setTimeout(resolve, 1000));`

Thoughts?

ArmandoGraterol:
@willm30 
Thank you so much for taking a look! It makes sense to me and it's a pain in the ass but I think I'm gonna go with this solution:

> 1. We pass the data in over the location object, given that we have it inside the saga. This would guarantee we have the latest data, independent of the race with the subgraph.

I remember that you implemented something like this before for another action, and I quite like that solution so I'm gonna try and do the same here.

willm30:
@ArmandoGraterol Nice one. You just need to add a state object as a param to the call to [`historyObject.push`](https://v5.reactrouter.com/web/api/history) on line 115 of `utils/saga/effects.ts`. And then you can access it in the motion page using `useLocation()`. The data won't always be available, e.g. if you access the url directly instead of being redirected from the saga, or if you arrive via the actions list in colony home, but that's not an issue since in these cases we assume the correct data will be returned from the resolver. It's just a matter of checking if the state object is defined and using it if so, or using the resolver data if not. 

I'll try and see if I can see why this is happening but in any case I think this is a reliable workaround.

willm30:
@ArmandoGraterol I was wrong in suggesting earlier that this doesn't seem to happen with other motions. I have in the past noticed that the annotation doesn't show up immediately when doing anything, action or motion, and am still noticing that (e.g. creating a payment, minting tokens... you need to refresh for the annotation to show up). It's just that we're relying on that annotation for additional data here, so it's more noticeable. It also _doesn't_ happen in the safeTransaction action (i.e the annotation does show up immediately), because there we don't need to query the subgraph to get the metadata hash. It's a param in the `annotateTransaction` call, so is available in the event data. 

So yeah, doesn't seem like there's much we can do. I think we've got a good workaround in using location state.

ArmandoGraterol:
> @ArmandoGraterol I was wrong in suggesting earlier that this doesn't seem to happen with other motions. I have in the past noticed that the annotation doesn't show up immediately when doing anything, action or motion, and am still noticing that (e.g. creating a payment, minting tokens... you need to refresh for the annotation to show up). It's just that we're relying on that annotation for additional data here, so it's more noticeable. It also _doesn't_ happen in the safeTransaction action (i.e the annotation does show up immediately), because there we don't need to query the subgraph to get the metadata hash. It's a param in the `annotateTransaction` call, so is available in the event data.
> 
> So yeah, doesn't seem like there's much we can do. I think we've got a good workaround in using location state.

Interesting insight! I didn' that about the other motions. Thank you for looking into it.

I've uploaded a possible solution for the safe transaction motion using location like we've talked about. I tried to get all of the data from the location state instead of the subgraph and it seems like it's working as expected, hopefully it works for you too!

willm30 (changes requested):
Nice work using the location state. Just need to update the widget in the bottom right also. 

generic

And perhaps add some additional space between the widget and the bottom of the page to account for open popovers like below (which is currently running off-screen): 

addpsace

src/modules/dashboard/sagas/motions/finalizeMotion.ts
```diff
@@ -26,39 +24,22 @@ function* finalizeMotion({
...
-    /*
-     * Increase the estimate by 100k WEI. This is a flat increase for all networks
-     *
-     * @NOTE This will need to be increased further for `setExpenditureState` since
-     * that requires even more gas, but since we don't use that one yet, there's
-     * no reason to account for it just yet
-     */
-    const estimate = bigNumberify(networkEstimate).add(bigNumberify(100000));
-
     const {
```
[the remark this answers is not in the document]
ArmandoGraterol: The previous solution wasn't optimal, and I think it was done as a quick workaround. 
 
The problem is that it's not a reliable way of estimating the gas to finalize the motion. For example, with the safe transactions, I couldn't finalize them because the estimated gas limit was lower than what it required.

Alex suggested that instead of estimating the action inside the finalize function and then manually adding more gas we could estimate the execution of the "finalizeMotion" transaction itself and that would give a better, more accurate gas limit.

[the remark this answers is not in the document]
willm30: Awesome, thanks

src/modules/dashboard/components/ActionsPage/ActionsComponents/DefaultMotion.tsx
```diff
@@ -143,6 +154,16 @@ const DefaultMotion = ({
...
   }, []);
+  const { formatMessage } = useIntl();
+  const { state: locationState } = useLocation<{
+    title: string;
+    transactions: SafeTransaction[];
+    annotationMessage: string | null;
+    safeData: {
+      contractAddress: string;
+      chainId: string;
+    };
```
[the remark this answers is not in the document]
ArmandoGraterol: Nice

ArmandoGraterol:
@willm30 

The widget should now display info correctly!

> And perhaps add some additional space between the widget and the bottom of the page to account for open popovers like below (which is currently running off-screen):

Unfortunately, it seems like any solution short of taking a deep dive into the popover is not gonna be good enough. We can increase the space between the widget and the end of the page but the popover doesn't really have a fixed height so a user with different tokens and reputations in several domains will still overflow, we need a dynamic solution. 

The ideal case would be that the initial styles of the popover take into that the popover is going to overflow and readjust the position. This dynamic repositioning is something that the library we are using for the popover _should_ allow already, however, I tried several things and it's not working, at least not in a straightforward way. Therefore I would suggest that we leave this be as there's no simple and good solution currently it seems, most likely we will revisit this in the CDapp thought.

willm30 (approved):
Nice work, all looking good from my testing 👍

Julianahlk (approved):
Looking good from my end as well, just one comment and a question:

1. Even though I closed each DetailsWidget item when I refreshed the page they come back as collapsed, is this intended behavior? Should they not stay closed?
multiple refresh

2. Every time after I click on the claim button I get the Error below, is it normal?
Screen Shot 2022-11-08 at 6 13 32 PM

ArmandoGraterol:
@Julianahlk Thank you for your feedback!

> 1. Even though I closed each DetailsWidget item when I refreshed the page they come back as collapsed, is this intended behavior? Should they not stay closed?

Yes, they are not expected to stay closed. The toggle state is stored inside an `useState`, for them to stay open we would have to store the state in the local storage or something like that.

> 2. Every time after I click on the claim button I get the Error below, is it normal?

 I don't think it's normal but also I think this error _may_ be on `master` as I think I might have seen it before through the time of using the Dapp (I don't think there's an issue for it tho), and also, we haven't touched the claim logic in this feature :)

PR #4019, opened 2022-10-20
Fix mobile menu not opening in production

willm30:
## Description

This PR fixes a bug in the mobile responsive menu where the popover wasn't displaying anything if the user was on the wrong network or logged out.

**Changes** 🏗

* Move "basic menu" out of logged-in control flow so it's always visible, no matter what

User will now see this: 

fixedmenu

Instead of this:

brokenmenu


rdig (approved):
Seems to be working now :+1:

jakubcolony (approved):
LGTM 👍

PR #4021, opened 2022-10-21
Fix selectedContractMethods when deleting a tab

### LINKED ISSUE #3944: Removing tab in SafeControl hides Contract Interaction's selected function inputs
## Steps to reproduce

1. Add two transactions in SafeControl, with the second one being a Contract Interaction.
2. Select a function and enter input data.
3. Delete the first transaction.
4. The input data has disappeared in the Contract Interaction tab. 

## Expected behaviour

Removing a tab should not affect the view of a different tab.

## Actual behaviour

lose-function-input

Julianahlk:
## Description

Currently, when you remove a tab, the keys are not being updated in the selectedContractMethods as these are dependent on the transactionFormIndex which changes when a tab is deleted.

The expected behavior is that when you delete a tab, the contract methods for a contract interaction transaction should not disappear.

**New stuff** ✨

* Updated `handleTabRemoval` to shift the keys of the selectedContractMethods when a tab is removed.

contractmethods

(Resolves #3944)

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeForm.tsx
@@ -163,14 +163,42 @@ const ControlSafeForm = ({
   }, [validateForm]);
 
   const handleTabRemoval = useCallback(
-    (arrayHelpers: FieldArrayRenderProps, index: number) => {
+    (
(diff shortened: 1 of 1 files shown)


willm30 (approved):
Nice fix, working for me.

ArmandoGraterol (approved):
Nice work fixing this Juliana :ok_hand:

chinins (approved):
Good to go!

PR #3852, opened 2022-09-14
Add locked funding source

agnieszkajarosikloj:
## Description

This PR adds locked funding sources for streaming payment.
Screenshot 2022-09-14 at 09 44 35
[Figma Link](https://www.figma.com/file/dCtvv76S8mk5HNApjwmBto/Expenditures?node-id=5541%3A513296)

Form section with settings will be added in different PR. 

## TODO

- Add locked state for the settings section after merging [PR](https://github.com/JoinColony/colonyDapp/pull/3833).

Resolves  #3846

PR #4022, opened 2022-10-24
Fix errors counter

### LINKED ISSUE #4024: Issue with error validation not updating correctly on continued use
## Description

Issue with validation discovered when testing https://github.com/JoinColony/colonyDapp/pull/3852. As described here - https://github.com/JoinColony/colonyDapp/pull/3852#pullrequestreview-1148851724

## Expected behaviour

* The validation should be accurate when fields are updated.

## Actual behaviour

* There are some inconsistencies in what triggers the validation (usually it's any change to the input but sometimes I had to manually click Fix it to see the errors number update) but I can't really determine the steps to reproduce. Here's a video of it happening (notice the first funding source updates the number of errors on change, whereas for the second one it doesn't)

https://user-images.githubusercontent.com/33682027/197577534-dc8bab53-6d90-4a24-8266-c0089c3e45c6.mov

agnieszkajarosikloj:
## Description

This PR fixes the bug in updating error counter. The counter was not correctly updating after adding multiple funding sources. It is shown [here](https://github.com/JoinColony/colonyDapp/pull/3852#pullrequestreview-1148851724).

**New stuff** ✨

Added focus in DatePicker component after clicking "Fix it" button.

Screenshot 2022-10-24 at 15 11 05

Resolves #4024

### DIFF
--- a/src/modules/core/components/Fields/DatePicker/DatePicker.tsx
+++ b/src/modules/core/components/Fields/DatePicker/DatePicker.tsx
@@ -33,6 +33,11 @@ export interface DatePickerOption {
   showDatePicker?: boolean;
 }
 
+interface DatePicerError {
+  date?: string | MessageDescriptor;
+  option?: string | MessageDescriptor;
+}
+
 interface Props {
   name: string;
   showTimeSelect?: boolean;
@@ -51,7 +56,7 @@ interface DateInputProps extends React.HTMLProps<HTMLButtonElement> {
   dateFormat: string;
   shouldShowDatePicker: boolean;
   selectedOption?: DatePickerOption | null;
-  error?: string;
+  error?: DatePicerError;
 }
 
 /** The component displaying the currently selected date / option */
@@ -63,6 +68,7 @@ const DateInput = (
     shouldShowDatePicker,
     selectedOption,
     error,
+    name,
   }: DateInputProps,
   ref: React.Ref<HTMLButtonElement>,
 ) => {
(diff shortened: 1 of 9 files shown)


src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -33,6 +34,11 @@ export interface DatePickerOption {
   showDatePicker?: boolean;
 }
 
+interface DatePicKerError {
```
jakubcolony, line 37:
```suggestion
interface DatePickerError {
```

src/modules/core/components/Fields/DatePicker/DatePicker.tsx
```diff
@@ -260,9 +274,12 @@ const DatePicker = ({
       ? `${DEFAULT_DATE_FORMAT}, ${DEFAULT_TIME_FORMAT}`
       : DEFAULT_DATE_FORMAT;
 
+  const { inputRef: omniInputRef } = useDateTriggerFocus(index, false);
```
jakubcolony, line 277: Not sure this should be called `omniInputRef`

jakubcolony (changes requested):
Hey @agnieszkajarosikloj, nice work with the counter, it updates as soon as you fix an invalid field now.
I left some comments about the code but I'm also wondering if the approach of identifying the invalid fields by their order rather than something like name is the right one. I don't think it's super clear you need to set the index prop manually for it to function whereas every field already has a name prop that's also its unique identifier. 
Would be good if someone else could weigh in on that

agnieszkajarosikloj:
@jakubcolony I have removed indexes. Now names are compared.

jakubcolony (approved):
Thanks for making the changes @agnieszkajarosikloj, LGTM 👍

chinins (approved):
Very nice - good job!

PR #4025, opened 2022-10-24
Fix Motions Query/Subscription `args` local resolver

rdig:
This PR fixes a issue reported in the internal `#tech-support` channel, where all motions on the `metacolony` would show up as _"forced"_ actions.

Turns out that while initially that looked so, in reality, the actions list would only show the actions triggered internally by the successful motions _(the initiator was actually the `VotingReputation` extension contract address)_

This was because the `motions` subgraph subscription (query) would fail, so the query would return an empty array. This also ties in with an implementation detail that we have, were we bring all the actions, but filter them out with the correct motions, if they were internal actions created by the extension contract. But since the dapp didn't know about any motions anymore, it would just show the actions.

Now, the `motions` query is "augmented" with a few local resolvers: `status`, `type`, `args`, etc. The `args` local resolver would fail in two places: first, if a _address null_ token was being used for a payment, second, if the return value for a simple decision was not the correct format.

Once the above two were fixed, everything started working correctly.

Besides the above, I've also put in a fix, that will prevent showing the _Simple Decisions_ Colony Home navigation entry, if the `VotingReputation` extension is not upgraded to version `7`, meaning the colony does not support them.

_Note: that this fix is already deployed into production_

### DIFF
--- a/src/data/generated.ts
+++ b/src/data/generated.ts
@@ -1354,6 +1354,7 @@ export type ActionsFilter = {
 export type MotionsFilter = {
   associatedColony?: Maybe<Scalars['String']>;
   extensionAddress?: Maybe<Scalars['String']>;
+  action_not?: Maybe<Scalars['String']>;
 };
 
 export type OneTxPayment = {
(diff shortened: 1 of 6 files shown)


willm30 (approved):
💪

chinins (approved):
All good!

PR #4026, opened 2022-10-25
Handle long strings in `Select` component

### LINKED ISSUE #3991: Handle long strings in Select
This issue generalises one of the cases mentioned by @willm30 in #3972 related to handling long strings in Select. Similarly to #3943, I think the fix should be apply to all Select instances. 

image

jakubcolony:
## Description

This PR adds a little fix to correctly apply ellipsis overflow to Select options:
Screenshot 2022-10-25 at 19 46 57

I have checked that it doesn't impact any Select that it shouldn't but would appreciate if you double check that.

Resolves #3991

### DIFF
--- a/src/modules/core/components/Fields/Select/Select.css
+++ b/src/modules/core/components/Fields/Select/Select.css
@@ -1,6 +1,5 @@
 .baseSelect {
   padding: 5px 10px;
-  width: auto;
   width: 100%;
   position: relative;
   border: 0;
--- a/src/modules/core/components/Fields/Select/SelectListBox.css
+++ b/src/modules/core/components/Fields/Select/SelectListBox.css
@@ -2,8 +2,7 @@
 .baseTheme {
   height: auto;
   max-height: 170px;
-  width: auto;
-  min-width: 100%;
+  width: 100%;
   position: absolute;
   top: calc(100% + 5px);
   z-index: var(--z-index-nav);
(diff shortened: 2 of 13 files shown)


chinins (approved):
It seems to be working well.

willm30 (changes requested):
I think some of the selectors may need to have a min-width added manually where they are used, e.g: 

no

Also, should the "selected" label remain visible in its entirety and the team name the thing that is shortened? Just thinking out loud here. An easy fix for selectors that contain domain names (which are capped at 20 chars) would be to remove the right hand side padding, but then you'd need to manually review all other instances and reapply padding where appropriate, so there may be more a general solution. That's if it's worth implementing at all. 

sel

willm30:
If the above is seems to much of a hassle, maybe this could be a safe control specific fix? Unless there are any selectors on master that would currently benefit, but I haven't noticed any so far. Domain names are capped at 20 chars and tokens at 5 so they aren't at risk of overflow.

jakubcolony:
@willm30 Thanks for spotting this mate, the issue was that I constrained all the Select option lists to 100% width while there were 4 cases in which the list had to be wider, I added an extra appearance prop to handle this:
Screenshot 2022-10-26 at 20 04 42
Screenshot 2022-10-26 at 20 04 36

>Also, should the "selected" label remain visible in its entirety and the team name the thing that is shortened?

I agree, updated:
Screenshot 2022-10-26 at 19 42 15

I've now manually gone over every (I think) `<Select>` in the dapp and everything seems to be working correctly but if you could have another look that'd be great 👍

willm30 (approved):
Nice one! From my testing, all is working as expected.

PR #4036, opened 2022-10-27
Fix SingleUserPicker invalid state

### LINKED ISSUE #4027: Safe control pickers styling breaks if safe transaction fails
I've seen this issue before but I wasn't sure how to reproduce it. I don't know if it's relevant since the transaction failed most likely because there was something wrong with my bridging environment so I'll let @ArmandoGraterol or @willm30 decide and picko or close this issue.

## Steps to reproduce

1. Create any safe transaction.
2. Make it fail after clicking submit (you'll probably know better what can cause that)
3. Go back and click on any of the pickers

## Expected behaviour

* Nothing happens

## Actual behaviour

* The styling of the clicked picker breaks:

https://user-images.githubusercontent.com/112586815/197875751-daa7628b-88b9-4c9f-b512-b02f76263299.mov

Screenshot 2022-10-25 at 21 26 15

willm30:
## Description

This PR fixes a bug in the SingleUserPicker, in which the touch state is not set and thus the "invalid" input state doesn't show up when the picker is empty. It also fixes a bug in the "invalid" input state, which wasn't composing from the input class. 

**Changes** 🏗

* Set touched to true on blur
* Have invalidInput inherit from input class

userpicker

## To test

Go to any instance of the Single User Picker. When no user is selected, and the input has been touched, the red error line should show up (wasn't happening before). 

Resolves #4027


jakubcolony (approved):
LGTM, nice fix @willm30
Before:
Screenshot 2022-10-27 at 19 40 41

After:
Screenshot 2022-10-27 at 19 43 24

Julianahlk (approved):
Working as expected 👍
Screen Shot 2022-10-27 at 6 11 43 PM

chinins (approved):
Very nice!

PR #4043, opened 2022-10-28
Refactor cicd

bogdan-1337:
## Description

- Adding github workflow for build and deploy CI/CD process. Moved Dapp dockerfile from colony-docker-images repo into this repo to centralise the process.

**New stuff** ✨

Github workflow
Dockerfile

**Changes** 🏗

* Changes go here

**Deletions** ⚰️

* Deletions go here

## TODO

- [ ] Add TODOs


.github/workflows/buildDeploy.yml
```diff
@@ -1,8 +1,79 @@
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v2
+        
+      - name: Build and push
+        uses: docker/build-push-action@v3
+        with:
+          context: ${{ github.workspace }}
+          file: ./Dockerfile
+          push: true
+          tags: 204031746016.dkr.ecr.eu-west-2.amazonaws.com/app-frontend:${{ env.IMAGE_TAG }}
```
rdig, line 62: Shouldn't the region and cluser id's _(I presume)_ be stored inside env variables / secretes ?

```suggestion
          tags: ${{ env.AWS_CLUSTER_ID }}.${{ env.AWS_REGION }}.amazonaws.com/app-frontend:${{ env.IMAGE_TAG }}
```

rdig (approved):
Seems fine from my _(limited)_ perspective

Since you're already an admin, feel free to bypass the two reviews requirement and merge it directly _(after you rebase it on the lastest changes, obviously)_

.github/workflows/buildDeploy.yml
bogdan-1337: fair suggestion and I did consider it. the reason why I left it hardcoded is because the ECR repo where all the images will be based in will always be based in that account/region, as we will only have 1 image version that gets deployed into all envs, so they are not really variables.

rdig: Is that "cluster id" (or whatever it is), part of the private repo name ? Since honestly that's what triggered my alarm bells initially, not necessarily the region :)

bogdan-1337: That number is the AWS account id, which is the account id of the QA account where the images will be stored. The prod environment and any other environment will be using that account to pull images from. 

I'll create a variable for it, just to not have it public in the pipeline.

PR #4044, opened 2022-10-28
Fix validation edge cases

### LINKED ISSUE #4042: Investigate validation edge cases
## _Originally posted by @Julianahlk in https://github.com/JoinColony/colonyDapp/pull/3821#pullrequestreview-1159327857_

**Besides, while testing I saw a few validation issues that perhaps you could look into:**
Raw Transaction - When typed a number such as 87,897,897,987,987,978 I received the following error: 
validation1
But with small numbers I am not seeing the issue above.

Contract Interaction - I used this verified contract address for Etherium `0x7301cfa0e1756b71869e93d4e4dca5c7d0eb0aa6`
And when trying to interact with the data (bytes) input field for the upgradeToAndCall function it seems that this field doesn't work no matter the length of the Hex value I typed.
testing1

willm30:
## Description

This PR fixes a couple of bugs in the Safe Control validation.

First, in the Contract Interaction section, dynamically-sized byte arrays were defaulting to bytes1 arrays, when actually, their size is arbitrary. 

Second, in the Raw Transaction section, inputting a wei amount over the max safe integer was causing ethers to error because the string was being coerced into a `number`. We can just pass it straight through as a string. 

**Changes** 🏗

* Add validation for dynamically sized byte arrays. 

Resolves #4042

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
@@ -35,8 +35,8 @@ export interface SafeTransaction {
   nft: SelectedNFT | null;
   nftData: NFT | null;
   tokenData: SafeBalanceToken | null;
-  amount: number | null;
-  rawAmount: number | null;
+  amount: string | null;
+  rawAmount: string | null;
   data: string;
   contract: AnyUser | null;
   abi: string;
--- a/src/modules/dashboard/sagas/utils/safeHelpers.ts
+++ b/src/modules/dashboard/sagas/utils/safeHelpers.ts
@@ -192,9 +192,13 @@ export const getRawTransactionData = (
     throw new Error('Transaction does not contain a recipient.');
   }
 
+  if (transaction.rawAmount === null) {
+    throw new Error('Transaction does not contain an amount');
+  }
+
   return zodiacBridgeModule.interface.functions.executeTransaction.encode([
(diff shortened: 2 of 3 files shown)


src/modules/dashboard/components/Dialogs/ControlSafeDialog/ControlSafeDialog.tsx
```diff
@@ -35,8 +35,8 @@ export interface SafeTransaction {
   nft: SelectedNFT | null;
   nftData: NFT | null;
   tokenData: SafeBalanceToken | null;
-  amount: number | null;
-  rawAmount: number | null;
+  amount: string | null;
```
willm30, line 38: These are coming direct from the input as strings, which is actually ok because ethers accepts number strings for uint input types. This is convenient in the raw transaction section for example, where the amount of wei entered can exceed the max safe integer in javascript. Passing it through straight as a string fixes overflow errors.

ArmandoGraterol (approved):
Working as expected comrade! Nice work.

Another day, another param type discovered :joy:

Julianahlk (approved):
I tested a few times with different verified contract addresses and different functions and things are looking good now. Thank you for addressing these issues Will. Excellent work!
Screen Shot 2022-10-28 at 9 24 13 PM
Screen Shot 2022-10-28 at 9 24 50 PM
Screen Shot 2022-10-28 at 9 32 21 PM

chinins (approved):
All good!

PR #4050, opened 2022-11-01
Fix amount flashing in Safe Transaction Preview

### LINKED ISSUE #4049: Fix amount flashing in Safe Transaction Preview
Watch the "Amount" field. When you type in the input, the component rerenders. This fires a useEffect in the `numeral` component, that causes the innerHTML of the `numeral` component to update. This causes an occasional, momentary loss of the value inside the numeral component, leading to the flashing shown below.

 
amount-flashing

willm30:
## Description

This PR implements a fix for the amount flashing in the Safe Transaction Preview. See the issue for an example of the problem.

It's an intentionally "quick and easy" fix. I envisioned the problem being fixed by https://github.com/JoinColony/colonyDapp/pull/3922, however, since it's been closed in favour of focusing on the CDapp, I've implemented something simple that solves the problem without impeding the display of engineering notation in other parts of the Dapp.

## To test

Go to safe control.
Select transfer funds and go to preview. 
No amount of moving to / from preview or inputting into the preview's inputs should cause any flashing in the amount line.

Numeral component should not be broken in any other part of the dapp.

**Changes** 🏗

* Only run the effect if we need to render the `<sup>` tags (as is the case only when displaying engineering notation). This is the only example I could see of the number formatter returning an html string.

Resolves #4049

### DIFF
--- a/src/modules/core/components/Numeral/Numeral.tsx
+++ b/src/modules/core/components/Numeral/Numeral.tsx
@@ -57,19 +57,30 @@ const Numeral = ({
     mantissa,
     totalLength,
   });
+  const prefixStr = prefix ? `${prefix} ` : '';
+  const suffixStr = suffix ? ` ${suffix}` : '';
+  const fullNumber = `${prefixStr}${formattedNumber}${suffixStr}`;
 
   useEffect(() => {
     if (outputRef.current) {
-      const prefixStr = prefix ? `${prefix} ` : '';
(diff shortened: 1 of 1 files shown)


jakubcolony (approved):
Nice workaround @willm30, not experiencing the flashing anymore 👍

Julianahlk (approved):
Nice fix! LGTM 👍

PR #4052, opened 2022-11-02
Don't abbreviate numbers less than 7 digits

### LINKED ISSUE #4051: Fix formatting of numbers with less than 7 digits
## Specification

The original number formatting issue #3733 has been moved over to CDapp and @arrenv asked to include this small fix in the Dapp for the time being.

Numbers below 1 million should not be abbreviated like in the screenshot:
image

jakubcolony:
## Description

This PR prevents numbers with less than 7 digits from being abbreviated:
Screenshot 2022-11-02 at 11 20 46

**Changes** 🏗

* New format config for medium numbers (1,000-999,999)
* Removed `totalLength` prop from formatting function and `Numeral` component. It makes more sense to control the precision of a number by setting its mantissa length rather than to restrict it by total length.

Resolves #4051

### DIFF
--- a/src/modules/core/components/Numeral/Numeral.tsx
+++ b/src/modules/core/components/Numeral/Numeral.tsx
@@ -27,9 +27,6 @@ export interface Props extends HTMLAttributes<HTMLSpanElement> {
   /** Number of mantissa digits to show */
   mantissa?: number;
 
-  /** Total length of number to show */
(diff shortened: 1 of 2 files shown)


chinins (approved):
Looking good!

Julianahlk (approved):
Well done Jakub! 👍

PR #4064, opened 2022-11-13
Add User Agent Data Types for Typescript

### LINKED ISSUE #4062: Dapp does not work in Firefox since Simple Decisions
## Steps to reproduce

1. Go to the DApp within the Colony browser.
2. You should see a white screen and the DApp won't load.
3. Check the console and you will see the error:
```
Uncaught TypeError: window.navigator.userAgentData is undefined
```

The line causing the issue - https://github.com/JoinColony/colonyDapp/blob/9af4ba9536bdcdd44eda852a29aa18b740905333/src/modules/core/components/RichTextEditor/Toolbar.tsx#L30

## Expected behaviour

* The DApp should work in Firefox

## Actual behaviour

* The DApp does not work in Firefox since Simple Decisions was released.

Screenshot 2022-11-09 203845

danbr:
## Description

This PR resolves the issue where the dapp doesn't load in FF/Safari.
Adding TypeScript definitions for navigator.userAgentData seems to resolve the issue. I have only tested on Safari (Mac).

Before:

Screenshot 2022-11-13 at 20 13 00

After:

Screenshot 2022-11-13 at 20 50 33

Resolves #4062


tsconfig.base.json
```diff
@@ -1,42 +1,47 @@
...
-    "resolveJsonModule": true,
-    "skipLibCheck": true,
-    "strictNullChecks": true,
-    "strictPropertyInitialization": true,
-    "target": "es6",
-    "typeRoots": ["src/typeDefs", "node_modules/@types"],
-    "types": ["node", "events", "jest"]
-  }
+    "compilerOptions": {
+        "baseUrl": ".",
```
chinins, line 3: It seems that there is an extra space in front of each line vs. the original file. Is it your linting? I think it's good to keep it consistent as in master, without extra space.

chinins (commented):
As far as I understand, `userAgentData` on navigator object is not supported on safari & firefox.
https://developer.mozilla.org/en-US/docs/Web/API/Navigator/userAgentData
The fix of adding `?` after `userAgentData` it works and it's a legitimate fix. However, I'm not sure of adding the types package. It removes the necessity of tslint ignore but adds additional package and most importantly if we ever use it again it won't trigger the alarm bell to use `?` after `userAgentData` and the bug might be repeated in a different part of code.

I hope I'm clear. I'm just not sure what's the right way here for types...

willm30 (approved):
Nice fix Dan 👍

willm30:
>  However, I'm not sure of adding the types package. 

Without the types package, TS doesn't acknowledge that `userAgentData` exists at all, which I believe is the source of my original error (i.e., knowing that it is provided by Chrome made me think TS is just wrong, instead of thinking it might _only_  be provided by Chrome, which is suggested by the types package's definition). 

Without:
useraentts

With:
withtypes

I would therefore be inclined to include it.

ArmandoGraterol (approved):
Yup, no longer crashes for me :ok_hand: Firefox on Linux

chinins (approved):
@willm30, thank you for your perspective!

It seems that this is the way to go indeed then. Nice fix!

dalovar:
I'm still experiencing this issue today on Firefox when loading:

https://xdai.colony.io/landing
https://xdai.colony.io/colony/meta

The errors in the console are:
`Uncaught TypeError: window.navigator.userAgentData is undefined

Firefox can’t establish a connection to the server at wss://xdai.colony.io/graph/subgraphs/name/joinColony/subgraph.

The connection to wss://xdai.colony.io/graph/subgraphs/name/joinColony/subgraph was interrupted while the page was loading.`

When will the solution be pushed to production? So I can simply load the website and have it fixed on firefox?
The interface is currently blank on Firefox due to this error

danbr:
> When will the solution be pushed to production? So I can simply load the website and have it fixed on firefox?
> The interface is currently blank on Firefox due to this error

Thank you for your comment. We're sorry that you're having issues with the Dapp when using Firefox.
We are hoping to move the fix into production next week.

firebomb1:
@danbr I still encounter this error and thus the dApp is still completely unusable in Firefox.

When will this fix be pushed in production?

nixtoshi:
> @danbr Hello, I still encounter this error and thus the dApp is still completely unusable in Firefox.
> 
> When will this fix be pushed in production?

Hey im another user, agree with this. Firefox is probably the best browser right now, so it's pretty bad that this is happening on Chrome.

So far the easies solution if you need to use Colony is just using a Chrome browser. I recommend regular Chrome, or Chrome beta that is fully destined to just crypto-related stuff.

arrenv:
Thanks @firebomb1 & @nixtoshi, this actually should be in production. I will follow up on this.

arrenv:
@firebomb1 & @nixtoshi, looks like the fix for Firefox is in production. Please let me know if you notice any issues though.

firebomb1:
@arrenv Thanks, the app works now on Firefox!

PR #4065, opened 2022-11-14
Add improvements and bug fixes to Safe Control

ArmandoGraterol:
Compilation of fixes and improvements as a result of testing on QA

**Changes** 🏗

* Added safeguards to IPFS metadata fetching
* Added metatransaction messages for gas stations related to Safe Control
* Fixed "Unknown contract" appearing in the event list of ActionPage as "Unknown contract contract"


chinins (approved):
Very good!

willm30 (approved):
LGTM bro

src/data/resolvers/motions.ts
```diff
@@ -1310,20 +1310,21 @@ export const motionsResolvers = ({
           apolloClient,
         );
 
-        const safeTxMetadata = JSON.parse(
-          (await getColonyMetadataIPFS(annotation?.values?.metadata || '')) ||
-            '[]',
-        );
+        if (annotation) {
+          const safeTxMetadata = await getColonyMetadataIPFS(
```
danbr, line 1314: I'm a bit surprised with getColonyMetadataIPFS It doesnt really make use of ColonyMetadataEventParser.
The func is returning (promise)any, but could be (type) ColonyMetadata as the name suggests.

Also, inside getColonyMetadataIPFS it doesnt handle different version of metadataVersion. I guess in theory we would not expect any safe data to be encoded in V1, everything should now be V2, but its good to have the framework so we can easily support V3.

Not using ColonyMetadataEventParser to parse V2 data means that yup validation is not be used.

What happens if the data is malformed?

If you're using the annotation to store data, why not model the function on: `ActionsPageFeedItemWithIPFS.tsx` around line 51, which already implements much of the parsing/validation?

danbr (commented):
Left a few comments, I dont know the details of the specific case, but perhaps you can make use of the ColonyEventMetadataParser?

ArmandoGraterol:
@danbr For some reason, when I merged #4067 this branch was also merged?? Don't know what happened there, but don't worry, I will address your feedback in another PR

Issue #4070, opened 2022-11-17
Awarding Reputation Motions not working


arrenv:
## Steps to reproduce

1. Ensure you have the Governance extension installed, a team and some reputation in the team.
2. Try to Award reputation with a Motion in a team.
      - The transaction will fail with the error `voting-rep-invalid-domain-id`.
3. Try to manually override the `Create a motion in Root` option.
      - The transaction will fail with the error `voting-rep-invalid-domain-id`.

The problem here from the contract side is that it is not possible to award reputation in teams. This is by design to prevent subdomains from awarding reputation that flows upwards and could take over a Colony. (Thanks @area)

## Expected behaviour

* The user should be able to Award reputation in a team.
* The Motion should only create the Motion in the Root domain.

## Actual behaviour

* You are unable to create a Motion to Award reputation in a team.
* The transaction will be reverted with the error `voting-rep-invalid-domain-id`.
* The same error happens if you manually select the `Create a motion in Root` option.

![create-in-root](https://user-images.githubusercontent.com/33682027/202496133-5731baf2-a675-4a67-b6bd-ff80de49bd10.gif)

PR #4075, opened 2022-11-22
Add more misc fixes and improvements to Safe Control

ArmandoGraterol:
**Changes** 🏗

* Restored some package-lock code that was causing some warnings in the console.
* Fixed not being able to create a motion on the root domain when rewarding reputation in a subdomain.
* Fixed the dapp crashing on the homepage when reloading after recompiling.
* Added filter function to `getContractUsefulMethods` to filter out read-only methods.
* Updated `getColonyMetadataIPFS` to `getSafeTransactionFromAnnotation`.

Resolves [#4070](https://github.com/JoinColony/colonyDapp/issues/4070)


chinins (approved):
Looking good!

PR #4079, opened 2022-11-24
Improve instructions in 2nd step of adding a safe

### LINKED ISSUE #4078: Update Instructions Modal for Adding a Safe
### Description

While going through the process of adding a Safe I found it confusing as to what I needed to do with the details. The following adds an update to have a better description for the link to set up the module. As well as generating the link for the user to be able to more easily navigate to add the module to their Safe.

1. Update the copy of the warning message to "Be aware. Cross-chain bridging can have an element of risk. " as of design.

2. Update the URL to the correct section of the setup document, for the moment before the official move to Colony docs, the URL should be the following and open in a new tab: 

     `https://colony.gitbook.io/colony/advanced-features/safe-control-gnosis-safe#step-2-connect-the-safe`

3. Add a URL direct to the apps page to install the Bridge Module. This needs to be generated for the user using the selected chain (with the Gnosis used prefix), and their Safe address (should open in a new tab).

     `https://app.safe.global/{{CHAIN_PREFIX}}:{{SAFE_ADDRESS}}/apps?appUrl=https%3A%2F%2Fzodiac.gnosisguild.org%2F`

     The chain prefixes that Safe uses are listed here - https://github.com/safe-global/web-core/blob/a0a0ca221f712a595b0016d7ed86932483730402/src/config/chains.ts

     Note: I do have a concern of having to maintain this link in the DApp, if you think it is fine, all good. Or, if you have any suggestions please let me know.

4. Add confirmation tooltip for copying value.

### Design

- [Figma link - modal updates](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=8118%3A22920&t=syWHwuSTVqyjgWOu-1)
- [Figma link - copied confirmation tooltip](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=8120%3A22976&t=syWHwuSTVqyjgWOu-1)

**Updated design**
Add Safe 3

**Copied tooltip**
Screenshot 2022-11-24 132614

### Implementation

- [x] Update the copy of the warning message.
- [x] Update the URL and text of the Learn more link.
- [x] Update the design to include the new headings.
- [x] Add in a link direct to the apps page for the Bridge Module.
- [x] Add confirmation for value copied

ArmandoGraterol:
Improvements were made specifically to the 2nd step of the "Add Safe" process per our observations in the live QA session.

**Note**: In case you are in need of a safe to test the steps: `0x3F107AFF0342Cfb5519A68B3241565F6FC9BAC1e` (Ethereum).

**Changes** 🏗

* Changed copy of warning in 2nd step
* Added a new subsection with a link to easily get to the Safe App screen where the user is supposed to install the bridge module.
  * This link is generated by using the selected safe chain and address
* Updated copy and link of "Learn more" in the 2nd step so that the user gets to the relevant section in our docs faster/easier.
* Added a tooltip to the clipboard copy icon so that it's easier to tell that something was copied onto the clipboard.

(It would be a "good to have" if we can find a Binance safe also, just to test the link when the safe is on Binance)

FireShot Capture 757 - Actions - Colony - wonker - localhost

Resolves #4078

### DIFF
--- a/src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/ConnectSafe.tsx
+++ b/src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/ConnectSafe.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react';
+import React from 'react';
 import { FormikProps } from 'formik';
 import {
   defineMessages,
@@ -6,7 +6,6 @@ import {
   MessageDescriptor,
   useIntl,
 } from 'react-intl';
-import copyToClipboard from 'copy-to-clipboard';
 
 import Button from '~core/Button';
 import DialogSection from '~core/Dialog/DialogSection';
@@ -15,10 +14,13 @@ import Icon from '~core/Icon';
 import { Tooltip } from '~core/Popover';
 
 import {
+  ETHEREUM_NETWORK,
   GNOSIS_AMB_BRIDGES,
   GNOSIS_NETWORK,
   SUPPORTED_SAFE_NETWORKS,
 } from '~constants';
+import { CONNECT_SAFE_INSTRUCTIONS, getModuleLink } from '~externalUrls';
+import { useClipboardCopy } from '~modules/dashboard/hooks';
 
 import { FormValues, AddExistingSafeProps } from './index';
 
@@ -63,7 +65,7 @@ const MSG = defineMessages({
   },
   copyDataTooltip: {
     id: 'InvisibleCopyableAddress.copyAddressTooltip',
-    defaultMessage: `{copied, select,
+    defaultMessage: `{isCopied, select,
       true {Copied}
       false {{tooltipMessage}}
     }`,
@@ -73,8 +75,6 @@ const MSG = defineMessages({
     defaultMessage: 'Click to copy',
   },
 });
-
-const instructionsHref = `https://colony.gitbook.io/colony/advanced-features/safe-control-gnosis-safe/adding-a-safe#step-2-connect-the-safe`;
 
 interface ConnectSafeProps
   extends Pick<AddExistingSafeProps, 'colonyAddress' | 'setStepIndex'> {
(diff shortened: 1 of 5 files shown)


src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/ConnectSafe.tsx
```diff
@@ -62,25 +91,54 @@ const ConnectSafe = ({
   values: { chainId },
   setStepIndex,
   colonyAddress,
+  safeAddress,
 }: ConnectSafeProps & FormikProps<FormValues>) => {
+  const selectedChain = SUPPORTED_SAFE_NETWORKS.find(
+    (network) => network.chainId === Number(chainId),
+  );
+  const { formatMessage } = useIntl();
+  const moduleHref = `https://app.safe.global/${selectedChain?.shortName.toLowerCase()}:${safeAddress}/apps?appUrl=https%3A%2F%2Fzodiac.gnosisguild.org%2F`;
```
willm30, line 100: Perhaps move this to a helper function like `getModuleLink()` that returns this href? I think that would be nicer to read, especially if you're just skimming through. 

And perhaps assign the constants that prepend and append the selectedChain/safeAddress to variables in `externalUrls` also?

willm30 (commented):
Looking good. Left some q's in the code. 

Re the safe on binance, perhaps worth deploying one yourself and then claiming related costs back from colony? Not sure exactly the protocol here but would seem reasonable, especially given we're only talking a few dollars. Essentially a business expense.

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/ConnectSafe.tsx
```diff
@@ -62,25 +91,54 @@ const ConnectSafe = ({
...
   const CopyableData = ({ label, text }: CopyableProps) => {
-    const { isCopied, handleClipboardCopy } = useClipboardCopy(text);
```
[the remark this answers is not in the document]
ArmandoGraterol: I didn't consider this approach, sounds good to me. I just did it how it was familiar to me from other places in the dapp

src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/ConnectSafe.tsx
ArmandoGraterol: I don't think putting this inside a function will make it more readable, however, I do have some alternatives in mind now that you have mentioned `externalUrls`

chinins (approved):
Good job!

PR #4090, opened 2022-11-29
Fix safe address validation on Add Safe process

### LINKED ISSUE #4080: Validate address on adding an address into the input
## Description
This was observed during the live QA session as not intuitive and resulted in the user going back and trying again as they thought it was broken.

## Steps to reproduce

1. Copy an address into the Safe address field.
2. Try to click on the input again.
3. Try to click on the Next button again.
4. You have to click anywhere else in the modal in order for it to validated

## Expected behaviour

* Address should try to validate straight away when enough characters have been entered.

## Actual behaviour

* When copying an address into the address input, it does not validate.
* When clicking in the input again with an address it doesn't validate.
* When clicking on the disabled Next button it does not validate

adding-address-in

ArmandoGraterol:
Fixed safe contract address field not validating after copy-pasting or inputting throught autocomplete an address. It should now validate instantly after those cases.

Resolves #4080

### DIFF
--- a/src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/CheckSafe.tsx
+++ b/src/modules/dashboard/components/Dialogs/AddExistingSafeDialog/CheckSafe.tsx
@@ -98,16 +98,14 @@ const CheckSafe = ({
   const [
     ,
     { error: addressError, touched: addressTouched, value: address },
-    { setError: setAddressError },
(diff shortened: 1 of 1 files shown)


chinins (approved):
Looking good!

willm30 (approved):
Working 4 me

PR #4095, opened 2022-12-01
Update gas station label for add/remove safes actions

### LINKED ISSUE #4084: Fix transaction label in GasStationPopover
## Steps to reproduce

1. Add an existing Safe to a Colony
2. On the final step of adding a Safe, notice the transaction label.

## Expected behaviour

* The transaction label should be accurate to the transaction, in this case it should be `Adding the Safe`.
* Check the same message being used when removing a Safe.

I'm assuming this is using the same saga to update the Colony's IPFS metadata, so, please confirm if this request is not possible.

## Actual behaviour

* The transaction for adding a Safe currently labeled with "Edit Colony Details".

34846676-01944e10f136e4fd78890cfa28ac8531

ArmandoGraterol:
FireShot Capture 759 - Add Safe from Ethereum - Action - Colony - wonker - localhost

FireShot Capture 760 - Remove Safe - Action - Colony - wonker - localhost

Resolves #4084

### DIFF
--- a/src/modules/dashboard/sagas/actions/manageExistingSafes.ts
+++ b/src/modules/dashboard/sagas/actions/manageExistingSafes.ts
@@ -127,15 +127,21 @@ function* manageExistingSafesAction({
       currentMetadataIPFSHash,
     );
 
-    const currentColonyMetadata = JSON.parse(currentMetadata);
+    const currentColonyMetadata = JSON.parse(currentMetadata || '{}');
+
+    if (!currentColonyMetadata?.data) {
+      throw new Error(
+        `There was an error wile fetching the current colony metadata. Please try again later.`,
+      );
+    }
 
     let updatedColonyMetadata: any = {};
 
     if (!isRemovingSafes) {
       updatedColonyMetadata = {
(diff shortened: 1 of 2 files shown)


willm30 (commented):
The second one should still refer to the annotation label?

annotati

willm30 (changes requested):
I'm also noticing this for the first time which is preventing me from adding the safe, so I guess we should check for null when fetching colonyMetadata in `manageExistingSafes.ts`? 

addExistingSafe

Related to this, it doesn't look like we're using the same metadata format in removing / adding. One is `.colonySafes` and one is `.data.colonySafes`.
remadd

ArmandoGraterol:
@willm30 It amazes me that you can find these bugs! It should be fixed

willm30 (approved):
Nice, can confirm the gas stn label has been updated. For whatever reason, my system is struggling to connect to ipfs and so haven't been able to actually add a safe. There may be a conversation to be had about how to correctly handle this in the UI, since currently you'll only see the gas station loading forever.

PR #4096, opened 2022-12-01
Update add/remove safe event copies

### LINKED ISSUE #4085: Incorrect transaction message in actions log
## Steps to reproduce

1. Add a Safe and navigate to the Actions screen for adding the Safe
2. Notice the first log entry in the logs list.

Strangely, I tested this a couple of times and it was inconsistent in replicating it. You can still see it here though - https://migration.colony.io/colony/metacooldao/tx/0x29f090c0adeb958ef5190a3787268d683156d79534c3b9f179bb3ecf63f8a200

Either way, the "funding" permission text should be removed as it is not accurate, and ensure the full stop is added.

## Expected behaviour

* The message should be: "{Username} added the Safe {Safe_Address} from the {Chain_Name} using the Administration permission."
* For removing, please correct to: "{Username} removed {Safe_Address} from the {Chain_Name} using the Administration permission."

[Figma Link - Actions page](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=727%3A15054&t=1MytesFhwrIVww6u-1)

## Actual behaviour

* The message is showing as "{Username} changed this colony's name to {Colony_Name}"

34846917-3154d1bc07dea2b2d24e97c32381a849

34847177-de89fd7639c6cb9f4a05e00d2c7871f4

ArmandoGraterol:
The issue mentions another bug in which the copy for the add safe transaction would be wrong, unfortunately, I haven't been able to replicate this as of now and my only suspicion is that it's due to IPFS shenanigans.

Resolves #4085


chinins (approved):
All good!

danbr (approved):
Its a thing of beauty.... What's not to like? 👍

PR #4098, opened 2022-12-01
Fix token avatar styles on Safe transactions

### LINKED ISSUE #4088: Token alignment issue in transaction details
## Description

When using a token that had a token, it caused the blockie to appear behind it as well as issues with the alignment.

## Expected behaviour

* The icon of the logo should align with the symbol and value.

## Actual behaviour

* Alignment issue when a token is picked up in the transaction

34848634-10ec38552fbb2bbc1f26088f3459fa9a

ArmandoGraterol:
For the avatar to be displayed you would need to access the avatar component inside the `isERC20Token(tokenData)` conditional on `Value`.

If you need a logoUri, you can use this one: `https://safe-transaction-assets.safe.global/tokens/logos/0x6810e776880C02933D47DB1b9fc05908e5386b96.png`

FireShot Capture 762 - asdasda - Action - Colony - wonker - localhost

Resolves #4088


arrenv (changes requested):
@ArmandoGraterol the alignment is looking better, it seems to still have the Blockie icon in the background through.

chinins (approved):
Nice one!

PR #4099, opened 2022-12-01
Update styles of To section in Safe transactions

### LINKED ISSUE #4087: Inconsistent layout of names and addresses
### Description

1. Ensure the layout of names and addresses are consistent. We had added the layout of names and addresses as stacked to avoid issues with addresses being hidden if the name is too long or cutting the username off after only a few characters. Due to this change not being reflected in the designs, it seems this was built again. It would be good to keep it consistent though.

2. There is also an issue with the alignment of the avatar.

3. Also, if there is no username, it is not needed to include the text "Address". This can be removed.

### Design

34847953-666b3ec7a03f9ed489316ca6a6b1a3a3 (2)

34847748-be27bb83180ce476f03931abbe0c396e

ArmandoGraterol:
The section should now be consistent with the other `To` section in other actions.

FireShot Capture 762 - asdasda - Action - Colony - wonker - localhost

Resolves #4087

### DIFF
--- a/src/modules/core/components/DetailsWidgetUser/DetailsWidgetUser.css
+++ b/src/modules/core/components/DetailsWidgetUser/DetailsWidgetUser.css
@@ -14,6 +14,10 @@
   overflow: hidden;
 }
 
+.textContainer > span {
+  height: 15px;
+}
+
 .textContainer > div {
   max-width: 100%;
 }
--- a/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidgetSafeTransaction/DetailsWidgetSafeTransaction.css
+++ b/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidgetSafeTransaction/DetailsWidgetSafeTransaction.css
@@ -59,12 +59,6 @@
   max-width: 250px;
 }
 
-.recipient div[class*='DetailsWidgetUser_textContainer'] {
-  flex-direction: row;
-  margin-left: 10px;
-  column-gap: 10px;
-}
-
 .recipient figure ~ span:nth-of-type(2) {
   margin-top: 4px;
   font-family: var(--family-techie);
--- a/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidgetSafeTransaction/components/Recipient.tsx
+++ b/src/modules/dashboard/components/ActionsPage/DetailsWidget/DetailsWidgetSafeTransaction/components/Recipient.tsx
@@ -3,7 +3,6 @@ import classnames from 'classnames';
 import { FormattedMessage } from 'react-intl';
 
 import { AnyUser, Colony } from '~data/index';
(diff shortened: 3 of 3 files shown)


willm30 (commented):
Looking good for me. 

However, your screenshot is alerting me to a different issue.

I implemented a fallback for the Avatar component because I noticed that if the url doesn't return the expected resource nothing is displayed. However, in this case, it's returning the Gnosis logo which is partially transparent, and the fallback is displayed behind it, which looks odd. 

What we really want is to hook into the request sent by the css and conditionally render the fallback based on whether the request returns a success status code. I looked at the Web Performance API but doesn't look that exposes enough data about requests being made to be useful here. I also skimmed through other Web APIs but didn't see anything that stood out. It can be done with the webRequest api but I believe that's only for use within browser extensions. 

If not, perhaps we just need to revert it so we don't display a fallback. What you think?

ArmandoGraterol:
> Looking good for me.
> 
> However, your screenshot is alerting me to a different issue.
> 
> I implemented a fallback for the Avatar component because I noticed that if the url doesn't return the expected resource nothing is displayed. However, in this case, it's returning the Gnosis logo which is partially transparent, and the fallback is displayed behind it, which looks odd.
> 
> What we really want is to hook into the request sent by the css and conditionally render the fallback based on whether the request returns a success status code. I looked at the Web Performance API but doesn't look that exposes enough data about requests being made to be useful here. I also skimmed through other Web APIs but didn't see anything that stood out. It can be done with the webRequest api but I believe that's only for use within browser extensions.
> 
> If not, perhaps we just need to revert it so we don't display a fallback. What you think?

@willm30 
This is something that came up in #4098 actually, you can check out what I did there but yeah the CSS styles are the ones that make the blockie display in the background. As far as my understanding of the avatar component goes, with the modification that I did in that PR, a fallback should always be available in the case of the avatar image not being available.

willm30:
@ArmandoGraterol Can you point me to where the fallback occurs? I can't see  it. If `avatarURL` is defined, but the css `url()` function fails to fetch the resource, then a div will render with nothing as the background image. Do I have that right?

If so, I don't think there's a lot we can do about it, as I don't think we're able to hook into the browser's network request data well enough to render a fallback in the event the `url()` fetching returns an error status code. And given the possibility of partially transparent avatars, a css-based fallback (like initially implemented) isn't suitable either.

ArmandoGraterol:
@willm30 You are right, we do have fallbacks for the cases that `avatarURL` is not defined but in the case that `url()` fails we don't have a fallback, and the fallbacks proposed across the internet are not good because they are not really fallbacks (If you define 2 backgrounds in `background-image`, both are loaded at the same time as we already noticed). I'm gonna try something on the other PR.

ArmandoGraterol:
@willm30 A possible solution to what we are talking should now be up on #4098

willm30 (approved):
Good job

chinins (approved):
Very nice!

PR #4107, opened 2022-12-14
Update Safe transaction used event

### LINKED ISSUE #4091: Better handle processing and status of transactions
### Description

When using the Safe Control feature you are required to manually use a 3rd-party service to process the actual transaction on the Safe's chain. This allows the users to cover the gas costs required to perform the actual transaction.

While this would ideally be built into the Colony UI, for the time being, we need to navigate to the 3rd-party service. For example, the URL for Gnosis Chain to Mainnet is `https://alm-bsc-xdai.herokuapp.com/100/{transaction_hash}`

So, we need to add in a callout banner in the URL to navigate to this service as well as status indicators to notify the user whether or not the transaction has been completed or not.

- There may be tips here on how to get the status of the transaction on the Safe's chain - https://github.com/omni/tokenbridge/blob/ff9f3fb7d6b4b8b484b4ef92427ce03411265714/alm/src/components/StatusContainer.tsx

Another addition is the status of the transaction in the Actions list and the name of the Safe interacted with.

### Design
 - [Figma link - Callout to process transaction](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=8155%3A23136&t=1MytesFhwrIVww6u-1)
 - [Figma link - Status of transaction](https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=8158%3A23544&t=1MytesFhwrIVww6u-1)
 - https://www.figma.com/file/LlVmeMQxHVA04evfrgZyuN/Safe-Control?node-id=8158%3A23961&t=1MytesFhwrIVww6u-1

transaction-callout-status

completed_transaction

action-item-status

ArmandoGraterol:
**NOTE: Don't be intimidated by the size of the PR, half of the new additions are contract ABIs I added as they are necessary for the new functionality. They can be skipped over when reviewing as they are just standard AMB ABIs I extracted from Token Bridge repo**

Updated safe transaction actions and motions so that they use the new `ArbitraryTransaction` event. The idea of this change is that the user can have a better UX when it comes to finding and executing the transaction when it's passed to the home bridge.

- When you create a safe transaction action, you should see a "Completed" tag instantly in the details widget. This only applies to the dev env as the transactions are automatically executed. To test how would the action page should look after creating the action in the QA/prod env, remove the `onLocalDev` in line 125 of `getTransactionStatuses`.

- When you create a safe transaction motion, the "Completed" tag should only show after you **finalize** a motion that has **passed**. The motion page should show the tags ("Pending" or "Complete") without reloading. To see how the motion page should look after finalizing in a "real" env, again, remove the `onLocalDev`.

- The banner shown when there's a pending transaction should show a button with a valid link to manually process a transaction (`{BRIDGE_MONITOR_BASE_URL}100/{CROSS_CHAIN_MESSAGE_TRANSACTION_HASH}`). In the case of the action page, the transaction hash should be the same hash we use to generate the action page and for motions it should use the transaction hash of the `finalizeMotion` transaction.

- Unfortunately, there's no way to _actually_ test manually executing safe transactions and changing dynamically the tags as we use a real API to check that the message was bridged (`etherscan/bscscan`), but, it should _theorically_ work. (I've tested the API endpoints generated by the function, but by plugging in real values and it should work. I choose to believe)

- Ideally, please check that every safe transaction type is working as expected, by doing single and multiple transactions as all of that should work as expected despite changing from using `Annotation` to `ArbitraryTransaction` event to show safe transactions.

- Safe I've been using for testing: `0x3F107AFF0342Cfb5519A68B3241565F6FC9BAC1e` (Ethereum)
- Related module address: `0x8dfABd117Db8BC64829Ce7C595F20998b054cDC2`

FireShot Capture 767 - Actions - Colony - wonker - localhost
FireShot Capture 768 - Safe transaction_ AAAASSSSSSSS - Action - Colony - wonker - localhost
FireShot Capture 769 - Safe transaction_ asdasda - Motion - Colony - wonker - localhost
FireShot Capture 770 - Safe transaction_ asdasda - Motion - Colony - wonker - localhost
FireShot Capture 771 - Safe transaction_ AAAASSSSSSSS - Action - Colony - wonker - localhost

## Bridging env instructions

0. have your normal dev env running. 
1. `cd src/lib/colonyNetwork` 
8. `git fetch` 
9. `git checkout maint/cross-chain-safe` 
10. `git submodule update --init` 
11. `yarn` 
12. `yarn provision:safe:contracts` 
13. `cd lib/safe-contracts` then `yarn` (ignore errors)
14. `cd ../..` (back into colonyNetwork) 
15. `bash scripts/start-bridging-environment.sh` 
16. `node scripts/start-bridging-environment.js` 
17. Copy the data in console into your .env and restart webpack (home bridge address, local safe address, local token address etc.)

```
# For testing Safe Control locally. Ensure you have a foreign chain running locally.

LOCAL_HOME_BRIDGE_ADDRESS=

LOCAL_FOREIGN_BRIDGE_ADDRESS=

LOCAL_SAFE_ADDRESS=

ZODIAC_BRIDGE_MODULE_ADDRESS=

LOCAL_ERC721_ADDRESS=

LOCAL_SAFE_TOKEN_ADDRESS=
```

Resolves #4091


chinins (approved):
Looking good!

src/data/resolvers/colonyActions.ts
```diff
@@ -9,6 +9,7 @@ import {
 import { BigNumberish } from 'ethers/utils';
 import { AddressZero } from 'ethers/constants';
 import { Resolvers } from '@apollo/client';
+import uniqWith from 'lodash/uniqWith';
```
willm30, line 12: This belongs in ~utils/lodash. Makes imports neater in the event you have two or more from lodash.

src/data/resolvers/colonyActions.ts
```diff
@@ -264,9 +295,7 @@ export const colonyActionsResolvers = ({
           isWhitelistActivated: false,
           verifiedAddresses: [],
           colonySafes: [],
-          safeData: null,
-          safeTransactions: [],
-          transactionsTitle: '',
+          safeTransactionStatuses,
```
willm30, line 298: This should default to empty array like one below?

src/i18n/en-events.ts
```diff
@@ -40,21 +40,21 @@ const eventsMessageDescriptors = {
   [`event.${ColonyAndExtensionsEvents.DomainMetadata}.name`]: `{initiator} changed teams's name from {oldName} to {domainName}`,
   [`event.${ColonyAndExtensionsEvents.DomainMetadata}.color`]: `{initiator} changed teams's color from {oldColor} to {domainColor}`,
   [`event.${ColonyAndExtensionsEvents.DomainMetadata}.fallback`]: `{initiator} changed this team, but values are the same`,
-  [`event.${ColonyExtendedActions.SafeTransactionInitiated}.rawTransaction`]: `{safeName} initiated a transaction with {isSafeTransactionRecipientUser, select, 
+  [`event.${ColonyAndExtensionsEvents.ArbitraryTransaction}.rawTransaction`]: `{safeName} initiated a transaction with {isSafeTransactionRecipientUser, select, 
     true {{safeTransactionRecipient}}
     other {{safeTransactionRawTransactionAddress}}
   }`, // @TODO Confirm w product
```
willm30, line 46: Assuming these have been a-oked by product now?

src/modules/dashboard/sagas/actions/initiateSafeTransaction.ts
```diff
@@ -1,4 +1,5 @@
-import { ClientType, ColonyClient } from '@colony/colony-js';
+import { ClientType } from '@colony/colony-js';
+import { fill } from 'lodash';
```
willm30, line 2: Also can be moved to and imported from ~utils/lodash

src/modules/dashboard/transformers/transformers.ts
```diff
@@ -1,6 +1,7 @@
 import { AddressZero, HashZero } from 'ethers/constants';
 import { bigNumberify } from 'ethers/utils';
 import Decimal from 'decimal.js';
+import uniqWith from 'lodash/uniqWith';
```
willm30, line 4: Same as above re utils/lodash

src/utils/safes/getTransactionStatuses.ts
```diff
@@ -0,0 +1,133 @@
+import { Contract, ethers, Event } from 'ethers';
+import isArray from 'lodash/isArray';
```
willm30, line 2: Same as above re utils/lodash

willm30 (commented):
Nice work dude. 

I was initially having trouble in the actions resolver, but this was because I needed to re-run `npm run provision` to use the updated network clients. After restarting my dev environment on this branch, by and large I'm seeing what I'm expecting to see based on your pr description. 

However, couple of observations.

This is what I see when transferring funds (action) from the ethereum safe. 

transfer-funds-br

When I do a refresh, it loads fine:

tf

I tried a few times and same result on each occasion. Passing state over the location object in actions as well as motions would solve what I am presuming is an issue with the subgraph not ingesting the data quickly enough after submitting the form.

I can confirm however that the completed/pending labels show up in the details widget, and in the actions list on the colony home page (once the data is present). 

This is not the case for motions. With motions, the completed/pending tag is not showing up in the actions list (but does appear in the details widgets).

Top safe tx is the finalized motion, bottom the action. 
mt

Aside from this, completed/pending tags are showing up as expected and the banner appears to have the correct tx hash for motions/actions.

Let me take another quick look once you address the above and then we should be good to go 👍

src/data/resolvers/colonyActions.ts
ArmandoGraterol: The default is already an empty array in a few lines above: `let safeTransactionStatuses: TRANSACTION_STATUS[] = [];`

src/i18n/en-events.ts
ArmandoGraterol: Yeah I think it's safe to remove these comments

ArmandoGraterol:
@willm30 Thank you very much for reviewing! The changes to the issues you pointed out should be up

willm30 (changes requested):
I do see the completed tag now on motions in the actions list; however, it's appearing before the motion has been finalised:
beforefina

Also, and perhaps I missed this before, but the completed tag in the details widget on the motion is requiring a refresh to appear:
tf

When I force a motion / create an action transaction, this is what I see:
forcing

Loads normally on reload.

ArmandoGraterol:
@willm30 Most of the stuff you pointed out should be fixed! I couldn't reproduce the transaction status tag not appearing after finalizing the motion without reloading, however, I did make some changes that could affect it so please test that again and let me know if it's still present

willm30 (commented):
When creating an action, most of the information is now showing up. The annotation is still generic though ("initiated a transaction").  
ban

Also, the "pending/completed" tag isn't showing up for me in motions until I refresh/navigate away and back again.

willm30 (approved):
Per our discord conversation, the statuses are showing up correctly in qa so approving!

PR #4109, opened 2022-12-14
Add warning to TransferFunds section on Control Safe

### LINKED ISSUE #4086: Ensure there is validation on the recipient address
## Steps to reproduce

1. With a Safe added on Colony, go to create a transaction request
2. Go to create a transaction and add in an address that is not on the Safe's chain. e.g. use a Gnosis Chain address, like a Colony address.
3. You will notice that it allows you to do this transaction without validation. This would result in lost funds.

## Expected behaviour

* The recipient field should validate that the address entered exists on the same chain as the Safe.
* This should be the case for all transaction types.

## Actual behaviour

* There is no validation on the recipient address.

34847489-7b474ad040cd1109b20a00c8ca0389c6

ArmandoGraterol:
Safe I used to test:
`0x3F107AFF0342Cfb5519A68B3241565F6FC9BAC1e` (Ethereum)
`0x8dfABd117Db8BC64829Ce7C595F20998b054cDC2` (Module address)

`0xe759f388c97674D5B8a60406b254aB59f194163d` (Binance Smart Chain)
`0xCC199aCAb2B38C6DC5f7304B29BDEb3a83E7d212` (Module address)

FireShot Capture 766 - Actions - Colony - wonker - localhost

Resolves #4086

### DIFF
--- a/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/TransactionTypesSection.css
+++ b/src/modules/dashboard/components/Dialogs/ControlSafeDialog/TransactionTypesSection/TransactionTypesSection.css
@@ -56,3 +56,25 @@
 .noUsefulMethods > section {
   padding: 1px 20px 20px;
 }
+
+.warningContainer {
+  display: flex;
+  align-items: center;
+  padding: 14px;
+  border-radius: 5px;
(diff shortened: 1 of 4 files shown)


willm30 (approved):
LGTM

PR #4144, opened 2023-01-23
Add remove nomination modal

### LINKED ISSUE #4121: Implement multicall transactions on a new colony creation
### Description

We would like to incorporate the new multicall functionality available on the latest [v11 of the contracts](https://github.com/JoinColony/colonyNetwork/releases/tag/glwss2) into the DApp with the next deployment. First off, we would like to implement this when creating a colony by reducing the required number of transactions.

In the case where you are creating a new colony with a new token (screenshot below) you can do  the transactions `Create Token`, `Create Colony`, and `Deploy Token Authority` into a (new) function called `createColonyForFrontend`. Then `setTokenAuthority` and `setTokenOwner` can be combined with multicall. The `Add extension` transaction remains its own, but the two giving permissions to the extension can be combined into one with multicall. This should result in 4 total transactions, down from the 8 as shown.

End result:
 - 1: Create new Colony
 - 2: Setup Colony's native token
 - 3: Add payment extension to Colony
 - 4: Setup permissions for payment extension

One thing to note for the extension steps, is that the last two need the address of the extension which is only certain after the extension has been has been mined.

For other combinations of colony creation transactions, the same logic can be applied i.e. to determine what can be combined through the contracts, it is wherever methods are in the same contract file, with the additional requirement that if the function ends up doing something that's needed for a second call, then they cannot be combined.

Screenshot 2023-01-17 163909

### Bonus points
Look into other actions where multiple transactions are required and implement multicall there as well.

agnieszkajarosikloj:
## Description

This PR adds a modal to confirm removing nomination (UI only).
No action is triggered when the confirm button is clicked.

To test the modal go to: **/incorporation/create** route and click on the remove button in the banner.

**New stuff** ✨

* remove nomination modal

**Figma link** 
https://www.figma.com/file/fmnKIUY0LGfEnsOOQloVVX/Korporatio?node-id=976%3A114808&t=0nX8WiVLTXFb5dxn-0

Screenshot 2023-01-23 at 23 16 16

Resolves #4121


arrenv (approved):
Thanks, looks good @agnieszkajarosikloj

PR #4175, opened 2023-02-03
Bug fix of spawn truffle

vporton:
## Description

Both
```
spawn('node_modules/.bin/truffle', ['migrate', '--reset', '--compile-all'], ...
```
and
```
spawn('./node_modules/.bin/truffle', ['migrate', '--reset', '--compile-all'], ...
```
fail on my Ubuntu Linux, even despite I create a symlink from `./node_modules/.bin/truffle` to truffle binary.

So, I fixed this bug.

**Changes** 🏗

* Add `npx` when spawning truffle.

### DIFF
--- a/scripts/deploy_contracts.js
+++ b/scripts/deploy_contracts.js
@@ -6,7 +6,7 @@ const { NETWORK_ROOT } = require('./paths');
 
 
 const deployContracts = () => {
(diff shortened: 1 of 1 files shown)


vporton:
I singed the CLA (even twice), but it shows "not signed yet".

PR #4186, opened 2023-02-08
Add old colony metadata version handling to manage safe saga

### LINKED ISSUE #4185: Unable to add a Safe with old colony metadata format
## Steps to reproduce

1. Try and add a Safe with the old colony metadata format. This is an example of the format:
```
{
  "colonyDisplayName":"Builders",
  "colonyAvatarHash":null,
  "verifiedAddresses":["0x00Da8d2F207E34B9523798FaF36fC9ff04cA9588"],
  "colonyTokens":[],
  "isWhitelistActivated":true
}
```
2. The gas station popover will continue to load with no error.
3. Checking the console, you should see the error `Error: There was an error wile fetching the current colony metadata. Please try again later.`

## Expected behaviour

* You should be able to use the feature even if you have the old colony metadata, by detecting the old format in the saga and then updating the format to the one that is required.
* While this error should not happen anymore, the typo in the error can be fixed as well, changing `wile` to `while`.

## Actual behaviour

* If your Colony is using the old colony metadata format, then interacting with the Safe control feature will not work at all, as the saga is not able to read the old format.

image

You can see an example of the old format here - https://cloudflare-ipfs.com/ipfs/QmPQCaSvTzdVbDkNHr2F9mVZwEUSYJf1t9hZzV6YVpzGbu

ArmandoGraterol:
Added old colony metadata version support by using: `https://cloudflare-ipfs.com/ipfs/QmPQCaSvTzdVbDkNHr2F9mVZwEUSYJf1t9hZzV6YVpzGbu` as an example.

To test it, I replaced `currentMetadata` with: 
`"{"colonyDisplayName":"Builders","colonyAvatarHash":null,"verifiedAddresses":["0x00Da8d2F207E34B9523798FaF36fC9ff04cA9588"],"colonyTokens":[],"isWhitelistActivated":true}"` and checked with `console.log` if the `updatedColonyMetadata` was being formatted the same way as the new colony metadata version.

Resolves #4185

### DIFF
--- a/src/modules/dashboard/sagas/actions/manageExistingSafes.ts
+++ b/src/modules/dashboard/sagas/actions/manageExistingSafes.ts
@@ -1,7 +1,10 @@
 import { call, fork, put, takeEvery } from 'redux-saga/effects';
 import { ClientType } from '@colony/colony-js';
 import isEmpty from 'lodash/isEmpty';
-import { getStringForMetadataColony } from '@colony/colony-event-metadata-parser';
+import {
+  getStringForMetadataColony,
+  getEventMetadataVersion,
+} from '@colony/colony-event-metadata-parser';
 
 import { ContextModule, TEMP_getContext } from '~context/index';
 import {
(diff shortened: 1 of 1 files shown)


arrenv (approved):
Fixes the issue on my end. Very nice @ArmandoGraterol

PR #4194, opened 2023-02-09
Update permissions for Safe Control actions

### LINKED ISSUE #4192: Safe Control Root Permissions issue
## Steps to reproduce

1. Have a user only have root permissions and try to use the Safe Control feature
2. Notice that you are not able to use the feature

## Expected behaviour

* You should be able to use the feature with only root permissions
* Permission updates
  * Interacting with arbitrary transactions requires root permissions
  * Changing colony metadata requires root permissions
  * This the permissions need to be that only root have the permissions to use Safe control thus the message should say the same.

## Actual behaviour

* You are not able to control a Safe with only root permissions

ArmandoGraterol:
Updated conditionals, events, and messages to reflect that the user would need only the root permission to make safe control related actions.

FireShot Capture 810 - Safe transaction_ asdasda - Action - Colony - wonker - localhost
FireShot Capture 811 - Safe transaction_ asdasda - Action - Colony - wonker - localhost
FireShot Capture 812 - Add Safe from Ethereum - Action - Colony - wonker - localhost
FireShot Capture 813 - Remove Safe - Action - Colony - wonker - localhost

User with no root permission:
FireShot Capture 814 - Actions - Colony - wonker - localhost

Resolves #4192


arrenv (approved):
Looks good and working as expected, thanks @ArmandoGraterol

PR #4196, opened 2023-02-10
Fix New Decision dialog scrolling up on state change

### LINKED ISSUE #4118: New Decision Dialog being scrolled to top prematurely
## Steps to reproduce

1. Try to create an agreement that is longer than the initial visible space.

## Expected behaviour

* I can scroll down and effect change to the agreement.

## Actual behaviour

* I cannot scroll past the initial visible line without being slung back to the top

[decisions-broken.webm](https://user-images.githubusercontent.com/32598350/212079700-95192be3-54ab-4db2-a596-4d3f0b4b90a1.webm)

ArmandoGraterol:
`editor.setEditable(!disabled)` was triggered every time the `editor` prop changed and that made the state of the scroll reset. The solution was just to move the culprit to their own `useEffect` function where we don't call it unnecessarily like before.

Resolves #4118

### DIFF
--- a/src/modules/core/components/RichTextEditor/RichTextEditor.tsx
+++ b/src/modules/core/components/RichTextEditor/RichTextEditor.tsx
@@ -4,6 +4,7 @@ import classnames from 'classnames';
 import { useField } from 'formik';
 
 import { InputStatus } from '~core/Fields';
+
 import Toolbar from './Toolbar';
 
 import styles from './RichTextEditor.css';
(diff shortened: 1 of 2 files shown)


willm30:
Shouldn't this be based on master? As the bug is currently in deployment I believe?

arrenv:
> Shouldn't this be based on master? As the bug is currently in deployment I believe?

Doing a `master` deployment seems unlikely to happen in the immediate, while, Safe Control will hopefully be deployed today, and I want this fix to be included.

arrenv (approved):
Looks to work for me, nice!

PR #4198, opened 2023-02-13
Update version warning on Control Safe

ArmandoGraterol:
Update colonyJS version on warning of Control Safe dialog

The latest colony network version shouldn't be included in this feature yet, so this should be easy to test (Just go to Control Safe)


arrenv (approved):
Tested on QA, working as expected.

PR #4199, opened 2023-02-13
Fix getAnnotationFromSubgraph crashing the ActionsPage

ArmandoGraterol:
Added conditional to check if the `address` prop is 1 level above on the event object or in the `values` prop. Previously this would crash the ActionsPage due to `values.address` being undefined ever since Metatransactions.

### DIFF
--- a/src/utils/events/getAnnotationFromSubgraph.ts
+++ b/src/utils/events/getAnnotationFromSubgraph.ts
@@ -35,12 +35,12 @@ export const getAnnotationFromSubgraph = async (
        * be filtering these out, and show the most recent annotation, no matter
        * who sent it
        */
-      .filter(({ address, values: { agent }, values }) => {
+      .filter(({ address, values: { agent, address: valuesAddress } }) => {
         const userAddressLowered = userAddress.toLowerCase();
         return (
-          agent.toLowerCase() === userAddressLowered ||
+          agent?.toLowerCase() === userAddressLowered ||
           address?.toLowerCase() === userAddressLowered ||
-          values.address?.toLowerCase() === userAddressLowered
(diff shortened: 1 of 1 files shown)


arrenv (approved):
Working as expected.

src/utils/events/getAnnotationFromSubgraph.ts
```diff
@@ -35,11 +35,13 @@ export const getAnnotationFromSubgraph = async (
...
-      .filter(
-        ({ values: { agent, address } }) =>
-          agent.toLowerCase() === userAddress.toLowerCase() ||
-          address.toLowerCase() === userAddress.toLowerCase(),
-      ) || [];
-
+      .filter(({ address, values: { agent }, values }) => {
+        const userAddressLowered = userAddress.toLowerCase();
+        return (
+          agent.toLowerCase() === userAddressLowered ||
```
jakubcolony, line 41: Are we positive this will never be `undefined`?

src/utils/events/getAnnotationFromSubgraph.ts
```diff
@@ -35,11 +35,13 @@ export const getAnnotationFromSubgraph = async (
        * be filtering these out, and show the most recent annotation, no matter
        * who sent it
        */
-      .filter(
-        ({ values: { agent, address } }) =>
-          agent.toLowerCase() === userAddress.toLowerCase() ||
-          address.toLowerCase() === userAddress.toLowerCase(),
-      ) || [];
-
+      .filter(({ address, values: { agent }, values }) => {
```
jakubcolony, line 38: We could also destructure the `address` from `values` (and rename it, so it doesn't conflict)

jakubcolony (approved):
All good 👍

PR #4212, opened 2023-03-06
Update regex on getChainNameFromSafe

### LINKED ISSUE #4208: Including brackets "()" in a Safe name breaks the control a safe
## Steps to reproduce

1. Add a Safe to a colony and include in the name brackets for e.g. `(on ETH)`.
2. Try to control a safe, you will receive a not able to connect message.

2023-02-28 11_31_13-Actions _ Colony - Ameva

You can see it working fine without the brackets.

37520885-ab419587db81c7ab2dc7955029cd1f88

## Expected behaviour

* You should be able to use brackets in a Safe name

## Actual behaviour

* Using brackets in a safe name breaks the `getTxServiceBaseUrl` function in `safeHelpers.ts`

ArmandoGraterol:
Updated regex on `getChainNameFromSafe` so that it only matches the last parenthesis on the string, which should always be the chain name.

To quickly test the regex with string, we can use this: https://regex101.com/r/CMMXjQ/1

FireShot Capture 816 - Actions - Colony - wonker - localhost

Resolves #4208

### DIFF
--- a/src/modules/dashboard/sagas/utils/safeHelpers.ts
+++ b/src/modules/dashboard/sagas/utils/safeHelpers.ts
@@ -393,5 +393,5 @@ export const getContractInteractionData = async (
 };
 
 export const getChainNameFromSafe = (safeDisplayName: string) => {
-  return safeDisplayName.match(/\((.*)\)/)?.pop() || '';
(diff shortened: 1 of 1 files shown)


willm30 (approved):
Looking good to me 😎

PR #4213, opened 2023-03-10
Fix `npm run provision` not building `colonyNetwork` successfully

ArmandoGraterol:
Self-explanatory title. Now in order to run the dev env of the Dapp we are going to require the npm version to specifically be `8.1.4` in order to avoid problems when installing packages inside `colonyNetwork`.

To test, make sure to delete the `colonyNetwork` lib folder if you already have it and go throught the normal dev process and see if everything works as expected. 

Example: `npm i` > `npm run provision` (check that `colonyNetwork` is built successfully with no leftover changes) > `npm run dev:heavy` > Create a user > Create a colony.

### DIFF
--- a/.github/workflows/gh-pages.yml
+++ b/.github/workflows/gh-pages.yml
@@ -25,7 +25,7 @@ jobs:
           ssh-private-key: ${{ secrets.DEPLOY_KEY }}
 
       - name: npm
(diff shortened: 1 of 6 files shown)


rdig (approved):
Yep, good to go.

I'm going to also drop the review requirements down to 1 in order for you be able to merge stuff faster

PR #4214, opened 2023-03-23
Fix users without permissions not being able to create Safe motions

ArmandoGraterol:
- Also, I've removed the `hasRegisteredProfile` conditional from the `AdvancedDialog` and `ManageSafeDialog` as it seems kind of redundant to have it there... If the user can access the UAC, it means that they have a registered profile already as the `New Action` button is disabled otherwise, and there's no other way for users to get to this flow.

### User with no permission in a colony without reputation:
FireShot Capture 819 - Actions - Colony - wonker - localhost

### User with permission in a colony without reputation:
FireShot Capture 820 - Actions - Colony - wonker - localhost

### User with permission in a colony with reputation:
FireShot Capture 821 - Actions - Colony - wonker - localhost

### User with no permission in a colony with reputation:
FireShot Capture 822 - Actions - Colony - wonker - localhost


rdig (approved):
Seems all good.

On Monday I'll do a new deployment anyway to fix a token claims bug, so this will get deployed as well.

PR #4215, opened 2023-03-27
Fix NFT transfers breaking Colony transfer claims

rdig:
This PR fixes a issue that showed up in production where, after a Colony received a NFT transfer, all the subsequent Token transfers would not show up.

This is because both ERC-20 and ERC-721 emit the same `Transfer` event, but with one of the arguments changed. Because of this the events would break when being parsed.

For more details see the comment I left in the resolver where the fix is applied.


willm30 (approved):
Nice work catching this 👏

PR #4218, opened 2023-08-01
Fix Transfer NFT section on Control Safe

ArmandoGraterol:
The API endpoint we used to get NFTs was updated (Both the URL and the data structure of the response.), therefore it wasn't working before. This should fix it and make the section available once more.


rdig (approved):
:+1:

